PR #17, opened 2015-01-03
Fixed build against rust nightly.

ktossell:
Added `use` statements for `cmp::Ordering`, `ops::{Add, Sub}`, `borrow::IntoCow`, `num::ToPrimitive`

Changed `deriving` -> `derive`. Changed `[x, ..N]` -> `[x; N]`.

Made `format(f: &fmt::Formatter)` return `fmt::Result` instead of
`IoResult`. Had to raise generic errors in a couple of cases where
`IoResult`'s fields were being set (but these errors were being thrown
out anyway).

Temporarily set `#![feature(old_orphan_check)]` because `#[deriving(Hash)]` needs it (see https://github.com/rust-lang/rust/issues/20477).

Thanks to @eddyb for explaining the format() situation.

### DIFF
--- a/src/date.rs
+++ b/src/date.rs
@@ -7,6 +7,8 @@
  */
 
 use std::{fmt, hash};
+use std::cmp::Ordering;
+use std::ops::{Add, Sub};
 
 use {Weekday, Datelike};
 use duration::Duration;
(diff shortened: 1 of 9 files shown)


lifthrasiir:
Thank you! Regarding `format`, I think it needs to verify the format string much earlier than today...

PR #24, opened 2015-01-27
Added parsing functions.

John-Nagle:
Added date/time parsing functions for RFC 2822 ("Tue, 20 Jan 2015 17:35:20 -0800", as used in email) and RFC3339/ISO8601 ("2015-01-20T17:35:20.001-0800", as used in Atom, microformats, JSON, etc.) date/time strings. Tests included and run successfully.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,4 +16,7 @@ name = "chrono"
 
 [dependencies]
 time = "0.1.14"
+regex = "*"
+regex_macros = "*"
+
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ name = "chrono"
 
 [dependencies]
 time = "0.1.14"
-regex = "*"
-regex_macros = "*"
+regex = "0.1.12"
+regex_macros = "0.1.6"
 
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -78,6 +78,7 @@ Addition and subtraction is also supported.
 The following illustrates most supported operations to the date and time:
 
 ~~~~ {.rust}
+
 # #![allow(unstable)]
 # /* we intentionally fake the datetime...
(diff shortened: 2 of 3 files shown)


lifthrasiir:
Oh, my... I was preparing my own (streamlined) parsing facility! It was being done in the private branch, which I just pushed to the public [`new-format` branch](https://github.com/lifthrasiir/rust-chrono/tree/new-format). For now there is no parsing code, but it does contain some preparation for getting it as smooth as possible.

Besides from this, I'm not sure what to do with this PR. I do think that there should be some shortcut formatting and parsing functions around `DateTime`, but which formats to support is unclear. I'm reasonably neutral on the additional `regex` dependency (though my branch tries to avoid that).

John-Nagle:
I'd suggest finishing your parser, then providing shortcut
functions for RFC3339/ISO8601 and RFC2822 dates.  Then take
the test cases from my code to check those.

    When do you think you will have this done?  I wrote these
because I'm writing an RSS/Atom feed reader and need those
functions.

			John Nagle

lifthrasiir:
@John-Nagle There is no hard deadline as I'm doing this in my spare time, but it is currently given the highest priority possible (next to the daily job). Please allow me a bit more, though. :)

lifthrasiir:
@John-Nagle Okay, I'm almost done with `new-format` branch. I cannot merge this PR as is, but considering that the parser for RFC 2822 and 3339 cannot be composed of other parsers (there are subtle differences), I want to reuse your code to get a working implementation while getting rid of regex codes (and to have the proper attribution). Are you fine with this?

John-Nagle:
Sure.  That is fine with me.

				John Nagle

lifthrasiir:
I was quite late, but this PR is now fully merged to `new-format` branch which will eventually become 0.2. Thank you!

PR #28, opened 2015-03-02
Remove old closure kind syntax

tshakah:
Chrono was broken by https://github.com/rust-lang/rust/pull/22884


lifthrasiir:
Thank you!

PR #33, opened 2015-04-24
Language update

guersam:
This PR introduces new dependency `num` as it took over many old features from `std::num`. Is it acceptable?


lifthrasiir:
Thank you! Actually I want to avoid `num` dependency, but getting it work is first.

PR #36, opened 2015-04-28
Add optional rustc-serialize support

guersam:
This PR support deriving `rustc-serialize` traits when `rustc-serialize` feature is enabled, as the same way as implemented in `time`.

Temporary or format-related structs and enums are excluded.


lifthrasiir:
I'm still not sure about the exact format (e.g. is `NaiveDate::ymdf` stable enough for exponsure?), but this is a lot better than nothing. Thank you!

PR #45, opened 2015-09-03
New formatters %.3f, %.6f, %.9f

tomgco:
I needed to print out a fixed precision to ensure that the output would always match no matter what the input was, for example `70ms` needs to output `.070` and `0ms` needs to print `.000`.

I am really new to rust and I hope that this is OK. 

I tried to create a macro that would construct Nanosecond{3,6,9} however to do that I needed to use `std::concat_idents!` which was unstable so you are left with a more verbose matching. If you have any improvements let me know on this PR let me know.

Thanks.

### DIFF
--- a/src/format/mod.rs
+++ b/src/format/mod.rs
@@ -125,6 +125,12 @@ pub enum Fixed {
     /// May print nothing, 3, 6 or 9 digits according to the available accuracy.
     /// See also `Numeric::Nanosecond`.
     Nanosecond,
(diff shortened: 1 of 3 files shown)


lifthrasiir:
@tomgco I was thinking hard about the generic width/precision specification (which I didn't seriously try to adapt, but glibc and some others have that and I assume we eventually need to do so), but #46 will force pushing a new version to crates.to anyway and I don't see any further reason to delay accepting this except that we will want to do this more generally. Thank you and sorry for delay!

Issue #42, opened 2015-08-08
Generated `RustcDecodable` impl is not very safe


lifthrasiir:
Since Chrono does not contain an unsafe code this is not a much threat, but it can destroy internal consistency in subtle ways (e.g. February 29 in the non-leap year). We may want to add a verification procedure to the `RustcDecodable` impl, or at least a method to do that posthumously.

PR #48, opened 2015-09-09
Add serde serialization support

polyfractal:
Adds optional serializaiton support through serde, basically mimicking rustc-serialization added in #36.  

Requires the nightly version of serde (e.g. compiler plugins and macros) instead of the stable version.  The stable version was just too invasive to easily add to Chrono without a lot of changes.

Because cargo cannot have a feature and a dependency groups share the same name, the feature for this is called "serde_support" instead of "serde".

Sorry about all the trailing new lines being removed in the diffs, not sure why Atom decided to eat those.  Lemme know and I can fix those if you want.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,10 @@ env:
 script:
   - cargo build -v
   - cargo build -v --features rustc-serialize
+  - cargo build -v --features serde_support
   - cargo test -v
   - cargo test -v --features rustc-serialize
+  - cargo test -v --features serde_support
   - cargo doc
 after_script:
   - cd target && curl http://www.rust-ci.org/artifacts/put?t=$RUSTCI_TOKEN | sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+language: rust
 rust:
     - stable
     - nightly
@@ -11,10 +12,12 @@ env:
 script:
   - cargo build -v
   - cargo build -v --features rustc-serialize
-  - [ "$TRAVIS_RUST_VERSION" = "nightly" ] && cargo build -v --features serde_support
   - cargo test -v
   - cargo test -v --features rustc-serialize
-  - [ "$TRAVIS_RUST_VERSION" = "nightly" ] && cargo test -v --features serde_support
+  - if [ "$TRAVIS_RUST_VERSION" = "nightly" ]; then
+      cargo build -v --features serde_support;
+      cargo test -v --features serde_support;
+    fi
   - cargo doc
 after_script:
   - cd target && curl http://www.rust-ci.org/artifacts/put?t=$RUSTCI_TOKEN | sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
-language: rust
+rust:
+    - stable
+    - nightly
 os:
   - linux
   - osx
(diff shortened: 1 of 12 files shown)


polyfractal:
Argh, I'm not very good at travis...how do we make the serde tests only run on nightly?

lifthrasiir:
@polyfractal According to [this Travis help](http://docs.travis-ci.com/user/languages/rust/), you may add the following to `.travis.yml`:

    rust:
        - stable
        - nightly

...then use `[ "$TRAVIS_RUST_VERSION" = "nightly" ] && ...` to guard the command from running in non-nightly versions.

lifthrasiir:
I have no particular objection here, but problems raised in #42 equally apply to this PR. We want to...

* Avoid spilling auxiliary information to the serialization (e.g. year flags in `NaiveDate` should not be encoded);
* Validate the decoded values (the generated implementation cannot); and
* Allow other formats like ISO 8601 to be decoded.

I'm not well versed in serde, so any guidance or even patch would be appreciated (not mandatory for this PR).

polyfractal:
I removed the auto-derivation from `YearFlags`.  Everything seems to work alright, but I'm new to Chrono so I dunno if this is alright?

I also added some *very* basic serialization tests for the main components.  It basically uses bincode to serialize a date/time to a Vec<u8> then back, and checks to make sure various formatting is consistent.  Lots of holes and edge cases, but it at least adds some kind of verification.

Also, theoretically, fixed the travis build ... except it is failing on the nightlies because it doesn't seem to be building correctly (none of the conditional flags seem to be triggered).  It's building/testing just fine on my machine, so I'm not sure what the problem is.  I'll work on it some more tomorrow.

lifthrasiir:
> I removed the auto-derivation from `YearFlags`. Everything seems to work alright, but I'm new to Chrono so I dunno if this is alright?

For example, `NaiveDate`s [keep additional information](https://github.com/lifthrasiir/rust-chrono/blob/master/src/naive/date.rs#L1455-L1463) to speed various operations up. The additional information inside `NaiveDate` (currently, lower 13 bits of `u32`) need not to be encoded.

> I also added some *very* basic serialization tests for the main components. It basically uses bincode to serialize a date/time to a Vec then back, and checks to make sure various formatting is consistent. Lots of holes and edge cases, but it at least adds some kind of verification.

What I'm particularly concerned is that one can modify the encoded data to break the internal invariant of Chrono. For example, say that we are encoding `NaiveDate::from_ymd(2015, 12, 30)` as two words `[2015, 364]` (because Dec 30 is 364th day in that year); if we instead decode `[2015, 368]`, there is no guarantee that this will fail upright or at least causes an early runtime error.

polyfractal:
Ah, ok, I understand better now.  Thanks.

> For example, NaiveDates keep some additional information to speed various operations up. The additional information inside NaiveDate (currently, lower 13 bits of u32) need not to be encoded.

Is it problematic if these extra bits *do* get encoded/serialized?  The alternative is to extract the valid 19 bits and serialize it as a u16 + u8 (with 5 unused bits).  But this seems awfully complicated if we don't really care if the lower 13 tag along for the ride.

> What I'm particularly concerned is that one can modify the encoded data to break the internal invariant of Chrono. For example, say that we are encoding NaiveDate::from_ymd(2015, 12, 30) as two words [2015, 364] (because Dec 30 is 364th day in that year); if we instead decode [2015, 368], there is no guarantee that this will fail upright or at least causes an early runtime error.

Yeah, this makes sense.  If we decode a corrupted (maliciously or otherwise) date, we bypass all the validation you have in place.  I think we'd need to write out the serde deserializer method by hand and force it to run Chrono-specific validation at the end.

I'm just as new to Serde as I am to Chrono, but I think this shouldn't be too terrible.  I can piggyback on the autogenerated code to use as boilerplate.

lifthrasiir:
> Is it problematic if these extra bits *do* get encoded/serialized?

Not directly, but I think it is better to decouple the serialization format from hairy internals of Chrono. The year-ordinal format is arguably a better format for interoperability.

lifthrasiir:
I was unable to concentrate on Chrono for months (sorry), but it seems that #51 has a better treatment on issues I've raised. I decided to close this in favor of #51, but thank you for your PR!

polyfractal:
++, not a problem for me!  I was quickly getting out of my depth here, happy someone else (who knew what they were doing) added it instead :)

PR #51, opened 2015-11-19
Added serde support for DateTime, NaiveDate, NaiveTime and NaiveDateTime.

martinrlilja:
I tried to take a stab at adding serde support as it seems as if the other attempt was abandoned. This one uses debug formatting for serialization and parse (FromStr) for deserialization. It also works in Rust stable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,10 @@ env:
 script:
   - cargo build -v
   - cargo build -v --features rustc-serialize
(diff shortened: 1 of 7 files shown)


lifthrasiir:
Thank you, I was unable to concentrate on Chrono for months, but I think this is better than #48 with respect to my concerns (including the validation problem).

PR #54, opened 2015-12-23
make ParseErrorKind public

KodrAus:
It'd be nice to be able to run a match over the ParseError returned from an attempted parse. Making the ParseErrorKind public means you can match `ParseError(ParseErrorKind::NotEnough))` for example.

### DIFF
--- a/src/format/mod.rs
+++ b/src/format/mod.rs
@@ -184,7 +184,7 @@ macro_rules! fix  { ($x:ident) => (Item::Fixed(Fixed::$x)) }
 pub struct ParseError(ParseErrorKind);
 
 #[derive(Debug, Clone, PartialEq, Copy)]
-enum ParseErrorKind {
+pub enum ParseErrorKind {
     /// Given field is out of permitted range.
     OutOfRange,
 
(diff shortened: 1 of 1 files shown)


lifthrasiir:
Hmm, I modelled `ParseError` after [`std::num::ParseFloatError`](http://doc.rust-lang.org/std/num/struct.ParseFloatError.html) which internal is private. I was reluctant to expose the error internals so far since it takes more time to stabilize errors. For the purpose of KodrAus/elasticsearch-rs#3, it might be better to implement a partial parsing function that will return a slice of the input instead of `TooShort` error; maybe `ParseError::is_partial()` method may be better for your use case.

KodrAus:
Ahk thanks, I'll give that a go :) Great work on this library, it's really nice to work with

KodrAus:
So I ended up using the `Parsed` struct which makes it easy to inspect the results of parsing, without having to let it fail first. Much better. Thanks!

lifthrasiir:
Glad to hear that! I originally intended to merge the PR (or work on my own) if something like `ParseError::is_partial()` is necessary, but it seems that `Parsed` is sort of a swiss-army knife :)

PR #56, opened 2016-01-06
Add `hours` offset creation method for FixedOffset

bfrog:
This adds a simple method to let me create FixedOffset values using whole simple timezone hour offsets.

Most common timezones are given as whole hour offsets from UTC making this a nice convenience over the seconds east/west methods.

### DIFF
--- a/src/offset/fixed.rs
+++ b/src/offset/fixed.rs
@@ -22,6 +22,16 @@ pub struct FixedOffset {
 }
 
 impl FixedOffset {
+    /// Makes a new `FixedOffset` using a relative hour value from UTC
+    ///
+    /// For example America/Chicago would be -6 hours from UTC (CST) and could
+    /// be created using FixedOffset::hours(-6).
+    ///
+    /// Panics when out-of-bound `hours`.
+    pub fn hours(hours: i32) -> FixedOffset {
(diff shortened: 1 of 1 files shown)


lifthrasiir:
Thank you for the PR. I intentionally have avoided a single set of methods for making `FixedOffset`, however, because both conventions for time zone offsets are commonly used. (`east` is common for written notation, `west` is common for internal notation.) I think `east_hours` and `west_hours` may be a better fit; what do you think?

bfrog:
I was actually very confused by the west/east methods as all the ISO
timezones are +/- hours from utc... I wasn't exactly sure what I would get
with the east/west until I took a closer look.

lifthrasiir:
@bfrog That sounds great, because I *wanted* any user using that API to take a closer look. (Hopefully the names `east` and `west`, once recognized, should be clear that the hemisphere in the name gets a positive sign. If it were not please let me know.)

bfrog:
Hmm... well I think if you have east/west methods then perhaps they should only allow positive values, and the documentation explain closer how they map to ISO string timezone offset.

lifthrasiir:
@bfrog Aha, that explains the issue. Maybe I should put an explicit example of negative inputs for `east` and `west`. I was also thinking about a small sized example per module for such cases. Thank you for the input!

PR #57, opened 2016-01-09
Don't depend on num's default features; removing rand requirement

dten:
requirement for `rand` comes from `num`'s default features which are all unused in `chrono`
not including the default features means that that `build` no longer requires `rand` (and hence does not pass it on to others importing `chrono`)

Not quite sure why `cargo test` still wants rand, but it's to do with the `serde_json` dev-dependency

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ name = "chrono"
 
 [dependencies]
 time = "0.1"
-num = "0.1"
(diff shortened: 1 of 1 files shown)


lifthrasiir:
Good observation. I never thought that Chrono transitively depends on `rand` until this point. Thank you for the PR!

PR #65, opened 2016-03-23
Clarify timezone conversion in docs

dbrgn:
I had to search for it for a while.

On a related note, why are traits like `std::convert::From<FixedOffset> for UTC` not implemented? Any specific reason, or could these be added?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -49,9 +49,10 @@
 //!
 //! Chrono provides a `DateTime` type for the combined date and time.
 //!
-//! `DateTime`, among others, is timezone-aware and
-//! must be constructed from the `TimeZone` object.
-//! `DateTime`s with different time zones do not mix, but can be converted to each other.
+//! `DateTime`, among others, is timezone-aware and must be constructed from
+//! the `TimeZone` object.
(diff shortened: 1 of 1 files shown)


lifthrasiir:
Thank you!

> On a related note, why are traits like `std::convert::From<FixedOffset> for UTC` not implemented? Any specific reason, or could these be added?

`From` implementation is absent simply because Chrono is much older than `From` (introduced in Rust 1.0). That said, I'm not sure how to convert from `FixedOffset` to `UTC`: `UTC` is a zero-sized type, so it is clear that multiple `FixedOffset` will map to `UTC` which sounds critically wrong. What conversion do you have in mind?

dbrgn:
Isn't that what `my_fixedoffset_datetime::with_timezone(UTC)` does?

lifthrasiir:
@dbrgn Well, that would be `From<DateTime<FixedOffset>> for DateTime<UTC>`. `FixedOffset` and `UTC` types are just for timezone information.

dbrgn:
Ah, that's actually what I meant :)

I can try to come up with a pull request in the next days if you want.

lifthrasiir:
@dbrgn Yeah, thank you again!

PR #66, opened 2016-03-28
Fix parsing LongWeekday when day is Sunday

jwilm:
I was working on parsing the `HTTP-date` format, and I discovered that the RFC850 variant was failing to parse! It turned out that this was an issue parsing Sunday at the front of the string. There's now a test that covers each day of the week. The test doubles as verification that chrono can parse RFC850 using the public API.

---
This change is [Reviewable](https://reviewable.io/reviews/lifthrasiir/rust-chrono/66)


lifthrasiir:
Wow, thank you for finding that! I'm late today so I'll issue a new release tomorrow.

jwilm:
Thanks!

PR #67, opened 2016-03-29
Add From impls for DateTime

dbrgn:
This is a first attempt at implementing the `std::convert::From` trait.

First of all, this only overwrites the timezone, but does not actually convert the datetime, right? If that's the case, I'm not sure if that's semantically the right thing to do in a `From` implementation. Are there already functions for real conversion?

Also, I did not manage to implement `From` generically for all timezones. I'm not sure that's possible.


quodlibetor:
This functionality was merged in pr #271

PR #77, opened 2016-06-14
Clean up Duration intro doc

JohnHeitmann:
• Make a Duration description the first thing mentioned instead of project history.
• Add "magnitude" to the description to disambiguate it from Interval (in the Joda sense).
• Brush up some awkward language.
• Add a doc link and illustrate the module namespace.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,10 +46,11 @@ extern crate chrono;
 
 ### Duration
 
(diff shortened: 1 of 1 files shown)


lifthrasiir:
Thank you!

PR #78, opened 2016-06-14
Minor DateTime intro doc update

JohnHeitmann:
• Put timezone in the lede since it's critical.
• Mention Instant and SystemTime to put DateTime in context


lifthrasiir:
Thank you!

Issue #74, opened 2016-05-13
Timestamp in milliseconds/nanoseconds


craftytrickster:
I would like to have the ability to get the amount of milliseconds (or maybe even nanoseconds) since unix epoch. 

Currently, this seems to only be available in seconds:
https://lifthrasiir.github.io/rust-chrono/chrono/datetime/struct.DateTime.html#method.timestamp

Is there a reason not to provide two extra methods like?
`timestamp_milliseconds`
`timestamp_nanoseconds`

If it is indeed desired, I can open a PR to add them.

Please let me know.

craftytrickster:
I would have closed this much earlier, but for some reason I was never received an email notification about the above PR.

Thanks for the work.

lifthrasiir:
Ah indeed, I guess it wasn't closed automatically since it lacked comments in the form of "Fixes #ISSUENO" or similar.

PR #81, opened 2016-07-13
Add functions to get milli/micro/nano-seconds from a DateTime

beneills:
Using the underlying naive::NaiveTime fractional part, we compute
the number of milli/micro/nano-seconds since the last second boundary.

The reason for not computing elapsed time since 1970 is because we
would hit potential issues of i64s not being large enough (the range
would be strictly smaller than the 64bit-timestamp range, causing
compatibility issues).

This was written in response to issue #74.

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -60,6 +60,30 @@ impl<Tz: TimeZone> DateTime<Tz> {
         self.datetime.timestamp()
     }
 
+    /// Returns the number of milliseconds since the last second boundary
+    ///
+    /// note: this is not the number of milliseconds since January 1, 1970 0:00:00 UTC
+    #[inline]
+    pub fn timestamp_milliseconds_part(&self) -> u32 {
+        self.datetime.timestamp_milliseconds_part()
+    }
+
+    /// Returns the number of microseconds since the last second boundary
+    ///
+    /// note: this is not the number of microseconds since January 1, 1970 0:00:00 UTC
+    #[inline]
+    pub fn timestamp_microseconds_part(&self) -> u32 {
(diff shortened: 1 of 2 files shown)


lifthrasiir:
I think it is a fine addition, but I have some questions:

- Shouldn't the names of those methods follow [`Duration::subsec_nanos`](https://doc.rust-lang.org/std/time/struct.Duration.html#method.subsec_nanos)? (My opinion: Yes.)

- How should they behave on the leap seconds? The nanosecond part may exceed 999,999,999 in that case, and you need a clear policy for that. (My opinion: Let they return the value exceeding 1s, and document that caveat in big letters. The proper formatting should go through `format` which already handles leap seconds correctly, and the accessors should be considered as a "raw" format.)

- Probably nitpicking, but please avoid adding spaces in irrelevant lines ;) You can surgery them with `git add -p` or similar tools.

beneills:
> Shouldn't the names of those methods follow Duration::subsec_nanos? (My opinion: Yes.)

Renamed according to this schema.

> How should they behave on the leap seconds? The nanosecond part may exceed 999,999,999 in that case, and you need a clear policy for that. (My opinion: Let they return the value exceeding 1s, and signal that caveat loudly. The proper formatting should go through format which already handles leap seconds correctly, and the accessors should be considered as a "raw" format.)

Added prominent Rustdoc warnings to all accessors.

> Probably nitpicking, but please avoid adding spaces in irrelevant lines ;) You can surgery them with git add -p or similar tools.

Fixed.  Apologies for this: was trying out a new editor.

lifthrasiir:
@beneills I've squashed and merged the PR. Thank you!

PR #84, opened 2016-07-30
Fix spilling of exceeding `tm_sec` into `tm_nsec` in `tm_to_datetime()`.

sergey-v-galtsev:


lifthrasiir:
Oh my, this was a very subtle typo. Thank you for pointing it out---How did you found it, by the way?

sergey-v-galtsev:
I've spotted it by an accident while browsing the source.
(I am deciding which library to use for formatting `std::time::SystemTime`.)

PR #86, opened 2016-08-02
Update serde

alexbool:
Well, `chrono` also works with older versions of `serde`, so I decided to use `<0.9` as version bound


lifthrasiir:
Thank you! (You are right that Chrono works with serde 0.7 as well. I personally haven't tried 0.8 yet, but CI is already green, so yeah.)

alexbool:
It would also be great to publish a release

lifthrasiir:
@alexbool I've taken this opportunity to make a quick fix for #42. Now done!

alexbool:
Thanks!

PR #87, opened 2016-08-02
upgrade serde deps

KodrAus:
Just updating `serde` to the latest version.


lifthrasiir:
Essentially same to #86 which has been now merged. Thank you nevertheless!

PR #89, opened 2016-08-16
Support deserializing from bincode

dtolnay:
Bincode is a minimal format that expects the Deserialize implementation to tell it what type of data it should expect to see.

Fixes https://github.com/TyOverby/bincode/issues/85.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,3 +22,4 @@ serde = { version = "<0.9", optional = true }
 
 [dev-dependencies]
 serde_json = { version = ">=0.7.0" }
+bincode = { version = "0.6", features = ["serde"], default-features = false }
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -496,7 +496,7 @@ mod serde {
         fn deserialize<D>(deserializer: &mut D) -> Result<Self, D::Error>
             where D: de::Deserializer
         {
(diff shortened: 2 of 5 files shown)


lifthrasiir:
@dtolnay Looks good. The test is missing though---would the following test be enough?

```rust
    #[test]
    fn test_serde_bincode() {
        use self::bincode::SizeLimit;
        use self::bincode::serde::{serialize, deserialize};

        let dt = UTC.ymd(2014, 7, 24).and_hms(12, 34, 6);
        let encoded = serialize(&dt, SizeLimit::Infinite).unwrap();
        let decoded: DateTime<UTC> = deserialize(&encoded).unwrap();
        assert_eq!(dt, decoded);
        assert_eq!(dt.offset(), decoded.offset());
    }
```

dtolnay:
Added tests for all 4 types.

lifthrasiir:
I don't know why Travis is blocked, but it is highly unlikely that this commit will have a different behavior per platform. Thank you, @dtolnay!

jprafael:
@lifthrasiir This fix has been landed after the release of 0.2.25 which is the version that is currently published on https://crates.io/crates/chrono. Could you make a 0.2.26 release? Right now working around the issue by compiling against the git repository.

Thanks for the awesome lib!

PR #97, opened 2016-10-01
Add support for assignment operators. Fixes #70

### LINKED ISSUE #70: add support for assignment operators
Rust 1.8 stabilized assignment operator such as += and -= 

I think they would fit the API nicely.

ijanos:
Because the non-naive dates are non-copy and the Add/Sub trait is not implemented for references, a clone is needed on on the non-naive types because Add/Sub consumes the value.

Requires Rust 1.8 or higher.

### DIFF
--- a/src/date.rs
+++ b/src/date.rs
@@ -8,7 +8,7 @@
 
 use std::{fmt, hash};
 use std::cmp::Ordering;
-use std::ops::{Add, Sub};
+use std::ops::{Add, Sub, AddAssign, SubAssign};
 
 use {Weekday, Datelike};
 use duration::Duration;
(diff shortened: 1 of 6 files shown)


lifthrasiir:
I'm currently undecided on adding `Add/SubAssign` variants to `DateTime`, mostly because the addition and subtraction themselves are undecided (keep in mind that Chrono is still at 0.2 :S). Naive types seem to fine to have them, though---I'm happy to accept the reduced version of this PR.

ijanos:
I just realized I committed wrong AddAssign implementations on the naive types here. I will soon create a new a pull request with updated code and tests on the naive types.

If you decide to keep Add and Sub then add the Assigns too, you can create more readable code with them in my opinion.

PR #101, opened 2016-10-04
Clarify that Duration is an "accurate" duration

### LINKED ISSUE #100: Clarify that `Duration` is for the accurate duration only
ISO 8601 distinguishes between "accurate duration" (exact amount of time) and "nominal duration" (relative to conceptual years/months/weeks/days). For example, adding an "accurate duration" of 24 hours to a datetime just before a shift in daylight savings time will give a different result than adding a "nominal duration" of 1 day to that datetime. The duration formats specified in ISO 8601 represent a combination of accurate (H/M/S) and nominal (W/Y/M/D) components: `PnnYnnMnnDTnnHnnMnnS` or `PnnW`.

`chrono`'s [`Duration`](https://lifthrasiir.github.io/rust-chrono/chrono/struct.Duration.html) is confusing in relation to ISO 8601 because ISO 8601 specifies days and weeks as "components with nominal duration", but the `Duration::weeks()` and `Duration::days()` functions are implemented as "accurate" durations.

I propose one of the following two solutions:

1. Change the behavior of `Duration` to match the ISO 8601 duration format: treating weeks and days as components with nominal duration and adding additional functions for the other components specified in ISO 8601 (year and month). Note that this is a breaking change from the current behavior and would require adding additional struct members.

2. Take Java's approach to the issue by using separate [`Period`](https://docs.oracle.com/javase/8/docs/api/java/time/Period.html) (nominal) and [`Duration`](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html) (accurate) types. This would be backwards-compatible but would introduce an extra type.

By the way, this issue is similar to #52.

For reference, ISO 8601:2004 says the following regarding durations:

> ### 4.4.3 Duration
>
> #### 4.4.3.1 General
>
> Duration can be expressed by a combination of components with accurate duration (hour, minute and second) and components with nominal duration (year, month, week and day). The term duration will be used to designate expressions containing components with accurate duration, with nominal duration, or both.
>
> * NOTE 1: Applications requiring accurate duration using the time unit day can make use of the provisions of ISO 31-1.
>
> ...

and some definitions from earlier in the standard:

> ### 2.1.6: duration
> non-negative quantity attributed to a time interval, the value of which is equal to the difference between the time points of the final instant and the initial instant of the time interval, when the time points are quantitative marks
> * NOTE 1: In the case of discontinuities in the time scale, such as a leap second or the change from winter time to summer time and back, the computation of the duration requires the subtraction or addition of the change of duration of the discontinuity.
> * NOTE 2: Duration is one of the base quantities in the International System of Quantities (ISQ) on which the International System of Units (SI) is based. The term "time" instead of "duration" is often used in this context.
> * NOTE 3: For the term "duration", expressions such as "time" or "time interval" are often used. The term "time" is not recommended in this sense and the term "time interval" is deprecated in this sense to avoid confusion with the concept "time interval".
> * NOTE 4: The SI unit of duration is the second.
>
> ### 2.1.7: nominal duration
> duration expressed amongst others in years, months, weeks or days
> * NOTE: The duration of a calendar year, a calendar month, a calendar week or a calendar day depends on its position in the calendar. Therefore, the exact duration of a nominal duration can only be evaluated if the duration of the calendar years, calendar months, calendar weeks or calendar days used are known.

jturner314:
The ISO 8601 format includes both "nominal" (year, month, week, and day) and "accurate" (hour, minute, and second) components. However, the `Duration` type only represents an "accurate" duration because arithmetic with nominal components is not defined in ISO 8601.

This PR fixes #100.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -49,11 +49,13 @@
 //!
 //! ### Duration
 //!
-//! [**`Duration`**](./struct.Duration.html)
-//! represents the magnitude of a time span. `Duration` used to be provided by Chrono.
-//! It has been moved to the `time` crate as the
-//! [`time::Duration`](https://doc.rust-lang.org/time/time/struct.Duration.html) type, but is
-//! still re-exported from Chrono.
+//! [**`Duration`**](./struct.Duration.html) represents the magnitude of a time
+//! span. Note that this is an "accurate" duration represented as seconds and
+//! nanoseconds and does not represent "nominal" components such as days or
+//! months. `Duration` used to be provided by Chrono. It has been moved to the
+//! `time` crate as the
+//! [`time::Duration`](https://doc.rust-lang.org/time/time/struct.Duration.html)
+//! type, but is still re-exported from Chrono.
 //!
 //! ### Date and Time
 //!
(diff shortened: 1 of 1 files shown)


lifthrasiir:
Thank you!

Issue #88, opened 2016-08-06
Easy way to convert timestamp to UTC object?


rory:
As near as I can tell, the only way to convert a timestamp (i.e. number of seconds since UNIX epoch¹) is with [`NaiveDateTime::from_timestamp()`](https://lifthrasiir.github.io/rust-chrono/chrono/naive/datetime/struct.NaiveDateTime.html#method.from_timestamp), however that returns a DateTime without a timezone... but if we know the unix epoch time, then we know exactly what second it is? a timestamp of 60 is 00:01 UTC on 1st Jan 1970, it's not like it can be in a some other timezone.

How about some method that'll take a unix epoch and return a DateTime in UTC timezone?

---

¹ I'm aware of leap seconds

lifthrasiir:
@rory Wouldn't `UTC.timestamp(secs, nanos)` be enough for your use case?

rory:
On 06/08/16 17:27, Kang Seonghoon wrote:
> @rory <https://github.com/rory> Wouldn't |UTC.timestamp(secs, nanos)| be
> enough for your use case?

D'oh! It looks like it might. I obviously didn't find that in the docs.
I was looking for `from_timestamp`, so that didn't come up. I'll try
that out. Thanks! 👍

lifthrasiir:
@rory Great! Please close the issue if it worked :)

> I was looking for `from_timestamp`, so that didn't come up.

Huh, that may be confusing indeed. I'll think more about that.

lolgesten:
Cross references in the docs between `DateTime::timestamp` and `Utc::timestamp` would be gold.

lolgesten:
uhm. strike that. can we actually go from Utc::<something> to a DateTime?

quodlibetor:
All the methods on [`TimeZone`](https://docs.rs/chrono/0.4.0/chrono/offset/trait.TimeZone.html) are implemented on `Utc`, anything that has a type signature that ends with `-> DateTime` will convert a `Utc` to a `DateTime`. Sometimes you need to go by way of dates. So: `Utc.ymd(2018, 1, 2).and_hms(1, 1, 1)` will result in a `DateTime`.

lolgesten:
Yes, I found it eventually. Thanks! In rust/chrono to convert millis UNIX epoch is:

```
    let millis = reader.get_time_stamp();
    let seconds = (millis / 1000) as i64;
    let nanos = ((millis % 1000) * 1_000_000) as u32;
[...]

PR #108, opened 2016-11-12
Improve docs around constructing DateTime objects

quodlibetor:
This provides examples for most of the constructor-like methods on
`TimeZone`, examples on the various `Offset` impls, and links
`NaiveDateTime` to `TimeZone` so that it's more obvious how you're
supposed to do things.

This is related to #88, which is something that I ran into when I
started using rust-chrono.


lifthrasiir:
I had a very minor adjustment in the mind, but changed my mind later. (My initial thought was that the documentation in `chrono::offset` is primarily intended for `TimeZone` implementers, but after some more thought I figured out that the suggested wording is less confusing for everyone.) Thank you so much!

PR #109, opened 2016-11-14
Added #[derive(Hash)] for Weekday enum

amikhalev:
This is so it can at least be used in a HashSet, if not in a BTreeSet
because it does not impl Ord.


lifthrasiir:
Thank you :)

amikhalev:
You're welcome. Thank you for the timely merge.

PR #113, opened 2016-12-07
Added impl FromStr, Serialize, Deserialize for Weekday

amikhalev:


amikhalev:
Bump? I think the failed test is a random failure

tekknolagi:
This would be great to have. I ran into this today.

lifthrasiir:
I had been busy finishing and publishing [Kailua](https://github.com/devcat-studio/kailua) for recent weeks. Sorry for delay!

I was initially going to close this PR, because Chrono is not a good place to put the localization stuff (what if I want "[星期一](https://zh.wikipedia.org/wiki/%E6%98%9F%E6%9C%9F%E4%B8%80)" for Monday or "[Freitag](https://de.wikipedia.org/wiki/Freitag)" for Friday?). However we already had some hard-coded localization stuffs in the formats, primarily because we have existing known formats mandating English. So provided that we adhere to the existing format (so making this different from the localization), I'm now going to merge this PR with a small change to make `thur` invalid, so that it can be parsed with `Fixed::LongWeekdayName` (`%A`) format.

Thank you again for the PR!

tekknolagi:
Awesome! Thanks for merging! :D

PR #116, opened 2016-12-28
Typo correction to documentation

rschulman:
I think the point here is the equality between `local_dt` and `fixed_dt`, unless I'm misunderstanding?


lifthrasiir:
The text is correct (sorry, I had exhausted energy to work on anything outside work for months). `local_dt` (09:10:11.012 at local time zone) can differ to `dt` (09:10:11.012 UTC) if your local time zone is not UTC, but `fixed_dt` (18:10:11.012+09:00) will be always identical to `dt`.

I guess Travis is running in UTC so this (wrong) change had not been caught---should I put a `TZ` override somewhere? ;)

PR #122, opened 2017-01-28
Update for serde 0.9

KodrAus:
This is a quick upgrade to `serde 0.9`. It depends on https://github.com/TyOverby/bincode/issues/92 ~~so I can't run tests yet~~.

Note: I've changed the version bound for `serde` to `^0.9` because it's not backwards compatible with previous versions.

CC: @lifthrasiir

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 rust:
   # 1.8.0 is the earliest known version that Cargo does work for the current crates.io-index repo.
   # probably older versions work, but we are forced to use this as the minimum for Cargo...
-  - 1.8.0
+  - 1.13.0
   - stable
   - beta
   - nightly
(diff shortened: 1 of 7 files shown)


alexbool:
Seems like we have to drop rust 1.8 support.

anowell:
fwiw, serde 0.9 only runs [CI for rustc 1.13](https://github.com/serde-rs/serde/blob/v0.9.0/.travis.yml#L4) and [has indicated that 0.9.x will be the final versions to support anything before 1.15](https://github.com/serde-rs/serde/issues/668#issuecomment-271234652)

KodrAus:
I'll update the build config to start at `1.13`.

lifthrasiir:
I have tested this PR against various stable versions. 1.12.0 seems to be the earliest version that this PR works, so using 1.13.0 as a baseline seems not too bad (according to @anowell's observation). Thank you (and sorry for much delayed response)!

alexbool:
A fresh release would be great! ;)

PR #124, opened 2017-02-05
Serde 0.8 support timestamps when deserializing

quodlibetor:
This is useful when using `derive(Deserialize)` on something that contains some sort of `DateTime` but the API you're interacting with returns integers instead of strings.

I've got another PR incoming that implements this for serde 0.9. I'm not sure if there are any plans to continue supporting 0.8, but I need it until the ecosystems finishes catching up to the 0.9 changes.


quodlibetor:
closing this since the 0.9 pr has been merged

PR #132, opened 2017-02-20
Corrected spelling error

ericfindlay:


lifthrasiir:
Thank you! I should mention that `README.md` is generated from `src/lib.rs`, so you also want to fix that. Feel free to put an additional commit if you want; I'll merge this in 12 hours.

PR #140, opened 2017-04-10
Update code snippets in readme to use syntax highlighting

TheZoq2:


quodlibetor:
This makes everything look better! That failure seems to be because of TyOverby/bincode@d8eac92 , it seems like the best thing to do is to import `Infinite` struct directly, since apparently a version 0.7.0 has been released of that crate with all Ty's recent changes.

lifthrasiir:
I have been so late, thank you for the patience! I've bumped the dev-dependency in the parent commit as well.

PR #145, opened 2017-05-04
date: mention that `NaiveDate` is better for serialization

### LINKED ISSUE #92: add serialization/deserialization for Date<UTC>
It'd be nice to be able to just use `Date<UTC>` with serde where only a date is given (no time).

mathstuf:
Fixes #92.


lifthrasiir:
Oh, great! Thank you for the PR!

PR #158, opened 2017-06-13
bump to 0.4.0 for correct semver

### LINKED ISSUE #156: Chrono 0.3.1 breaks backwards compatibility with Chrono 0.3.0
I am using Rocket 0.2.8, and for that, I need to use Serde 0.9 (for now). It worked great with chrono 0.3.0, but chrono 0.3.1 updates to Serde 1.0, which breaks backwards compatibility.

durka:
Fixes #156.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "chrono"
-version = "0.3.1"
+version = "0.4.0"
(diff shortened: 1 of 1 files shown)


durka:
When this is merged, 0.3.1 should be yanked from crates.io.

Issue #159, opened 2017-06-14
Proposal: Release chrono 0.4 supporting serde 1, dropping rustc-serialize and rust < 1.15


quodlibetor:
Serde is growing more and more features and rustc-serialize has been dead for a long time. Keeping rustc-serialize support causes a significant amount of work, makes tests harder to write, and I still can't figure out how to make doctests for custom-derive work in travis in #125.

I think we could clean up a bunch of code and fix the annoyance that we're causing to the community in #146 and #156 by changing serde version and dropping old support in one fell swoop, rather than having to do another breaking change at a later date.

quodlibetor:
Also bincode 1.0 does not support rustc-serialize, so if we ever want to support that we will need to drop support or go through some (probably extremely) uncomfortable contortions to maintain old and new bincodes.

quodlibetor:
Oh, I was wrong, it's _going_ to drop support for rustc-serialize, [once it's released](https://github.com/TyOverby/bincode/blob/master/changelist.org)

lifthrasiir:
I personally have no problem in keeping Rustc-serialize support (they essentially have the same code as Serde support for now, and can be left as is without breaking others). As a data point, Rust-url 1.5.0 (released last week) still continues to support it.

dtolnay:
I agree with @quodlibetor and I would recommend dropping rustc-serialize.

The `url` crate is in a totally different situation because they are preserving compatibility back to their 1.0.0 release in April 2016. If they were releasing an 0.4.0 this week, there would be no rustc-serialize in it.

If somebody needs to continue using an ancient serialization library, they can easily use an ancient version of chrono as well. Cargo makes that easy. (Well, except when 0.3.1 is a breaking change.)

quodlibetor:
@lifthrasiir Rustc-Serialize has the same code as Serde now, but when I tried to implement #125 there are a couple of meaningful differences, Serde makes it [trivial](https://github.com/chronotope/chrono/pull/125/files#diff-8daac88609c52af7718d9f334e7a2bf7R729) to support deserializing from multiple types (positive/negative integers) trivially, I _think_ it's possible to do correctly in Rustc-Serialize but it's a bit more work, based on [the generic Decodable example](https://docs.rs/rustc-serialize/0.3.24/rustc_serialize/trait.Decodable.html#examples). 

[...]

Issue #146, opened 2017-05-05
Better usage of SemVer


heartsucker:
The upgrade from 0.3.0 to 0.3.1 broke several of my projects because the dependency on `serde` 0.9 to 1.0 caused API incompatibilities. It would be nice if you guys were more mindful of this in the future.

sgrif:
Getting lots of reports about this on Diesel as well. It'd be great if 0.3.1 could be yanked and the change could be re-released as 0.4.0 instead.

lifthrasiir:
I apologize for any breakage, but this is intentional (and very exceptional---this should be the only occurrence of this kind of release). Please let me explain why.

I have received a lot of complaints when Chrono went from 0.2.25 to 0.3.0; the crates are (temporarily) split around the version of popular dependencies they depend on, and you cannot (currently) have multiple dependencies of the same crate with differing versions---if this were possible I would have immediately done this! This split-brain situation has occurred several times in the past, but Chrono 0.3.0 was particularly problematic because it coincided with Serde 0.9. I monitor the IRC for any mention of Chrono and the resulting chaos was quite annoying.

After 0.3.0 I have been convinced that the split-brain situation is hard to reconcile with the semantic versioning alone, especially because Chrono *is* also a relatively popular crate (so making a minor release makes the split-brain worse). A new release was made with this in mind: It is better to make everyone use Serde 1.0 (after a short burst of breakage) than keeping the semantic versioning, and technically Chrono does not have to strictly adhere to the semver as it's pre-1.0. I know, this is probably the one-off chance, but Serde 1.0 is the only significant dependency that Chrono supports, so it is a kind of reasonable bet.

To the end my current subjective experience suggests that the resulting breakage is no worse than the previous release. But I might be missing something---for example I can't imagine if the place where people ask about Chrono has been changed in mere 3 months, but this is technically possible. I want to hear about other crate authors' experience, especially comparing with Serde 0.9.

dtolnay:
For `url` we are planning to support Serde 1.0 in a non-breaking release while preserving compatibility with old Serde: https://github.com/servo/rust-url/pull/327. It is using [these Serde shims](https://github.com/serde-rs/legacy).

lifthrasiir:
@dtolnay Very interesting. I wasn't aware of those shims; I'll look for using them in the future release.

golddranks:
[...]

PR #160, opened 2017-06-16
bincode doesn't have serde feature

ignatenkobrain:
It is built-in.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,4 +27,4 @@ serde = { version = "1", optional = true }
 
 [dev-dependencies]
 serde_json = { version = "1" }
-bincode = { version = "0.8.0", features = ["serde"], default-features = false }
+bincode = { version = "0.8.0" }


quodlibetor:
It looks like [this is only true for bincode 1.0](https://github.com/TyOverby/bincode/blob/ff3dab845a6d2904b009455583da4e67ce3a8a2c/changelist.org) (and [the commit that removed the feature](https://github.com/TyOverby/bincode/commit/ff3dab845a6d2904b009455583da4e67ce3a8a2c#diff-90796f738364edb526dfa13994d2f801).

If you bump the bincode dependency I think we can include this in a forthcoming (hopefully: #146 #159 ) 0.4 release.

quodlibetor:
Actually, reading the git history of bincode does not match the release history at all, [there is no 0.8.0 tag](https://github.com/TyOverby/bincode/tags), even though it was released [on May 2nd](https://crates.io/crates/bincode/0.8.0).

@ignatenkobrain what caused you to notice that the feature is missing?

ignatenkobrain:
@quodlibetor so, I'm packaging crates for Fedora..

I didn't want to package alpha version of bincode, so I took latest "stable" release on crates.io (which is 0.8.0). Then when I tried to build chrono, RPM showed me missing dependency on crate(bincode/serde) and then I just unpacked archive downloaded from crates.io and found that there are no features at all...

You know, people mess up with their git repos, so.... but believe me, features do not exist on 0.8.0 ;)

dtolnay:
I filed https://github.com/TyOverby/bincode/issues/132 to tag the bincode releases. 0.8.0 is indeed the newest one.

lifthrasiir:
Indeed, the PR compiles just fine. I'll merge this. Thank you!

Issue #175, opened 2017-07-13
impossible to parse Unix timestamp with subsecond precision.


jonasbb:
I need to parse a Unix timestamp with subsecond precision, like `1441497364.649`. This should equal the format string `%s%.3f`. I am only able to parse such a string, if the subsecond part is equal to 0. Otherwise I only get an `Impossible` error, which I don't understand, because there are not inconsistencies in the timestamp.

I tested it with this program
```
fn main() {
    let fmt = "%s%.3f";

    let dt = NaiveDate::from_ymd(2015, 9, 5).and_hms_milli(23, 56, 4, 649);
    println!("{}", dt.format(fmt).to_string());
    let s = dt.format(fmt).to_string();
    let dt2 = NaiveDateTime::parse_from_str(&s, fmt);
    println!("{:?}", dt2);
    if let Err(e) = dt2 {
        println!("{}", e);
    }

    let dt = NaiveDate::from_ymd(2015, 9, 5).and_hms_milli(23, 56, 4, 0);
    println!("{}", dt.format(fmt).to_string());
    let s = dt.format(fmt).to_string();
    let dt2 = NaiveDateTime::parse_from_str(&s, fmt);
    println!("{:?}", dt2);
    if let Err(e) = dt2 {
        println!("{}", e);
    }
}
```

It prints (and the first error is unexpected):
```
1441497364.649
Err(ParseError(Impossible))
no possible date and time matching input
1441497364.000
Ok(2015-09-05T23:56:04)
```

quodlibetor:
This is surprising, but I am able to reproduce and I would expect your code to work. I have limited internet/computer time for the next few days, but I will take a look asap.

 If this is blocking you, a temporary workaround would be splitting on the dot, using %s and then adding a duration equal to the subsec part :-(

PR #176, opened 2017-07-14
Fix Allow parsing of timestamps with additional subsecond precision

jonasbb:
Fixes https://github.com/chronotope/chrono/issues/175


quodlibetor:
Nice! Could you add a test to `chrono::naive::datetime::test_datetime_parse_from_str` that verifies the behavior as well?

quodlibetor:
actually that method could use both a `%S.%f` and plain `%S` test, if you're willing.

jonasbb:
I have added additional test for each of the %f variants.

quodlibetor:
Awesome, thanks!

PR #185, opened 2017-09-04
Add doc comment to lib.rs: Conversion from and to EPOCH timestamps

kosta:
This might fix https://github.com/chronotope/chrono/issues/88

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -288,6 +288,17 @@
 //! assert!(Utc.datetime_from_str("Sat Nov 28 12:00:09 2014", "%a %b %e %T %Y").is_err());
 //! ```
 //!
+//! ### Conversion from and to EPOCH timestamps
+//!
+//! Use [`Utc.timestamp(seconds, nanoseconds)`](./offset/trait.TimeZone.html#method.timestamp) 
+//! to construct a [`DateTime<Utc>`](./struct.DateTime.html) from a UNIX timestamp 
+//! (seconds, nanoseconds that passed since January 1st 1970).
+//!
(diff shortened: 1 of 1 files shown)


kosta:
Did you get a chace to take a look? :)

quodlibetor:
Sorry I just got back from travels. This looks great, thanks!

kosta:
I noticed it's been a year and a few releases since this got merged, but it still doesn't show up in the Readme.md (it's on docs.rs, though). 

Maybe a `make readme` is missing? 

Shall I run that and send a PR?

quodlibetor:
Oh wow, I didn't realize that that was necessary. There has now been enough drift that it's not _just_ a matter of running `make README.md`, so I started adding something to CI using [cargo readme](https://crates.io/crates/cargo-readme) to guarantee that the real readme is always up to date (like [here](https://github.com/quodlibetor/diesel-derive-newtype/blob/d0de6c4827e429fd6acd4eba4fe531ca865ba17b/.travis.sh#L43-L49)) but got a little delayed. If you'd like to either fix it or add the automation that would be welcome, otherwise I'll get to it this weekend.

Thanks for pointing this out!

kosta:
I tried cargo-readme, but first it complained about a missing key in Cargo.toml, then it generated an empty readme. So it didn't work for me (but I took no time to debug).

`make readme` works well, and it seems to me Readme and lib.rs doc comments didn't diverge much, I think it should only take a few minutes to clean this up. I hope I can do this either this weekend or on Monday.

kosta:
Thank you! :)

PR #186, opened 2017-09-19
fix rfc2822 parsing to allow omitting second

upsuper:


upsuper:
ping?

quodlibetor:
Thanks! Just verified that this is in fact correct behavior according to RFC2822. (And sorry for the delay, I've been traveling and am catching up on everything.)

upsuper:
@quodlibetor would you mind releasing a new version which includes this fix?

quodlibetor:
I don't have ownership on crates.io, only @lifthrasiir can do a release. I agree that now would be a reasonable time for a 0.4.1.

quodlibetor:
For now you ought to be able to use [`[patch]`](http://doc.crates.io/manifest.html#the-patch-section) for your development/applications like so:

```toml
[patch.crates-io]
chrono = { git = "https://github.com/chronotope/chrono" }
```

Issue #178, opened 2017-08-05
formatting in RFC3339 has a different format on different platforms


Thomasdezeeuw:
On MacOS the nanoseconds are printed with 6 numbers after the decimal, while Linux and Window prints it with 9 numbers.

```
2017-08-04T13:57:08.427902+00:00 // MacOS
2017-08-04T13:57:08.427902838+00:00 // Linux, Windows
```

Any change we can change this to have a more constant experience cross platform?

Thomasdezeeuw:
I think I found the cause here: https://github.com/chronotope/chrono/blob/42a7c8e5899b1e0087294779df27c8ecd095fa8c/src/naive/time.rs#L1144-L1164. Would a change be accept that always print 6 numbers after the decimal for the nanoseconds?

quodlibetor:
I wouldn't really want to drop 3 significant digits on Linux and Windows just because MacOS only provides microsecond precision. Chrono also generally uses the number of digits of precision (0, 3, 6, 9) to signify how much precision is available, so defaulting to 9 digits seems wrong.

If you want deterministic output I'd recommend using a format string with the `%6f` format specifier. I'm open to more conversation around this, but changing the output of a method that's used by so many crates would require a pretty strong motivation, IMO.

Thomasdezeeuw:
I'm currently using the following code:

```
    const FORMAT_ITEMS: &'static [Item<'static>; 1] = &[Item::Fixed(Fixed::RFC3339); 1];
    let timestamp = chrono::Utc::now()
        .format_with_items(FORMAT_ITEMS.iter().cloned());
```

How would I change the format to deterministic in that case?

quodlibetor:
I think this does what you want:

```rust
#![allow(dead_code)]
extern crate chrono;

use chrono::format::StrftimeItems;
use chrono::format::Item;

fn format() {
    let res = StrftimeItems::new("%Y-%m-%dT%H:%M:%S%.6fZ")
        .map(|item| match item {
            Item::Error => Err(item),
            _ => Ok(item),
        })
        .collect::<Result<Vec<_>, _>>()
        .unwrap();
    println!("parsed out: {}", chrono::Utc::now().format_with_items(res.iter().cloned()));
}

fn without_parse() {
    use chrono::format::Item::*;
    use chrono::format::Numeric::*;
    use chrono::format::Pad::*;
    use chrono::format::Fixed::*;

    const ITEMS: &[Item<'static>] =  &[
[...]

Issue #157, opened 2017-06-08
Serialise UTC timezone as “Z” in .to_rfc3339()


linkmauve:
When a DateTime<FixedOffset> has the offset being UTC, serialise it into “Z” instead of “+00:00” as it is the canonical way.

quodlibetor:
[Section 4.3 of RFC 3339](https://tools.ietf.org/html/rfc3339#section-4.3) pretty explicitly states that `Z` and `+00:00` are semantically equivalent:

> ... an offset of "Z" or "+00:00", which imply that UTC is the preferred reference point for the specified time.

Sometimes I prefer `Z` for my timezones, but normally for human consumption, `+00:00` is more consistent with other offsets. The [`format`](https://docs.rs/chrono/0.3.0/chrono/datetime/struct.DateTime.html#method.format) method is the right thing to use if you're displaying for humans.

We don't seem to have something that will automatically emit either a Z or an offset, is that something that comes up for you often?

linkmauve:
It was just that I see it used way more often, and it is also shorter, I don’t really mind it either way but it seemed more sensible to use Z than +00:00 or -00:00.
I just had to change my testcases in xmpp-parsers to use +00:00 compared to the specification examples I was importing, but that’s not much of a justification: https://hg.linkmauve.fr/xmpp-parsers/file/tip/src/delay.rs#l105

quodlibetor:
Fair enough, I also see Z used more often.

I'm going to close this, but I would accept a pr that implements a `.to_rfc3339z()` method, since I agree it's a nice shorthand and, since the current behavior is perhaps a bit surprising I expect a different method would highlight the output format of the existing method.

quodlibetor:
Oh, nvm I don't have rights to close tickets. @lifthrasiir how do you feel about my proposed resolution, above?

quodlibetor:
aaaannnnd now I understand how github works.

Issue #171, opened 2017-07-07
Give `to_rfc3339` a param: number of fractional second digits


sanmai-NL:
https://docs.rs/chrono/0.4.0/chrono/struct.DateTime.html#method.to_rfc3339 has no parameters, but with some applications too long a string of digits isn't expected (e.g., [on the web](https://www.feedvalidator.org/docs/error/InvalidRFC3339Date.html)), however, `to_rfc3339` currently emits such long strings (e.g. `959686092`).

https://www.ietf.org/rfc/rfc3339.txt seems to prescribe one digit only for the fractional seconds, so perhaps the current implementation is in error? If for practical reasons the number of fractional second digits should be tunable irrespective of the official grammar, then I propose that the range be between 0 (in which case the decimal separator isn't emitted) and, say, 16 (to be determined).

quodlibetor:
I think that you're reading the ABNF specified in RFC 3339 incorrectly, [it says that the internet date/time format is](https://tools.ietf.org/html/rfc3339#section-5.6):

```
...
   time-secfrac    = "." 1*DIGIT
...
   partial-time    = time-hour ":" time-minute ":" time-second
                     [time-secfrac]
...
   full-time       = partial-time time-offset
...
   date-time       = full-date "T" full-time
```

That `1*DIGIT` _looks_ like it's saying that there should be one digit, but it's actually a [variable repetition rule](https://tools.ietf.org/html/rfc2234#section-3.6):

```
   The operator "*" preceding an element indicates repetition. The full
   form is:

        <a>*<b>element

   where <a> and <b> are optional decimal values, indicating at least
   <a> and at most <b> occurrences of element.
```

So the current implementation is definitely correct, and in fact the feed validator [link you provided](https://www.feedvalidator.org/docs/error/InvalidRFC3339Date.html) includes an example with multiple digits in the fraction of the datetime `2002-10-02T15:00:00.05Z`.

This is, though, I think related to #165, in that we'd like to be able to have multiple ways of displaying the fractional portion of a time.

sanmai-NL:
Well I hoped to convery I wasn't so sure about the grammar at all, only took a quick look.

Some background into the jungle of date time notations ...
> Several standards and profiles have been derived from ISO 8601, including RFC 3339 and a W3C note on date and time formats.

(Source: https://www.cl.cam.ac.uk/~mgk25/iso-time.html#software)

[...]

Issue #165, opened 2017-06-25
X.690/ASN.1 compatible subsecond formatting


rubdos:
Hello there,

I'm currently working on a ASN.1 DER serializer. This standard has support for sub-second representation, but they (foolishly) drop significant zeros in their representation.  
I currently use

```Rust
let mut time = format!("{}", time.format("%Y%m%d%H%M%S%f"));
let (prefix, fraction) = time.split_at_mut(14);       
let time = match fraction.as_bytes().iter().enumerate().rfind(|&(_, &x)| x != b'0') {
    Some((offset, _)) => prefix.to_owned() + "." + &fraction[0..offset+1] + "Z",
    None => prefix.to_owned() + "Z",                  
};                 
```

to generate my output (which is both ugly and inefficient). What it basically does is it prints out the `"%f"` (nanoseconds) and truncates its zeroes. If there's nothing left, the `"."` is also omitted.

**TL;DR**: So: I basically wonder whether it'd be possible for a (preferably more efficient) `"%S"` like formatter that outputs the seconds, and omits zeroes and the `"."` when "possible".

quodlibetor:
Can't you do something like:

```
if time.nanosecond() == 0 {
   // fmt without frac
} else {
   // fmt with frac
}
```
Unless I'm missing something that's what this hypothetical format specifier would do anyway, and increasing the size of the format mini-language for this use case seems like overkill.

rubdos:
> increasing the size of the format mini-language for this use case seems like overkill.

roger that!

It's not only that though, it's also trimming the trailing zeros of the nanosecond. It's stupid, but the standard requires it.

quodlibetor:
Does the if-statement I suggested not fix your performance concerns?

Also, actually, the `%.f` format (note the `.`) should already do what you want. I just checked and this passes for me:

    assert_eq!(FixedOffset::east(34200).ymd(2001, 7, 8).and_hms_nano(0, 34, 59, 0)
               .format("%.f").to_string(),
               "");

Otherwise I don't actually understand what you want.

quodlibetor:
Oh, wait, you want something that can result in `.7` instead of currently you'd end up with `.700`? So, print the smallest amount possible?

If so I take it back that sounds pretty reasonable. I think maybe some sort of `%.-f` format would work?

lifthrasiir:
[...]

PR #205, opened 2018-01-12
Add DateTime to_rfc3339p(z) methods, tests

dekellum:
These additions allow convenient control of RFC 3339 formatted output:

 * Number of subsecond digits to display

 * Whether to use the 'Z' variant, instead of "+00:00" for TZ offset
   0, UTC.

...while remaining faithful to the RFC 3339. The implementation uses the existing formatting Item mechanism.

The 'Z' variant was originally requested in #157 where it was suggested a PR was welcome.   Control of subsecond digits was requested in #178 and @quodlibetor gave me a starting point for using the formatting Item mechanism.

Tests and rustdoc are included. Please consider merging this or let me know if it has any deficits, thanks!

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -287,7 +287,7 @@ impl<Tz: TimeZone> DateTime<Tz> where Tz::Offset: fmt::Display {
         use format::Fixed::*;
         use format::Fixed::Nanosecond;
 
-        static PREFIX: &[Item<'static>] = &[
+        const PREFIX: &'static [Item<'static>] = &[
             Numeric(Year, Zero),
             Literal("-"),
             Numeric(Month, Zero),
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -309,10 +309,13 @@ impl<Tz: TimeZone> DateTime<Tz> where Tz::Offset: fmt::Display {
             }
         };
 
-        let tzitem = Item::Fixed(match use_z {
-            true  => Fixed::TimezoneOffsetColonZ,
-            false => Fixed::TimezoneOffsetColon
-        });
+        let tzitem = Item::Fixed(
+            if use_z {
+                Fixed::TimezoneOffsetColonZ
+            } else {
+                Fixed::TimezoneOffsetColon
+            }
+        );
 
         match ssitem {
             None =>
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -246,6 +246,23 @@ impl<Tz: TimeZone> DateTime<Tz> where Tz::Offset: fmt::Display {
         self.format_with_items(ITEMS.iter().cloned()).to_string()
     }
 
+    /// Return an RFC 3339 and ISO 8601 date and time with subseconds formatted
+    /// as per a Fixed variant or None for no subseconds.  The Fixed
+    /// variants supported are: Nanosecond, Nanosecond3, Nanosecond6
+    /// and Nanosecond9. Other values will panic!
+    pub fn to_rfc3339p(&self, subform: Option<Fixed>) -> String {
+        self.rfc3339_via_items(subform, false)
+    }
+
+    /// Return an RFC 3339 and ISO 8601 date and time with subseconds formatted
+    /// as per a Fixed variant or None for no subseconds.  The Fixed
+    /// variants supported are: Nanosecond, Nanosecond3, Nanosecond6
+    /// and Nanosecond9. Other values will panic!
+    /// If the timezone is UTC (offset 0), use 'Z'.
+    pub fn to_rfc3339pz(&self, subform: Option<Fixed>) -> String {
+        self.rfc3339_via_items(subform, true)
+    }
+
     /// Formats the combined date and time with the specified formatting items.
     #[inline]
     pub fn format_with_items<'a, I>(&self, items: I) -> DelayedFormat<I>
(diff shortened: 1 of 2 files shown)


quodlibetor:
Awesome, thanks! I'll review this over the weekend, I want to do a survey of our existing API to make sure that we are working towards something consistent.

dekellum:
Thanks. While Travis CI is delayed waiting for osx builds, there are two failures so far:

 1. I used a static for the PREFIX rfc3339 Item array, but this has only been supported since rustc 1.17 (rust-lang/rust#40027).  Shall I change this to const?  Alternately could make 1.17 the minimum version, but then there are probably other cases to make static or other bigger changes associated with that.

2. The build with clippy is failing on existing code, not these changes, but its somewhat related to the above, e.g: 

~~~
error: Constants have by default a `'static` lifetime
   --> src/datetime.rs:463:23
         const ITEMS: &'static [Item<'static>] = &[
                       ^^^^^^^ help: consider removing `'static`
~~~

I assume that lint is new since the last successful build a few months ago.  I'm willing to try and fix that as well, depending on how to solve (1), but that might be better in its own PR?

quodlibetor:
We want to preserve compat back to rust 1.13, so we should disable that clippy lint. Feel free to do it in this PR but in a separate commit, thanks! By the same logic, please change it to const.

I've enabled weekly builds so that we'll get alerted if clippy changes cause things to fail and future contributors won't have to deal with this.

dekellum:
@quodlibetor: Tests are passing. Did you have a chance to survey and consider this regarding API consistency?  

In case it helps: I'm not attached to the introduced -`p(z)` suffix naming.  The 'z' was based on comments in #157.  I'm not sure if there is a better Rust style available here vs this sort of libc naming style.  

Also this currently uses the Fixed enum for the sub-second formatting parameter.   That's arguable consistent, but at the expense of a potential panic to enforce RFC 3339 constraints at runtime. An alternative would be to introduce a new enum for just the sub-second format supported.  

Please let me know your preferences or other concerns.

Perhaps I didn't make a strong enough case for this addition in the first place: I believe the 'Z' and fixed sub-second format variants of RFC-3339 are too commonly needed to not include in this crate, particularly when considering the amount of code involved (see #178 example) to correctly implement it externally.

quodlibetor:
Two things:

1. I definitely want to get this merged, there's little debate in my mind that we need something like this
2. My concern was with the `Option<Fixed>` enum, I don't believe that there are any other examples of us taking such a complex parameter to something
3. Sorry, my day job is day, night and weekend job for the last month or so which is why this is languishing

I _think_ that taking a new enum (especially one that statically prevents errors) will be more ergonomic, despite the extra import.

My larger concern is with formatting in general -- the fact that this is complex to handle this in an external crate is a bug which should be fixed, but I _think_ that we basically just need to create a new formatting API to manage it.

If you want to help me out, the survey that I wanted to do was mostly just to look at our various other formatting APIs and see:

1. Do any methods take enums? If so, how complex are they?
2. Do the [Rust API Guidelines](https://rust-lang-nursery.github.io/api-guidelines/) have anything to say about API's like this?
3. Should we be documenting these more complex formatting methods in terms of the underlying format method that takes a slice of `Format::Item` variants so that external users are less dependent on us to handle this for them?

dekellum:
In the latest I've introduced a SecondsFormat enum.  This simplifies the function enough that I also went with a single method, named `to_rfc3339_opts` (which I think is a more rust-style name) and use a single bool for use of the `Z`-variant.  The new enum is included in datetime.rs and pub export at `chrono::` top level.  I didn't put this in `chrono::format` because its role is now more as a façade over the format interfaces.

In my evolving understanding of the formatting API, I share your larger concern.  Error handling appears problematic with parsing formats from strings.  Working with the format Items directly is complicated by enum/variant name collisions and the need for cloned value iteration?  It would be nice if there was some easy way to create a static or constant _Format(-ter)_ ,  for this application from a create_rfc_3339_format(args) kind of function, and then simple use that repeatedly over many `DateTime` values.

But this PR makes chrono more usable now and in the interim.  When there is a a better format module, `DateTime::to_rfc_3339` and `to_rfc_3339_opt` can be reimplemented in terms of that, and possibly also deprecated.

src/datetime.rs
```diff
@@ -1059,6 +1133,28 @@ mod tests {
                    Ok(EDT.ymd(2015, 2, 18).and_hms_micro(23, 59, 59, 1_234_567)));
     }
 
+    #[test]
+    fn test_rfc3339_opts() {
```
quodlibetor: Could you add a couple of these asserts to the docs for `to_rfc3339_opts`? Maybe just all 4 (pst/utc,true/false) of the `Millis` asserts?

src/datetime.rs
```diff
@@ -246,6 +270,56 @@ impl<Tz: TimeZone> DateTime<Tz> where Tz::Offset: fmt::Display {
...
+
+        let tzitem = Item::Fixed(
+            if use_z {
+                Fixed::TimezoneOffsetColonZ
+            } else {
+                Fixed::TimezoneOffsetColon
+            }
+        );
+
+        match ssitem {
```
quodlibetor: this match could probably be something like:

`PREFIX.iter().chain(s.iter()).flat_map(|s| s).chain([tzitem].iter())` I'm not sure off the top of my head what the correct nesting of flat_map with option would be (maybe just sticking the option in the map chain would do it?) but if you're willing to give it a shot I think it'd be cleaner.

src/datetime.rs
```diff
@@ -16,6 +16,30 @@ use Date;
 use format::{Item, Numeric, Pad, Fixed};
 use format::{parse, Parsed, ParseError, ParseResult, DelayedFormat, StrftimeItems};
 
+/// Specific formatting options for seconds
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum SecondsFormat {
```
quodlibetor: Actually: we should probably name this `FractionFormat` and also add a [`#[doc(hidden)] __Nonexhaustive` variant](https://github.com/diesel-rs/diesel/blob/5435e30f1dc28008a8bef34c889998a0b009f0c4/diesel/src/result.rs#L10) so that we can expand this enum in the future, without it being a breaking change?

dekellum: Naming: I went through some pre-PR iteration on this. Calling it SubSecFormat or FractionFormat makes one want to use something like `Omit` for the `Secs` variant and then it doesn't fit together as well.   Its really intended to be one parameter that specifies the seconds format including the `seconds ['.' subseconds]`.  Added more doc to try and clarify this.

If I add `__Nonexhaustive` in this, then I need to add a `panic!()` in `to_rfc3339_opts` to catch the case where it doesn't cover a new variant.   The `std` `ErrorKind` example of this being used is for a case where the enum in question is being returned and consumed by external code.  I don't think there is a valid use case for external code matching on `SecondsFormat` but I added a warning in rustdoc not to do this. I hope that is sufficient.

src/datetime.rs
dekellum: I tried this, but I couldn't get it to compile.  I'm glad we agree that the format API is hard to use though. :)  I also agree the current impl looks redundant, but in my own attempts to refactor, I ran into various forms of trouble with the barrow-checker.  _Non-Lexical Lifetimes_ might make that possible in the future?

src/datetime.rs
dekellum: Yes, included in latest.

dekellum:
Also regarding your comment on future expansion and #165 (as well as linked #171), I renamed the one variant to `AutoSi` and improved rustdoc to indicate what it does.  Should make it easy to add additional variants in the future for other number of digits, e.g. `Centis`-econds or `NoPad`.

quodlibetor:
Thank you! Sorry for the slow turnaround, but this is great!

dekellum:
Thanks!  Was just looking at the accumulated changes since 0.4. At some point, would be good to know if there as anything else that is particularly release gating?

Issue #208, opened 2018-01-16
Implement Iterator API extensions for NaiveDate days iterator


gnzlbg:
NaiveDate days iterator can implement FusedIterator, TrustedLen, and Step. These should be implemented as they become stable.

ImmemorConsultrixContrarie:
`std::iter::FusedIterator` was stabilised in 1.26

PR #209, opened 2018-01-16
Add a days iterator for NaiveDate

### LINKED ISSUE #152: implement Iterator
Hi,
I'm new at Rust so I might have missed something here.
I have two NaiveDate and I'd like to write:
```
for d in start..end {
            
        }
```

but it needs the Iterator Trait.

gnzlbg:
This PR implements an `iter_days()` method for `NaiveDate` that returns an iterator from the current date until `MAX_DATE` with a step-size of `1` day. It closes #152 .

The `Iterator` is an `ExactSizeIterator` because the distance in days from any date until `MAX_DATE` can be computed in `O(1)`. 

Since `Step`, `TrustedLen`, and `FusedIterator` are unstable, I have filled issue #208 to track implementing these as they become stable. The iterator can support them.

To iterate over shorter date ranges users can use `.take`, `.take_while`, ... on this iterator. 

Once we implement `Step` for it, users will be able to iterate over weeks by using a step-size of `7` days. 

Iterating over months is hard, and this PR explicitly does not address this issue. I've argued in #152 why I think that we shouldn't implement any iterator over months in the library. But if someone wants to discuss that, I've implemented some and could contribute to that discussion (but please open an issue just for that and ping me).

### DIFF
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -958,6 +958,32 @@ impl NaiveDate {
     pub fn format<'a>(&self, fmt: &'a str) -> DelayedFormat<StrftimeItems<'a>> {
         self.format_with_items(StrftimeItems::new(fmt))
     }
+
+    /// Returns an iterator that steps by days until the last representable date.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use chrono::NaiveDate;
+    ///
+    /// let expected = [
+    ///     NaiveDate::from_ymd(2016, 2, 27),
+    ///     NaiveDate::from_ymd(2016, 2, 28),
+    ///     NaiveDate::from_ymd(2016, 2, 29),
+    ///     NaiveDate::from_ymd(2016, 3, 1),
+    /// ];
+    ///
+    /// let mut count = 0;
+    /// for (idx, d) in NaiveDate::from_ymd(2016, 2, 27).iter_days().take(4).enumerate() {
+    ///    assert_eq!(d, expected[idx]);
+    ///    count += 1;
+    /// }
+    /// assert_eq!(count, 4);
+    /// ```
+    #[inline]
+    pub fn iter_days(&self) -> NaiveDateDaysIterator {
+        NaiveDateDaysIterator { value: *self }
+    }
 }
 
 impl Datelike for NaiveDate {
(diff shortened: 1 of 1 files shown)


src/naive/date.rs
```diff
@@ -958,6 +958,32 @@ impl NaiveDate {
...
+    ///
+    /// let mut count = 0;
+    /// for (idx, d) in NaiveDate::from_ymd(2016, 2, 27).iter_days().take(4).enumerate() {
+    ///    assert_eq!(d, expected[idx]);
+    ///    count += 1;
+    /// }
+    /// assert_eq!(count, 4);
+    /// ```
+    #[inline]
+    pub fn iter_days(&self) -> NaiveDateDaysIterator {
```
gnzlbg: I called the method `iter_days` but I don't know whether this is a good name. I don't actually like it and would prefer something like just `.days()` or `.days_until_max_date()` or similar (`.until_end_of_days()`? :D ).

gnzlbg:
@quodlibetor ?

gnzlbg:
@kennytm ?

quodlibetor:
@gnzlbg Thanks for this! The iterator is definitely desired, I haven't responded to this because I've been trying to clear up what the chrono API *should* look like, which has pushed me away from giving this a thorough review.

I do really want this, though, sorry for the delay.

kennytm:
@gnzlbg  Why am I involved? 😅

gnzlbg:
@quodlibetor I personally like C++'s approach to date and time quite a bit. It has three libraries:

 * [`chrono`](http://en.cppreference.com/w/cpp/chrono): clocks, time points, durations, and time units (using compile-time rational numbers)
 * [`date`](https://github.com/HowardHinnant/date/blob/master/include/date/date.h): very fast time-zone unaware date algorithms
 * [`tz`](https://github.com/HowardHinnant/date/blob/master/include/date/tz.h): time-zone support

and each library builds on top of the previous one (date builds on chrono, tz builds on date). There were some design mistakes that could be fixed in Rust:

* making minutes use a 64-bit integer instead of a 32-bit integer
* exposing only a system clock and a monotonic clock (no high-resolution clock) 

Some literature:

* lots of blog post explaining date and tz's design: https://howardhinnant.github.io/
* chrono papers:
   * http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2008/n2661.htm
   * http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2008/n2498.html
   * http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/p0092r1.html
* current date and tz proposal: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0355r5.html

quodlibetor:
This was merged in #457

PR #211, opened 2018-01-21
Fix a markdown rendering difference warning.

kennytm:
Rust is going to change its Markdown rendering engine from hoedown to pulldown-cmark. In pulldown, a table row starting with just whitespaces will cause that whole cell disappeared. This causes rendering difference between the two engines.

To fix this, we add leading and trailing `|` to the rows so that empty cells are correctly rendered.

cc rust-lang/rust#47501.

### DIFF
--- a/src/format/strftime.rs
+++ b/src/format/strftime.rs
@@ -8,77 +8,77 @@
 
 The following specifiers are available both to formatting and parsing.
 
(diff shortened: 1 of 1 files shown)


quodlibetor:
Thanks!

Issue #212, opened 2018-01-24
README incorrectly outputting format


lancecarlson:
Currently the README says format should output a string with the seconds without decimal values:

```rust
assert_eq!(format!("{:?}", dt), "2014-11-28T12:00:09Z");
```

However, when I print out the same thing I get:

```
2018-01-23T15:52:12.922267839Z
```

Is this intentional? Am I missing something?

lancecarlson:
Perhaps this is a serde default and out of scope of the chrono package?

quodlibetor:
nope our README _should_ be correct. I would like to migrate to [cargo-readme](https://github.com/livioribeiro/cargo-readme) so that we don't duplicate top-level docs, but I don't recall how much of a difference our README/lib docs diverge. I'm not even sure off the top of my head if they're _similar_.

If you want to fix it to match the actual output that would be cool.

pitdicker:
Code examples have been removed from the README in https://github.com/chronotope/chrono/pull/787.

PR #213, opened 2018-01-24
Update README.md

lancecarlson:
Include the decimal values in the default output format

Mentioned here: https://github.com/chronotope/chrono/issues/212

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,7 +235,7 @@ assert_eq!(dt.format("%a %b %e %T %Y").to_string(), dt.format("%c").to_string())
 assert_eq!(dt.to_string(), "2014-11-28 12:00:09 UTC");
 assert_eq!(dt.to_rfc2822(), "Fri, 28 Nov 2014 12:00:09 +0000");
 assert_eq!(dt.to_rfc3339(), "2014-11-28T12:00:09+00:00");
-assert_eq!(format!("{:?}", dt), "2014-11-28T12:00:09Z");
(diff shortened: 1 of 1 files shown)


quodlibetor:
hmmm I can't actually reproduce that formatting, even though I asked you to change it: https://play.rust-lang.org/?gist=113202e1c9e9790000351650f2fc7ff0&version=stable

lancecarlson:
Ah so, I did some experimenting, and I figured out that it prints out the additional formatting when using now. Try this:

```rust
let dt2 = Utc::now();
assert_eq!(format!("{:?}", dt2), "");
```

lancecarlson:
This came up recent btw when I was trying to make a datetime format to a urlencoded version that an API endpoint was expecting. The endpoint gave me back a 500 error because it wasn't expecting nanosecond values. It kind of threw me off that it would format with the nanoseconds on a Utc.now() given the readme was formatting it without.

quodlibetor:
I would take a PR that updates the README with another code block that demonstrates the `.now()` debug display, because I agree that is confusing and probably wrong.

lancecarlson:
@quodlibetor Added a couple of lines with more detail. This acceptable or would you like something more?

lancecarlson:
@quodlibetor are my latest changes satisfactory?

quodlibetor:
Yup! This is kind of a sad part of the API, and we should document more clearly why it happens, but it's good to have the current behavior documented at least. Thanks!

Issue #201, opened 2018-01-10
Replace weekday-number methods with a single days_since method


clarfonthey:
Essentially:

```
number_from_monday() => days_since(Monday) + 1
number_from_sunday() => days_since(Sunday) + 1
num_days_from_monday() => days_since(Monday)
num_days_from_sunday() => days_since(Sunday)
```

If we want to replace `number_from` with something I'd probably go with `days_since_plus_one` or something more explicit.

PR #216, opened 2018-01-29
Add Weekday::days_since.

clarcharr:
This adds the method described in #201.


src/lib.rs
```diff
@@ -539,6 +539,25 @@ impl Weekday {
...
+    /// The number of days since the given day.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use chrono::Weekday::*;
+    /// assert_eq!(Mon.days_since(Mon), 0);
+    /// assert_eq!(Sun.days_since(Tue), 5);
+    /// ```
+    pub fn days_since(&self, other: Weekday) -> u32 {
```
pickfire: Should this be inlined?

Issue #195, opened 2017-12-05
Chrono stability?


coder543:
Looking at [the contributions](https://github.com/chronotope/chrono/graphs/contributors), things seem to have been pretty quiet lately. [This comment thread about Diesel 1.0](https://www.reddit.com/r/rust/comments/7hkw6c/announcing_diesel_100_beta_1/dqs46df/?context=10) brings up some concern about Chrono's lack of confidence in its own stability.

> Chrono is probably the most concerning here, TBH I'm not sure why they aren't on the path to 1.0.

and

> So I'm not sure what to do about this problem. Rust clearly has a 1.0-phobia issue, but Diesel is ready to stabilize it's API, and doing so will be huge for Diesel and for Rust.

and

> I guess when Chrono releases a breaking change we'll have to deal with it. :/

are relevant quotes.

Is there any chance that Chrono is ready to update to v1.0? Or is the API still so unstable?

quodlibetor:
I [responded in that comment thread](https://www.reddit.com/r/rust/comments/7hkw6c/announcing_diesel_100_beta_1/dqsa8ug/).

tl;dr is that, yes, there are known breaking changes desired. I've been sick since that comment thread happened, but I've been thinking and writing notes on *my* known blockers to 1.0. @lifthrasiir probably has others.

Some obvious ones:

* Trait revamp -- the existing trait collection doesn't really make sense, a
  bunch of things are inherent methods on a bunch of different concrete types
  instead of being in a trait, the specific way associated types are used makes
  generic programming unecessarily difficult (that's why I haven't implemented
  `From<Timelike> for Utc`, for example), etc.

  This could be backwards-compatible, but I really want to deprecate/delete the
  existing footgun methods.
* Duration is only positive, making implementing `Add`/`Sub` and various
  iterators extremely unergonomic.
* Only one calendar system is supported (although I have no idea how multiple
  calendar systems would be implemented, maybe this would be the domain of a
  different library or would be backwards compatible)

So there are a grab-bag of API changes, new types, and possibly truly
fundamental design decisions that we aren't sure about.

Chrono hasn't been getting a lot of commits because:

* there aren't lots of bugs
* there aren't feature requests, except those ones which require API revamps I
  mentioned above.
* Neither lifthrasiir nor myself have lots of time to idly tool on chrono --
  it's basically good enough for everyone's purposes, while still, sadly,
  containing many deficiencies that we'd like to fix before releasing a 1.0

[...]

PR #221, opened 2018-02-23
Use no_std-compatible num dependencies

cuviper:
Rather than the `num` meta-crate, use `num-integer` and `num-traits`
without default features to make them `#[no_std]`.  `num-iter` is
just a dev-dependency now for a few test cases.

The only public change is the `impl FromPrimitive for Weekday`, but this
is still the same exact trait that `num` re-exports, so this is not a
breaking change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,8 @@ name = "chrono"
 
 [dependencies]
 time = "^0.1.36"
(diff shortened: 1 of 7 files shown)


cuviper:
I saw that #195 was contemplating no-std support.

PR #236, opened 2018-04-02
Optionally remove time crate dependency

jethrogb:
Continuation of #137. I had to keep the old branch around for... reasons.

Original PR:

---
I would like to use the chrono crate to process time information on systems that don't have a clock. The time crate, which is used by chrono to obtain and represent the local time, can't be made to work on such a system. This PR adds a default-enabled "local" feature to the crate and moves all code that is dependent on the time crate behind a cfg-gate. Current users of chrono need to do nothing. Users of chrono that would like to use it without the time crate can set `default-features = false` in their Cargo.toml.

**Open questions**:
* *Test coverage* I have fixed all unit tests such that they will pass with or without the local feature. There are however 31 doc tests that use some code that has been cfged-out. I'm not sure how to selectively disable them.
* *UTC* `UTC::now` and `UTC::today` also depend on the time crate. I have put them behind the same cfg-gate, but that might be a misnomer. We could leave this as is, or I could add another feature with a different name or rename the current feature.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -33,8 +33,10 @@ build_and_test() {
   TZ=ACST-9:30 channel test -v --lib
   channel build -v --features rustc-serialize
   TZ=EST4 channel test -v --features rustc-serialize --lib
-  channel build -v --features 'serde bincode'
-  TZ=UTC0 channel test -v --features 'serde bincode'
+  channel build -v --features serde
+  TZ=UTC0 channel test -v --features serde --lib
+  channel build -v --features serde,rustc-serialize
+  TZ=Asia/Katmandu channel test -v --features serde,rustc-serialize
 }
 
 build_only() {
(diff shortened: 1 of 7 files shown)


jethrogb:
> Ah, I see we're not running with cargo test --no-default-features at all. I agree that that is unfortunate. There's no way to just not run the doctests either, afaict. I would strongly prefer to test with --no-default-features, I'll do some more research.

@quodlibetor What do you mean? I'm changed the way CI is run in https://github.com/chronotope/chrono/pull/236/commits/5e68c60bb2c5cde7e9432d10b0bebb211b712b24#diff-b4553af39acae935d505c76e139d4aa3

quodlibetor:
Oh I'm just confused. Yeah what you've currently got seems like as ideal as we're going to get.

quodlibetor:
I'm really busy this week, but will release chrono 0.4.2 with this feature by the weekend.

Thanks so much for your patience!

quodlibetor:
This is now in crates.io as chrono 0.4.2.

Thanks again for your patience!

PR #237, opened 2018-04-03
impl Sub<T> for T for Date/Time types

### LINKED ISSUE #168: Add support for subtracting DateTime<Tz>
At the moment the [DateTime documentation](https://docs.rs/chrono/0.4.0/chrono/struct.DateTime.html) shows me that there is an impl to subtract `DateTime<FixedOffset>` instances from one another to obtain a `Duration` instance that describes the time delta.

What doesn't seem to be possible is to do something like this:
```` rust
extern crate chrono;
use chrono::prelude::*;

fn foo() {
    let start: DateTime<Local> = Local::now();
    let end: DateTime<Local> = Local::now();
    let dur = end - start; // Error: `std::ops::Sub` not implemented for `chrono::DateTime<chrono::Local>`
}
````

What I would like to be able to do is such a subtraction in general for `DateTime<Tz>`, for any `Tz` type param. Ideally it would even be possible to subtract e.g. a `DateTime<Local>` from a `DateTime<Utc>` and have the time zone conversion done internally in the `std::ops::Sub`.

Could this be implemented, or is there some kind of restriction that prevents this?

**UPDATE**: I just found the `dt0.signed_duration_since(dt1)` method, which does what I need.  However, if the possibility exists to implement the `Sub` trait, I'd still like that to happen as it would be a more intuitive way to access that functionality.

tobz1000:
Fixes #168.

I've added simple wrappers around the `.signed_duration_since` methods for `Date`, `DateTime`, `NaiveDate`, `NaiveTime`, and `NaiveDateTime`.

I've added docs to match what's already there - just altered copies of the `.signed_duration_since` docs. I'm not sure if it would be preferable to move the bulk of the detail in those docs to one of the two places, with a "for more info..." message in the other, for maintainability's sake.

The `Date`/`DateTime` implementations currently support all timezones (as long as they match between operands); although @quodlibetor mentions maybe it should only be implemented for Utc? Let me know if this should be changed. (I'm personally a fan of allowing for all time zones as long as they match.)

### DIFF
--- a/src/date.rs
+++ b/src/date.rs
@@ -360,6 +360,15 @@ impl<Tz: TimeZone> Sub<OldDuration> for Date<Tz> {
     }
 }
 
+impl<Tz: TimeZone> Sub<Date<Tz>> for Date<Tz> {
+    type Output = OldDuration;
+
+    #[inline]
+    fn sub(self, rhs: Date<Tz>) -> OldDuration {
+        self.signed_duration_since(rhs)
+    }
+}
+
 impl<Tz: TimeZone> fmt::Debug for Date<Tz> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "{:?}{:?}", self.naive_local(), self.offset)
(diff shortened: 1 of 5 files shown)


quodlibetor:
This seems great! And, yeah, as long as we're not doing _conversions_ between timezones during subtraction I don't see any reason to block subtraction based on it.

PR #242, opened 2018-04-25
Add %#z as "Permissive" timezone parsing

### LINKED ISSUE #219: supported timezone syntax for DateTime from string
I'm trying to accommodate the default timezone syntax returned from postgresql.  It seems to provide an abbreviated form that chrono may not support-- it omits 00 minutes.  Is there support for this in Chrono?  

http://play.integer32.com/?gist=b6394d764deb6a4bcf7b1b5ebd038969&version=stable

```
extern crate chrono;
use chrono::prelude::*;
use chrono::DateTime;

fn main() {
    let dt = "2018-02-06 11:17:09.534889-05";
    let test = DateTime::parse_from_str(dt, "%Y-%m-%d %H:%M:%S%.6f-%z");
    println!("{:?}", test);                                        
}

```

quodlibetor:
This allows you to parse a timezone that:

* Is either `Z` or an actual offset
* Contains no minutes, just the hour

Fixes #219 

I'm not sure if this is ready for merge, conceptually the `chrono::format::Fixed::Internal` variant is meant to ensure that we can add new variants, but it doesn't do it via the now-standard `#[doc(hidden)] __internal` variant.

Probably I should move the new variant into the hidden internal enum, but I'm not sure that it's worth it, because then users won't be able to statically construct parse lists. I'm open to opinions here.

### DIFF
--- a/src/format/mod.rs
+++ b/src/format/mod.rs
@@ -200,6 +200,15 @@ pub enum Fixed {
     /// Same to [`TimezoneOffsetColonZ`](#variant.TimezoneOffsetColonZ) but prints no colon.
     /// Parsing allows an optional colon.
     TimezoneOffsetZ,
+    /// Same as [`TimezoneOffsetColonZ`](#variant.TimezoneOffsetColonZ), but
+    /// allows missing minutes (per [ISO 8601][iso8601]).
+    ///
+    /// # Panics
+    ///
+    /// If you try to use this for printing.
+    ///
+    /// [iso8601]: https://en.wikipedia.org/wiki/ISO_8601#Time_offsets_from_UTC
+    TimezoneOffsetPermissive,
     /// RFC 2822 date and time syntax. Commonly used for email and MIME date and time.
     RFC2822,
     /// RFC 3339 & ISO 8601 date and time syntax.
@@ -491,6 +500,8 @@ pub fn format<'a, I>(w: &mut fmt::Formatter, date: Option<&NaiveDate>, time: Opt
                         off.map(|&(_, off)| write_local_minus_utc(w, off, false, false)),
                     TimezoneOffsetZ =>
                         off.map(|&(_, off)| write_local_minus_utc(w, off, true, false)),
+                    TimezoneOffsetPermissive =>
+                        panic!("Do not try to write %#z it is undefined"),
                     RFC2822 => // same to `%a, %e %b %Y %H:%M:%S %z`
                         if let (Some(d), Some(t), Some(&(_, off))) = (date, time, off) {
                             let sec = t.second() + t.nanosecond() / 1_000_000_000;
@@ -609,4 +620,3 @@ impl FromStr for Weekday {
         }
     }
 }
-
--- a/src/format/mod.rs
+++ b/src/format/mod.rs
@@ -200,15 +200,6 @@ pub enum Fixed {
     /// Same to [`TimezoneOffsetColonZ`](#variant.TimezoneOffsetColonZ) but prints no colon.
     /// Parsing allows an optional colon.
     TimezoneOffsetZ,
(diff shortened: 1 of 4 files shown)


Dowwie:
This request began when I was trying to convert a timestamp provided by a recent version of Postgres to a native Chrono type.  As my database (for testing) is in the EST time zone, I found that my timestamps omitted tz minutes.  However, I suspect that Postgres *may* include minutes, depending on TZ.  With this given, I recommend a single Chrono format that supports optional tz minutes -- in other words, a format that supports +RR:OO   where R: Required , O: Optional

gnzlbg:
@Dowwie 

>  recommend a single Chrono format that supports optional tz minutes -

This is exactly what this commit should do.

quodlibetor:
@Enet4 yeah the only issue with breaking changes is that breaking chrono breaks a *lot* of code. It's not as bad as libc, but it's still pretty annoying for the entire ecosystem. So ideally we'll have an absolute max of one per year, which means I'd rather give warning and bend over backwards to preserve backcompat.

I think that means that I should add this as a variant of that internal enum, and document that the various enums will become expandable in the next breaking change.

I agree with you about the fact that we use the same items for parsing/writing. I've mentioned before that I don't like the error messages generated by our parse code, but this is another aspect of it that needs to be cleaned up.

Good catch with the docs!

quodlibetor:
@Dowwie gnzlbg is correct, you can try this branch out in your project with the following cargo config:

```toml
[dependencies]
chrono = { git = "https://github.com/quodlibetor/rust-chrono/", branch = "add-permissive-tz-parsing" }
```

and use the `%#z` format specifier and see if it works for you?

Dowwie:
I've run the following simple test and found that optional timezone minutes work as expected:

```rust
extern crate chrono;

use chrono::prelude::*;


fn main() {
    let dt_values = vec!["2018-04-30 11:55:14.847928+00",
                         "2018-04-30 11:55:14.847928+00:00",
                         "2018-04-30 11:55:14.847928+01:00",
                         "2018-04-30 11:55:14.847928+01:30"];

    for value in dt_values {
        let my_dt = DateTime::parse_from_str(value, "%Y-%m-%d %H:%M:%S%.6f%#z")
                              .unwrap().with_timezone(&Utc);

        println!("my_dt is: {:?}", my_dt);
    }

}

```

Dowwie:
@quodlibetor is anything keeping this change from the next release?

quodlibetor:
Yes I need to refactor it to obey the semver-privacy rules that chrono set up for itself @Dowwie . It should be a mechanical transformation from the current design. I've been busy but have a bit more time for OSS this week. I'm intending to get it out ASAP -- depending on free time, the next couple days to this weekend.

quodlibetor:
Okay this code now appropriately uses the private `Internal` struct. I think.

As I wrote that I realized that with the code the way that it was before it would have been completely impossible for a user to match on the `Fixed` enum without having a default case, and probably I can just get rid of my most recent uglifying commit. I'll ponder it, but if anyone has opinions about whether or not c3207a5 is necessary I am interested.

quodlibetor:
Since it's still possible to match on the Internal variant as `Internal(_)`, and the docs don't explicitly state that the enum is intended to be `open` I don't think that it's possible to backwards-compatibly add the parsing without doing the ugliness in my last commit. In 0.5 we'll make it a "truly" open enum, but for now I'll get this released.

quodlibetor:
this is now on crates.io as [0.4.3](https://crates.io/crates/chrono/0.4.3)

Dowwie:
@quodlibetor great job!  thanks, all for adding this!

Dowwie:
@quodlibetor Last meetup was my first RustNYC meetup.  I didn't realize who you were but we sat right next to each other.  Hah!    I'll be sure to introduce myself at the next event (after your talk!)

quodlibetor:
Oh, awesome! Looking forward to it!

PR #245, opened 2018-05-10
ISO-8601 Duration parser?

FauxFaux:
I'd like to see an ISO-8601 Duration (e.g. `P1DT5S`) parser... somewhere. It appears that the `time` crate is deprecated, and that `chrono` is where we're prototyping things. I realise that `chrono` is trying to rid itself of its `OldDuration`, however, and .. the docs say, move to `time`'s. Presumably that would now be "move to `std`'s".

I have written a comically sloppy parser for this format, attached to `OldDuration`. This isn't super useful, as `OldDuration` is disabled by default by features. The tests aren't even run. (They don't pass.)

The limitations in the parser (not accepting Months or Years) mirror the [Java Duration parser](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#parse-java.lang.CharSequence-), and would easily satisfy my usecase. In fact, I've previously been using the Java parser for years and had no idea that the restrictions existed.

Is there anywhere this can usefully be inserted into `chrono`? Should this be a separate crate, as an extension to `std`, or an RFC for `std`? Any thoughts on a better way to do this kind of parser without loading a proper parser like `nom`?

Any thoughts on how much validation you'd expect from a sloppy parser? I'd like it to reject 'M' on the left of 'T', I think.

(I do not expect this pull request to be merged as-is.)

### DIFF
--- a/src/oldtime.rs
+++ b/src/oldtime.rs
@@ -277,6 +277,56 @@ impl Duration {
         }
         Ok(StdDuration::new(self.secs as u64, self.nanos as u32))
     }
+
+    /// Creates an `OldDuration` from an ISO-8601-like duration expression.
+    ///
+    /// These duration expressions look like `P4DT3M1S` for "four days,
+    /// three minutes and one second". The "accurate" definition of days and weeks
+    /// is used, not the human concept.
+    ///
+    /// Accepts only Weeks, Days, Hours, Minutes and Seconds.
+    ///
+    /// Very little validation is done. `P3.1MT3M` will be parsed as
+    /// "six minutes, six seconds", not "three and a bit months, three minutes"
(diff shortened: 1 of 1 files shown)


FauxFaux:
I wrote a `nom` and a better by-`hand` parser, based on these, as a separate crate. The `nom` parser is shorter and faster, at the expense of error reporting.

https://github.com/FauxFaux/time-parse

All take a `&str`-like and return a `Duration`-like. I dropped support for negative durations.

quodlibetor:
Awesome, thanks!

pickfire:
Doesn't this implementation to match the command allows `P3S2M`?

Farkal:
Any news on this PR ? I need this feature and it would be easier if it was merged :wink:

PR #247, opened 2018-05-18
Nanosecond Timestamps

HarkonenBade:
This merge requests adds a function for retrieving a DateTime as nanoseconds elapsed since the Unix Epoch. It also adds serde functions for serialising/serialising DateTime objects from nanosecond timestamps.

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -992,6 +992,167 @@ pub mod serde {
         }
     }
 
+    /// Ser/de to/from timestamps in nanoseconds
+    ///
+    /// Intended for use with `serde`'s `with` attribute.
+    ///
+    /// # Example:
+    ///
+    /// ```rust
+    /// # // We mark this ignored so that we can test on 1.13 (which does not
+    /// # // support custom derive), and run tests with --ignored on beta and
+    /// # // nightly to actually trigger these.
+    /// #
+    /// # #[macro_use] extern crate serde_derive;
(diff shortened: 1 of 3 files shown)


HarkonenBade:
hmm, that clippy failure seems to be due to clippy adding new checks given that it is not in any code that I have touched.

src/datetime.rs
```diff
@@ -968,6 +992,167 @@ pub mod serde {
...
+    ///
+    /// let as_string = serde_json::to_string(&my_s)?;
+    /// assert_eq!(as_string, r#"{"time":1526522699918355733}"#);
+    /// let my_s: S = serde_json::from_str(&as_string)?;
+    /// assert_eq!(my_s.time, time);
+    /// # Ok(my_s)
+    /// # }
+    /// # fn main() { example().unwrap(); }
+    /// ```
+    pub mod ts_nano_seconds {
```
quodlibetor: I think we should just treat this as a single word `ts_nanoseconds`

quodlibetor:
Oh could you add yourself to the CHANGELOG, too?

HarkonenBade:
What additions to the changelog file should I make?

HarkonenBade:
Ok, renamed and added a few lines to CHANGELOG.md

quodlibetor:
Great, thanks! I'll publish this in a release by this weekend.

Issue #250, opened 2018-06-08
Sub-second format without dot (.)


emschwartz:
I'm working on a project that uses an RFC 3339-like format without the symbols (`-`, `:`, `.`) and with milliseconds cut/padded to 3 . As far as I can tell, chrono allows me to parse everything except the milliseconds on their own. However, I think the only way to encode and decode milliseconds (or other fixed-length fractions of a second) is with a leading dot (`.070`). 

Would it be possible to add a way to parse milliseconds without the dot (e.g. `%3f`)?

quodlibetor:
Yup, that seems like a reasonable request. If you want to implement it I'm happy to mentor, otherwise I'll try and build it this weekend.

emschwartz:
Excellent. I'm a rust newbie but I can take a stab at it this afternoon

quodlibetor:
cool, feel free to ask questions here or in https://gitter.im/chrono-rs/chrono

PR #251, opened 2018-06-08
Add nanoseconds without dots

emschwartz:
Resolves https://github.com/chronotope/chrono/issues/250

### DIFF
--- a/src/format/parse.rs
+++ b/src/format/parse.rs
@@ -308,11 +308,16 @@ pub fn parse<'a, I>(parsed: &mut Parsed, mut s: &str, items: I) -> ParseResult<(
                         s = &s[2..];
                     }
 
-                    Nanosecond | Nanosecond3 | Nanosecond6 | Nanosecond9=> {
+                    Nanosecond | Nanosecond3 | Nanosecond6 | Nanosecond9 => {
                         if s.starts_with('.') {
                             let nano = try_consume!(scan::nanosecond(&s[1..]));
                             try!(parsed.set_nanosecond(nano));
                         }
+                    }
+
+                    Nanosecond3NoDot | Nanosecond6NoDot | Nanosecond9NoDot => {
+                        let nano = try_consume!(scan::nanosecond(&s[1..]));
+                        try!(parsed.set_nanosecond(nano));
                     }
 
                     TimezoneName => return Err(BAD_FORMAT),
@@ -584,6 +589,11 @@ fn test_parse() {
             num!(Hour), lit!(":"), num!(Minute), lit!(":"), num!(Second), fix!(TimezoneOffset)];
            year: 2015, month: 2, day: 4, hour_div_12: 1, hour_mod_12: 2,
            minute: 37, second: 5, offset: 32400);
+    check!("20150204143705567",
+            [num!(Year), num!(Month), num!(Day),
+            num!(Hour), num!(Minute), num!(Second), num!(Nanosecond)];
+            year: 2015, month: 2, day: 4, hour_div_12: 1, hour_mod_12: 2,
+            minute: 37, second: 5, nanosecond: 567000000);
     check!("Mon, 10 Jun 2013 09:32:37 GMT",
            [fix!(ShortWeekdayName), lit!(","), sp!(" "), num!(Day), sp!(" "),
             fix!(ShortMonthName), sp!(" "), num!(Year), sp!(" "), num!(Hour), lit!(":"),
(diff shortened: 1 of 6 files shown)


quodlibetor:
Looking great!

Regarding your questions:

* Yeah, I think we should duplicate the nanosecond parse tests, all of them. You can verify that you are only finding nanoseconds without dots by doing something like:

  ```rust
  check!(".400", [fix!(Nanosecond3NoDot)]; INVALID);  // probably not actually INVALID
  check!("400", [fix!(Nanosecond3NoDot)]; nanosecond: 400_000_000);
  ```

  Does that make sense?
* Only other place to update is [the strftime docs](https://github.com/emschwartz/chrono/blob/efafef42e30dd3a919d9be85dbf45c2dcc65947a/src/format/strftime.rs#L57-L60) and the CHANGELOG.md

emschwartz:
Good catch. It was an issue with using the `try_consume` macro with a slice of the string. `try_consume` updates `s` in place, but if you pass a slice it forgets about the whole rest of the string

src/format/parse.rs
```diff
@@ -532,6 +557,46 @@ fn test_parse() {
...
+
+    check!("",             [fix!(Nanosecond9NoDot)]; ); // no field set, but not an error
+    check!("0",            [fix!(Nanosecond9NoDot)]; nanosecond: 0);
+    check!("4",            [fix!(Nanosecond9NoDot)]; nanosecond: 400_000_000);
+    check!("42",           [fix!(Nanosecond9NoDot)]; nanosecond: 420_000_000);
+    check!("421",          [fix!(Nanosecond9NoDot)]; nanosecond: 421_000_000);
+    check!("42195",        [fix!(Nanosecond9NoDot)]; nanosecond: 421_950_000);
+    check!("421950803",    [fix!(Nanosecond9NoDot)]; nanosecond: 421_950_803);
+    check!("421950803547", [fix!(Nanosecond9NoDot)]; nanosecond: 421_950_803);
+    // check!("421950803547", [fix!(Nanosecond9NoDot), num!(Second)]; nanosecond: 421_950_803, second: 54); // don't skip digits that come after the 9
```
quodlibetor: did you want to uncomment this?

emschwartz: Yeah, I'm fixing the test. I'll add another commit to the PR tomorrow

emschwartz:
Yeah, I'll make those changes tomorrow.

Would you prefer me to squash all the commits into one or leave them as they are?

quodlibetor:
The commits are nicely separated as is, no need to squash. I would also appreciate the refactoring to use `InternalInternal` to be its own commit, I hope to revert it with the next breaking change.  

Speaking of it, though, if you are going to be editing your test commit *and* you use git from the command line I have [a *very* new project](https://github.com/quodlibetor/git-fixup) that I'm looking for beta testers/bug reporters/contributors on that is supposed to make editing old commits a bit nicer.

emschwartz:
> Could you move your new variants into the InternalInternal struct, the equivalent of this commit for your changes? It didn't exist when you started work, but I think that, sadly, it's necessary for strict semver compat.

I'm not sure I understand that. Is the idea to make it so this feature can't be accessed by the user? What is the semver issue (it's a non-breaking change)?

quodlibetor:
Adding new variants to a public enum [is a major breaking change](https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md#enums). The idea isn't to prevent usage of the feature, it's to require that the only usage of the feature is through the `parse` method, specifically so that we don't break semver.

The hack used by the ecosystem to mark enums as open is to have a `#[doc(hidden)] __DoNotMatchAgainstMe` variant and to include in the docs for an enum that additions to it are expected. However, chrono was created before that convention existed, and so it created the separation of internal and external enums. The internal enum is not public, and so it can be expanded without trouble. The external enum, though, is public and so technically someone could be doing an exhaustive match against the whole thing.

For most libraries I think that just having bad docs would mean that you can just fix them and improve life, but chrono is one of the most depended-on crates in the ecosystem and so if any crate is going to have someone doing crazy things against massive enums then chrono might be it.

quodlibetor:
specifically, the only things that you should need to change should be to move your variants to the `InternalInternal` enum and to change your uses of `fix!` to `internal_fix!`, all tests (in particular tests that verify `parse`) should continue to work without any other changes.

emschwartz:
Gotcha. Thanks for the explanation. Should it be `InternalInternal` or `InternalFixed`?

emschwartz:
Made the change. Ready for re-review

quodlibetor:
LGTM, I'm going to do one more review pass this evening, I don't recall any other possible concerns so hopefully I'll get this published tonight.

emschwartz:
Sounds good. Thanks for the help!

PR #252, opened 2018-06-13
Add tests and handling for negative timestamps

quodlibetor:


quodlibetor:
Currently this is failing with:

```text
---- src/naive/datetime.rs - naive::datetime::NaiveDateTime::timestamp_millis (line 281) stdout ----
        thread 'rustc' panicked at 'test executable failed:

thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `-43200900`,
 right: `-900`', src/naive/datetime.rs:13:1
note: Run with `RUST_BACKTRACE=1` for a backtrace.

', librustdoc/test.rs:341:17
note: Run with `RUST_BACKTRACE=1` for a backtrace.

---- src/naive/datetime.rs - naive::datetime::NaiveDateTime::timestamp (line 250) stdout ----
        thread 'rustc' panicked at 'test executable failed:

thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `-43201000`,
 right: `-1`', src/naive/datetime.rs:13:1
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

quodlibetor:
🤕 the tests were failing because I was using `11` to refer to `11pm`, when obviously that is `11am`. 😢 

On the plus side, this has always worked, and now it's documented better and the variables I think make the code a bit more clear.

PR #257, opened 2018-06-27
bincode="0.8.0" -> bincode="1.0.1"

bcko:


quodlibetor:
It looks like bincode has changed its API enough that this will require more meaningful changes. We mostly use bincode internally as a smoke test, what caused you to even notice that the wrong version was here?

I'd be happy to take a change that does do the upgrade, of course.

PR #258, opened 2018-07-04
Extra serde modules

novacrazy:
Sorry this took so long; it kind of got pushed to the side for a while.

This fleshes out the `serde` modules for seconds, milliseconds and nanoseconds for both UTC `DataTime`s and `NaiveDateTime`s.

It also simplifies a few things that were shared across the modules.

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -998,24 +998,117 @@ pub mod serde {
         }
     }
 
-    /// TODO
+    /// Ser/de to/from timestamps in milliseconds
+    ///
+    /// Intended for use with `serde`s `with` attribute.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # // We mark this ignored so that we can test on 1.13 (which does not
+    /// # // support custom derive), and run tests with --ignored on beta and
+    /// # // nightly to actually trigger these.
+    /// #
+    /// # #[macro_use] extern crate serde_derive;
+    /// # #[macro_use] extern crate serde_json;
+    /// # extern crate chrono;
+    /// # use chrono::{TimeZone, DateTime, Utc};
+    /// use chrono::serde::ts_milliseconds;
+    /// #[derive(Deserialize, Serialize)]
+    /// struct S {
+    ///     #[serde(with = "ts_milliseconds")]
+    ///     time: DateTime<Utc>
+    /// }
+    ///
+    /// # fn example() -> Result<S, serde_json::Error> {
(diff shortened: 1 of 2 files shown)


src/datetime.rs
```diff
@@ -926,77 +948,61 @@ pub mod serde {
...
             }
 
             /// Deserialize a timestamp in seconds since the epoch
-            fn visit_i64<E>(self, value: i64) -> Result<DateTime<FixedOffset>, E>
+            fn visit_i64<E>(self, value: i64) -> Result<DateTime<Utc>, E>
                 where E: de::Error
             {
-                from(FixedOffset::east(0).timestamp_opt(value, 0), &value)
+                serde_from(Utc.timestamp_opt(value / 1_000_000_000,
+                                            (value % 1_000_000_000) as u32),
```
jonasbb: This fails if value is negative, because the modulus will return a negative number and casting it to u32 will make it exceed the maximal value of 999_999_999.
The same applies for the other deserializations of i64.

novacrazy: I mean, that's how it was before. I didn't actually change that code, Git just says I did.

Would it be better to call `abs` on `value` before the modulus?

jonasbb: I know that it is a previous issue. I just though I mention it, if the whole serialization part is changed here anyways.

I do not think `abs` alone works right. Suppose value is `-1`, then you want the two values passed to `timestamp_opt` to be `-1` and `999_999_999`, but just `abs` would give you `0` and `1`.

novacrazy: As the nanosecond argument of `timestamp_opt` is unsigned, there isn’t a way to directly pass a negative timestamp correctly anyway. 

What I could do is check for a negative timestamp and offset the seconds by negative one second, then pass one second subtract the positive nanoseconds part for the nanoseconds argument? That would give the desired results.

novacrazy: Same with the other signed timestamps.

quodlibetor: Sorry about the delay reviewing this!

That does seem reasonable, not handling negative timestamps is a... sad mistake, that I think might be duplicated through more of the codebase. Since this isn't you're fault I'm happy to merge this and fix the negative timestamp issue in a separate PR. Let me know what you think.

quodlibetor: Relatedly, I started trying to deal with negative timestamps in #252 but haven't had the time to finish it.

quodlibetor:
Thanks for the patience, I'm fine keeping the existing bug and merging this.

PR #261, opened 2018-07-09
Tried to make it easier to find specifiers

storyfeet:
I was looking to find the specifiers for ages. I realise the now the pointer was there, I just didn't see it.  I think it's likely other people will miss it too, as there is a lot to find your way through.

I've added a title for Formatting and Parsing, and slightly tweaked the wording removing ()brackets as it is useful information for library users.

Hope this helps.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -207,10 +207,13 @@
 //! # }
 //! ```
 //!
+//! ### Formatting and Parsing
+//!
 //! Formatting is done via the [`format`](./struct.DateTime.html#method.format) method,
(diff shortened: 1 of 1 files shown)


quodlibetor:
Sorry about the clippy failures, feel free to add a clippy allow of `trivially_copy_pass_by_ref` to the lib.rs file (there are a few other clippy lints allowed there which should serve as an example.)

I'll take a look at the generated docs, thanks for this!

quodlibetor:
Thanks!

PR #270, opened 2018-08-10
Construct a NaiveDate by specifying eg. "the 2nd Friday of March 2017"

asayers:
This PR adds the `NaiveDate::from_ymwd` constructor, which allows you to make a `NaiveDate` by specifying eg. "the 2nd Friday of March 2017".  It contains a couple of panics, but these are consistent with the behaviour of the other `NaiveDate` constructors.

* Is the name any good?
* Should it be 0-indexed or 1-indexed? (I prefer 1-indexing in this case.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ Versions with only mechnical changes will be omitted from the following list.
 
 ### Features
 
-* Added `NaiveDate::from_ymwd` for getting eg. the 2nd Friday of March 2017.
+* Added `NaiveDate::from_ymwd{,_opt}` for getting eg. the 2nd Friday of March 2017.
 
 ## 0.4.5
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ Versions with only mechnical changes will be omitted from the following list.
 
 ### Features
 
-* Added `NaiveDate::from_ymwd{,_opt}` for getting eg. the 2nd Friday of March 2017.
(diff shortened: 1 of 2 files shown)


quodlibetor:
As a first pass, regarding the name, are there any examples of this in other languages/libraries? This seems useful, but also seems like it interacts with general date iteration support.

Before we consider adding this I would definitely like to see a `from_ymwd_opt` so that it's possible to handle bad input, the current version can just call and unwrap that.

asayers:
> are there any examples of this in other languages/libraries?

Haskell's thyme library calls this [`WeekdayOfMonth`](http://hackage.haskell.org/package/thyme-0.3.5.5/docs/Data-Thyme-Calendar-WeekdayOfMonth.html). It's a constructor which takes the same parameters as `NaiveDate::from_ymwd`, and then you convert it to a `Day` using an `Iso`.

Perhaps `NaiveDate::from_weekday_of_month` would be more descriptive? (If a bit unwieldy.)

> seems like it interacts with general date iteration support

I guess you're imagining something like this?

```rust
(NaiveDate::from_ymd(2017,3,1)..).filter(|x| x.weekday() == WeekDay::Friday).nth(3)
```

This seems like a nice way to get the functionality I want; but does chrono provide an iterator over consecutive ` NaiveDate`s?

> `from_ymwd_opt`

:+1:

asayers:
@quodlibetor Ping?

asayers:
Hi @quodlibetor, should I close this PR?

PR #273, opened 2018-08-14
`https://doc.rust-lang.org/time/time/struct.Duration.html` does not exist

kud1ing:


quodlibetor:
Thanks! Chrono actually uses the `time` crate from crates.io, I've merged another PR that fixes that.

kud1ing:
Ok, thanks.

PR #298, opened 2019-01-23
Adds quickcheck::Arbitrary support for DateTime

purew:
Available under feature "quickcheck-enabled".

Test with:
  cargo test --features quickcheck-enabled qc_

Notes:
I'm not sure about the best way to include the quickcheck dependency. It now lives under `[dependencies]` as an optional feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ name = "chrono"
 [features]
 default = ["clock"]
 clock = ["time"]
+quickcheck-enabled = ["quickcheck"]
 
 [dependencies]
 time = { version = "0.1.39", optional = true }
@@ -33,6 +34,7 @@ num-integer = { version = "0.1.36", default-features = false }
 num-traits = { version = "0.2", default-features = false }
 rustc-serialize = { version = "0.3.20", optional = true }
 serde = { version = "1", optional = true }
+quickcheck = { version = "*", optional = true }
 
 [dev-dependencies]
 serde_json = { version = "1" }
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -1402,6 +1402,20 @@ pub mod serde {
     }
 }
 
+#[cfg(feature = "quickcheck-enabled")]
+impl quickcheck::Arbitrary for DateTime<Utc> {
+    fn arbitrary<G: quickcheck::Gen>(g: &mut G) -> Self {
+        Utc.timestamp(i64::arbitrary(g), u32::arbitrary(g))
+    }
+}
+
+#[cfg(feature = "quickcheck-enabled")]
+impl quickcheck::Arbitrary for DateTime<Local> {
+    fn arbitrary<G: quickcheck::Gen>(g: &mut G) -> Self {
+        Local.timestamp(i64::arbitrary(g), u32::arbitrary(g))
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::DateTime;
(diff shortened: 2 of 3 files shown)


quodlibetor:
Hey, awesome! Thanks!

Do you know what I should expect from running this? Based on my understanding and on eyeballing this PR I would kind of expect quickcheck to run for a long time and occasionally/hopefully/hopefully not discover panics or unsafety? When I run it it is pretty fast and just says "okay":

```text
$ time cargo test --features quickcheck-enabled -- qc_
    Finished dev [unoptimized + debuginfo] target(s) in 0.08s
     Running target/debug/deps/chrono-eb41b2e7e3c98354

running 2 tests
test datetime::tests::qc_test_quickcheck ... ok
test datetime::tests::qc_test_quickcheck ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 93 filtered out

   Doc-tests chrono

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 212 filtered out

cargo test --features quickcheck-enabled -- qc_  0.32s user 0.19s system 94% cpu 0.539 total
```

PureW:
The default number of tests run is pretty low for the simple test being done. You can increase the number of tests by setting `QUICKCHECK_TESTS` env-var higher.

For example:
```
QUICKCHECK_TESTS=10000000 cargo test --features=quickcheck-enabled qc_
```

quodlibetor:
Ah, neat! That does work for me.

Do you have an opinion on what reasonable next steps would be? Anything in quickcheck I should look at in particular to improve our likelihood of finding bugs?

PureW:
I started using quickcheck this week so I don't know too much about best ways of implementing it or verifying such functionality other than what I've glanced from the docs.

Perhaps @burntsushi has pointers for how client libraries best would add quickcheck support?

BurntSushi:
I think there are two orthogonal concerns here:

1. The testing of chrono itself with quickcheck.
2. The availability of `Arbitrary` implementations on chrono data types for others to use in their own quickcheck tests.

For (1), you generally want to look for universal properties of your data types, generally held under some assumption, and then write tests for those properties. The test in this PR is one such example: given any UTC timestamp, it should be possible to extract its component parts, reconstruct it, and assert that the reconstructed value is the same as the original. I'm sure there are other relevant properties, for example, "adding a date with a naive time always results in something at least as big as the original date."

For (2), the thing you need to be careful about is that it introduces quickcheck as a public dependency of your crate. So for example, if quickcheck has a semver incompatible release, then generally any public dependents also need to have a semver incompatible release once they want to update to the new version of quickcheck. With that said, this PR puts the `Arbitrary` impls behind a non-default feature, so you might be able to get away with ignoring the requirement for doing a semver bump in order to upgrade quickcheck. Ideally, quickcheck would have a 1.0 release, but it in turn has a public dependency on `rand` which is itself not at 1.0 yet. Once that happens, it should be more safe to use quickcheck as a public dependency.

@PureW Note that you can work around the lack of `Arbitrary` impls in this crate by new-typing `DateTime` and writing your own private `Arbitrary` impl in your crate. This may be annoying to various degrees depending on your situation though.

PureW:
@BurntSushi Thanks for the informative reply!

> @PureW Note that you can work around the lack of Arbitrary impls in this crate by new-typing DateTime and writing your own private Arbitrary impl in your crate. This may be annoying to various degrees depending on your situation though.

Aye, that's how I've proceeded in my own project for now.

Issue #301, opened 2019-01-27
Add `Option<DateTime>` outputs for various serde helpers (ts_seconds, etc)


manifest:
The following doesn't work now:

```rust
#[derive(Debug, Serialize, Deserialize)]
struct Object {
    #[serde(with = "ts_seconds")]
    time: Option<DateTime<Utc>>,
}
```
```rust
error[E0308]: mismatched types
  --> src/db/rtc.rs:36:17
   |
36 | #[derive(Debug, Serialize, Deserialize)]
   |                 ^^^^^^^^^ expected struct `chrono::datetime::DateTime`, found enum `std::option::Option`
   |
   = note: expected type `&chrono::datetime::DateTime<chrono::offset::utc::Utc>`
              found type `&'__a std::option::Option<chrono::datetime::DateTime<chrono::offset::utc::Utc>>`

error[E0308]: match arms have incompatible types
  --> src/db/rtc.rs:36:28
   |
36 | #[derive(Debug, Serialize, Deserialize)]
   |                            -----------
   |                            |
   |                            expected enum `std::option::Option`, found struct `chrono::datetime::DateTime`
   |                            match arm with an incompatible type
   |                            in this macro invocation
   |
   = note: expected type `std::option::Option<chrono::datetime::DateTime<_>>`
              found type `chrono::datetime::DateTime<_>`

error[E0308]: match arms have incompatible types
  --> src/db/rtc.rs:36:28
   |
36 | #[derive(Debug, Serialize, Deserialize)]
   |                            -----------
   |                            |
   |                            expected enum `std::option::Option`, found struct `chrono::datetime::DateTime`
   |                            match arm with an incompatible type
   |                            in this macro invocation
   |
   = note: expected type `std::option::Option<chrono::datetime::DateTime<_>>`
              found type `chrono::datetime::DateTime<_>`
```

manifest:
I've just added a PR with these new modules:
- ts_nanoseconds_option
- ts_milliseconds_option
- ts_seconds_option

They all reuse current Serializers and Deserializers' Visitors.

elpiel:
Is this still on the roadmap/workings?

quodlibetor:
This is now merged.

PR #302, opened 2019-01-27
Add Serialize/Deserialize for Option<DateTime>

manifest:
This PR for the [issue](https://github.com/chronotope/chrono/issues/301).

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -835,6 +835,18 @@ pub mod serde {
     use offset::{LocalResult, TimeZone, Utc, FixedOffset};
     use serdelib::{ser, de};
 
+    #[doc(hidden)]
+    #[derive(Debug)]
+    pub struct SecondsTimestampVisitor;
+
+    #[doc(hidden)]
+    #[derive(Debug)]
+    pub struct NanoSecondsTimestampVisitor;
+
+    #[doc(hidden)]
+    #[derive(Debug)]
+    pub struct MilliSecondsTimestampVisitor;
+
     // try!-like function to convert a LocalResult into a serde-ish Result
     fn serde_from<T, E, V>(me: LocalResult<T>, ts: &V) -> Result<T, E>
         where E: de::Error,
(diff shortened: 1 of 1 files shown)


manifest:
I've just added a PR with these new modules:
- ts_nanoseconds_option
- ts_milliseconds_option
- ts_seconds_option

They all reuse current Serializers and Deserializers' Visitors.

manifest:
One test for Rust 1.13.0 fails on

```rust
error: unions are unstable and possibly buggy (see issue #32836)
   --> /Users/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.48/src/macros.rs:42:13
    |
42  |             pub $t $i { $($field)* }
    |             ^^^^^^^^^^^^^^^^^^^^^^^^
    | 
   ::: /Users/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.48/src/lib.rs
    |
177 | cfg_if! {
    | - in this macro invocation
Build failed, waiting for other jobs to finish...
error: Could not compile `libc`.
```

I'm not sure where it comes from. It seems unrelated to this PR.

quodlibetor:
yeah that error is coming from libc, the most recent version doesn't support rust 1.13. I'm not sure what the best way to handle that is, but it's not something you have to worry about.

manifest:
@quodlibetor Is there anything that need to be done to have this one merged?

promethe42:
+1 can we merge this?

quodlibetor:
Sorry about the delay, I feel like there ought to be a better way to do this, since currently using an option can be done pretty trivially https://github.com/chronotope/chrono/pull/125#issuecomment-388342995 .

So, (A) I would rather this was already more ergonomic, and (B) maybe it would be better to do something that doesn't add quite as much API surface/code that requires maintenance.

promethe42:
@quodlibetor maybe I missed something but the code in https://github.com/chronotope/chrono/pull/125#issuecomment-388342995 doesn't work for me.

It throws with `Json(Error("missing field `revisionDate`"))` when the field is missing.

IMHO because the proposed code in your comment does not actually handle the case when the field is not here. Which is the whole point behind `Option<DateTime<Utc>>`.

So AFAIK this PR is still required and very useful.

promethe42:
My bad: the suggested solution in https://github.com/chronotope/chrono/pull/125#issuecomment-388342995 works if I add `#[serde(default)]` to the corresponding field.

quodlibetor:
Looks good enough to merge, we can make the code-size smaller in a future commit.

PR #305, opened 2019-02-22
Added `Offset` support to `Local`

araspik:
This would be really useful when trying to convert `DateTime<Local>` to `DateTime<FixedOffset>`, allowing this simple code:
```rust
local_datetime.with_timezone(Local.fix())
```
instead of the much longer current solutions.
Note that no tests are added; This was a quick solution (even though I dug through both `time` and `chrono`'s docs for a bit).
Improvements are definitely possible, but this is a good start.

### DIFF
--- a/src/offset/local.rs
+++ b/src/offset/local.rs
@@ -10,6 +10,7 @@ use naive::{NaiveDate, NaiveTime, NaiveDateTime};
 use {Date, DateTime};
 use super::{TimeZone, LocalResult};
 use super::fixed::FixedOffset;
+use super::offset::Offset;
 
 /// Converts a `time::Tm` struct into the timezone-aware `DateTime`.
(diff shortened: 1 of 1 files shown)


araspik:
It seems that Travis CI tests on Mac are broken (something about unions)?

mqudsi:
@araspik #271 fixes the root issue by adding `Into` and `From` for all the datetime combinations, that should probably suffice for your needs?

ErichDonGubler:
Wait, #271 actually just sets the seconds offset of the `FixedOffset` to be 0, doesn't it? Is that actually what we want here?

quodlibetor:
Yeah #271 allows converting `Local` to e.g. `Utc`, but this would allow converting to a `FixedOffset` that is the same as the Local Offset. If this got some tests I'd be happy to merge it.

Rebasing on top of master will fix the 1.13 tests.

PR #312, opened 2019-04-21
derive {Copy, Eq, Hash} for LocalResult.

kennytm:
Include the additional traits so that e.g. `LocalResult<FixedOffset>` can be Copy/Eq/Hash'ed.

### DIFF
--- a/src/offset/mod.rs
+++ b/src/offset/mod.rs
@@ -26,7 +26,7 @@ use Weekday;
 use {Date, DateTime};
 
 /// The conversion result from the local time to the timezone-aware datetime types.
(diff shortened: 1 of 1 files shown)


quodlibetor:
Do you know off the top of your head if the derive for Copy/Eq/Hash does a e.g. `Copy where T: Copy` for the generic bound? I suppose it must otherwise this shouldn't compile.

Thanks!

kennytm:
@quodlibetor Yes, if you `#[derive(X)]` it will automatically add the `T: X` bound to the `impl`.

PR #320, opened 2019-06-21
DelayFormat now works with alignment and width

### LINKED ISSUE #233: DelayedFormat does not respect format items
[This code](https://play.rust-lang.org/?gist=b9a3b5a1faaf2042346d1c269022b6f9&version=stable):

```rust
extern crate chrono;

fn main() {
    let date = chrono::Utc::now();
    let tz = date.format("%Z");
    println!("left {:>10} right", tz.to_string());
    println!("left {:>10} right", tz)
}
```

Outputs:

```text
left        UTC right
left UTC right
```

But both lines should probably be the same.

cc #230

SamokhinIlya:
Fix #233

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -1721,4 +1721,34 @@ mod tests {
         assert_eq!(SystemTime::from(epoch.with_timezone(&FixedOffset::east(32400))), UNIX_EPOCH);
         assert_eq!(SystemTime::from(epoch.with_timezone(&FixedOffset::west(28800))), UNIX_EPOCH);
     }
+
+    #[test]
+    fn test_datetime_format_alignment() {
+        let datetime = Utc.ymd(2007, 01, 02);
+
+        // Item::Literal
+        let percent = datetime.format("%%");
+        assert_eq!("  %", format!("{:>3}", percent));
+        assert_eq!("%  ", format!("{:<3}", percent));
+        assert_eq!(" % ", format!("{:^3}", percent));
+
+        // Item::Numeric
+        let year = datetime.format("%Y");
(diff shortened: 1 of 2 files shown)


quodlibetor:
Thanks!

Could you add a couple of tests that demonstrate the behavior and correctness?

SamokhinIlya:
Sure. Also I have question about try! macro in format function. Is it used there intentionally or just hasn't been updated to question mark operator? If it's the latter, can I replace it?

quodlibetor:
Thanks! As you noticed, chrono still supports ancient rust that doesn't have the question mark operator.

These new tests are confusing, in them you're creating a string without alignment, and then aligning that string, which isn't related to the chrono formatting.

This does match my expectation of what would happen, since I don't think of the chrono formatter as knowing about padding.

SamokhinIlya:
> These new tests are confusing, in them you're creating a string without alignment, and then aligning that string, which isn't related to the chrono formatting.

Do you mean it'll be better to inline it like this?
```rust        
assert_eq!(format!("  2007 January 02", ymd_formatted), format!("{:>17}", ymd));
assert_eq!(format!("2007 January 02  ", ymd_formatted), format!("{:<17}", ymd));
assert_eq!(format!(" 2007 January 02 ", ymd_formatted), format!("{:^17}", ymd));
```

quodlibetor:
Oh darn, I misunderstood what the code was doing. I saw:

```rust
        let percent = datetime.format("%%");
```

as

```rust
        let percent: String = datetime.format("%%");
```

but it's actually

```rust
       let percent: ::format::DelayedFormat<_> = datetime.format("%%");
```

That makes me pretty happy with this, thanks!

PR #341, opened 2019-09-07
Implement Support for no_std

### LINKED ISSUE #336: no_std support
mentioned in [195](https://github.com/chronotope/chrono/issues/195#issuecomment-527201627), we need a good story around no_std/alloc support.

The parts of this that jump right to mind:

* Breaking up the core mathy parts of chrono into their own crates/features
* Refactoring various stringy bits to be in terms of `Write`
* parts of chrono around parsing allocate, too, I believe.

CryZe:
This adds a new `std` feature to chrono that is enabled by default. By deactivating this feature via `default-features = false` you can now use chrono in applications that don't use the standard library. The `serde` feature is supported as well.

Resolves #336


CryZe:
Well looks like I definitely need help with that CI script.

quodlibetor:
Thank you! I will look at the CI script.

ci/travis.sh
```diff
@@ -119,3 +126,6 @@ build_and_test
 
 CHANNEL=1.13.0
 build_only
+
+CHANNEL=stable
+build_core_test
```
quodlibetor: nit: I think I'd prefer `build_core_test` to just go inline between lines 125 and 126 instead of reactivating stable

src/datetime.rs
```diff
@@ -3,12 +3,16 @@
...
-use std::ops::{Add, Sub};
+use core::{str, fmt, hash};
+use core::cmp::Ordering;
+use core::ops::{Add, Sub};
+#[cfg(any(feature = "std", test))]
 use std::time::{SystemTime, UNIX_EPOCH};
 use oldtime::Duration as OldDuration;
 
+#[cfg(not(any(feature = "std", test)))]
+use alloc::string::{String, ToString};
```
quodlibetor: It seems like it might be worth sticking this (and what seems like the two functions that return strings) behind a separate `alloc` feature flag?

CryZe: I added an `alloc` feature now. I'm kind of sad how much is missing without it though. And it's mostly only because the timezone offset is stored as a String in the DelayedFormat.

src/format/mod.rs
```diff
@@ -17,20 +17,54 @@
...
+#[cfg(any(feature = "alloc", feature = "std", test))]
+use alloc::boxed::Box;
+#[cfg(any(feature = "alloc", feature = "std", test))]
+use alloc::string::{String, ToString};
+
+#[cfg(not(any(feature = "alloc", feature = "std", test)))]
+mod core_only {
+    /// Core only
+    #[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+    pub struct Box<T: ?Sized>(core::marker::PhantomData<T>);
```
CryZe: This thing is kind of hacky, but it allows at least partially supporting some of the format module stuff.

CryZe:
Also with alloc this is now slightly a breaking change as the serde feature is now called `serde-1`, because you can't have feature dependencies on optional crates themselves (for whatever reason).

quodlibetor:
> Also.. this is now slightly a breaking change

Hm that's a good point, I'll need to think about that. Does SemVer apply to compilation flags? I've taken things out of the --no-default-features install without bumping the patch version before and I'm not sure that I've gotten bug reports about it. This is a much more dramatic change than before, though, so maybe not.

> as the serde feature is now called serde-1, because you can't have feature dependencies on optional crates themselves.

Yeah that's true, but you can rely on optional dependencies as though they're features, so I think that it's fine to just use `serde`, and it seems to work for me.

CryZe:
I believe if you activate the serde feature without the alloc feature now, you get weird compilation errors. I guess if it's properly documented that you need to also activate alloc then, it might not be a problem.

quodlibetor:
Yeah that's a good point, I didn't quite realize that it was possible for people to want to be able to use serde without std, but in retrospect it makes sense.

src/format/mod.rs
quodlibetor: Can you explain how/if this actually works/what the runtime experience would be? Or is there some explanation of this pattern somewhere?

Alternatively, would you mind adding some small code to the core test file that demonstrates what this ends up doing?

CryZe: It's only so the Item enum actually compiles as there's Box<str> in there. I'm not sure if we even need this enum when alloc isn't compiled in. I honestly don't know what even works in that module at all if DelayedFormat isn't compiled in.

CryZe: Does the timezone for the DelayedFormat really need to be a String or could it be turned into a &'static str? (I'm not sure if there ever are any truly dynamic timezones that would need String)

CryZe: Wait, what is OwnedLiteral and OwnedSpace even used for? I don't see them being used anywhere.

quodlibetor:
Just to be clear, you mean that `serde` requires the alloc feature, so doing `--no-default-features --features serde` that will start failing, but I would expect just doing `--features serde` would work. And that's what appears to be the case, but I might be missing something.

CryZe:
I'd be surprised if it works, but I guess it might? I'll check tomorrow.

quodlibetor:
I just pushed a change that should remove all intermediary allocation even when the serde feature is enabled. I didn't look into how that plays with the core::box placeholder.

I tried to simplify the features, as well, let me know how it feels to you to use this way.

CryZe:
That does not seem to build if you build with serde and no alloc:
```
   Compiling chrono v0.4.9 (https://github.com/CryZe/chrono?branch=no-std#e5bbc94c)
error[E0432]: unresolved import `alloc`
    --> /home/user/.cargo/git/checkouts/chrono-5169c63406b50cd5/e5bbc94/src/naive/datetime.rs:1761:13
     |
1761 |         use alloc::format;
     |             ^^^^^
     |             |
     |             unresolved import
     |             help: a similar path exists: `core::alloc`

error[E0432]: unresolved import `alloc`
    --> /home/user/.cargo/git/checkouts/chrono-5169c63406b50cd5/e5bbc94/src/naive/datetime.rs:1908:13
     |
1908 |         use alloc::format;
     |             ^^^^^
     |             |
     |             unresolved import
     |             help: a similar path exists: `core::alloc`

error[E0432]: unresolved import `alloc`
    --> /home/user/.cargo/git/checkouts/chrono-5169c63406b50cd5/e5bbc94/src/naive/datetime.rs:2055:13
     |
2055 |         use alloc::format;
     |             ^^^^^
     |             |
     |             unresolved import
     |             help: a similar path exists: `core::alloc`

error: cannot find macro `format` in this scope
    --> /home/user/.cargo/git/checkouts/chrono-5169c63406b50cd5/e5bbc94/src/naive/time.rs:1448:51
     |
1448 |             value.parse().map_err(|err| E::custom(format!("{}", err)))
     |                                                   ^^^^^^

error: cannot find macro `format` in this scope
    --> /home/user/.cargo/git/checkouts/chrono-5169c63406b50cd5/e5bbc94/src/naive/datetime.rs:1711:51
     |
1711 |             value.parse().map_err(|err| E::custom(format!("{}", err)))
     |                                                   ^^^^^^

error: cannot determine resolution for the macro `format`
    --> /home/user/.cargo/git/checkouts/chrono-5169c63406b50cd5/e5bbc94/src/naive/datetime.rs:1857:46
     |
1857 |                     .ok_or_else(|| E::custom(format!("value is not a legal timestamp: {}", value)))
     |                                              ^^^^^^
     |
     = note: import resolution is stuck, try simplifying macro imports

error: cannot determine resolution for the macro `format`
    --> /home/user/.cargo/git/checkouts/chrono-5169c63406b50cd5/e5bbc94/src/naive/datetime.rs:1865:46
     |
1865 |                     .ok_or_else(|| E::custom(format!("value is not a legal timestamp: {}", value)))
     |                                              ^^^^^^
     |
     = note: import resolution is stuck, try simplifying macro imports

error: cannot determine resolution for the macro `format`
    --> /home/user/.cargo/git/checkouts/chrono-5169c63406b50cd5/e5bbc94/src/naive/datetime.rs:2004:46
     |
2004 |                     .ok_or_else(|| E::custom(format!("value is not a legal timestamp: {}", value)))
     |                                              ^^^^^^
     |
     = note: import resolution is stuck, try simplifying macro imports

error: cannot determine resolution for the macro `format`
    --> /home/user/.cargo/git/checkouts/chrono-5169c63406b50cd5/e5bbc94/src/naive/datetime.rs:2012:46
     |
2012 |                     .ok_or_else(|| E::custom(format!("value is not a legal timestamp: {}", value)))
     |                                              ^^^^^^
     |
     = note: import resolution is stuck, try simplifying macro imports

error: cannot determine resolution for the macro `format`
    --> /home/user/.cargo/git/checkouts/chrono-5169c63406b50cd5/e5bbc94/src/naive/datetime.rs:2150:46
     |
2150 |                     .ok_or_else(|| E::custom(format!("value is not a legal timestamp: {}", value)))
     |                                              ^^^^^^
     |
     = note: import resolution is stuck, try simplifying macro imports

error: cannot determine resolution for the macro `format`
    --> /home/user/.cargo/git/checkouts/chrono-5169c63406b50cd5/e5bbc94/src/naive/datetime.rs:2157:46
     |
2157 |                     .ok_or_else(|| E::custom(format!("value is not a legal timestamp: {}", value)))
     |                                              ^^^^^^
     |
     = note: import resolution is stuck, try simplifying macro imports

error: aborting due to 11 previous errors
???

quodlibetor:
Oh, I thought i had fixed that correctly, I'll get it working.

quodlibetor:
This seems like it's working on stable, now. The failures had to do with using core in 1.13, which I'm not super interested in debugging unless someone actively wants it.

CryZe:
Yeah it seems to be working now across all the different combinations :)

quodlibetor:
WooHoo!

Okay I'd like to see what we can do about the stub core module that you created, ideally we can make it unnecessary. I doubt I'll have any time to look at it before friday, if you're interested.

quodlibetor:
I am happy with this as is now, we can add and improve things as they come up. Thanks for this!

src/format/mod.rs
quodlibetor: Yes I ended up putting these variants behind the same feature gate, I'm not sure that it will have any effect on usability. _Potentially_ if people were compiling their own `format::Item` lists it could have been a thing?

Issue #347, opened 2019-10-06
Is `Datelike::num_days_from_ce` off-by-one?


gThorondorsen:
The documentation of `Datelike::num_days_from_ce` currently says:

> Returns the number of days since January 1, Year 1 (aka Day 1) in the proleptic Gregorian calendar.

Given such a description, I would expect the following test to pass, which it currently does not.

```rust
use chrono::{NaiveDate, Datelike};
assert_eq!(NaiveDate::from_ymd(1, 1, 1).num_days_from_ce(), 0); // the reference point
assert_eq!(NaiveDate::from_ymd(2, 1, 1).num_days_from_ce(), 365); // since year 1 is not a leap year
```

And the year 0, *a.k.a.* 1 BCE, is considered a leap year (in the proleptic Gregorian calendar) but the doctest says that `NaiveDate::from_ymd(0, 1, 1).num_days_from_ce()` gives -365 and not -366.

- - -
The fix to `Datelike::num_days_from_ce` on its own is just replacing the call to `Datelike::ordinal` by a call to `Datelike::ordinal0`. That of course breaks many tests throughout the code base. Is it worth trying to fix all of that, or is it better to adjust the relevant documentation to shift the reference date to December 31, 1 BCE?

gThorondorsen:
> The documentation of Datelike::num_days_from_ce currently says:
>> Returns the number of days since **January 1, Year 1 (aka Day 1)** in the proleptic Gregorian calendar.

(emphasis mine)

https://github.com/chronotope/chrono/blob/9e50bfe03460036dbd752440851d924be3730b08/src/naive/date.rs#L402

- - -
So I think I misunderstood the documentation. To me "the number of days since \<some-date>" can only mean that \<some-date> returns 0, the next day returns 1, and so on.

What about the following?
> Counts the days in the proleptic Gregorian calendar, with January 1, Year 1 (CE) as day 1.

I would even argue to renaming the method to `num_days_in_ce`, but that is a breaking change of course.

- - -
The same remarks also apply to `NaiveDate::from_num_days_from_ce` and its `_opt` version: the 1-based indexing is not clear enough.

quodlibetor:
As you pointed out the implementation is correct, it just starts at 1.

PR #348, opened 2019-10-06
Test `Datelike::num_days_from_ce` against an alternative implementation

gThorondorsen:
I find the alternative implementation clearer (obviously, I came up with it) but it may not be better than the current one, which I did not change. In particular, the current implementation may be slightly faster because it only does one or two divisions.

If you disagree that this is simpler, or a test worth having, feel free to just close the pull request.

P.S.: I uncovered #347 while working on this PR.

### DIFF
--- a/benches/chrono.rs
+++ b/benches/chrono.rs
@@ -3,7 +3,7 @@
 extern crate chrono;
 extern crate criterion;
 
-use criterion::{black_box, criterion_group, criterion_main, Criterion};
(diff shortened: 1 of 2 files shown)


quodlibetor:
Thanks for this! I agree that this implementation is more clear. If you're interested in running a benchmark I'd be happy to replace the current implementation with yours if it's as fast.

quodlibetor:
Finally got around to benching this, it does unfortunately seem to be twice as slow, but the test is indeed useful for future optimization attempts, thanks!

image

PR #349, opened 2019-10-08
Improve documentation of `num_days_from_ce` methods

gThorondorsen:
The new wording tries to make clearer that those methods use a 1-based
numbering scheme. I did not go with the renaming suggestion I made in #347 because I want more discussion to happen first.

The commit also includes a couple of drive-by cosmetic changes.


quodlibetor:
Thanks! This looks great! I definitely agree that this makes it more clear.

Issue #202, opened 2018-01-10
Replace month numbers with enum


clarfonthey:
Also allow determining its corresponding number similar to #201.

jam1garner:
I'd like to second this. I think some good reasoning for this would be:
* Compile time ensuring all months are accounted for
* Cleaner pattern matching
* More inline with Rust idioms

At minimum providing both an enum option and a numerical option would be a good start, albeit I don't know what the deprecation/compatibility guarantees of the project are. I would be willing to work on this issue but I'm struggling to find anything on what contributing looks like. @lifthrasiir or @quodlibetor do you have any thoughts? I would rather not put effort into an implementation only for it to not even receive a comment (see #353), but certainly don't mind putting in the leg work behind whatever you guys would like!

pitdicker:
A `Month` enum was added in https://github.com/chronotope/chrono/pull/461.

PR #353, opened 2019-10-28
Adds a public Month enum.

hhamana:
Implementation for a Month enum, allowing -ser|de-rialization.
The implementation mostly mirrors the Weekday enum.
It only adds the enum with minimal integration so as to
not cause any backwards compatibility issue.
A deeper integration should be up to discussion.
Suggested in Issue #202

### DIFF
--- a/src/format/mod.rs
+++ b/src/format/mod.rs
@@ -21,7 +21,7 @@ use std::fmt;
 use std::str::FromStr;
 use std::error::Error;
 
-use {Datelike, Timelike, Weekday, ParseWeekdayError};
+use {Datelike, Timelike, Weekday, ParseWeekdayError, Month, ParseMonthError};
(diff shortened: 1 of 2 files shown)


hhamana:
In hindsight, maybe I rushed a bit too much just tacking it on the lib, just because I felt it was something a time management library should have.
But it it would make more sense if that was an integral part of the date structure. There will only ever be those 12 months, and it would certainly benefit development to enforce it at compile time, for pretty much everything related to months in dates.
But just doing that properly enum-based instead of u32 based seems like a much more intense refactoring. I would be happy to help though.
I'm leaving this open for your consideration in this current light-handed implementation.

Arcovv:
Looks great!

quodlibetor:
Thanks @hhamana , I've rebased this into #461

PR #355, opened 2019-11-15
Remove or clarify references to time crate

jhpratt:
The time crate will _hopefully_ have a v0.2 release coinciding with the
release of Rust 1.40. As such, all references to the time crate make
clear that it is the v0.1 crate, not the soon-to-be-released v0.2 crate.
The first line of the README was also changed, as it's not a strict
superset of the new version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,9 +18,7 @@
 [gitter-image]: https://badges.gitter.im/chrono-rs/chrono.svg
 [gitter]: https://gitter.im/chrono-rs/chrono
 
-It aims to be a feature-complete superset of
-the [time](https://github.com/rust-lang-deprecated/time) library.
-In particular,
+Chrono aims to be a feature-complete time library. In particular,
 
 * Chrono strictly adheres to ISO 8601.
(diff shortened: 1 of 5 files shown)


quodlibetor:
I am happy to merge this or something like this once the time 0.2 crate has been released, I would be particularly interested in making sure that all the conversions work well. A lot of the decisions you're making in time-0.2 look great to me and will make the api a lot nicer.

If you wanted to update this PR to work well with time-0.2 -- even just one or two compat methods -- I'd be a bit happier to merge it.

jhpratt:
By update to work well with 0.2, you mean updating all of the examples? If so, that's definitely something I can do at some point. 0.2 is meant for wide compatibility, so it should be doable.

quodlibetor:
I would say probably updating or duplicating a couple of examples rather than all of them would be fine. Just enough to be sure that everything actually works correctly.

jhpratt:
Closing in favor of #389, which is a more general upgrade request.

PR #358, opened 2019-11-22
Optimize parsing

michalsrb:
Hi, I am working on a project that parses lots of json with `DateTime`s in it. The time spent parsing in chrono appeared quite high compared to other things in profiler, so I tried to optimize it.

Commit c12e4be adds benchmarks. Here is the resulting diff:
```
 name                                                 before ns/iter  after ns/iter  diff ns/iter   diff %  speedup 
 datetime::tests::bench_datetime_from_str             712              337                      -375  -52.67%   x 2.11 
 datetime::tests::bench_datetime_parse_from_rfc2822   252              181                       -71  -28.17%   x 1.39 
 datetime::tests::bench_datetime_parse_from_rfc3339   242              142                      -100  -41.32%   x 1.70
```

Individual commits have their diffs in them.

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -622,17 +622,17 @@ impl str::FromStr for DateTime<FixedOffset> {
 
     fn from_str(s: &str) -> ParseResult<DateTime<FixedOffset>> {
         const ITEMS: &'static [Item<'static>] = &[
-            Item::Space(""), Item::Numeric(Numeric::Year, Pad::Zero),
+                             Item::Numeric(Numeric::Year, Pad::Zero),
             Item::Space(""), Item::Literal("-"),
-            Item::Space(""), Item::Numeric(Numeric::Month, Pad::Zero),
+                             Item::Numeric(Numeric::Month, Pad::Zero),
             Item::Space(""), Item::Literal("-"),
-            Item::Space(""), Item::Numeric(Numeric::Day, Pad::Zero),
+                             Item::Numeric(Numeric::Day, Pad::Zero),
             Item::Space(""), Item::Literal("T"), // XXX shouldn't this be case-insensitive?
-            Item::Space(""), Item::Numeric(Numeric::Hour, Pad::Zero),
+                             Item::Numeric(Numeric::Hour, Pad::Zero),
             Item::Space(""), Item::Literal(":"),
-            Item::Space(""), Item::Numeric(Numeric::Minute, Pad::Zero),
+                             Item::Numeric(Numeric::Minute, Pad::Zero),
             Item::Space(""), Item::Literal(":"),
-            Item::Space(""), Item::Numeric(Numeric::Second, Pad::Zero),
+                             Item::Numeric(Numeric::Second, Pad::Zero),
                              Item::Fixed(Fixed::Nanosecond),
             Item::Space(""), Item::Fixed(Fixed::TimezoneOffsetZ),
             Item::Space(""),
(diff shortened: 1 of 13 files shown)


michalsrb:
Looks like it is failing with rust 1.13.0, I'll try to fix it...

quodlibetor:
niiice, thanks!

src/format/parsed.rs
```diff
@@ -112,6 +112,7 @@ pub struct Parsed {
 
 /// Checks if `old` is either empty or has the same value to `new` (i.e. "consistent"),
 /// and if it is empty, set `old` to `new` as well.
+#[inline(always)]
```
quodlibetor: hm could we check what the bench results look like with just `inline` instead of `inline(always)`? My understanding is that that gives llvm more flexibility to decide whether to inline, although AIUI it's still just a hint anyway.

src/datetime.rs
```diff
@@ -1753,4 +1753,53 @@ mod tests {
         assert_eq!(format!("{}  ", ymd_formatted), format!("{:<17}", ymd));
         assert_eq!(format!(" {} ", ymd_formatted), format!("{:^17}", ymd));
     }
+
+    #[cfg(feature = "bench")]
+    #[bench]
+    fn bench_datetime_parse_from_rfc2822(bh: &mut test::Bencher) {
+        bh.iter(|| {
+            let str = test::black_box("Wed, 18 Feb 2015 23:16:09 +0000");
```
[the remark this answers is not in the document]
michalsrb: I didn't realize when I wrote it, but yeah, it is strange as identifier.

src/format/parsed.rs
michalsrb: I remeasured again and got these results:

No attribute:
```
test datetime::tests::bench_datetime_from_str            ... bench:         347 ns/iter (+/- 14)
test datetime::tests::bench_datetime_parse_from_rfc2822  ... bench:         197 ns/iter (+/- 121)
test datetime::tests::bench_datetime_parse_from_rfc3339  ... bench:         166 ns/iter (+/- 8)
```

With `#[inline]`:
```
test datetime::tests::bench_datetime_from_str            ... bench:         344 ns/iter (+/- 9)
test datetime::tests::bench_datetime_parse_from_rfc2822  ... bench:         192 ns/iter (+/- 11)
test datetime::tests::bench_datetime_parse_from_rfc3339  ... bench:         167 ns/iter (+/- 15)
```

With `#[inline(always)]`:
```
test datetime::tests::bench_datetime_from_str            ... bench:         369 ns/iter (+/- 33)
test datetime::tests::bench_datetime_parse_from_rfc2822  ... bench:         175 ns/iter (+/- 9)
test datetime::tests::bench_datetime_parse_from_rfc3339  ... bench:         138 ns/iter (+/- 5)
```

Interestingly the `bench_datetime_from_str` got worse, unlike in my previous measurement. The gains may be too random. I am fine with omitting this commit, it may be too risky to force inlining.

quodlibetor:
So running it locally it seems like the `always` has no meaningful effect on runtime, if you'd like to verify and remove it that would be fine, otherwise I'll merge this as-is this weekend.

michalsrb:
Ok, just to clarify, should the `#[inline]` stay without the `always`, or should it go away completely?

quodlibetor:
I'd prefer it to stay without the always unless there's evidence that losing the always makes it slower.

michalsrb:
Rebased on master to fix conflicts and changed the `#[inline(always)]` to `#[inline]`.

quodlibetor:
This looks great! The issues are coming from the new "std is behind a feature gate" thing that caused conflicts, I can fix them but if you want to I think everything that you're doing can either switch to `core` or put things behind a feature gate. Also if you want to deal with it you can feel free to stick an `#![allow(deprecated)]` in the crate root to make the `try!` warnings go away, I've got a commit to transition to `?` that I didn't want to conflict with your changes, ironically.

quodlibetor:
Thanks so much!

PR #362, opened 2019-11-24
Implement rockets FromFormValue for naive types

7omb:
Adds a new rocket feature that implements FromFormValue for
NaiveTime, NaiveDate and NaiveDateTime.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ num-integer = { version = "0.1.36", default-features = false }
 num-traits = { version = "0.2", default-features = false }
 rustc-serialize = { version = "0.3.20", optional = true }
 serde = { version = "1.0.99", default-features = false, optional = true }
+rocket = { version = "0.4", default-features = false, optional = true}
 
 [target.'cfg(all(target_arch = "wasm32", not(target_os = "emscripten")))'.dependencies]
 wasm-bindgen = { version = "0.2", optional = true }
--- a/Makefile
+++ b/Makefile
@@ -26,4 +26,4 @@ test:
 
 .PHONY: doc
 doc: authors readme
-	cargo doc --features 'serde rustc-serialize bincode'
+	cargo doc --features 'serde rustc-serialize bincode rocket'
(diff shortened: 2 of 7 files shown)


quodlibetor:
What is the current recommended way to implement `FromFormValue` for external (e.g. chrono) types in rocket codebases in the face of orphan impls? Is everybody just newtyping things?

I'm both a bit nervous about embedding any definition of time formats, and feel kind of like this is the wrong place for this impl, but [@ SergioBenitez has already said they're not going to implement these](https://github.com/SergioBenitez/Rocket/issues/602).

I might be willing to accept this under a renamed `unstable-rocket` feature flag documented as not having any guarantees of continuing to exist between different chrono releases, but that also feels ugly.

Something that might be better would be to create a `rocket-chrono` (or `chrono-rocket`, depending on which organization it goes in) crate that has some `RocketNaiveTime(chrono::NaiveTime)` embedded in them with some reasonable `Into<NaiveDateTIme>` etc impls to at least make the implementations slightly less annoying. My intuition is that this would belong in the rocket org since I can imagine several more of these growing, but I'm open to debate.

7omb:
Yes, I think newtype wrappers are what everybody is probably doing right now and if you're just start a new project this feels quite clunky. This commit was just a little experiment over the weekend that I want to share and discuss. Since it doesn't fit into rocket right now the obvious/only alternative was chrono. But you're right, it feels misplaced and rocket would definitely be a better place. Even the `unstable-rocket` feature does't feel right. So it's ok for me to close this issue. I think I'll give your idea of the `rocket-chrono` wrapper a shot. Maybe this will converge to a better solution in the near future.

quodlibetor:
@7omb Thanks for the PR! I agree that the current situation with orphan impls in rust is pretty annoying.

If you want to do something generic I can imagine creating a `rocket-formvalue-interop` crate with a bunch of `FromFormValue<___>` newtype wrappers, maybe people want it badly enough that they'll contribute their own newtype wrappers.

7omb:
That's indeed a good idea. At least support for money and numeric types is missing as well right now.

PR #365, opened 2019-11-27
Add support for compilation on wasm32-wasi

### LINKED ISSUE #352: Support for wasm32-wasi
It seems like wasm32-wasi isn't supported, but could very easily be supported by chrono (although I don't quite know what the situation with the time crate is nowadays).

coolreader18:
This just disables `wasmbind` from taking effect on WASI. Resolves #352.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ num-traits = { version = "0.2", default-features = false }
 rustc-serialize = { version = "0.3.20", optional = true }
 serde = { version = "1.0.99", default-features = false, optional = true }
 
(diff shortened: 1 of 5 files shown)


CryZe:
This doesn‘t really resolve my issue. It‘s a good change, but the idea was to make now() work with WASI.

coolreader18:
It does compile without this, but the presence of wasm-bindgen in the dependency tree at all would add a bunch of `wbindgen` imports that runtimes like wasmtime and wasmer don't provide. I think I did misunderstand the issue, though.

quodlibetor:
I agree that this is an improvement, but we still want some more features to fully close out #352

quodlibetor:
Also thank you very much for the contribution @coolreader18 !

PR #374, opened 2019-12-13
Change docs and playground features so that it compiles

### LINKED ISSUE #373: Chrono is no longer able to be compiled by the playground
```toml
[package.metadata.playground]
all-features = true
```

```
error[E0554]: `#![feature]` may not be used on the stable release channel
   --> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/chrono-0.4.10/src/lib.rs:386:32
    |
386 | #![cfg_attr(feature = "bench", feature(test))] // lib stability features as per RFC #507
    |                                ^^^^^^^^^^^^^
```

quodlibetor:
All features includes some things that are nightly only.

Fixes #373

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,6 @@ alloc = []
 std = []
 clock = ["time", "std"]
 wasmbind = ["wasm-bindgen", "js-sys"]
-__internal_bench = ["criterion"]
 
 [dependencies]
 time = { version = "0.1.39", optional = true }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -55,10 +55,10 @@ doc-comment = "0.3"
 wasm-bindgen-test = "0.2"
 
 [package.metadata.docs.rs]
-all-features = true
+features = ["serde"]
```
shepmaster: To the best of my knowledge, docs.rs always builds the docs with the nightly compiler. This probably doesn't need to be changed for *functional* reasons.

quodlibetor: fair! I mostly just didn't want to accidentally document too much, as the later commits in this PR help prevent.

PR #375, opened 2019-12-16
Support comparison between dts with different timezones. Fixes #354

### LINKED ISSUE #354: Implement comparison for DateTimes with different TimeZones
PartialEq is implemented for different timezones:
`impl<Tz: TimeZone, Tz2: TimeZone> PartialEq<DateTime<Tz2>> for DateTime<Tz>`
but PartialOrd is only implemented for the same timezone:
`impl<Tz: TimeZone> PartialOrd<DateTime<Tz>> for DateTime<Tz>`

Example:
```
use chrono::{Local,Utc};
if Local::now() == Utc::now() { }; // Compiles
if Local::now() < Utc::now() { }; // Does not compile
if Local::now() < Utc::now().into() { } // Compiles
```

dlalic:
Fixes #354.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Versions with only mechanical changes will be omitted from the following list.
 
 * Support "negative UTC" in `parse_from_rfc2822` (@quodlibetor #368 reported in
   #102)
+* Support comparisons of DateTimes with different timezones (@dlalic in #375)
 
 ### Internal improvements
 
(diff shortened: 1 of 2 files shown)


quodlibetor:
Thanks this looks great! I wouldn't mind adding some more tests to verify edge-case behavior, but if you're not interested in adding that then I will be happy to!

dlalic:
> Thanks this looks great! I wouldn't mind adding some more tests to verify edge-case behavior, but if you're not interested in adding that then I will be happy to!

Thanks, TBH didn't know where to put the test in the first place, as test fn names seem to be quite generic with lots of asserts (not a criticism, just wanted to keep the things consistent). Feel free to improve on it yourself and maybe I can add more, or name some of the edge cases you'd like to see covered and where to put them.

PR #376, opened 2019-12-17
Remove deprecated Error::description

AnderEnder:
`Error::description` has been documented as soft-deprecated since 1.27.0 (17 months ago). It is going to be hard-deprecated soon.

This PR:
- Removes all implementations of `description` in all error types

Related PR: https://github.com/rust-lang/rust/pull/66919

### DIFF
--- a/src/format/mod.rs
+++ b/src/format/mod.rs
@@ -318,32 +318,22 @@ enum ParseErrorKind {
 /// Same to `Result<T, ParseError>`.
 pub type ParseResult<T> = Result<T, ParseError>;
 
-impl ParseError {
-    fn description(&self) -> &str {
+impl fmt::Display for ParseError {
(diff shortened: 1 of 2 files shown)


quodlibetor:
It's still required in rust 1.13, I think, we should maybe add `#[allow(deprecated)]` to description until chrono 0.5 and maybe create an issue to remove it in the next release.

AnderEnder:
@quodlibetor, I think the minimal version should be increased to 1.27.0. I do not think so ancient version should be still supported.

quodlibetor:
@AnderEnder the next minor release of chrono will definitely pin to a newer compiler, and possibly update the backcompat policy. For now, though, I do take back-compat very seriously and still support ancient rust in the 0.4 branch.

PR #378, opened 2019-12-27
support round tripping display <-> datetime

### LINKED ISSUE #147: Consider including the `T` in Display of DateTime
Currently, you cannot roundtrip a DateTime to a String through `ToString` and back with `FromStr`, because `FromStr` rejects strings without the `T` between the date and time, and `ToString` does not include it. This is pretty surprising behavior IMO, and it doesn't seem harmful to include the `T` in the displayed output anyway.

quodlibetor:
This extens `FromStr` to allow either a `T` or a ` ` (space) as the delimiter
between the date and the time, and, because of the fact that the `Z`
parser-specifier is shared with the Fixed notation, extends the fixed notation
to support `UTC` in addition to `Z` as the zero-offset.

IMO this Fixes #147

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ Versions with only mechanical changes will be omitted from the following list.
 
 ### Improvements
 
+* Support a space or `T` in `FromStr` for `DateTime<Tz>`, meaning that e.g.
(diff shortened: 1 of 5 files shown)


quodlibetor:
Interestingly enough perf with this is pretty much completely unchanged.

image

PR #379, opened 2019-12-27
Format code using 'cargo fmt'

Atul9:


quodlibetor:
thanks @Atul9 ! I am holding off on formatting the code right now because there are a couple big PRs hopefully landing soon, and I am trying to find some rustfmt configs that will make the transition a tiny bit less disruptive.

Issue #380, opened 2019-12-27
Feature request: Add a `Period` data structure


jwir3:
A `Period` data structure is similar to a `Duration`, except that it has a specific start date. For example, one might have a period of 21 days, starting from January 1, 2020, which would have a start date/time of `2020-01-01 00:00:00` and an end date/time of `2020-01-22 00:00:00`.  (Obviously, these don't have timezones/offsets, so they are more likely to be `NaivePeriod` objects.

I am working on an implementation for this in a branch of chrono. The reason I am thinking this might be useful within the chrono library itself is that, given a `NaivePeriod` object, we can then determine if there is overlap with another `NaivePeriod`, or whether it contains a `NaiveDateTime`. I personally am working on a scheduling app, and thus this is a useful use case for me. That said, I'm not sure it belongs within the chrono library - perhaps a separate, add-on library would be more useful.

Thanks for your feedback!

liquidscorpio:
Sounds interesting! Have you thought about the public API interface of the feature? If yes, would it be possible for you to share the same?

quodlibetor:
Yes, a `Period` (built on top of some sort of `TimeSpan` that is like a duration but can be negative) is something that is very much wanted inside of chrono!

If you feel like putting in-progress work up I would be happy to review/collaborate on it before it's ready for inclusion.

jwir3:
@liquidscorpio :

> Sounds interesting! Have you thought about the public API interface of the feature? If yes, would it be possible for you to share the same?

Yes. For my purposes, a `NaivePeriod` (i.e. a `Period` without a timezone/offset) fulfills my needs. I wrote a fairly quick add-on library here: www.github.com/jwir3/chrono_period. Specifically, the API looks something like this:

https://docs.rs/chrono_period/0.1.0/chrono_period/struct.NaivePeriod.html

Right now, I've only added the method implementations for `get_intersection_with()` and `intersects_with()` (the latter being a call that utilizes the former to return a simple true/false value).

@quodlibetor :

> Yes, a Period (built on top of some sort of TimeSpan that is like a duration but can be negative) is something that is very much wanted inside of chrono!

[...]

PR #384, opened 2020-01-04
:sparkles: Add a NaivePeriod data structure.

jwir3:
**NOTE**: This is a work in progress, it's not quite ready to be merged yet I don't think.

A NaivePeriod covers the period of time between two NaiveDateTime(s),
inclusive. This data structure can be used to intersect two NaivePeriods to
determine if they overlap.

This is in partial fulfillment of #380, but does not provide necessary code
to handle DateTime objects (with offsets).

Refs #380.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -440,7 +440,7 @@ pub use oldtime::Duration;
 #[cfg(feature="clock")]
 #[doc(no_inline)] pub use offset::Local;
 #[doc(no_inline)] pub use offset::{TimeZone, Offset, LocalResult, Utc, FixedOffset};
-#[doc(no_inline)] pub use naive::{NaiveDate, IsoWeek, NaiveTime, NaiveDateTime};
+#[doc(no_inline)] pub use naive::{NaiveDate, IsoWeek, NaiveTime, NaiveDateTime, NaivePeriod};
 pub use date::{Date, MIN_DATE, MAX_DATE};
 pub use datetime::{DateTime, SecondsFormat};
 #[cfg(feature = "rustc-serialize")]
@@ -455,7 +455,7 @@ pub mod prelude {
     #[cfg(feature="clock")]
     #[doc(no_inline)] pub use Local;
     #[doc(no_inline)] pub use {Utc, FixedOffset};
-    #[doc(no_inline)] pub use {NaiveDate, NaiveTime, NaiveDateTime};
+    #[doc(no_inline)] pub use {NaiveDate, NaiveTime, NaiveDateTime, NaivePeriod};
     #[doc(no_inline)] pub use Date;
     #[doc(no_inline)] pub use {DateTime, SecondsFormat};
     #[doc(no_inline)] pub use SubsecRound;
@@ -482,11 +482,14 @@ pub mod naive {
     mod isoweek;
     mod time;
     mod datetime;
+    mod period;
 
     pub use self::date::{NaiveDate, MIN_DATE, MAX_DATE};
     pub use self::isoweek::IsoWeek;
     pub use self::time::NaiveTime;
     pub use self::datetime::NaiveDateTime;
+    pub use self::period::NaivePeriod;
+
     #[cfg(feature = "rustc-serialize")]
     #[allow(deprecated)]
     pub use self::datetime::rustc_serialize::TsSeconds;
--- a/src/naive/period.rs
+++ b/src/naive/period.rs
@@ -1,0 +1,279 @@
+use naive::NaiveDateTime;
+use oldtime::Duration;
+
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -440,7 +440,7 @@ pub use oldtime::Duration;
 #[cfg(feature="clock")]
 #[doc(no_inline)] pub use offset::Local;
 #[doc(no_inline)] pub use offset::{TimeZone, Offset, LocalResult, Utc, FixedOffset};
-#[doc(no_inline)] pub use naive::{NaiveDate, IsoWeek, NaiveTime, NaiveDateTime};
+#[doc(no_inline)] pub use naive::{NaiveDate, IsoWeek, NaiveTime, NaiveDateTime, NaivePeriod};
```
quodlibetor: I'm tempted to keep this out of the root namespace and put the period module behind an `unstable` feature flag for a month or two before it's released with back-compat guarantees. I might change my mind much faster than that, though, or be convinced otherwise.

src/lib.rs
```diff
@@ -455,7 +455,7 @@ pub mod prelude {
     #[cfg(feature="clock")]
     #[doc(no_inline)] pub use Local;
     #[doc(no_inline)] pub use {Utc, FixedOffset};
-    #[doc(no_inline)] pub use {NaiveDate, NaiveTime, NaiveDateTime};
+    #[doc(no_inline)] pub use {NaiveDate, NaiveTime, NaiveDateTime, NaivePeriod};
```
quodlibetor: Same here keeping it out of the prelude at least at the beginning.

src/naive/period.rs
```diff
@@ -0,0 +1,279 @@
...
+/// As such, if you need to interset two time periods that are of differing time zones, you're out
+/// of luck (patches accepted, though).
+#[derive(PartialEq, Eq, PartialOrd, Ord, Copy, Clone, Debug)]
+pub struct NaivePeriod {
+    /// The date at which the period begins.
+    pub start: NaiveDateTime,
+
+    /// The date at which the period ends. This is inclusive, meaning that the period runs up to
+    /// and including this date/time.
+    pub end: NaiveDateTime
```
quodlibetor: We should keep these private, and require a constructor.

src/naive/period.rs
```diff
@@ -0,0 +1,279 @@
...
+/// This is similar to a [`Duration`](Duration), except that it has a fixed start
+/// date/time (and thus a fixed end date/time), allowing clients to determine if specific
+/// segments of time intersect.
+///
+/// # Notes
+/// This period is not reliant on time zones. For the time being, offsets aren't considered at all.
+/// As such, if you need to interset two time periods that are of differing time zones, you're out
+/// of luck (patches accepted, though).
+#[derive(PartialEq, Eq, PartialOrd, Ord, Copy, Clone, Debug)]
+pub struct NaivePeriod {
```
quodlibetor: Can we accomplish a bit more future compatibility and smaller api surface with a `Period<T: Datelike + Timelike>`? I'd rather reduce the number of places that chrono distinguishes between Naive and tz-aware types. Putting this behind an unstable flag will allow a punt on this for awhile.

src/naive/period.rs
```diff
@@ -0,0 +1,279 @@
...
+    /// let start = NaiveDateTime::new(NaiveDate::from_ymd(2020, 1, 1), NaiveTime::from_hms(0, 0, 0));
+    /// let end = NaiveDateTime::new(NaiveDate::from_ymd(2021, 1, 1), NaiveTime::from_hms(0, 0, 0));
+    ///
+    /// let np1 = NaivePeriod::new(start, end);
+    ///
+    /// assert_eq!(np1.duration(), Duration::days(366));
+    /// ```
+    #[inline]
+    pub fn new(start: NaiveDateTime, end: NaiveDateTime) -> Self {
+        NaivePeriod { start: start, end: end }
```
quodlibetor: Is it reasonable for us to verify that `start < end` or `start <= end`? Maybe rename to `left, right` and just put them in the right order?

src/naive/period.rs
```diff
@@ -0,0 +1,279 @@
+use naive::NaiveDateTime;
+use oldtime::Duration;
+
+use std::cmp::{min, max};
+
+/// A period of time between two ISO 8601 dates/times ([NaiveDateTime](NaiveDateTime)s).
```
quodlibetor: fwiw I think docs links only work on nightly, and docs.rs builds on nightly, so all links could be changed from the `[..](..)` to be just `[..]`. Separately, stylistically, I slightly prefer all docs links to just be simple names with backticks: e.g. 
```markdown
[`NaiveDateTime`]
```

PR #408, opened 2020-04-08
extract format_item function

sergeyt:
I'm writing custom date formatter on top of chrono. I'd like to reuse this function and later add additional formatting items.


quodlibetor:
This is great, thanks!

PR #409, opened 2020-04-08
[SergeyT] add additional fixed formatting items

sergeyt:
This change is [Reviewable](https://reviewable.io/reviews/chronotope/chrono/409)

### DIFF
--- a/src/format/mod.rs
+++ b/src/format/mod.rs
@@ -249,6 +249,9 @@ pub enum FixedExt {
     ///
     /// Prints only first letter.
     OneLetterUpperAmPm,
+    // Do not match against this.
+    #[doc(hidden)]
+    __DoNotMatch,
 }
 
 /// An opaque type representing fixed-format item types for internal uses only.
@@ -652,6 +655,7 @@ pub fn format_item<'a>(
                         result.push_str(if t.hour12().0 {"P"} else {"A"});
                         Ok(())
                     }),
+                _ => return Err(fmt::Error),
             };
 
             match ret {
--- a/src/format/mod.rs
+++ b/src/format/mod.rs
@@ -164,6 +164,10 @@ pub enum Fixed {
     ///
     /// Prints a full name in the title case, reads either a short or full name in any case.
     LongMonthName,
+    /// One letter month names.
+    ///
+    /// Prints only first letter.
+    OneLetterMonthName,
     /// Abbreviated day of the week names.
     ///
     /// Prints a three-letter-long name in the title case, reads the same name in any case.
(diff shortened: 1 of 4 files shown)


quodlibetor:
Thanks for this! I'm a bit hesitant to add this to chrono directly -- one letter month and weekday names are ambiguous, and even though this PR doesn't add them to the public formatting language, they're still footguns.

I would be more comfortable adding a `Fixed::Custom<UserFormatter>` enum variant and associated trait that allows users to extend the format spec with custom behavior. 

Does that seem reasonable to you? Would you be interested in working on that?

sergeyt:
@quodlibetor sounds reasonable since parser cannot be implemented for one-letter months/weekdays. one-letter am/pm is not ambiguous.

Just FYI these new items are needed for excel-style formatting.

sergeyt:
@quodlibetor what if we move ambiguous or unparsable items to separate enum like FixedUnparsable. I'd like to keep one-letter month/weekday in chrono since I am going to add locales for them.

Also I'd like to add 'Era' item existing in excel, I don't know maybe it exists in chrono already.

quodlibetor:
> 'Era' item existing in excel, I don't know maybe it exists in chrono already.

what is `Era` in excel?

Compat with excel OOTB seems worthwhile, so yeah I'm happy adding this with appropriate documentation, which I'll be fine adding.

In terms of implementation, to maintain backwards compatibility I think that these should go into a new `FixedExt` enum so that we can add a `From<FixedExt> for InternalFixed`. We can use the `#[doc(hidden)] __DoNotMatch` extensible enum trick for this new `FixedExt` enum so that we can continue adding public fixed formats for it.

sergeyt:
@quodlibetor sorry I still don't get idea behind adding From :smile: for FixedExt. why is this needed? it works without it

as I understand I should rebase with main branch, right?

quodlibetor:
> I still don't get idea behind adding From 😄 for FixedExt

Ah so currently you have expanded `Item` with an extra variant:

```rust
pub enum FixedExt {...}

pub enum Item {
    ...
    FixedExt(FixedExt),
}
```

but this breaks backwards compatibility, because `Item` is not an extensible enum and users are
allowed to do exhaustive matches on it.

I would prefer to have this:

```rust
pub enum FixedExt {...}

pub enum Fixed {
    ...
    Internal(InternalFixed)
}

// note not pub
enum InternalFixed {
    ...
    FixedExt(FixedExt)
}
```

If you do that, then I believe it will be required to add a `From` implementation to make it
possible to construct the list.

```rust
impl From<FixedExt> for Fixed {
    fn from(self) -> Fixed {
        Fixed::Internal(InternalFixed(self))
    }
}
```

But if you can add a test that uses `format_with_items` using your new items and the above
implementation (instead of adding it to `Item` directly) then that's fine.

> as I understand I should rebase with main branch, right?

Yes please. Sorry about the possible merge conflicts :-/

sergeyt:
> but this breaks backwards compatibility, because `Item` is not an extensible enum and users are
> allowed to do exhaustive matches on it.

got it, the same applies to `Numeric::ShortYear`, right? can it be resolved by versioning saying that there are breaking changes in new version? I don't expect that most users do match on the `Item`

sergeyt:
closing in favor of rebased version in PR #455

PR #412, opened 2020-04-10
Use js runtime for Local offset in js environment

### LINKED ISSUE #411: DateTime.with_timezone() seems not working with chrono::offset::Local in wasm32
Here is a simple test for native x86:

```rust
fn main() {
    let dt = chrono::DateTime::parse_from_rfc3339("2020-04-07T14:07:52+00:00").unwrap();

    println!("dt: {}", dt);
    println!("dt: {}", dt.with_timezone(&chrono::Local));
}
```

When I run it in my local timezone (UTC+2) it correctly prints:
```
dt: 2020-04-07 14:07:52 +00:00
dt: 2020-04-07 16:07:52 +02:00
```

But in a wasm project, this code:
```rust
        log!(d);
        log!(chrono::Local::now());
        log!(d.with_timezone(&chrono::FixedOffset::east(3600)));
        log!(d.with_timezone(&chrono::Local));
```

prints in Firefox:
```
2020-04-07T14:07:15+00:00
2020-04-10T15:36:23.396+02:00
2020-04-07T15:07:15+01:00
2020-04-07T14:07:15+00:00
```

You can see that with `FixedOffset` I get correct date, but not with `Local`.

quodlibetor:
Possibly Fixes #411

### DIFF
--- a/src/offset/local.rs
+++ b/src/offset/local.rs
@@ -157,6 +157,14 @@ impl TimeZone for Local {
         Date::from_utc(*utc, *midnight.offset())
     }
 
+    #[cfg(all(target_arch = "wasm32", not(target_os = "wasi"), feature = "wasmbind"))]
+    fn from_utc_datetime(&self, utc: &NaiveDateTime) -> DateTime<Local> {
+        // Get the offset from the js runtime
+        let offset = FixedOffset::west((js_sys::Date::new_0().get_timezone_offset() as i32) * 60);
+        DateTime::from_utc(*utc, offset)
+    }
+
+    #[cfg(not(all(target_arch = "wasm32", not(target_os = "wasi"), feature = "wasmbind")))]
     fn from_utc_datetime(&self, utc: &NaiveDateTime) -> DateTime<Local> {
         let timespec = datetime_to_timespec(utc, false);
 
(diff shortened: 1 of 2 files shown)


src/offset/local.rs
```diff
@@ -151,6 +151,14 @@ impl TimeZone for Local {
         Date::from_utc(*utc, *midnight.offset())
     }
 
+    #[cfg(all(target_arch = "wasm32", not(target_os = "wasi"), feature = "wasmbind"))]
+    fn from_utc_datetime(&self, utc: &NaiveDateTime) -> DateTime<Local> {
+        // Get the offset from the js runtime
+        let offset = FixedOffset::west((js_sys::Date::new_0().get_timezone_offset() as i32) * 60);
+        DateTime::from_utc(utc, offset)
```
ip1981, line 158: `utc` is a reference, but [NaiveDateTime is expected](https://docs.rs/chrono/0.4.11/chrono/struct.DateTime.html#method.from_utc).

quodlibetor: Thanks, you noticing that this was wrong pointed out that travis wasn't running tests correctly _at all_, I've migrated to gh actions which are correctly finding this bug. _Finally_ merging it in.

PR #418, opened 2020-05-04
Implement .abs() for Duration

abreis:
This PR adds a method to get the absolute (modulus) of a `Duration`.

It is useful for checking how much two `DateTime`s diverge without having to fall back to `i64`s.

### DIFF
--- a/src/oldtime.rs
+++ b/src/oldtime.rs
@@ -229,6 +229,12 @@ impl Duration {
         if d < MIN || d > MAX { None } else { Some(d) }
     }
 
+    /// Returns the duration as an absolute (non-negative) value.
+    #[inline]
+    pub fn abs(&self) -> Duration {
+        Duration { secs: self.secs.abs(), nanos: self.nanos }
+    }
+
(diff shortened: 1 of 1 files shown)


quodlibetor:
Thanks!

src/oldtime.rs
```diff
@@ -229,6 +229,12 @@ impl Duration {
         if d < MIN || d > MAX { None } else { Some(d) }
     }
 
+    /// Returns the duration as an absolute (non-negative) value.
+    #[inline]
+    pub fn abs(&self) -> Duration {
+        Duration { secs: self.secs.abs(), nanos: self.nanos }
```
ComputerDruid, line 235: This seems like it doesn't handle nanos properly; -1s + .5s is -0.5s which should have an absolute value of .5s not 1.5s

abreis: You're correct: I didn't realize at the time that the `nanos` field always 'adds' to the `secs` field (which makes sense since you can't store `-0` in the `secs` field).

abreis: I submitted #734.

PR #420, opened 2020-05-07
Replace "cfg(test)" with "cfg(doctest)" for readme testing

GuillaumeGomez:
Rustdoc now passes "doctest" when running in test mode.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ std = []
 clock = ["time", "std"]
 wasmbind = ["wasm-bindgen", "js-sys"]
 __internal_bench = []
+__doctest = []
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


GuillaumeGomez:
Question: I see you support rust 1.13 but `cfg(doctest)` appeared in `1.40`. Should I add a feature to disable it automatically for older versions?

quodlibetor:
Thanks, yes I would prefer it if you could put this behind a feature!

You'll also need to rebase off of master to fix a build problem that got into master.

Hey, sorry it took me awhile

GuillaumeGomez:
Sorry for the missing `fi` (not enough bash script in my life haha).

quodlibetor:
You're very lucky

PR #423, opened 2020-05-08
Fix format padding for multiple format::Item

ar3s3ru:
Latest master has failed the following test:
```
---- datetime::tests::test_datetime_format_alignment stdout ----

thread 'datetime::tests::test_datetime_format_alignment' panicked at 'assertion failed: `(left == right)`

  left: `"  2007 January 02"`,

 right: `"             2007                           January                                02"`', src/datetime.rs:2245:9

note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
_(Link: https://travis-ci.org/github/chronotope/chrono/jobs/682148123)_

The reason for the failure is that padding is applied in the `chrono::format::format()` function for each item:

https://github.com/chronotope/chrono/blob/f450ae32377c0cd425cd08f5bfd6e37c23931196/src/format/mod.rs#L600

This solution doesn't work as this test is asserting:

https://github.com/chronotope/chrono/blob/89da02ebd5e9fd14faa0bbeff5ed579b9804cd46/src/datetime.rs#L2242-L2247

because the same padding is applied to each `chono::format::Item` in `format_item`, instead of being applied to the whole output of `chrono::format::format`.

### DIFF
--- a/src/format/mod.rs
+++ b/src/format/mod.rs
@@ -17,29 +17,29 @@
 
 #![allow(ellipsis_inclusive_range_patterns)]
 
+#[cfg(feature = "alloc")]
+use alloc::boxed::Box;
+#[cfg(feature = "alloc")]
+use alloc::string::{String, ToString};
 use core::borrow::Borrow;
 use core::fmt;
 use core::str::FromStr;
 #[cfg(any(feature = "std", test))]
 use std::error::Error;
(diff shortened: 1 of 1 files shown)


quodlibetor:
Thanks for this! I fixed it separately, not sure how it snuck into master.

PR #425, opened 2020-05-22
Update patch version of time to add Redox support

AdminXVII:
The 0.1 version of the time crates supports Redox up from version 0.1.43.
Update the dependency to allow redox to use chrono.

From what I can gather, chrono will either no longer use the time crate or rely on version 0.2, but on the mean time this should allow easy inclusion of the Redox target.


quodlibetor:
So to be clear this should have been fine since cargo should choose the newest version anyway, but if this makes it easier for redux to depend on chrono 👍

PR #426, opened 2020-05-26
add secs_nanos function to support nanoseconds > max i64 in constructor

sploiselle:
`std::time::Duration`'s signature permits you to use a total number of nanoseconds > `i64`'s max value by "compressing" some of those nanos into seconds. This PR simply attempts to add a function mimicking that API.

### DIFF
--- a/src/oldtime.rs
+++ b/src/oldtime.rs
@@ -132,6 +132,19 @@ impl Duration {
         Duration { secs: secs, nanos: nanos as i32 }
     }
 
+    /// Makes a new `Duration` with the given number of seconds and nanoseconds.
+    /// 
+    /// `nanos` in excess of `999_999_999` are carried over into seconds.
+    #[inline]
+    pub fn seconds_nanos(secs: i64, nanos: i32) -> Duration {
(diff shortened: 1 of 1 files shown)


src/oldtime.rs
```diff
@@ -132,6 +132,16 @@ impl Duration {
         Duration { secs: secs, nanos: nanos as i32 }
     }
 
+    /// Makes a new `Duration` with the given number of seconds and nanoseconds.
+    pub fn secs_nanos(secs: i64, nanos: i32) -> Duration {
```
quodlibetor, line 136: This should probably be marked inline so that it's available for inlining across crate boundaries, otherwise since there are no generics rustc will generate the object code ahead of time.

Also since this is the first multi-parameter constructor for duration I think it could be named `seconds_nanos`, to keep it in-line with the single-parameter constructor naming.

Could this also document that if the number of nanoseconds is greater than 1B that the nanos will carry over into the seconds?

sploiselle:
@quodlibetor Can you ping me when you determine where the CI on this repo is broken?

sploiselle:
@quodlibetor Would you mind merging this?

PR #428, opened 2020-05-27
Fix build

quodlibetor:
A previous PR updated the README instead of src/lib.rs, which should have been
caught by CI but was not.


quodlibetor:
Seems like there was some merge skew, or something that allowed some bad stuff to get in.

PR #429, opened 2020-06-01
Fix serde with `_option` macro bug.

fx-kirin:
`ts_seconds_option` like macros have not been applied to `serialize_some`. Some library like `bincode` cannot deserialize those values because of this problem.

To reproduce such bug, i leave a github repository [here](https://github.com/fx-kirin/chrono-bincode-bug).

https://github.com/chronotope/chrono/pull/302

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -1184,7 +1184,7 @@ pub mod serde {
             where S: ser::Serializer
         {
             match *opt {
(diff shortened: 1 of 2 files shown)


quodlibetor:
Thanks for the contribution!

Could you add a test that verifies that this fixes something?

And you'll need to rebase on top of master to get a build fix.

quodlibetor:
I've added the test and verified. Thanks!

PR #433, opened 2020-06-16
Fix comment: correct the comment of RFC2822 format

eaufavor:
d.day() is u32 so the padding here is '0'(%d) not ' '(%e).


quodlibetor:
Nice catch!

PR #436, opened 2020-06-21
Migrate to github actions from travis + appveyor

quodlibetor:
Travis has simply not been running tests reliably.


quodlibetor:
I'm merging this without getting the wasm tests running because at least with this PR tests _run_, unlike what seemed to be happening before.

PR #445, opened 2020-07-02
Add rounding and truncating by Duration

### LINKED ISSUE #280: Add round/truncate functions for any specified Duration
It would be very useful for some use cases to have a round or truncate function for any specified Duration.

For now, I'm doing this to truncate some DateTime to a 5 minutes interval:
```rust
    let dt = Utc::now();
    let round_down = dt.timestamp() % (5 * 60);
    let dt_rounded = dt.with_nanosecond(0).unwrap() - time::Duration::seconds(round_down);
```

It would be nice to have something like this:
```rust
    let dt = Utc::now();
    let dt_rounded = dt.truncate(time::Duration::minutes(5));
```

I can submit a PR if there is some basic mentoring available -- I'm still inexperienced in Rust.

robyoung:
Resolves #280

This adds a `DurationRound` trait modelled after `SubsecRound`. It turned out to be a bit more complex that I had anticipated to handle all possible `DateTime`s and `Duration`s so I chose not to.

Both rounding and truncating are done via `Duration::num_nanoseconds` and `DateTime::timestamp_nanos`. This means that they will fail if either the `Duration` or the `DateTime` are too big to represented as nanoseconds. They will also fail if the `Duration` is bigger than the timestamp.

### DIFF
--- a/src/round.rs
+++ b/src/round.rs
@@ -1,8 +1,15 @@
 // This is a part of Chrono.
 // See README.md and LICENSE.txt for details.
 
+use core::cmp::Ordering;
+use core::fmt;
+use core::marker::Sized;
 use core::ops::{Add, Sub};
+use datetime::DateTime;
 use oldtime::Duration;
+#[cfg(any(feature = "std", test))]
+use std;
+use TimeZone;
 use Timelike;
 
 /// Extension trait for subsecond rounding or truncation to a maximum number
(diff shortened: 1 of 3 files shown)


src/round.rs
```diff
@@ -85,14 +88,105 @@ fn span_for_digits(digits: u16) -> u32 {
...
+/// Extension trait for rounding or truncating a DateTime by a Duration.
+///
+/// # Limitations
+/// Both rounding and truncating are done via [`Duration::num_nanoseconds`] and
+/// [`DateTime::timestamp_nanos`]. This means that they will fail if either the
+/// `Duration` or the `DateTime` are too big to represented as nanoseconds. They
+/// will also fail if the `Duration` is bigger than the timestamp.
+///
+/// [`Duration::num_nanoseconds`]: struct.Duration.html#method.num_nanoseconds
+/// [`DateTime::timestamp_nanos`]: struct.DateTime.html#method.timestamp_nanos
```
quodlibetor, line 100: This shouldn't be necessary: docs are built with nightly and should get auto-linkified if the item is in scope. You can check by running `cargo +nightly doc --open` -- it should run without warnings, and hte link should work without this.

It's also better to use the auto-linkification, because it is robust to the fact that items can appear in multiple places in docs, and relative paths aren't always correct.

robyoung: Ace, I've removed them.

robyoung:
I've just squashed the little fixup commits I did on my phone last night.

src/round.rs
```diff
@@ -85,14 +88,102 @@ fn span_for_digits(digits: u16) -> u32 {
...
 }
 
+/// Extension trait for rounding or truncating a DateTime by a Duration.
+///
+/// # Limitations
+/// Both rounding and truncating are done via [`Duration::num_nanoseconds`] and
+/// [`DateTime::timestamp_nanos`]. This means that they will fail if either the
+/// `Duration` or the `DateTime` are too big to represented as nanoseconds. They
+/// will also fail if the `Duration` is bigger than the timestamp.
+pub trait DurationRound {
```
quodlibetor, line 98: Overall this trait looks great, thanks!

A couple notes, though:

* I think that it should take an associated type for errors instead of panicking
* The methods should follow the same name order as the trait

Leading to:

```rust
pub trait DurationRound {

    type Err: std::error::Error;

    fn duration_round(&self, duration: Duration) -> Result<Self, Self::Err>;
    fn duration_trunc(&self, duration: Duration) -> Result<Self, Self::Err>;
}
```

There are possibly other naming schemes that we could consider:

* `trait RoundToDuration` gives `fn round_to_duration` and `fn trunc_to_duration` which is maybe
  worse.
* `trait RoundTo<T>` gives some worse ergonomics in some cases, but it may allow us to do something
  like:

  ```rust
  struct Seconds;

  impl<Tz> RoundTo<Seconds> for DateTime<Tz> { .. }
  ```

  which would remove the limitation of dealing with nanoseconds.

Actually, could we remove the `nanoseconds` limitation in specific cases by checking what the
smallest precision in the duration is? Even just deciding to do seconds vs nanoseconds by checking
if `Duration.nanos_mod_sec()` is non-zero should dramatically reduce the cases where this
could error.

I think leaving this as `DurationRound` but adding the check for precision is probably the way to go, but I'm curious what your use-case is, and if that makes sense for it?

robyoung: > A couple notes, though:
> 
> - I think that it should take an associated type for errors instead of panicking
> - The methods should follow the same name order as the trait
Totally agree. I've taken a stab at it.

> Actually, could we remove the `nanoseconds` limitation in specific cases by checking what the smallest precision in the duration is? Even just deciding to do seconds vs nanoseconds by checking if `Duration.nanos_mod_sec()` is non-zero should dramatically reduce the cases where this could error.
Yes. I had tried this originally but found that `Duration.nanos_mod_secs` is private. It then took me quite a while to discover that `Duration` sometimes comes from the `time` crate so I couldn't even make it crate public. Can you think of a way around that? I thought about converting to `std::time::Duration` and then looking at `subsec_nanos` but then it's not available on `no_std`.

>  but I'm curious what your use-case is, and if that makes sense for it?
My use case is pretty simple. Bucketing events that happened at most a year ago into durations of between a minute and a week. I will never come close to the limitations. The `RoundTo<T>` trait would be a little less easy to work with for me.

quodlibetor: > I had tried this originally but found that Duration.nanos_mod_secs is private.

Ah. Yeah this will be resolved by #286 , which is continuously becoming higher priority and I'm going to make time for. 

For now we can just accept the limited range (pretty sure it's still 1970±500 years) and file an issue to expand it either when it's a problem for folks or when I finally finish that.

src/round.rs
```diff
@@ -85,14 +92,155 @@ fn span_for_digits(digits: u16) -> u32 {
...
+/// `Duration` or the `DateTime` are too big to represented as nanoseconds. They
+/// will also fail if the `Duration` is bigger than the timestamp.
+pub trait DurationRound: Sized {
+    /// Error that can occur in rounding or truncating
+    #[cfg(any(feature = "std", test))]
+    type Err: std::error::Error;
+
+    /// Error that can occur in rounding or truncating
+    #[cfg(not(any(feature = "std", test)))]
+    type Err;
```
quodlibetor, line 109: I think it's probably worth adding a trait bound here of `Debug + Display`, [anyhow uses that bound in no std](https://github.com/dtolnay/anyhow/blob/1089f316136a9718b6e9b5943772822958cd268d/src/error.rs#L70) so it seems not insane. We can compatibly remove trait bounds down the line, but not add them, so I'd rather be as conservative as possible here, but I'm not familiar with no_std in general.

src/round.rs
```diff
@@ -176,4 +324,113 @@ mod tests {
         assert_eq!(dt.trunc_subsecs(0).nanosecond(), 1_000_000_000);
         assert_eq!(dt.trunc_subsecs(0).second(), 59);
     }
+
+    #[test]
+    fn test_duration_round() {
+        let dt = Utc.ymd(2016, 12, 31).and_hms_nano(23, 59, 59, 1_75_500_000);
```
quodlibetor, line 330: `1_75_500_000` is a bit of a weird spacing? Is it intentional?

src/round.rs
```diff
@@ -176,4 +324,113 @@ mod tests {
...
+
+        assert_eq!(
+            dt.duration_round(Duration::milliseconds(10)).unwrap().to_string(),
+            "2016-12-31 23:59:59.180 UTC"
+        );
+
+        let dt = Utc.ymd(2012, 12, 12).and_hms(18, 24, 12);
```
quodlibetor, line 337: Could you add some DTs that exactly test the borders? e.g. `Utc.ymd(2012, 12, 12).and_hms_milis(12, 30, 30, 500)` and `Utc.ymd(2012, 12, 12).and_hms_milis(11, 29, 29, 499)`?

src/round.rs
```diff
@@ -85,14 +92,155 @@ fn span_for_digits(digits: u16) -> u32 {
...
+}
+
+/// An error from rounding by `Duration`
+///
+/// See: [`DurationRound`]
+#[derive(Debug, Clone, PartialEq, Eq, Copy)]
+pub enum RoundingError {
+    DurationExceedsTimestamp,
+    DurationExceedsLimit,
+    TimestampExceedsLimit,
```
quodlibetor, line 209: Could you add docs/examples for these? You can feel free to move the tests for them up into the docstrings.

src/round.rs
robyoung: whoops, no, typo

quodlibetor:
Thanks!

Tom1380:
So how do you truncate `NaiveDateTime`? Looking at the docs, I only see `DateTime<Tz>` as an implementor for `DurationRound`.

PR #452, opened 2020-07-10
Add MIN and MAX DateTime, Time and NaiveDateTime

### LINKED ISSUE #386: Add MIN_DATETIME and MAX_DATETIME
Also add the naive versions. These would just be useful to have for completeness in the library.

robyoung:
Resolves #386

### DIFF
--- a/src/naive/time.rs
+++ b/src/naive/time.rs
@@ -16,6 +16,9 @@ use format::{parse, ParseError, ParseResult, Parsed, StrftimeItems};
 use format::{Fixed, Item, Numeric, Pad};
 use Timelike;
 
+pub const MIN_TIME: NaiveTime = NaiveTime { secs: 0, frac: 0 };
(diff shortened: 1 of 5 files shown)


src/naive/time.rs
```diff
@@ -16,6 +16,9 @@ use format::{parse, ParseError, ParseResult, Parsed, StrftimeItems};
 use format::{Fixed, Item, Numeric, Pad};
 use Timelike;
 
+pub const MIN_TIME: NaiveTime = NaiveTime { secs: 0, frac: 0 };
+pub const MAX_TIME: NaiveTime = NaiveTime { secs: 23 * 3600 + 59 * 60 + 59, frac: 1_999_999_999 };
```
quodlibetor, line 20: I think that this frac should be just `999_999_999` since it is a bit unlikely (but not impossible) that the very last second will be a leap second.

robyoung: Ah, I see, thanks.

quodlibetor:
Thanks!

PR #453, opened 2020-07-11
Locales

### LINKED ISSUE #153: Add support for locales in format
Hello.
It would be nice to be able to format a date in other languages than english.
Thanks to add this feature.

### LINKED ISSUE #199: Localized date/time
I would be nice to have a ability to get localized date/time. For example in C++ std::strftime function has format %x/%X that displays date/time in format appropriate to current locale.
P.S. If it's there already sorry for creating this issue I couldn't find it.

cecton:
Closes #199
Closes #153 

I made a pure Rust lib without any dependency that has the locales from glibc. It does not depend on glibc. It's using the locale data files from glibc.

```
% This file is part of the GNU C Library and contains locale data.
% The Free Software Foundation does not claim any copyright interest
% in the locale data contained in this file.  The foregoing does not
% affect the license of the GNU C Library as a whole.  It does not
% exempt you from the conditions of the license if your use would
% otherwise be governed by that license.
```

I contacted the FSF and they seem to allow the use of these data:

> In short, we believe these files aren't copyrightable; for more background,
> please check https://sourceware.org/ml/libc-locales/2013-q1/msg00048.html
>
> --
> I am not a lawyer, this is not a legal advice.

The PR is a draft, I'm waiting for your feedback before finishing it.


Cargo.toml
```diff
@@ -24,11 +24,12 @@ appveyor = { repository = "chronotope/chrono" }
 name = "chrono"
 
 [features]
-default = ["clock", "std"]
+default = ["clock", "std", "locales"]
```
quodlibetor, line 27: I don't know that this should be a default feature, but I _do_ think that it should be called out in the README as a major feature, there isn't a clear section for features right now, I'll add one and make sure that this is prominent.

Cargo.toml
```diff
@@ -38,6 +39,7 @@ num-integer = { version = "0.1.36", default-features = false }
 num-traits = { version = "0.2", default-features = false }
 rustc-serialize = { version = "0.3.20", optional = true }
 serde = { version = "1.0.99", default-features = false, optional = true }
+pure-rust-locales = "0.2.1"
```
quodlibetor, line 42: this needs to be `optional = true` and `locals = ["pure-rust-locals"]` above in order for everything to work correctly.

src/format/mod.rs
```diff
@@ -386,10 +437,12 @@ pub fn format_item<'a>(
     time: Option<&NaiveTime>,
     off: Option<&(String, FixedOffset)>,
     item: &Item<'a>,
-) -> fmt::Result {
+    locale: &str,
+) -> FormatResult<()> {
```
quodlibetor, line 441: we should create a new `fn format_item_localized` function to preserve backwards compatibility. I also think that for backcompat the default locale should be `"en_US"` or `"en_GB"`.

That also allows us to return the improved `FormatResult` from that function, but not change the error type here (which is also backwards-incompatible)

quodlibetor:
Also, unrelated to my willingness to merge this (I am happy about it!), I glanced at `pure-rust-locales` and it depends on `nom` at build time, which will unfortunately, I believe, meaningfully contribute to compilation time even though it doesn't affect final library size. It is a little complicated whether that counts as "without any dependency".

cecton:
> I glanced at `pure-rust-locales` and it depends on `nom` at build time, which will unfortunately, I believe, meaningfully contribute to compilation time even though it doesn't affect final library size. It is a little complicated whether that counts as "without any dependency".

Thanks! :grin: Yes I have a bit exaggerated on that. The data file for the locales come directly from glibc and I made `pure-rust-locales` for the sole purpose of offering these `const` with the locales. Therefore the data files are converted to a rust file during the build but it doesn't add nom as a dependency of the whole project (only as a build dependency).

The only faster+lighter alternative would be to publish the library with the file already generated. I'm not sure if there is an automatic way to do that...

> I'm not super familiar with localization in general. From looking at other programming languages it does seem standard to hook this directly into the datetime formatting, instead of providing an api that something like fluent could hook into. Honestly I have no idea what that would look like, do you have thoughts? Is it worth us pinging them to ask for input?

I don't know if fluent is going to provide a date and time formatting function. Usually those libraries are more focused on the translations. But I'm not sure, it's worth to ask. I know for example that in Python the date and time formatting depends on libc. I actually made a Rust crate specialized for that: https://crates.io/crates/libc-strftime (doing the same for chrono would be annoying as it adds a dependency to libc and the behavior is not consistent across all platforms).

The approach I have here is very minimalist and will allow the user to either use a default locale implementation consistent across platforms or using fluent (or anything specialized) instead. I believe I can even let the user select at compilation time what languages they want to support which should reduce the size of the binary.

src/format/mod.rs
cecton: I think we should actually use POSIX for the default https://docs.rs/pure-rust-locales/0.2.1/pure_rust_locales/POSIX/LC_TIME/index.html ... but maybe it doesn't change anything with en_US https://docs.rs/pure-rust-locales/0.2.1/pure_rust_locales/en_US/LC_TIME/index.html

quodlibetor: Oh yes POSIX makes more sense! Aggreed.

quodlibetor:
I agree with your points! I'll be happy to merge this with my requests applied.

cecton:
```
 error: expected identifier, found keyword `crate`
 --> /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/ryu-1.0.5/src/buffer/mod.rs:1:5
  |
1 | use crate::raw;
  |     ^^^^^
```
@quodlibetor the macos build is failing... do you have a requirement to support older versions of Rust?

quodlibetor:
> do you have a requirement to support older versions of Rust?

Yes, but not with all features: it is expected to not try to build or test the `locales` feature on 1.13.0.

quodlibetor:
In terms of 1.13.0, if you fix [this comment](https://github.com/chronotope/chrono/pull/453/files#r453217437) the tests should not try to build on 1.13.0 (there's an explicit allowlist of features that we test on 1.13.0 in `ci/github.sh`)

src/format/mod.rs
```diff
@@ -399,7 +460,7 @@ fn format_inner<'a>(
     time: Option<&NaiveTime>,
     off: Option<&(String, FixedOffset)>,
     item: &Item<'a>,
-) -> fmt::Result {
+) -> FormatResult<()> {
```
cecton, line 463: I changed it for consistency with the localized function but that breaks the API. I'm not sure what is best

cecton: nah there is way too much duplicated code, I will correct this

src/format/mod.rs
cecton: Done! More or less... I did duplicate some function but I broke the compatibility since I changed the result `fmt::Result` to `FormatResult`

Cargo.toml
cecton: done

quodlibetor:
Are you waiting on me for further action on this one @cecton , or is it still wip?

cecton:
@quodlibetor ah! I was going to say I would like your feedback already on what is done thanks

Cargo.toml
```diff
@@ -50,6 +52,7 @@ bincode = { version = "0.8.0" }
 num-iter = { version = "0.1.35", default-features = false }
 criterion = { version = "0.3" }
 doc-comment = { version = "0.3" }
+pure-rust-locales = { version = "0.2.1" }
```
quodlibetor: I don't think that this is necessary here, if chrono fails to build without the `locales` feature then that probably means that it's not correctly configured out (missing some `#[cfg(locales)]` functionality).

src/format/mod.rs
```diff
@@ -386,10 +446,11 @@ pub fn format_item<'a>(
...
```
quodlibetor, line 449: This should continue returning a regular `fmt::Result`

src/date.rs
```diff
@@ -10,6 +10,8 @@ use core::ops::{Add, Sub};
 use core::{fmt, hash};
 use oldtime::Duration as OldDuration;
 
+#[cfg(any(all(feature = "locales", any(feature = "alloc", feature = "std")), test))]
+use format::DelayedFormatLocalized;
```
quodlibetor, line 14: This should just be `#[cfg(feature = "locales")]`, anything else probably suggests that something is awry.

src/format/mod.rs
```diff
@@ -719,6 +796,81 @@ impl<'a, I: Iterator<Item = B> + Clone, B: Borrow<Item<'a>>> DelayedFormat<I> {
...
         format(f, self.date.as_ref(), self.time.as_ref(), self.off.as_ref(), self.items.clone())
+            .map_err(|_| fmt::Error)
+    }
+}
+
+/// A *temporary* object which can be used as an argument to `format!` or others.
+/// This is normally constructed via `format` methods of each date and time type.
+#[cfg(all(feature = "locales", any(feature = "alloc", feature = "std", test)))]
+#[derive(Debug)]
+pub struct DelayedFormatLocalized<'a, I> {
```
quodlibetor, line 807: I believe that you don't need to create this struct, you can just create a `Option<local>` field in the existing DelayedFormat and default it to `None` in the existing constructors, adding the new constructors under the feature flag.

src/format/mod.rs
```diff
@@ -719,6 +796,81 @@ impl<'a, I: Iterator<Item = B> + Clone, B: Borrow<Item<'a>>> DelayedFormat<I> {
...
+    /// The date view, if any.
+    date: Option<NaiveDate>,
+    /// The time view, if any.
+    time: Option<NaiveTime>,
+    /// The name and local-to-UTC difference for the offset (timezone), if any.
+    off: Option<(String, FixedOffset)>,
+    /// An iterator returning formatting items.
+    items: I,
+    /// Locale used for text.
+    locale: &'a str,
```
quodlibetor, line 817: Something that might be really nice would be to create an `enum Locale` in `pure-rust-locales`, that would remove the lifetime here, it could also be the target of the `match` that I earlier said might be better in `pure-rust-locales`

src/format/mod.rs
```diff
@@ -378,6 +381,63 @@ const TOO_SHORT: ParseError = ParseError(ParseErrorKind::TooShort);
...
+/// The category of format error.
+#[cfg(any(feature = "alloc", feature = "std", test))]
+#[derive(Debug, Clone, PartialEq, Eq, Copy)]
+enum FormatErrorKind {
+    /// The locale is unknown or not available.
+    #[cfg(feature = "locales")]
+    UnknownLocale,
+
+    /// Format error.
+    Format,
```
quodlibetor, line 398: This should definitely include the source/cause format error so that we don't lose that.

src/format/locales.rs
```diff
@@ -0,0 +1,361 @@
+use super::{FormatError, FormatErrorKind, FormatResult};
+
+macro_rules! locale_match {
```
quodlibetor, line 3: I think it would be nice if this function lived in pure-rust-locales instead of here.

src/format/mod.rs
quodlibetor: since `format_inner` is not pub its API can change, sorry if I said something incorrect about that.

A new function `pub format_item_localized` and possibly localized variants of the functions that call `format_item` will be necessary.

Cargo.toml
cecton: yes sorry that was a leftover, fixed

quodlibetor:
ah you did a fair amount while I was reviewing. not sure if I responded to old code, feel free to ping me when it's stabilized a bit.

src/format/mod.rs
cecton: There will be duplicated code if we do that. Or I can use fmt::Result everywhere but then we have no clear error message when the locale that was provided is unknown or not. Are you ok with that?

quodlibetor: it's fine to have a bit of duplicated code, but if the feature set is the same shouldn't it be guaranteed that the only possible error in `format_inner` is the `FormatResult::Format`? The other error variants can only come from locale handling?

src/date.rs
cecton: I copied DelayFormatLocalized from DelayFormat which has all of this. I just added the condition.

https://github.com/chronotope/chrono/blob/febb8dc168325ac471b54591c925c48b6a485962/src/format/mod.rs#L683-L694

I can put it only on locales if you think it's correct (I'm really not sure)

src/format/mod.rs
cecton: No actually you have a few more errors that are not fmt::Error:
https://github.com/chronotope/chrono/blob/febb8dc168325ac471b54591c925c48b6a485962/src/format/mod.rs#L492
https://github.com/chronotope/chrono/blob/febb8dc168325ac471b54591c925c48b6a485962/src/format/mod.rs#L643
https://github.com/chronotope/chrono/blob/febb8dc168325ac471b54591c925c48b6a485962/src/format/mod.rs#L647

Since I made a Locale enum as you suggested, the local handling is not done in chrono anymore so it is infallible (well, at least here but I think we still should re-export this Locale enum). I reverted the code to fmt::Error.

(I will continue tomorrow, I have not done the re-export yet)

src/format/mod.rs
cecton: This has been reverted

src/format/locales.rs
cecton: Done

src/format/mod.rs
```diff
@@ -38,9 +38,18 @@ use offset::{FixedOffset, Offset};
...
+
 pub use self::parse::parse;
 pub use self::parsed::Parsed;
 pub use self::strftime::StrftimeItems;
+#[cfg(all(feature = "locales", any(feature = "alloc", feature = "std", test)))]
+pub use pure_rust_locales::Locale;
+
+#[cfg(not(feature = "locales"))]
+#[allow(dead_code)]
+struct Locale;
```
cecton, line 52:
```suggestion
#[cfg(not(feature = "locales"))]
#[allow(dead_code)]
#[derive(Debug)]
struct Locale;
```

cecton:
@quodlibetor can you have a look again? The lint is failing but I doubt it is related to my changes

src/format/mod.rs
cecton: done

src/format/mod.rs
cecton: done

src/format/mod.rs
```diff
@@ -665,9 +691,10 @@ where
 {
     let mut result = String::new();
     for item in items {
-        format_inner(&mut result, date, time, off, item.borrow())?;
+        format_inner(&mut result, date, time, off, item.borrow(), None)?;
     }
-    w.pad(&result)
+    w.pad(&result)?;
+    Ok(())
```
cecton, line 697:
```suggestion
    w.pad(&result)
```

src/format/mod.rs
```diff
@@ -711,14 +739,54 @@ impl<'a, I: Iterator<Item = B> + Clone, B: Borrow<Item<'a>>> DelayedFormat<I> {
...
+            locale: Some(locale),
+        }
     }
 }
 
 #[cfg(any(feature = "alloc", feature = "std", test))]
 impl<'a, I: Iterator<Item = B> + Clone, B: Borrow<Item<'a>>> fmt::Display for DelayedFormat<I> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         format(f, self.date.as_ref(), self.time.as_ref(), self.off.as_ref(), self.items.clone())
+            .map_err(|_| fmt::Error)
```
cecton, line 789:
```suggestion
```

src/format/mod.rs
```diff
@@ -759,3 +826,42 @@ impl FromStr for Weekday {
...
+    date: Option<&NaiveDate>,
+    time: Option<&NaiveTime>,
+    off: Option<&(String, FixedOffset)>,
+    item: &Item<'a>,
+    locale: Locale,
+) -> fmt::Result {
+    let mut result = String::new();
+    format_inner(&mut result, date, time, off, item, Some(locale))?;
+    w.pad(&result)?;
+    Ok(())
```
cecton, line 843:
```suggestion
    w.pad(&result)
```

src/format/mod.rs
```diff
@@ -759,3 +826,42 @@ impl FromStr for Weekday {
...
+where
+    I: Iterator<Item = B> + Clone,
+    B: Borrow<Item<'a>>,
+{
+    let mut result = String::new();
+    for item in items {
+        format_inner(&mut result, date, time, off, item.borrow(), Some(locale))?;
+    }
+    w.pad(&result)?;
+    Ok(())
```
cecton, line 866:
```suggestion
    w.pad(&result)
```

README.md
```diff
@@ -220,12 +220,21 @@ Chrono also provides [`to_rfc2822`](https://docs.rs/chrono/0.4/chrono/struct.Dat
...
 for well-known formats.
 
+Chrono now also provides date formatting in almost any language without the
+help of an additional C library. This functionality is under the feature
+`locales`:
+
+```
+chrono { version = ..., features = ["locales"]
+```
+
```
cecton, line 230: I have added some doc...

README.md
```diff
@@ -220,12 +220,21 @@ Chrono also provides [`to_rfc2822`](https://docs.rs/chrono/0.4/chrono/struct.Dat
...
 ```rust
 use chrono::prelude::*;
 
 let dt = Utc.ymd(2014, 11, 28).and_hms(12, 0, 9);
 assert_eq!(dt.format("%Y-%m-%d %H:%M:%S").to_string(), "2014-11-28 12:00:09");
 assert_eq!(dt.format("%a %b %e %T %Y").to_string(), "Fri Nov 28 12:00:09 2014");
+assert_eq!(dt.format_localized("%A %e %B %Y, %T", Locale::fr_BE).to_string(), "vendredi 28 novembre 2014, 12:00:09");
```
cecton, line 237: ...and an example

src/format/mod.rs
```diff
@@ -388,8 +398,9 @@ pub fn format_item<'a>(
     item: &Item<'a>,
 ) -> fmt::Result {
     let mut result = String::new();
-    format_inner(&mut result, date, time, off, item)?;
-    w.pad(&result)
+    format_inner(&mut result, date, time, off, item, None)?;
+    w.pad(&result)?;
+    Ok(())
```
cecton, line 403:
```suggestion
    w.pad(&result)
```

src/format/strftime.rs
```diff
@@ -518,3 +518,42 @@ fn test_strftime_docs() {
...
+    // date specifiers
+    assert_eq!(dt.format_localized("%b", Locale::fr_BE).to_string(), "jui");
+    assert_eq!(dt.format_localized("%B", Locale::fr_BE).to_string(), "juillet");
+    assert_eq!(dt.format_localized("%h", Locale::fr_BE).to_string(), "jui");
+    assert_eq!(dt.format_localized("%a", Locale::fr_BE).to_string(), "dim");
+    assert_eq!(dt.format_localized("%A", Locale::fr_BE).to_string(), "dimanche");
+    assert_eq!(dt.format_localized("%D", Locale::fr_BE).to_string(), "07/08/01"); // TODO
+    assert_eq!(dt.format_localized("%x", Locale::fr_BE).to_string(), "07/08/01"); // TODO
+    assert_eq!(dt.format_localized("%F", Locale::fr_BE).to_string(), "2001-07-08"); // TODO
+    assert_eq!(dt.format_localized("%v", Locale::fr_BE).to_string(), " 8-jui-2001"); // TODO
```
cecton, line 539: While writing the test I realized there are still a lot of things wrong. Not all languages format their date and time the same way than in english. Those should use D_FMT, T_FMT and D_T_FMT https://docs.rs/pure-rust-locales/0.5.2/pure_rust_locales/POSIX/LC_TIME/index.html

ci/lint.sh
```diff
@@ -7,7 +7,7 @@ source "${BASH_SOURCE[0]%/*}/_shlib.sh"
 
 main() {
     # rustflags are present because of: https://github.com/rust-lang/rust-clippy/issues/5749
-    runt env RUSTFLAGS="-Dwarnings" cargo clippy --color=always -- -D warnings
+    runt env RUSTFLAGS="-Dwarnings" cargo clippy --color=always -- -D warnings -A clippy::manual-non-exhaustive
```
cecton, line 10: I had to add this because I'm getting this error:
```
error: this seems like a manual implementation of the non-exhaustive pattern
  --> src/datetime.rs:38:1
   |
38 |   pub enum SecondsFormat {
   |   ^---------------------
   |   |
   |  _help: add the attribute: `#[non_exhaustive] pub enum SecondsFormat`
   | |
39 | |     /// Format whole seconds only, with no decimal point nor subseconds.
40 | |     Secs,
41 | |
...  |
61 | |     __NonExhaustive,
62 | | }
   | |_^
   |
   = note: `-D clippy::manual-non-exhaustive` implied by `-D warnings`
help: remove this variant
  --> src/datetime.rs:61:5
   |
61 |     __NonExhaustive,
   |     ^^^^^^^^^^^^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_non_exhaustive

error: this seems like a manual implementation of the non-exhaustive pattern
   --> src/format/parsed.rs:27:1
    |
27  |   pub struct Parsed {
    |   ^----------------
    |   |
    |  _help: add the attribute: `#[non_exhaustive] pub struct Parsed`
    | |
28  | |     /// Year.
29  | |     ///
30  | |     /// This can be negative unlike [`year_div_100`](#structfield.year_div_100)
...   |
110 | |     _dummy: (),
111 | | }
    | |_^
    |
help: remove this field
   --> src/format/parsed.rs:110:5
    |
110 |     _dummy: (),
    |     ^^^^^^^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_non_exhaustive
```

This is not related to my change. It's just a new rule of clippy in nightly (probably added yesterday).

src/date.rs
cecton: I believe this is really necessary as the feature locales won't work without alloc or std. If we reduce to only `feature = "locales"` it means that building with `--no-default-features --features locales` will fail with non really descriptive error messages

cecton:
rust 1.13 + nostd + clippy = :fire: :fire: :fire: 

I had to add a lot of compiler exceptions everywhere. It felt like doing C programming

cecton:
@quodlibetor hey o/ OK NOW I believe it is purrfect. CI is green, all things are translated, backward compatibility is preserved, doc and tests have been added... let me know if there is still anything you would like to change.

src/format/strftime.rs
```diff
@@ -518,3 +641,43 @@ fn test_strftime_docs() {
...
+    // time specifiers
+    assert_eq!(dt.format_localized("%P", Locale::fr_BE).to_string(), "");
+    assert_eq!(dt.format_localized("%p", Locale::fr_BE).to_string(), "");
+    assert_eq!(
+        dt.with_nanosecond(1_026_490_000)
+            .unwrap()
+            .format_localized("%.f", Locale::fr_BE)
+            .to_string(),
+        ".026490"
+    );
```
cecton, line 672:
```suggestion
```

quodlibetor:
Thanks! I'll look at this as soon as I can, but I might be busy for a few days.

src/datetime.rs
```diff
@@ -493,6 +495,41 @@ where
...
+            &self.offset,
+            items,
+            locale,
+        )
+    }
+
+    /// Formats the combined date and time with the specified format string and locale.
+    /// See the [`format::strftime` module](./format/strftime/index.html)
+    /// on the supported escape sequences.
+    #[cfg(all(feature = "locales", any(feature = "alloc", feature = "std", test)))]
```
quodlibetor, line 524: same regarding `cfg`

README.md
quodlibetor: nice!

src/format/mod.rs
```diff
@@ -38,9 +38,19 @@ use offset::{FixedOffset, Offset};
 use {Datelike, Timelike};
 use {ParseWeekdayError, Weekday};
 
+#[cfg(all(feature = "locales", any(feature = "alloc", feature = "std", test)))]
+pub(crate) mod locales;
```
quodlibetor, line 42: huh, I didn't realize that we could use new syntax if it was cfg'd out and not in a different module, nice. (`pub(crate)` was [stabilized in 1.18.0](https://github.com/rust-lang/rust/blob/stable/RELEASES.md#version-1180-2017-06-08))

src/format/mod.rs
```diff
@@ -38,9 +38,19 @@ use offset::{FixedOffset, Offset};
...
+
 pub use self::parse::parse;
 pub use self::parsed::Parsed;
 pub use self::strftime::StrftimeItems;
+#[cfg(all(feature = "locales", any(feature = "alloc", feature = "std", test)))]
+pub use pure_rust_locales::Locale;
+
+#[cfg(not(feature = "locales"))]
+#[allow(dead_code)]
```
quodlibetor, line 51: why is this `allow(dead_code)` necessary?

src/format/strftime.rs
```diff
@@ -38,7 +38,7 @@ The following specifiers are available both to formatting and parsing.
 | `%j`  | `189`    | Day of the year (001--366), zero-padded to 3 digits.                       |
 |       |          |                                                                            |
 | `%D`  | `07/08/01`    | Month-day-year format. Same as `%m/%d/%y`.                            |
-| `%x`  | `07/08/01`    | Same as `%D`.                                                         |
+| `%x`  | `07/08/01`    | Locale's date representation (e.g., 12/31/99).                        |
```
quodlibetor, line 41: Oh darn why did we ever support this incorrectly 😢 

For all three of the ones that you changed the output of, I think we need to add a new format specifier `%#x` (and below `%#X` `%#c`) for the new Localized format, see e.g. `z` for the way that this is handled, it shouldn't be much more code:

https://github.com/chronotope/chrono/blob/130253ef2b914d300c1a0c6cff36cd3a5d846767/src/format/strftime.rs#L306-L312

README.md
cecton: Merci!

src/format/mod.rs
cecton: Yes! :sweat_smile: working around the 1.18 was really tricky sometimes. The CI works great fortunately

src/format/strftime.rs
cecton: :scream: this would make things incompatible with glibc's format. I don't think that's a big deal though... :thinking: 

(Note that if the feature "locales" is not used, the original behavior is kept. I don't know if that makes it more okay with the compatibility you seek)

Cargo.toml
cecton: done

README.md
```diff
@@ -220,12 +220,21 @@ Chrono also provides [`to_rfc2822`](https://docs.rs/chrono/0.4/chrono/struct.Dat
...
 for well-known formats.
 
+Chrono now also provides date formatting in almost any language without the
+help of an additional C library. This functionality is under the feature
+`locales`:
+
+```text
+chrono { version = ..., features = ["locales"]
+```
+
```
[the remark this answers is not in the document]
cecton: done (I also replaced the ... by "0.4")

Cargo.toml
```diff
@@ -29,6 +29,7 @@ alloc = []
 std = []
 clock = ["time", "std"]
 wasmbind = ["wasm-bindgen", "js-sys"]
+locales = ["pure-rust-locales"]
```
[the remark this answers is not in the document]
cecton: done

src/date.rs
```diff
@@ -10,6 +10,8 @@ use core::ops::{Add, Sub};
 use core::{fmt, hash};
 use oldtime::Duration as OldDuration;
 
+#[cfg(all(feature = "locales", any(feature = "alloc", feature = "std", test)))]
+use format::Locale;
```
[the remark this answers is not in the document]
cecton: done

src/date.rs
```diff
@@ -295,6 +297,40 @@ where
     pub fn format<'a>(&self, fmt: &'a str) -> DelayedFormat<StrftimeItems<'a>> {
         self.format_with_items(StrftimeItems::new(fmt))
     }
+
+    /// Formats the date with the specified formatting items and locale.
+    #[cfg(all(feature = "locales", any(feature = "alloc", feature = "std", test)))]
```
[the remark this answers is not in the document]
cecton: done

src/datetime.rs
```diff
@@ -19,6 +19,8 @@ use std::string::ToString;
 use core::borrow::Borrow;
 #[cfg(any(feature = "alloc", feature = "std", test))]
 use format::DelayedFormat;
+#[cfg(all(feature = "locales", any(feature = "alloc", feature = "std", test)))]
```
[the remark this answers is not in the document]
cecton: done

src/datetime.rs
```diff
@@ -493,6 +495,41 @@ where
     pub fn format<'a>(&self, fmt: &'a str) -> DelayedFormat<StrftimeItems<'a>> {
         self.format_with_items(StrftimeItems::new(fmt))
     }
+
+    /// Formats the combined date and time with the specified formatting items and locale.
+    #[cfg(all(feature = "locales", any(feature = "alloc", feature = "std", test)))]
```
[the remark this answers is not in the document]
cecton: done

src/datetime.rs
cecton: done

src/format/strftime.rs
quodlibetor: Oh that's a great point!

Yes, could you just change the documentation string to say something like:

> With the `locales` feature this is the Locale's date representation, otherwise same as...

src/format/mod.rs
cecton: No. It's probably with the long `#[cfg]` I had that there was an issue

cecton: done

src/format/strftime.rs
cecton: done

cecton:
@quodlibetor ready for review :grin:

src/format/strftime.rs
cecton: Oh! I just saw your message. Nevermind then, I will revert to the other state and update the doc as you said (I like it better personally!)

src/format/strftime.rs
```diff
@@ -227,9 +314,16 @@ impl<'a> Iterator for StrftimeItems<'a> {
                 }
 
                 macro_rules! recons {
-                    [$head:expr, $($tail:expr),+] => ({
-                        const RECONS: &'static [Item<'static>] = &[$($tail),+];
-                        self.recons = RECONS;
+                    [$head:expr, $($tail:expr),+ $(,)*] => ({
+                        #[cfg(feature = "locales")]
+                        {
+                            self.recons = vec![$($tail),+];
```
quodlibetor, line 320: We shouldn't stick a `vec` call in here, it will cause a new vec to be allocated every iteration of the parse loop, it should be possible to do the same trick as below? Actually I'm not sure why this branching on whether the locales feature is enabled is required?

src/format/strftime.rs
```diff
@@ -169,18 +174,93 @@ pub struct StrftimeItems<'a> {
     /// If the current specifier is composed of multiple formatting items (e.g. `%+`),
     /// parser refers to the statically reconstructed slice of them.
     /// If `recons` is not empty they have to be returned earlier than the `remainder`.
+    #[cfg(feature = "locales")]
+    recons: Vec<Item<'a>>,
+    #[cfg(not(feature = "locales"))]
     recons: &'static [Item<'static>],
```
quodlibetor, line 180: why does this need to be dynamically sized now? 

Ideally that shouldn't happen, it will cause a new vec to be allocated every iteration of the parse loop. If the vecs are definitely required then we need to find a way to re-use them, instead of creating new ones. It would be better to do `self.recons.clear(); self.recons.extend(...);` in all the places below where new ones are created. 

But I would like to be sure that that's actually necessary, it doesn't seem like it should be but I could definitely be missing something!

quodlibetor:
I added documentation of features to `lib.rs`, if you feel like documenting `locales` in that section you should feel free, otherwise I'll add it after merge.

src/format/strftime.rs
cecton: haha :grin: I don't spontaneously think from that perspective :thinking: but I guess you're right, there are a lot of allocations in there.

It's for the date, time and date&time formats. They are dependent on the locale. Example [here](https://docs.rs/pure-rust-locales/0.5.2/pure_rust_locales/en_GB/LC_TIME/constant.D_FMT.html) in British UK where the day is first, then the month, then the year). And yes it is absolutely necessary.

These strings needs to be parsed first, which I do here: https://github.com/chronotope/chrono/blob/ea3d20e11e5c3e0420a04d9b88578aa6944bf08a/src/format/strftime.rs#L249-L251

Though these are constants so maybe there is a way to load them statically? Like with lazy_static? Also the length of the parsed iterator may vary per locale. I'm not sure this is a great idea though, it's a lot of memory used, there are a lot of locales.

cecton: I think I got an idea of an alternative solution after looking at other crates.

Maybe we can split chrono into 2 crates: 1 for internals, 1 for chrono itself (or make a fake internals crate like serde does). Then I can probably build new `const` that use the internal representation of chrono  (`Item` and the other internal objects of chrono) using build.rs.

This would be like a conversion from the format of pure-rust-locales to chrono. Since they are const and they would be used only once, there would be no impact on memory and minimal impact on the binary size (correct me if I'm wrong).

src/format/strftime.rs
```diff
@@ -169,18 +174,93 @@ pub struct StrftimeItems<'a> {
...
+        #[cfg(feature = "locales")]
+        {
+            StrftimeItems {
+                remainder: s,
+                recons: Vec::new(),
+                d_fmt: D_FMT.to_vec(),
+                d_t_fmt: D_T_FMT.to_vec(),
+                t_fmt: T_FMT.to_vec(),
+            }
+        }
```
quodlibetor, line 232: I prefer `cfgs` to be out at the function level, could you make a pair of cfg-different private `StrftimeItems::with_remainder` functions?

src/format/strftime.rs
quodlibetor: > Then I can probably build new const that use the internal representation of chrono (Item and the other internal objects of chrono) using build.rs.
>
> This would be like a conversion from the format of pure-rust-locales to chrono. Since they are const and they would be used only once, there would be no impact on memory and minimal impact on the binary size (correct me if I'm wrong).

Is it possible to do this in an internal private module in chrono? I agree that it seems like this should work.

That said, I'd be happy to consider that a future optimization, if you feel like it we can land the change to chrono without that and follow up later. If we go that route I'd definitely appreciate an issue explaining what you think the right way to do it is, so that we know where to go.

That said, there are some things that can be done without that constification to reduce both code complexity and allocations:

* For being able to read the code, create a single type alias that is the only place that
  we need to do the `cfg(locales)` (not sure about the exact syntax for the lifetime off
  the top of my head):

  ```rust
  #[cfg(feature = "locales")]
  type Fmt<'a> = Vec<Item<'a>>;
  #[cfg(not(feature = "locales"))]
  Type Fmt = &'static [Item<'static>]
  ```

  which allows the `StrftimeItems` to not have any cfg's inside of it.

* Maybe make a pair of cfg-alterate functions that generate the `Fmt` types, instead of
  needing to put the `cfg` in so many of the functions.
* That also makes it easier to centralize the logic of replacing the vec, so that it's
  either a strict overwrite or a clear + extended.

cecton: > Is it possible to do this in an internal private module in chrono? I agree that it seems like this should work.

I believe it is **possible**. I gave a quick try to see how it goes and it seems it's going to be difficult: I didn't realize this but the format parser (parsing "%Y %d %m" to internal objects) is in the same module than parsing the date and time according to a format (`Utc.datetime_from_str("Fri Nov 28 12:00:09", "%a %b %e %T %Y"`). I'm not sure how to circumvent this issue without moving the modules and the components around... especially with the 1.13.

> I'd be happy to consider that a future optimization, if you feel like it we can land the change to chrono without that and follow up later.

If we go for the Vec for now we can release this as a new feature.

I really want to do the consts but I think the change is going to be intense (but it is definitely possible). If you think the implementation with clear&extend is good for now, then I will go for that for now so people can already enjoy the localization.

src/format/strftime.rs
cecton: done

src/format/strftime.rs
cecton: done

src/format/strftime.rs
cecton: done!

I moved the common statics to the module level to avoid duplication.

cecton:
@quodlibetor this is ready for review

quodlibetor:
This looks great! I've updated the feature name to `unstable-locales` until we have more experience with it. Hopefully this will make it easier to update the API in pure-rust-locales without forcing chrono to issue a breaking change.

Relatedly, how likely do you think it is that pure-rust-locales will require further breaking changes? It looks like it's had several, which makes sense, but do you have a sense on if it's converging on something that feels future-proof or is more experimentation expected? No real rush with the feature named `unstable-locales`, but releasing a 1.0 version will make me more comfortable renaming the feature to just `locales`.

quodlibetor:
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 

Thank you so much for this!

ammgws:
On phone at the moment but if I wanted to use the system's locale (i.e. LC_TIME from /etc/locale.conf), I would need to do the work to grab that info separately to use with the new `_localized` functions?

quodlibetor:
> if I wanted to use the system's locale (i.e. LC_TIME from /etc/locale.conf), I would need to do the work to grab that info separately to use with the new _localized functions?

Correct. I'm open to PRs (especially cross-platform PRs) that make that easier, but _probably_ the right place for that functionality would be in `pure-rust-locales`.

cecton:
> On phone at the moment but if I wanted to use the system's locale (i.e. LC_TIME from /etc/locale.conf), I would need to do the work to grab that info separately to use with the new `_localized` functions?

@ammgws Yes I'm afraid.

This is something that I could add in pure-rust-locales but I bet it will vary a lot per platform and environment. It would be better to do it in another crate.

The strong advantage of importing the locale data as we did is to not depend on libc, so you can have locales even with musl. (Also it's 100% consistent across platforms obviously.)

But the big inconvenient is that it is totally separate from existing mechanisms for localization. So yeah if you want to use the locales of the process user you need to get them first.

On the bright side, you can convert a &str like "fr_BE" to Locale:fr_BE (and all the others) easily.

cecton:
> Relatedly, how likely do you think it is that pure-rust-locales will require further breaking changes? It looks like it's had several, which makes sense, but do you have a sense on if it's converging on something that feels future-proof or is more experimentation expected? No real rush with the feature named `unstable-locales`, but releasing a 1.0 version will make me more comfortable renaming the feature to just `locales`.

Initially it should have been a single commit. Most of the changes come from its first use here in chrono. I imagine it is possible that others could ask new features but the format of the API shouldn't change. The data will also evolve a bit (I see that the last change is from 2019-05-08, so every year I guess).

ohk2kt3t4:
@cecton Do you have an example of deserialising a string to a Locale? The enum had try_from but not FromStr

cecton:
@ohk2kt3t4 hmmm :thinking: I should probably have use FromStr instead of TryFrom. I'm not sure what's the difference tbh. Do you have an issue with TryFrom?

The basic usage is:

```rust
use std::convert::TryFrom; // make sure this trait is in scope

let locale: Locale = "fr_BE".try_into()?;
```

If you have a String:

```rust
let locale_string: String = "fr_BE".to_string();
let locale: Locale = locale_string.as_str().try_into()?;
```

cecton:
If I understand correctly: FromStr is just the old one. I don't think I will make a release for such a minor issue. I'm not even sure it is an issue at all. The API of `from_str` is not more convenient (you still need FromStr in scope).

The feature `locales` of chrono is not even meant to be used on older versions of Rust. It requires stable.

If I make a release of pure-rust-locales I might just add it but I have really no reason to touch it for now. I'm open to any argument that would convince me otherwise.

leontoeides:
I was really excited to discover this `format_localized` feature. I didn't realize it existed. It appears only to be implemented for `DateTime` and not for `NaiveDateTime`, `NaiveDate`, `NaiveTime`. I'm not familiar with this source here but it looks like a matter of copying and pasting (?)

cecton:
> I was really excited to discover this `format_localized` feature. I didn't realize it existed. It appears only to be implemented for `DateTime` and not for `NaiveDateTime`, `NaiveDate`, `NaiveTime`. I'm not familiar with this source to chrono so I'm not qualified to make any changes, but it looks like a matter of copying and pasting with a few tweaks (?)

strftime does have notion of timezone. You can display it with `%Z`/`%z`.

If you do have a timezone unaware object (`Naive*`) and want to use strftime, you need to attach a timezone onto it even if you don't use it.

**I'm not a library maintainer** but I don't personally see much value into adding strftime to `Naive*` types because: 1. it's kinda breaking (what do you do when the user provides `%z`?) and 2. it can be easily workedaround by using the not naive types.

djc:
> I'm not familiar with this source to chrono so I'm not qualified to make any changes, but it looks like a matter of copying and pasting with a few tweaks (?)

This isn't quite how chrono development works. I'm happy to provide you with some pointers if you have a hard time making sense of the source code, but it is unlikely that I'll be motivated to go and make these changes myself.

(But @cecton makes a good point that it might not make sense anyway.)

PR #455, opened 2020-07-13
SergeyT: add additional fixed formatting items

sergeyt:
rebasing PR #409

TODO:
* [ ] fix build
* [ ] consider fixing backward compatibility proposed in PR #409 

---
This change is [Reviewable](https://reviewable.io/reviews/chronotope/chrono/455)

### DIFF
--- a/src/format/mod.rs
+++ b/src/format/mod.rs
@@ -168,10 +168,6 @@ pub enum Fixed {
     ///
     /// Prints a full name in the title case, reads either a short or full name in any case.
     LongMonthName,
-    /// One letter month names.
-    ///
-    /// Prints only first letter.
-    OneLetterMonthName,
     /// Abbreviated day of the week names.
     ///
     /// Prints a three-letter-long name in the title case, reads the same name in any case.
@@ -180,10 +176,6 @@ pub enum Fixed {
     ///
     /// Prints a full name in the title case, reads either a short or full name in any case.
     LongWeekdayName,
-    /// One letter day of the week names.
-    ///
-    /// Prints only first letter.
-    OneLetterWeekdayName,
     /// AM/PM.
     ///
     /// Prints in lower case, reads in any case.
@@ -192,14 +184,6 @@ pub enum Fixed {
     ///
     /// Prints in upper case, reads in any case.
     UpperAmPm,
-    /// a/p.
-    ///
-    /// Prints only first letter.
-    OneLetterLowerAmPm,
-    /// A/P.
-    ///
-    /// Prints only first letter.
-    OneLetterUpperAmPm,
     /// An optional dot plus one or more digits for left-aligned nanoseconds.
     /// May print nothing, 3, 6 or 9 digits according to the available accuracy.
     /// See also [`Numeric::Nanosecond`](./enum.Numeric.html#variant.Nanosecond).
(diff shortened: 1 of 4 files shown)


src/format/mod.rs
```diff
@@ -638,6 +672,41 @@ fn format_inner<'a>(
...
+                    time.map(|t| {
+                        result.push_str(if t.hour12().0 {"p"} else {"a"});
+                        Ok(())
+                    }),
+                &OneLetterUpperAmPm =>
+                    time.map(|t| {
+                        result.push_str(if t.hour12().0 {"P"} else {"A"});
+                        Ok(())
+                    }),
+                _ => return Err(fmt::Error),
```
quodlibetor, line 707: Could you explicitly match on `__DoNotMatch`? We're doing this internally, and are guaranteed to update in lockstep, we just want downstream users to know that they shouldn't.

src/format/mod.rs
```diff
@@ -75,6 +75,10 @@ pub enum Numeric {
     /// Full Gregorian year (FW=4, PW=∞).
     /// May accept years before 1 BCE or after 9999 CE, given an initial sign.
     Year,
+    /// Full Gregorian year (FW=2, PW=∞). Can be negative.
+    ///
+    /// Prints last two digits for year >= 1000 or year <= -1000.
```
quodlibetor, line 80: Could you explicitly mention how ambiguous this is? E.g. that the years `1120` and `2020` will print the same year? And that it errors in years outside of the mentioned range?

Also, why does this error for N `-1000 < N < 1000`?

src/format/parsed.rs
```diff
@@ -166,6 +166,14 @@ impl Parsed {
         set_if_consistent(&mut self.year, value.to_i32().ok_or(OUT_OF_RANGE)?)
     }
 
+    /// Tries to set the [`year`](#structfield.year) field from given value.
+    #[inline]
+    pub fn set_year_short(&mut self, value: i64) -> ParseResult<()> {
```
quodlibetor, line 171: Is this used? I don't see it anywhere else in this PR. (possibly my search is broken)

src/format/mod.rs
```diff
@@ -268,6 +297,8 @@ pub enum Item<'a> {
     Numeric(Numeric, Pad),
     /// Fixed-format item.
     Fixed(Fixed),
+    /// Extra fixed-format item.
+    FixedExt(FixedExt),
```
quodlibetor, line 301: could you remove this from `Item`

sergeyt:
@quodlibetor i'll try to fix this PR soon. thanks for detailed review notes attached to the code, it is timesaver

sergeyt:
@quodlibetor I am not sure I understand how to integrated these `Fixed` formats if I remove `FixedExt` from `Item`. Should I extend `InternalFixed`?

quodlibetor:
> I am not sure I understand how to integrated these Fixed formats if I remove FixedExt from Item. Should I extend InternalFixed?

Exactly, yes.

src/format/mod.rs
pitdicker: Agreed. Adding `FixedExt` would be a breaking change. We (now) have `FixedInternal` to but new variants in.

pitdicker (commented):
It has been three years, sorry for the hiatus!

@sergeyt are you still interested in continuing this work?

To give my personal opinion: I don't think single-letter month and day names are useful enough to add to chrono. They are ambiguous, uncommon, and can't be used for parsing.

sergeyt:
> It has been three years, sorry for the hiatus!
> 
> @sergeyt are you still interested in continuing this work?
> 
> To give my personal opinion: I don't think single-letter month and day names are useful enough to add to chrono. They are ambiguous, uncommon, and can't be used for parsing.

no problem. it is not terribly important now

PR #458, opened 2020-07-21
Allow clippy::manual-non-exhaustive

robyoung:
These are needed in order to compile on 1.13.0. It seems unwise to have different non-exhaustive measures for different versions.

This should fix the broken linting.

### DIFF
--- a/ci/lint.sh
+++ b/ci/lint.sh
@@ -7,7 +7,7 @@ source "${BASH_SOURCE[0]%/*}/_shlib.sh"
 
 main() {
     # rustflags are present because of: https://github.com/rust-lang/rust-clippy/issues/5749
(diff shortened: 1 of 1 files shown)


quodlibetor:
Thanks so much!

Rust 1.13 doesn't allow `::` nested attributes, which is why that allow is failing, essentially we are forced to use the `#[cfg_attr(cargo-clippy, manual_non_exhaustive)]`. I would also be fine moving clippy lints out into a script that runs clippy with CLI allow flags.

robyoung:
That is a good point and I've just seen that #453 does it that way so I'm going to change so it's an easier rebase for them.

cecton:
I don't think I will even need to rebase. Since it's the exact same fix, there is no conflict

PR #463, opened 2020-07-25
Added a fuzzer as integration to OSS-Fuzz.

DavidKorczynski:


quodlibetor:
Thanks for this!

Could you add a readme to the fuzz directory that documents how to use it and where to look for the oss-fuzz integration? It's fine is it's pretty trivial.

DavidKorczynski:
I added a short README for the fuzzing set up. Could we merge this in before documenting how to run it by way of OSS-Fuzz, as the integration first need to be accepted on the OSS-Fuzz repo (https://github.com/google/oss-fuzz/pull/4192), and in order to pass the integration tests on the OSS-Fuzz repo we need to have this PR accepted.

quodlibetor:
Yup works for me! I'll definitely appreciate the docs PR after the integration is set up.

Thanks again!

PR #464, opened 2020-08-03
docs: fix link to cargo docs

ammgws:
As per title


quodlibetor:
Thanks!

If you run `make readme` in the root of the repo the lint should pass.

quodlibetor:
Oh wait no I have it backwards: you should modify the doc-comment at the top of lib.rs and then it should pass.

PR #465, opened 2020-08-05
Fix Local.from_local_datetime method for wasm

ShadoySV:
      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md
Fix Local.from_local_datetime method for wasm

### DIFF
--- a/src/offset/local.rs
+++ b/src/offset/local.rs
@@ -3,16 +3,21 @@
 
 //! The local (system) time zone.
 
+#[cfg(not(all(target_arch = "wasm32", not(target_os = "wasi"), feature = "wasmbind")))]
 use oldtime;
 
 use super::fixed::FixedOffset;
 use super::{LocalResult, TimeZone};
-use naive::{NaiveDate, NaiveDateTime, NaiveTime};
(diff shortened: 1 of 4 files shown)


src/offset/local.rs
```diff
@@ -7,12 +7,16 @@ use oldtime;
 
 use super::fixed::FixedOffset;
 use super::{LocalResult, TimeZone};
-use naive::{NaiveDate, NaiveDateTime, NaiveTime};
+#[cfg(not(all(target_arch = "wasm32", not(target_os = "wasi"), feature = "wasmbind")))]
+use naive::NaiveTime;
+use naive::{NaiveDate, NaiveDateTime};
 use {Date, DateTime};
+#[cfg(not(all(target_arch = "wasm32", not(target_os = "wasi"), feature = "wasmbind")))]
 use {Datelike, Timelike};
```
quodlibetor, line 15: I'm a little confused why we need to add these cfg flags here? Is it that this changes the methods to have unused imports with the new flags?

ShadoySV: ... to have only corresponding used imports for corresponding targets. Without the new flags, there will be unused imports for wasm32 target

quodlibetor: Thanks!

PR #473, opened 2020-08-16
Skip IANA TZ names when parsing %Z specifier

abashurov:

      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md

This PR prevents `%Z` specifier from causing an instant parsing failure, instead, skipping chars until the next whiltespace-like char, as implemented in glibc strptime.


abashurov:
> Thanks for this! Could you also modify the %Z row in `src/format/strftime.rs` to state that in parsing mode this discards all non-whitespace characters?

Done and done

> Aside: it is kind of wild that that is the behavior of glibc, I had to confirm it myself.

Well, it's a lot of work with so little payoff. Also, I probably should have provided a link to the glibc source code to simplify the verification, sorry for that

PR #483, opened 2020-09-25
The wasm-bindgen code refers to oldtime::Duration and is failing

quodlibetor:
wasm only works on new-enough rust that we can unconditionally use the crate
keyword in that method, so let's try that.

https://github.com/chronotope/chrono/pull/286#issuecomment-699146233

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,12 +3,20 @@ ChangeLog for Chrono
 
 This documents all notable changes to [Chrono](https://github.com/chronotope/chrono).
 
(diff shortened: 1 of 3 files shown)


CryZe:
Looks like this fixes my build issues 👍

quodlibetor:
@CryZe thanks for verifying, and sorry it broke your build!

jhpratt:
Looks like the comment regarding semver isn't actually implemented?

benesch:
If we're going to get pedantic about SemVer, breaking changes are permitted in minor _or_ patch releases pre 1.0. It's Cargo that violates the SemVer spec on this point.

PR #491, opened 2020-09-29
Fix call of extern tzset function

### LINKED ISSUE #490: #478 breaks illumos
After this PR was merged chorno no longer compiles on illumos:
```
error[E0425]: cannot find function `tzset` in the crate root
  --> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/chrono-0.4.18/src/sys/unix.rs:81:15
   |
81 |             ::tzset();
   |               ^^^^^ not found in the crate root
   |
help: consider importing this function
   |
11 | use oldtime::tzset;
   |

   Compiling tempfile v3.1.0
   Compiling http-body v0.3.1
```

quodlibetor:
Fixes #490

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,4 @@
-name: test
+name: All Tests and Builds
 
 on:
   push:
@@ -116,3 +116,19 @@ jobs:
         env:
           RUST_VERSION: stable
           WASM: wasm_simple
+
+  cross-targets:
+    strategy:
+      matrix:
+        target:
+          - x86_64-sun-solaris
+
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install cross
+        run: bash ci/install-cross.sh
+
+      - name: Build static library
+        run: cross check --target ${{ matrix.target }}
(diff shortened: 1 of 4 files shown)


benesch:
Oof, the conditional compilation in time v0.1 was so hard to follow. Sorry about this.

Any idea why `timegm` reinvents the wheel here? https://github.com/chronotope/chrono/blob/c324d0ea08d1c9032563ae77a0f6e325dda5ba6e/src/sys/unix.rs#L50-L52 Seems like that function should just use the `tzset` declaration you're adding? (Not your code! Just trying to make sense of the time v0.1 source code.)

quodlibetor:
> Oof, the conditional compilation in time v0.1 was so hard to follow. Sorry about this.

Yeah it's the worst, not your fault I explicitly tried to track this.

No I couldn't figure out why it was written the way that it was. The commit (https://github.com/time-rs/time/commit/4155027da8f1ce9466c54457e9e7288b61076d4b) isn't super clear on why it didn't use the already-existing timegm.

quodlibetor:
> Seems like that function should just use the tzset declaration you're adding?

Once we're running the test suite on all the targets that it refers to I'd like to delete that internal tzset definition.

PR #492, opened 2020-09-30
Fix a small typo in NaiveDateTime's Display impl

alexander-jackson:
Fix a small typo in the documentation of NaiveDateTime's Display
implementation to say Display instead of Debug.


quodlibetor:
Thanks!

PR #496, opened 2020-10-28
Remove redundant field name

luisbg:
There are more of these in the code. I am sending the fix for one just to check if this is accepted.

- [ ] Have you added yourself and the change to the [changelog]? No. I can do that in a second revision after review.

[changelog]: ../CHANGELOG.md

### DIFF
--- a/src/sys.rs
+++ b/src/sys.rs
@@ -121,6 +121,6 @@ impl Tm {
             0 => inner::utc_tm_to_time(self),
             _ => inner::local_tm_to_time(self),
         };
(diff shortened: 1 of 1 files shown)


luisbg:
The failing PR checks are from other areas of the code.

luisbg:
I just noticed your repo is not in Hacktoberfest.

Can you add the `hacktoberfest-accepted` label to this PR?

quodlibetor:
Sorry, unfortunately chrono still supports very old rust, which means that we cannot use the field-init shorthand.

PR #506, opened 2020-11-24
Add Default implementation for naive date types

DiD92:
This PR aims to add some [Default](https://doc.rust-lang.org/std/default/trait.Default.html) implementations for some of the naive types. The currently added defaults are the following:

* Default `NaiveDateTime` has been set to `1970-01-01 00:00:00`
* Default `NaiveDate` has been set to `1970-01-01`
* Default `NaiveTime` has been set to `00:00:00`

The main reason for the PR is to help in struct definitions that wish to simply `#derive[Default]` but currently are not able to because on of their attributes is of the previously mentioned types, adding this change would help simplify the code in that particular scenario.

### DIFF
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -1719,6 +1719,8 @@ impl str::FromStr for NaiveDate {
 /// # Example
 ///
 /// ~~~~
+/// use chrono::NaiveDate;
+///
 /// let default_date = NaiveDate::default();
 /// assert_eq!(default_date, NaiveDate::from_ymd(1970, 1, 1));
 /// ~~~~
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -1714,16 +1714,16 @@ impl str::FromStr for NaiveDate {
     }
 }
 
-/// The default for a NaiveDate is 1st of January 1970.
+/// The default value for a NaiveDate is 1st of January 1970.
 ///
 /// # Example
 ///
-/// ~~~~
+/// ```rust
 /// use chrono::NaiveDate;
 ///
 /// let default_date = NaiveDate::default();
 /// assert_eq!(default_date, NaiveDate::from_ymd(1970, 1, 1));
-/// ~~~~
+/// ```
 impl Default for NaiveDate {
     fn default() -> Self {
         NaiveDate::from_ymd(1970, 1, 1)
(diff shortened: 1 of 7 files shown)


quodlibetor:
Sorry for the delay! I agree that this is just kind of annoying to have not-implemented, and just using the unix epoch makes sense as a reasonably-surprising value.

If you're willing to fix the tests and add implementations of this for the non-naive `Datetime<Tz>` (where it just defaults to the same date/time in the user-requested timezone) I would be happy to merge this!

DiD92:
Hi @quodlibetor , sorry for my late reply, yes I'm more than happy to do that! I'll update the PR once I have the changes.

DiD92:
Hi @quodlibetor, the test now pass and I've added `Default` implementations for `DateTime` as you requested, if any more changes are needed please feel free to request them.

src/naive/date.rs
```diff
@@ -1714,6 +1714,22 @@ impl str::FromStr for NaiveDate {
     }
 }
 
+/// The default for a NaiveDate is 1st of January 1970.
+///
+/// # Example
+///
+/// ~~~~
```
quodlibetor, line 1721: we recently (in between your submission of this PR and now) swapped all the doctests to use backticks to be more inline with other rust code. If you feel like changing these that'd be appreciated, but your call.

src/naive/time.rs
```diff
@@ -1341,6 +1341,20 @@ impl str::FromStr for NaiveTime {
     }
 }
 
+/// The default for a NaiveTime is midnight, oo:oo:00 exactly.
```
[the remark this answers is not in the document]
DiD92: Fixed!

src/naive/date.rs
DiD92: I've changed the docs from the `Default` impl to use backticks!

DiD92:
Hello @quodlibetor hope you had a nice holiday, I simply wanted to know if anything more was required in order to have the PR merge onto master, happy to make any changes if needed!

quodlibetor:
Thank you!

malthe:
I think `DateTime::now()` is a more useful default – and arguably a lot less surprising. If you want an essentially "unset" datetime as the default value, then `Option<DateTime>` is a better choice.

PR #509, opened 2020-11-30
Feature/duration round for naive datetime

robyoung:
Depends on #508 

This is off the back of [this comment](https://github.com/chronotope/chrono/pull/445#issuecomment-717323407).


quodlibetor:
Thanks very much!

PR #510, opened 2020-11-30
Use backticks for documentation examples

robyoung:
The rustdoc documentation and the book only show examples as escaped with backticks.

I could not find any reference to tilde escaped examples, although tildes are a bit of a hard character to search for. They seem to work but for the sake of consistency and following the documentation I feel like the change is worth it.

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -83,12 +83,12 @@ impl<Tz: TimeZone> DateTime<Tz> {
     ///
     /// # Example
     ///
-    /// ~~~~
+    /// ```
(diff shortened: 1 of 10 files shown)


quodlibetor:
Thanks! I've been thinking of doing this for awhile.

FWIW the tildes are just from whatever version of markdown is implemented, it's documented (identically) as part of both [GFM](https://github.github.com/gfm/#fenced-code-blocks) and [CommonMark](https://spec.commonmark.org/0.29/#fenced-code-blocks).

PR #511, opened 2020-12-02
Simplify conversion between JS Date and DateTime<Utc>

ObsidianMinor:
When using this library with wasm, it was difficult to test the result of what came out of chrono in JS since the result always had the milliseconds of whatever Date.now() was. This change makes it so the conversion is simpler and based on `DateTime::timestamp_millis`, which will make our milliseconds match up.

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -872,30 +872,18 @@ impl From<js_sys::Date> for DateTime<Utc> {
 #[cfg(all(target_arch = "wasm32", not(target_os = "wasi"), feature = "wasmbind"))]
 impl From<&js_sys::Date> for DateTime<Utc> {
     fn from(date: &js_sys::Date) -> DateTime<Utc> {
-        let millisecs_since_unix_epoch: u64 = date.get_time() as u64;
-        let secs = millisecs_since_unix_epoch / 1000;
-        let nanos = 1_000_000 * (millisecs_since_unix_epoch % 1000);
-        let naive = NaiveDateTime::from_timestamp(secs as i64, nanos as u32);
-        DateTime::from_utc(naive, Utc)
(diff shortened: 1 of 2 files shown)


quodlibetor:
It seems like this would make more sense if it set the JS subsecond portion to the subsec portion of the datetime?

ObsidianMinor:
Thinking about it, shouldn't this conversion just use `timestamp_millis` with `Date::new_0` anyway? Seems like it'd save a lot on JS<->Rust calls (since each "set utc part" call is a separate function generated with wasm-bindgen).

quodlibetor:
yeah that does seem much better, although it's probably worth validating what the maximum value that _can_ be set that way is, and falling back to the current behavior if it's possible to create timestamps larger than that value.

ObsidianMinor:
The max/min `Date` value is equal to ±8,640,000,000,000,000 milliseconds. Any attempt to make a date more than that (including through setUTC functions) results in an `Invalid Date`.
image

quodlibetor:
Plausibly we should care about MAX_SAFE_INTEGER instead of the max date? I'm not actually sure whether the wasm Date API needs to go through floating point conversion? https://stackoverflow.com/a/57225494/25616

ObsidianMinor:
It'll need to go through float conversions, yes, but ±8,640,000,000,000,000 can be represented as a f64 safely, so this shouldn't be an issue I believe. We will need to document its limits.

quodlibetor:
Oh, yes I see. Yeah that sounds good!

ObsidianMinor:
I also fixed up the JS `Date` -> `DateTime<Utc>` so it's a much simpler conversion.

quodlibetor:
Thanks for this!

PR #516, opened 2020-12-21
Add documentation for DateTime FromStr impl

### LINKED ISSUE #505: Missing documentation for FromStr impl for DateTime
`DateTime<Tz>` implements `FromStr`, which means `str::parse` can parse datetimes. This is very convenient, and I'd like to use it to parse datetimes provided on the command line. But I can't find documentation for the accepted format - the [docs](https://docs.rs/chrono/0.4.19/chrono/struct.DateTime.html#impl-FromStr) only contain the generic `FromStr` description, "Parses a string `s`...".

As determined by looking at source code and experimenting, `FromStr` seems to accept a slight superset of RFC3339, allowing space to separate the date and time parts. This behavior makes `FromStr` more useful to me than `DateTime::parse_from_rfc3339()`, which only accepts `T` as separator. Since some sources of dates, such as `date --rfc-3339=seconds`, use space as separator, I'd like to support both.

Is the `FromStr` parsing format documented somewhere? If not, can I rely that its behavior won't change in a future release?

robyoung:
Resolves #505

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -802,6 +802,17 @@ where
     }
 }
 
+/// Accepts a relaxed form of RFC3339.
+/// A space or a 'T' are acepted as the separator between the date and time
+/// parts. Additional spaces are allowed between each component.
+///
+/// All of these examples are equivalent:
+/// ```
+/// # use chrono::{DateTime, Utc};
+/// "2012-12-12T12:12:12Z".parse::<DateTime<Utc>>();
+/// "2012-12-12 12:12:12Z".parse::<DateTime<Utc>>();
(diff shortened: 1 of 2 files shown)


quodlibetor:
Thanks for this! The docs look great.

Did you check the generated documentation? My recollection is that trait docs need to be on the methods in order for them to show up in the generated docs, but my recollection could be wrong or things could have changed.

robyoung:
Yes, it shows up when you expand the trait.

robyoung:
Sorry, I wasn't at a computer yesterday. It comes out looking like this.

Documentation of FromStr on DateTime

quodlibetor:
Thanks!

PR #527, opened 2021-01-19
 Fix possible panic in NaiveDate::from_num_days_from_ce_opt

b3t33:
Add some test including panic triggering example
`NaiveDate::from_num_days_from_ce_opt(i32::MAX)`.
As the function already yields `None` for `i32::Max - 365`, this fix
does not change the behaviour, while preventing a panic.

      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md


quodlibetor:
Thanks for this! If you rebase on main the clippy lints should go away.

Just curious, did you actually run into this issue or did you just notice it?

quodlibetor:
Actually I was able to rebase for you.

b3t33:
Thanks for rebasing.
Well, actually using [`quickcheck`](https://github.com/BurntSushi/quickcheck) 1.0 (combined with a very simplistic construction of `NaiveDate`) escalated this for me and inspecting the source revealed the problematic `days + 365`.

pitdicker:
This fixes a potential panic, similar to those found in https://github.com/chronotope/chrono/issues/1010.

@b3t33 Could you rebase this PR and target the 0.4.x branch?

src/naive/date.rs
```diff
@@ -2027,6 +2029,14 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_date_from_num_days_from_ce_opt() {
+        let ndays_from_ce = |days| NaiveDate::from_num_days_from_ce_opt(days);
+        ndays_from_ce(i32::MAX);
```
jtmoon79, line 2035: Can these two checks be moved into `test_date_from_num_days_from_ce` below?
Also, why not check the return value of the closure ?

jtmoon79 (changes requested):

djc:
This was taken care of in #1052. Thanks for doing the work!

PR #304, opened 2019-02-21
Microseconds Timestamps, clone of #247

jaredvann:
This adds a function for retrieving a DateTime as microseconds like https://github.com/chronotope/chrono/pull/247 did for nanoseconds. The equivalent serde functions for serialising/deserialising are included as well.

PR #533, opened 2021-01-31
Add serde functions for optional microsecond timestamps

nickelc:
Adds optional microsecond timestamp support in addition to #304


Folyd (approved):
LGTM

PR #539, opened 2021-02-25
impl Error trait for ParseWeekdayError and ParseMonthError

mike-kfed:
Those two structs were missing Error implementation which made using `?` kinda hard in my code.

I also removed the dummy field, as it is not needed IMO.

I added feature="std" for the error and Display implementations, I hope that is okay like that.

ready to adapt the PR to your liking, just let me know.

### DIFF
--- a/src/month.rs
+++ b/src/month.rs
@@ -235,6 +235,16 @@ pub struct ParseMonthError {
     pub(crate) _dummy: (),
 }
 
+#[cfg(feature = "std")]
+#[cfg_attr(docsrs, doc(cfg(feature = "std")))]
+impl std::error::Error for ParseMonthError {}
+
+impl fmt::Display for ParseMonthError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "ParseMonthError {{ .. }}")
+    }
+}
+
 impl fmt::Debug for ParseMonthError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "ParseMonthError {{ .. }}")


pitdicker:
For 0.5 we should rework the plan is to rework te `Error` type and use it for most places in chrono that now return a custom error of `Option`. But for the 0.4.x branch this seems reasonable.

@mike-kfed would you still be available to work on this? Starting with a rebase against the 0.4.x branch?

mike-kfed:
Hi @pitdicker I am still available for working on this. I will rebase to 0.4.x branch and apply my changes, that is all that is needed correct?

src/month.rs
```diff
@@ -231,8 +231,16 @@ impl Months {
 
 /// An error resulting from reading `<Month>` value with `FromStr`.
 #[derive(Clone, PartialEq, Eq)]
-pub struct ParseMonthError {
-    pub(crate) _dummy: (),
```
pitdicker, line 235: Why are the `_dummy` fields removed? This makes it possible for library users to construct the type, and we can no longer add fields in the future (not that I think we will...). Can you revert this part?

pitdicker (changes requested):
Great, thank you! I can't give the final approval, that would be @djc.

Also I don't know how to get the CI to run when the targeted branch is switched. Maybe another push would do it?

src/month.rs
mike-kfed: The dummy field is not needed, constructing works by doing `let e = ParseMonthError;`, and why can't you add fields at a later time? I fail to see the issues you raised, sorry, especially since `pub struct ParseWeekdayError {}` is defined the same (which was not added by me).

About running CI, not sure how to do that either, sorry. A push might work, but that implies a further needed commit.

pitdicker: A private field is a common future proofing trick. See the somewhat related https://rust-lang.github.io/api-guidelines/future-proofing.html#structs-have-private-fields-c-struct-private. But yes, it is not used consistently at the moment, and not all that interesting for this type.

But I wanted to call it out, as it is an unrelated change.

mike-kfed: I don't see it as an unrelated change, as it makes it consistent with ParseWeekdayError. And even with your future proofing argument I still find this field unnecessary for this very application. However I can add it back, also to ParseWeekdayError if desired.

pitdicker:
Sorry! I could trigger the CI with
```
git commit --no-edit --amend
git push -f
```

We already had a green mark.

src/month.rs
pitdicker: If you could please add it back and not change `ParseWeekdayError` than this PR will just be simple to approve :smile:.

mike-kfed:
> We already had a green mark.

Sorry! but is rebuilding now anyways due to leaving dummy alone

pitdicker (approved):
Thank you!

djc:
Thanks!

PR #541, opened 2021-03-01
#494 export Duration's OutOfRangeError

### LINKED ISSUE #494: OutOfRangeError appears in public signatures but is a private type
Without the `time` feature, chrono uses its own definition of `Duration` in `oldtime.rs` (a private module), re-exporting it from the root of the crate. `Duration.to_std` can return an `OutOfRangeError`. However, this type is not exposed, making it impossible to match on.

This is *not* a problem with the `time` feature, as the time library exposes its version of `OutOfRangeError`.

For API parity, could the error type be exposed by chrono, please? Ideally, chrono would re-export `time::OutOfRangeError` or `oldtime::OutOfRangeError`, to allow code to be written that will deal with whatever chrono's feature set is.

elpiel:
Resolves #494 
Exporting the error so that reliance on `mod oldtime` can be fixed and for the `time` crate to be unneeded when handling this error

      about adding the PR number)

[changelog]: ../CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Versions with only mechanical changes will be omitted from the following list.
 
 ## 0.4.23
 * Impl Add<Month> + Sub<Month> for NaiveDateTime
+* Export `oldtime::OutOfRangeError` to the public API
 
 ## 0.4.22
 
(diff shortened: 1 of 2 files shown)


ayrat555:
@djc Hello! Can you please take a look

djc:
I'd be okay with this, just needs to be rebased.

elpiel:
Let me take a look at this PR again

elpiel:
@djc would you mind taking a look at this?

djc (approved):
This seems reasonable to me.

esheppa (approved):

PR #542, opened 2021-03-07
Add pyo3 integration

pickfire:
Some parts taken from https://github.com/kangalioo/pyo3-chrono

Note, I will continue working on this in few more days, I have something I wanted to experiment in the meantime.

cc @kangalioo

## Implemented

- [x] `PyDate <-> NaiveDate`
- [x] `PyTime <-> NaiveTime`
- [x] `PyDateTime <-> NaiveDateTime`
- [x] `PyDelta <-> Duration` (only implemented for `oldtime` feature)
- [x] ~~`PyDate <-> Date`~~ (may confuse people since `PyDate` does not have timezone)
- [x] `PyTzInfo <-> Utc`
- [x] `PyTzInfo <-> FixedOffset`
- [x] `PyDateTime <-> DateTime`

      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md

### DIFF
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -832,7 +832,7 @@ impl NaiveDate {
 
     /// Returns the packed month-day-flags.
     #[inline]
-    fn mdf(&self) -> Mdf {
+    pub(crate) fn mdf(&self) -> Mdf {
         self.of().to_mdf()
     }
 
--- a/src/naive/internals.rs
+++ b/src/naive/internals.rs
@@ -371,7 +371,7 @@ impl fmt::Debug for Of {
 /// (month, day of month and leap flag),
 /// which is an index to the `MDL_TO_OL` lookup table.
 #[derive(PartialEq, PartialOrd, Copy, Clone)]
-pub(super) struct Mdf(pub(super) u32);
+pub(crate) struct Mdf(pub(super) u32);
 
 impl Mdf {
     #[inline]
@@ -419,7 +419,7 @@ impl Mdf {
     }
 
     #[inline]
-    pub(super) fn month(&self) -> u32 {
+    pub(crate) fn month(&self) -> u32 {
         let Mdf(mdf) = *self;
         mdf >> 9
     }
@@ -432,7 +432,7 @@ impl Mdf {
     }
 
     #[inline]
-    pub(super) fn day(&self) -> u32 {
+    pub(crate) fn day(&self) -> u32 {
         let Mdf(mdf) = *self;
         (mdf >> 4) & 0b1_1111
     }
--- a/src/naive/time/mod.rs
+++ b/src/naive/time/mod.rs
@@ -795,7 +795,7 @@ impl NaiveTime {
     }
 
     /// Returns a triple of the hour, minute and second numbers.
-    fn hms(&self) -> (u32, u32, u32) {
+    pub(crate) fn hms(&self) -> (u32, u32, u32) {
         let (mins, sec) = div_mod_floor(self.secs, 60);
         let (hour, min) = div_mod_floor(mins, 60);
         (hour, min, sec)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ Versions with only mechanical changes will be omitted from the following list.
 * Add support for getting week bounds based on a specific `NaiveDate` and a `Weekday` (#666)
 * Remove libc dependency from Cargo.toml.
 * Add the `and_local_timezone` method to `NaiveDateTime`
+* Add `pyo3` integration (@pickfire @kangalioo #542)
 
 ## 0.4.19
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ serde = { version = "1.0.99", default-features = false, optional = true }
 pure-rust-locales = { version = "0.5.2", optional = true }
 criterion = { version = "0.3", optional = true }
 rkyv = {version = "0.7", optional = true}
+pyo3 = { version = "0.16.4", optional = true }
 
 [target.'cfg(all(target_arch = "wasm32", not(any(target_os = "emscripten", target_os = "wasi"))))'.dependencies]
 wasm-bindgen = { version = "0.2", optional = true }
@@ -54,6 +55,7 @@ serde_derive = { version = "1", default-features = false }
 bincode = { version = "1.3.0" }
 num-iter = { version = "0.1.35", default-features = false }
 doc-comment = { version = "0.3" }
+pyo3 = { version = "0.16.2", features = ["auto-initialize"] }
 
 [target.'cfg(all(target_arch = "wasm32", not(any(target_os = "emscripten", target_os = "wasi"))))'.dev-dependencies]
 wasm-bindgen-test = "0.3"
--- a/Makefile
+++ b/Makefile
@@ -26,4 +26,4 @@ test:
 
 .PHONY: doc
 doc: authors readme
-	cargo doc --features 'serde rustc-serialize bincode'
+	cargo doc --features 'serde rustc-serialize bincode pyo3'
--- a/README.md
+++ b/README.md
@@ -62,11 +62,13 @@ Optional features:
 
 - `wasmbind`: Enable integration with [wasm-bindgen][] and its `js-sys` project
 - [`serde`][]: Enable serialization/deserialization via serde.
+- [`pyo3`][]: Enable integration with pyo3.
 - `unstable-locales`: Enable localization. This adds various methods with a
   `_localized` suffix. The implementation and API may change or even be
   removed in a patch release. Feedback welcome.
 
 [`serde`]: https://github.com/serde-rs/serde
+[`pyo3`]: https://github.com/PyO3/pyo3
 [wasm-bindgen]: https://github.com/rustwasm/wasm-bindgen
 
 See the [cargo docs][] for examples of specifying features.
--- a/ci/github.sh
+++ b/ci/github.sh
@@ -6,7 +6,7 @@ set -euo pipefail
 source "${BASH_SOURCE[0]%/*}/_shlib.sh"
 
 TEST_TZS=(ACST-9:30 EST4 UTC0 Asia/Katmandu)
-FEATURES=(std serde clock "alloc serde" unstable-locales)
+FEATURES=(std serde clock "alloc serde" unstable-locales pyo3)
 CHECK_FEATURES=(alloc "std unstable-locales" "serde clock" "clock unstable-locales")
 RUST_132_FEATURES=(rustc-serialize serde)
 
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -84,7 +84,7 @@ pub enum SecondsFormat {
 #[derive(Clone)]
 #[cfg_attr(feature = "rkyv", derive(Archive, Deserialize, Serialize))]
 pub struct DateTime<Tz: TimeZone> {
-    datetime: NaiveDateTime,
+    pub(crate) datetime: NaiveDateTime,
     offset: Tz::Offset,
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -49,11 +49,13 @@
 //!
 //! - `wasmbind`: Enable integration with [wasm-bindgen][] and its `js-sys` project
 //! - [`serde`][]: Enable serialization/deserialization via serde.
+//! - [`pyo3`][]: Enable integration with pyo3.
 //! - `unstable-locales`: Enable localization. This adds various methods with a
 //!   `_localized` suffix. The implementation and API may change or even be
 //!   removed in a patch release. Feedback welcome.
 //!
 //! [`serde`]: https://github.com/serde-rs/serde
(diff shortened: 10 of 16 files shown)


pickfire:
Not sure how can we add `pyo3` integration for `Duration` from `time` crate since we cannot `impl` for another crate, do we need to do a wrapper type or just remove support for pyo3 for time crate?

src/naive/date.rs
```diff
@@ -762,7 +762,7 @@ impl NaiveDate {
 
     /// Returns the packed month-day-flags.
     #[inline]
-    fn mdf(&self) -> Mdf {
+    pub(crate) fn mdf(&self) -> Mdf {
```
pickfire, line 765:
```
error: `pub(restricted)` syntax is experimental (see issue #32409)
   --> src/naive/date.rs:765:9
    |
765 |     pub(crate) fn mdf(&self) -> Mdf {
    |         ^^^^^
```
What should we do here?

src/offset/fixed.rs
```diff
@@ -242,3 +242,68 @@ mod tests {
...
+        let py = gil.python();
+        let offset = FixedOffset::east(3600).to_object(py);
+        let py_module = py.import("datetime").unwrap();
+        let py_timezone = py_module.getattr("timezone").unwrap();
+        let py_timedelta = PyDelta::new(py, 0, 3600, 0, true).unwrap();
+        let py_timezone = py_timezone.call1((py_timedelta,)).unwrap();
+        // assert_eq!(offset.compare(py_timezone).unwrap(), Ordering::Equal);
+        // TODO: not sure how to convert this to timezone struct to
+        // compare since we don't really have a struct to downcast
+        assert_eq!(offset.as_ref(py), py_timezone);
```
pickfire, line 294: @davidtwco Do you know if there is any way that we are able to compare timezone? I tried comparing with `Eq` but even though both have the same value it didn't seemed to pass.

pickfire:
Sorry for the long wait, I was waiting for 2018 edition move but it didn't seemed to happen (and probably not anytime soon). Currently I a couple of types that involves timezone (`DateTime`, `Utc`, `FixedOffset`) cannot be converted back to rust, one part is because pyo3 does not seemed to provide access to timezone information, another reason is I think it probably is not that useful as naive date/time, maybe that can come later.

src/naive/date.rs
pickfire: Will migration to rust 2018 happening anytime soon?

pickfire:
Rebased to latest master with 2018 edition and got it compiling, need to double check where I left off.

djc:
What's the use case for this? Why would it be helpful/good for this to live inside chrono instead of in a separate crate? I like PyO3 and have worked with it before, but it will take a while before I'm comfortable adding a bunch of extra stuff to chrono given the lack of maintenance this crate has been seeing for a while.

pickfire:
> What's the use case for this? Why would it be helpful/good for this to live inside chrono instead of in a separate crate? I like PyO3 and have worked with it before, but it will take a while before I'm comfortable adding a bunch of extra stuff to chrono given the lack of maintenance this crate has been seeing for a while.

So next time when users use pyo3, chrono integration would be seamless, adding it in this crate allows using the internal representation of chrono and allows to convert directly to python. An alternative would be to add it to pyo3 but that have the same limitation as adding it as a separate crate, at least the integration is just like serde where you can just add a feature and use it rather than having to look up a separate crate, the pyo3 stuff is pretty much separated so it does not add that much of maintenance.

djc:
Okay, that does make some sense.

In terms of how we implement this, I think I'd prefer to have a single top-level `pyo3` module with a module hierarchy within that mirrors the internal chrono hierarchy (instead of scattering pyo3 code over the internal hierarchy).

pickfire:
I moved most parts into `pyo3` except `oldtime` since there are some internal functions within oldtime.

Also figured out how to convert from python utc.

pickfire:
Weird, not sure why stuff used to passed and now starts failing due to associated functions visibility.

pickfire:
Now that pyo3 supports tzinfo access, I can continue working on datetime, not sure why it fails spuriously and even segfaults sometimes.
```
thread 'pyo3::datetime::test_pyo3_topyobject' panicked at 'Failed to construct datetime: PyErr { type: <class 'TypeError'>, value: TypeError("tzinfo argument must be None or of a tzinfo subclass, not type 'datetime.time'"), traceback: None }', src/pyo3/datetime.rs:31:10
stack backtrace:
   0: rust_begin_unwind
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/panicking.rs:143:14
   2: core::result::unwrap_failed
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/result.rs:1749:5
   3: core::result::Result<T,E>::expect
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/result.rs:1022:23
   4: chrono::pyo3::datetime::<impl pyo3::conversion::ToPyObject for chrono::datetime::DateTime<Tz>>::to_object
             at ./src/pyo3/datetime.rs:19:24
   5: chrono::pyo3::datetime::test_pyo3_topyobject::{{closure}}
             at ./src/pyo3/datetime.rs:87:24
   6: chrono::pyo3::datetime::test_pyo3_topyobject
             at ./src/pyo3/datetime.rs:106:5
   7: chrono::pyo3::datetime::test_pyo3_topyobject::{{closure}}
             at ./src/pyo3/datetime.rs:79:1
   8: core::ops::function::FnOnce::call_once
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/ops/function.rs:227:5
   9: core::ops::function::FnOnce::call_once
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

pickfire:
@davidhewitt Can you please help to review this pull request? Not sure if it's due to `acquire_gil` in tests that causes the spurious failures.

pickfire:
Do I need to bump the python version in CI? The pull request is 3.6 but pyo3 only supports 3.7 and above.

davidhewitt:
If it's running on 3.6, yes it'll need a bump. Are you sure about that though? PyO3 has a check which should prevent you from being able to do that: https://github.com/PyO3/pyo3/blob/1d71d173bb2544156348b86c2ae31288172f78df/pyo3-ffi/build.rs#L86

If it really is running on 3.6, can you file a bug report please? You can check using `PYO3_PRINT_CONFIG=1` environment variable.

pickfire:
> If it's running on 3.6, yes it'll need a bump. Are you sure about that though? PyO3 has a check which should prevent you from being able to do that: https://github.com/PyO3/pyo3/blob/1d71d173bb2544156348b86c2ae31288172f78df/pyo3-ffi/build.rs#L86

Ah, sorry. That was a separate issue, I mean the CI which I wanted to ask chrono maintainers.

But I am only not sure about the spurious failures (`cargo test` success roughly like half the time and the other half fail partially or segmentation fault), I did try looking into it but not sure why yet, maybe I will double check on those `unsafe` parts.

davidhewitt:
👍 I'm happy to investigate, however please allow time for me to get around to this. My OSS queue is long and my time limited at the moment!

davidhewitt:
Sorry for the long delay, finally took a quick look at this.

Regarding segfaults, looks like you've got incorrect ownership. The below:

```rust
        unsafe {
             PyDateTime_IMPORT();
             PyObject::from_owned_ptr(py, PyDateTime_TimeZone_UTC())
         }
```

should be

```rust
        unsafe {
             PyDateTime_IMPORT();
             PyObject::from_borrowed_ptr(py, PyDateTime_TimeZone_UTC())
         }
```

Because `PyDateTime_TimeZone_UTC()` returns a borrowed pointer (not owned).

@djc - I wonder if you have an opinion where this integration should live? Given that `chrono` is actively maintained again I'm reasonably comfortable with having `chrono` as an optional dependency in PyO3, although I see comments above imply there are internal `chrono` APIs used here for performance. Moving this to PyO3 may not be so good for performance in that case?

davidhewitt:
(I resurrected https://github.com/PyO3/pyo3/pull/1588 which I'll probably merge shortly for PyO3 0.17, which would remove the need for the `unsafe` code here.)

djc:
> @djc - I wonder if you have an opinion where this integration should live? Given that `chrono` is actively maintained again I'm reasonably comfortable with having `chrono` as an optional dependency in PyO3, although I see comments above imply there are internal `chrono` APIs used here for performance. Moving this to PyO3 may not be so good for performance in that case?

I think it would make more sense for the PyO3/chrono integration to live in the PyO3 repository than in the chrono repository: (1) the chrono side of this is relatively low-complexity compared to the PyO3 integration bits, (2) PyO3 is arguably a higher-level dependency than chrono and (3) I think a chrono integration provides more value to the average PyO3 user than a PyO3 integration would provide to the average chrono user (maybe this is a restatement of 2?).

If we identify extra API surface that we need to expose to make that work well, that's certainly a worthy trade-off.

davidhewitt:
"owned" form basically means it will decrease the python reference count later. TBH the only way to know is to look at the docs for the C API - most return owned pointers, in this case it's a bit murky because the C API is a macro which just expands to the static item and PyO3 has to make a function instead.

@pickfire, given the preference above to have this in PyO3, would you be interested in moving this to the PyO3 repo, perhaps after https://github.com/PyO3/pyo3/pull/1588 merges?

src/naive/internals.rs
```diff
@@ -371,7 +371,7 @@ impl fmt::Debug for Of {
 /// (month, day of month and leap flag),
 /// which is an index to the `MDL_TO_OL` lookup table.
 #[derive(PartialEq, PartialOrd, Copy, Clone)]
-pub(super) struct Mdf(pub(super) u32);
+pub(crate) struct Mdf(pub(super) u32);
```
pickfire, line 374: The thing I see is that `Mdf` needs to be exposed, so I am not sure if it is a good tradeoff is the internals are being exposed.

src/naive/time/mod.rs
```diff
@@ -795,7 +795,7 @@ impl NaiveTime {
     }
 
     /// Returns a triple of the hour, minute and second numbers.
-    fn hms(&self) -> (u32, u32, u32) {
+    pub(crate) fn hms(&self) -> (u32, u32, u32) {
```
pickfire, line 798: And these to get internal values.

djc:
I'd be happy to add API surface if that materially speeds up PyO3 access.

davidhewitt:
@pickfire I'll try to merge my PyO3 branch this weekend, it would then be possible for you to open both PyO3 and chrono PRs at the same time by using a git dependency on chrono.

pickfire:
I have something that I wanted to do so I think I will only do this next month.

davidhewitt:
No problem, this has been blocked on me for like a year already 😅

Psykopear:
Hi everybody. I'm interested in an integration between pyo3 and chrono (and possibly chrono-tz too).

If I understand correctly, your current idea is to add the integration in PyO3 rather than here, doing something similar to what was done in this PR, and if needed add things to the public api here, is that right?

I'd like to help move things forward. I can start porting this PR to the PyO3 codebase (under a feature flag, in a `chrono` module inside `src/convertions`), see what's needed on the chrono side, and open other PRs/issues here to discuss changes to the public api.

@davidhewitt did I get this right? I read a lot of discussions/issues/PRs both here and on the PyO3 repo, I might have missed something
@pickfire are you currently working on this? If so, let me know if I can help

pickfire:
I tried this again last week, but I noticed that this needs changes on both chrono and **time** crate, @djc do we have the luxury to be able to change time crate? We need `pub const SECS_PER_DAY` from time crate.

Thanks for the ping @Psykopear, but I only get to work on this once last week and haven't touch it yet, but I will continue on this for now, not sure what help I need yet, but I guess maybe help me follow up on time crate?

Psykopear:
good to hear that @pickfire .
I'm sorry I'm not sure I understand the problem here, does chrono (or the chrono integration for PyO3) requires the *time* crate?
Are we talking about this one: https://github.com/time-rs/time ?

I know chrono used to be built upon time v0.1, but I don't see any reference of that in the current codebase (I ported the PR linked here on top of the current `main` branch of PyO3).

> Now I am just slightly stuck at figuring out how to convert from FixedOffset to PyTimeZone because the API changed a bit, previously I only have to convert from FixedOffset to PyObject, but I believe I can figure this out in a moment.

Yes I was working on that too while porting the PR, I ended up converting it where used (not sure it's the best way, but it works):
```rust
        let tz = self.offset().fix().into_py(py);
        let tz: &PyTzInfo = tz.extract(py).unwrap();
```

djc:
@pickfire on main we no longer depend on time, on the 0.4.x branch we still do. I don't think there's a way to get new time 0.1 releases out, no.

PR #546, opened 2021-03-16
Rust 2018 Edition, MSRV = 1.40 and Windows MSRV tests

spire-ffoston:

      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md

This PR updates chrono to use the Rust 2018 edition. 99% of the code changes are to fix breakages due to 2018's module path specification changes. 

The edition update requires bumping the MSRV version of chrono as `1.16` was released long before the 2018 edition was a thing (it was released on 2016/11/10 according to the blog post). `1.40` was the oldest version I could get working without further changes as some dependencies would fail to compile on older versions.

Finally the new MSRV means that Windows MSRV tests can be reenabled.

### DIFF
--- a/.github/workflows/test-release.yml
+++ b/.github/workflows/test-release.yml
@@ -31,10 +31,11 @@ jobs:
         - os: ubuntu-latest
           rust_version: nightly
         - os: ubuntu-16.04
-          rust_version: 1.13.0
+          rust_version: 1.40.0
         - os: macos-latest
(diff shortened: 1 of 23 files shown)


pickfire:
Isn't 1.40 a bit too recent? Serde still supports 1.13 at this point of time.

djc:
A version of this was done in #653.

Issue #531, opened 2021-01-31
chrono panics on a valid input string


omayerg:
```rust
 use chrono::format;

fn main() {
    let p = &mut format::Parsed::new();
    format::parse(p, "MONYAŽA Šander`",
    format::StrftimeItems::new("%A, %e-%h-%y %H:%M:%S"));
    
    println!("No panic");
}
```

([Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f184e5164c1f41834f9c122991e1932d))

Expected null result, but actually it panics.
It seems chrono parser misses the case where a string starts with a valid datetime part, e.g. SUN, MON, TUE... but then have non-ascii characters. For example, running the above string without letting it start with any three letters that represents a day of week doesn't panic.

pitdicker:
This is fixed by https://github.com/chronotope/chrono/pull/1024.

PR #554, opened 2021-04-14
Check short/long weekday/month by char instead of byte

UnicodingUnicorn:

      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md

---

When looking at Issue #531, I saw that for some reason when parsing short/long months/weekdays, chrono checks the string byte by byte instead of char by char, which gives up the ghost when Unicode gets involved. I don't know if it is some weird optimisation thing I don't get, but I changed it to check by char instead.


pitdicker:
An alternative to this PR was merged in https://github.com/chronotope/chrono/pull/1024.

jtmoon79:
@UnicodingUnicorn you're welcome to add the two test assertions.  But otherwise, according to @pitdicker this can be closed.

PR #555, opened 2021-04-15
Add flexible format supports to DateTime::parse_from_rfc3339

foxhlchen:
DateTime::parse_from_rfc3339 lacks of some iso-8601 format supports which
makes it very hard to collaborate with tools such as GNU date.

This commit adds support for thoses formats generated by GNU data:

date --iso-8601=ns
2021-04-15T09:49:00,321433476+08:00

date --iso-8601=minutes
2021-04-15T09:49+08:00

date --iso-8601=hours
2021-04-15T09+08:00

Signed-off-by: foxhlchen <foxhlchen@gmail.com>

      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -2575,11 +2575,23 @@ mod tests {
             Ok(FixedOffset::east(0).ymd(2015, 2, 18).and_hms(23, 16, 9))
         );
         assert_eq!(
+            DateTime::parse_from_rfc3339("2015-02-18T23+00:00"),
+            Ok(FixedOffset::east(0).ymd(2015, 2, 18).and_hms(23, 0, 0))
+        );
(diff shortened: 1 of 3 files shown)


pitdicker:
I has been two years, but let me leave a comment...

Parsing almost any dates that are basically anything except the RFC3339 format is a pain right now, I am hitting it too. The best you can currently do is define multiple format strings and try until one succeeds.

But modifying RFC3339 does not seem like an option: there are other uses that rely on the strict parsing/validation of this format.

djc:
Going to close this. RFC 3339 is not the same as ISO 8601, and we shouldn't act like it is.

Issue #416, opened 2020-05-02
Implement Duration::num_years()


clawoflight:
It's not entirely trivial to calculate without time zone information, but it would be nice to have.

azzamsa:
`num_years`, `num_months` is nice to have.

So we could have this:

```python
# python
print(f"Your age: {diff.years} years, {diff.months} months, {diff.days} days")
```

quodlibetor:
I would definitely take a PR for this.

azzamsa:
Thank you for taking the step.

I've asked about the situation [here](https://discord.com/channels/442252698964721669/448238009733742612/743020012142461010) and [here](https://discord.com/channels/442252698964721669/448238009733742612/743020838231605249).

Some people in the discord said that chrono doesn't implement this because there is no single answer for `how many days in year`, and `how many days in months`. But, other popular datetime libraries handle them already. Such python datetime std or Pendulum of Python, Moment of js, or Joda of Java.


---

I agree we should have `num_years` and `num_months`. let the chrono handle this. because writing them manually requires so much time and knowledge (such as 356 or 366 days in a year, 28 or 30 or 31 in a year).

Yesterday I try to port some of my daily utilities to Rust. One of them is birthday utilities.

``` Python
#!/usr/bin/env python3
import datetime
from dateutil.relativedelta import relativedelta

current_dt = datetime.datetime.now()
birth_date = datetime.datetime.strptime("2000-11-28", "%Y-%m-%d")
birth_date_formatted = birth_date.strftime("%A, %d %B %Y ")

diff = relativedelta(current_dt, birth_date)  # <---
print(f"Your age: {diff.years} years, {diff.months} months, {diff.days} days") # <---
```
Take a look at the two last lines. I need more than 2 hours to figure it out in rust. If only we had `num_years` and `num_months`. it will take only less than half an hour.

Thank you

arossbell:
Would `num_months` and `num_years` both return i64 types (rather than float types)?
(It appears that `dateutil.relativedelta` acts like that, but making certain for this ticket.)

gpgkd906:
It hard to calculate months and years with only ```Duration``` because  leap years...
so I just create a simple crate for this...
maybe helps?

https://crates.io/crates/date_component

```
use chrono::prelude::*;
use date_component::date_component;

fn main() {
    let date1 = Utc.ymd(2015, 4, 20).and_hms(0, 0, 0);
[...]

PR #557, opened 2021-04-27
Add `elapsed_years` to `Date` and `DateTime`

yozhgoor:

      about adding the PR number)
      we can't reintroduce it?

Related to #416

I have make this PR for retrieve easily an age from a birthdate
The API is similar to [`std::time::Instant`](https://doc.rust-lang.org/std/time/struct.Instant.html#method.elapsed)

[changelog]: ../CHANGELOG.md


quodlibetor:
It seems like this would work better as a method on `chrono::Duration`?

yozhgoor:
I need to use a date here, because years are not constant. A year is on average equals to 52.1428571 weeks so this will be really less accurate on `chrono::Duration`.

src/lib.rs
```diff
@@ -594,6 +594,20 @@ mod datetime;
 pub mod format;
 mod round;
 
+/// Constants that can be used by all components.
+pub mod consts {
+    /// Constants of type `f32`.
+    pub mod f32 {
+        /// Number of weeks in a year.
+        pub const WEEK_PER_YEAR: f32 = 52.1775;
```
cecton, line 602:
```suggestion
        pub const WEEKS_PER_YEAR: f32 = 52.1775;
```
The comment is correct though :sweat_smile:

src/lib.rs
```diff
@@ -594,6 +594,20 @@ mod datetime;
...
+    }
+    /// Constants of type `f64`.
+    pub mod f64 {
+        /// Number of weeks in a year.
+        pub const WEEK_PER_YEAR: f64 = 52.1775;
```
cecton, line 607:
```suggestion
        pub const WEEKS_PER_YEAR: f64 = 52.1775;
```

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Versions with only mechanical changes will be omitted from the following list.
 * Add support for microseconds timestamps serde serialization/deserialization (#304)
 * Fix `DurationRound` is not TZ aware (#495)
 * Implement `DurationRound` for `NaiveDateTime`
+* Add a function that calculate the years elapsed between now and a given `Date` or `DateTime`
```
cecton, line 20:
```suggestion
* Add a function that calculates the number of years elapsed between now and a given `Date` or `DateTime` (#557)
```

yozhgoor:
Thanks @cecton for the spell checking :D

yozhgoor:
I don't know what I do with this PR btw :thinking: 
It's not a problem if this don't fit with the API, at this time we can just close.
I just wanted to solve a problem I had and some others too (see https://github.com/chronotope/chrono/issues/416#issuecomment-687885743).
I don't have the level to trying to calculate the elapsed years on `Duration`. This will make big changes to the API too
This PR is the most accurate i can do for the moment :smile:

cecton:
@Milo123459 this seems ready now :rocket:

yozhgoor:
I fix a conflict in CHANGELOG.md and i think i resolved the issues with `Utc::now` and `Utc::today`.
(I discovered the script to launch the CI locally too)

yozhgoor:
@Milo123459 Can we give it another shot before new changes ?
Thanks

yozhgoor:
So, this seems to be blocked again and I'm not sure why. What needs to be done to move forward? 

This feature cannot be made on Duration because the exact date needs to be known to calculate the number of years exactly. Otherwise it would be an estimation of the numbers of years. My code here gives the exact number of years. 

Do we want this feature or we don't want this feature? I would prefer to get a clear answer if possible. 
Otherwise it feels like I'm being ignored without any reason provided.

CHANGELOG.md
```diff
@@ -744,4 +745,3 @@ and replaced by 0.2.25 very shortly. Duh.)
 ## 0.1.0 (2014-11-20)
 
 The initial version that was available to `crates.io`.
-
```
Milo123459, line 747: Please remove this useless change

src/date.rs
```diff
@@ -275,6 +276,20 @@ impl<Tz: TimeZone> Date<Tz> {
     pub fn naive_local(&self) -> NaiveDate {
         self.date
     }
+
+    /// Retrieve the elapsed years from now to the given [`Date`].
```
cecton, line 280: consistency with other's function docs 
```suggestion
    /// Retrieves the elapsed years from now to the given [`Date`].
```

djc:
Unfortunately `std::convert::TryFrom` is not available in our MSRV target, so this needs some work.

djc:
Sorry, can you rebase this again to pull in the fixes from #706?

PR #559, opened 2021-05-04
Optionally implement Arbitrary for Naive{Date,Time,DateTime}

asayers:
I'm not sure if you guys are interested in taking a change like this.  If you are, I'd be happy to fill in more impls.

### DIFF
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -105,6 +105,15 @@ pub const MIN_DATE: NaiveDate = NaiveDate { ymdf: (MIN_YEAR << 13) | (1 << 4) | 
 /// The maximum possible `NaiveDate` (December 31, 262143 CE).
 pub const MAX_DATE: NaiveDate = NaiveDate { ymdf: (MAX_YEAR << 13) | (365 << 4) | 0o17 /*F*/ };
 
+#[cfg(feature = "arbitrary")]
+impl arbitrary::Arbitrary<'_> for NaiveDate {
+    fn arbitrary(u: &mut arbitrary::Unstructured) -> arbitrary::Result<NaiveDate> {
+        let year = u.int_in_range(MIN_YEAR..=MAX_YEAR)?;
+        let ord = u.int_in_range(1..=366)?;
+        NaiveDate::from_yo_opt(year, ord).ok_or(arbitrary::Error::IncorrectFormat)
+    }
+}
+
 // as it is hard to verify year flags in `MIN_DATE` and `MAX_DATE`,
 // we use a separate run-time test.
 #[test]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ Versions with only mechanical changes will be omitted from the following list.
 * Add support for microseconds timestamps serde serialization/deserialization (#304)
 * Fix `DurationRound` is not TZ aware (#495)
 * Implement `DurationRound` for `NaiveDateTime`
+* Implement `arbitrary::Arbitrary` for `NaiveDate`, `NaiveTime`, and
+  `NaiveDateTime`.  You need to enable the "arbitrary" feature to get these
+  impls.
 
 ## 0.4.19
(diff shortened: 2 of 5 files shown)


quodlibetor:
It seems like `arbitrary` is intended for use by fuzzers, are there examples of its use elsewhere? I don't have a lot of context from this PR.

greyblake:
@quodlibetor I'd love to use it.
It's not only for crazy fuzzers but also for something what is rather a property based testing.
E.g. particulary in my case: I have number of relative big domain models, which then transformed into DB records.
I have also transformer functions that works in the opposite direction.
Currently if I want to unit test it, the amount of tests I have to write is proportional to the complexity of the structures I have.
I could use Arbitrary it would be just a couple lines of code:
* Generate a random domain model
* Transform it to DB records
* Transform DB records back
* Compare the original model with that one we got

This test would be much more smaller and, what is also important, not biased (unlike me, a developer).

djc:
If someone is willing to rebase or resubmit it, I would certainly consider it worthwhile.

greyblake:
@djc Hi! The PR was rebased (unfortunately I also missed that).
Would you consider merging the PR or is there something else is required to get it merged?

src/naive/time/mod.rs
```diff
@@ -190,6 +190,15 @@ pub struct NaiveTime {
     frac: u32,
 }
 
+#[cfg(feature = "arbitrary")]
+impl arbitrary::Arbitrary<'_> for NaiveTime {
+    fn arbitrary(u: &mut arbitrary::Unstructured) -> arbitrary::Result<NaiveTime> {
+        let secs = u.int_in_range(0..=86_400)?;
```
djc, line 196: If I understand correctly these ranges should not be inclusive... a day has up to 86399 seconds, and `frac` is supposed to be smaller than `2_000_000_000`.

See https://github.com/chronotope/chrono/blob/1281ce01e56caf39a1c902086707efd796f6f2f6/src/naive/time/mod.rs#L391.

src/naive/date.rs
```diff
@@ -191,6 +191,15 @@ pub const MIN_DATE: NaiveDate = NaiveDate::MIN;
 #[deprecated(since = "0.4.20", note = "Use NaiveDate::MAX instead")]
 pub const MAX_DATE: NaiveDate = NaiveDate::MAX;
 
+#[cfg(feature = "arbitrary")]
+impl arbitrary::Arbitrary<'_> for NaiveDate {
+    fn arbitrary(u: &mut arbitrary::Unstructured) -> arbitrary::Result<NaiveDate> {
+        let year = u.int_in_range(MIN_YEAR..=MAX_YEAR)?;
+        let ord = u.int_in_range(1..=366)?;
```
conradludgate, line 198: Could we construct the YearFlags and determine the exact ndays in the year so constructing the date wouldn't fail

djc: That doesn't seem like it makes a particularly large difference?

src/naive/time/mod.rs
```diff
@@ -190,6 +190,15 @@ pub struct NaiveTime {
...
+        let frac = u.int_in_range(0..=2_000_000_000)?;
+        Ok(NaiveTime { secs, frac })
```
5225225, line 198: This feels like something that might be better to use `from_num_seconds_from_midnight_opt` (and unwrap it with a clear error that it is a `chrono` bug if it fails). That way you can't silently make invalid `NaiveTime`s.

And yeah, a secs of 86_400 *will* be rejected.

greyblake:
@asayers @djc @conradludgate 

I am very interested in this support of Arbitrary. Since @asayers apparently has no time to address the code review, I've done it on my own and opened a new PR: https://github.com/chronotope/chrono/pull/848    (which is alternative to this one)

greyblake:
@djc This one can be closed, since https://github.com/chronotope/chrono/pull/849 was merged.

Issue #560, opened 2021-05-06
Unable to parse millisecond unix timestamps


jszwedko:
Hey all,

I might be missing something, but I'm not seeing a way to use `parse_from_str` with unix timestamps that include milliseconds.

I would have expected something like:

```rust
fn main() {
    use chrono::TimeZone;
    let dt = chrono::offset::Utc.datetime_from_str("1620309579123", "%s%3f");
    println!("{:?}", dt);
}
```

to work, but it gives:

```
Err(ParseError(TooShort))
```

Is this possible or could it be added? I could try taking a shot at a PR.

jgoday:
@jszwedko Seems that Timestmap in fn parse_internal (format/parse.rs), try to consume all digits without taking into account next items/formats. I have changed it to allow for the next fixed item length, in case there is no separator between them (Nanosecond3NotDot in this case).

```rust
let dt = chrono::offset::Utc.datetime_from_str("1620309579123", "%s%3f");
```

should return now
```
Ok(2021-05-06T13:59:39.123)
```

Can make a PR if this is a valid approach.

jszwedko:
Nice, thanks @jgoday ! That seems like a reasonable approach to me.

yalinglee:
Hi! Would it be possible to include this fix in the next release? Thanks!

PR #561, opened 2021-05-14
#560: Unable to parse millisecond unix timestamps

jgoday:

- [X] Have you added yourself and the change to the [changelog]? (Don't worry
      about adding the PR number)
- [X] If this pull request fixes a bug, does it add a test that verifies that
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md

This PR tries to solve #560.
When parsing Timestamps, it allows to check the next item fixed length to avoid Timestamp consuming all the remain digits (see #560  example with nanoseconds **"%s%3f"**).
It groups items in tuples (current and next item) inside parse_internal, using slice windows.

### DIFF
--- a/src/format/parse.rs
+++ b/src/format/parse.rs
@@ -215,6 +215,20 @@ fn parse_rfc3339<'a>(parsed: &mut Parsed, mut s: &'a str) -> ParseResult<(&'a st
     Ok((s, ()))
 }
 
+fn get_fixed_item_len<'a, B>(item: Option<&B>) -> Option<usize>
+where
+    B: Borrow<Item<'a>>,
+{
+    use super::Fixed::*;
+
+    item.map(|i| match *i.borrow() {
+        Item::Fixed(Internal(InternalFixed { val: InternalInternal::Nanosecond3NoDot })) => 3,
+        Item::Fixed(Internal(InternalFixed { val: InternalInternal::Nanosecond6NoDot })) => 6,
+        Item::Fixed(Internal(InternalFixed { val: InternalInternal::Nanosecond9NoDot })) => 9,
+        _ => 0,
+    })
+}
+
 /// Tries to parse given string into `parsed` with given formatting items.
 /// Returns `Ok` when the entire string has been parsed (otherwise `parsed` should not be used).
 /// There should be no trailing string after parsing;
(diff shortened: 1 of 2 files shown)


jszwedko:
Circling back to this. This seems like a reasonable approach to me.

src/format/parse.rs
```diff
@@ -316,8 +340,10 @@ where
                     Minute => (2, false, Parsed::set_minute),
                     Second => (2, false, Parsed::set_second),
                     Nanosecond => (9, false, Parsed::set_nanosecond),
-                    Timestamp => (usize::MAX, false, Parsed::set_timestamp),
-
+                    Timestamp => {
+                        let next_size = get_fixed_item_len(next_item).unwrap_or(0);
+                        (s.len() - next_size, false, Parsed::set_timestamp)
```
meiamsome, line 345: Using `s.len()` here is problematic for parsing any formats where the timestamp does not terminate at the end of the string. 

For example, I have seen a format that is like `%s%3f%z` - so `123456789+0000` - in this case you hit `%s` consuming all of `123456789` because it is limited to the max of `123456789+0`.

Another case I've seen is `Date(%s%3f)` - so `Date(123456789)` - in this case you hit `%s` consuming all of `1234567` because that is the max, but `%3f` is left with three extra characters.

It seems the only real way would be to do something like `scan::number` to read however many chars `Timestamp` would capture and then go from there. Sadly, the return of `scan::number` doesn't currently include the number of characters read, instead returning a new slice pointer so that is not a trivial fix.

I don't know if it's really necessary to fix this for those formats, if they should just be a separate issue, or if they should just not be fixed by the library and left to consumers to implement tricky logic.

pitdicker:
I agree this is worth fixing. Combining `%s%3f` to parse a millisecond timestamp is clever. On the other hand it feels just a bit too magical.

An alternative is to add new specifiers to allow parsing millisecond, microsecond, and nanosecond timestamps as a single integer. @djc any preferences?

Also @meiamsome raised a good issue about the implementation here. @jgoday This PR is near the end of its second year... Would you still be available to push it forwards?

pitdicker:
The approach in this PR seems wordt persuing to me, but in a more general way.

It would be nice to be able to parse `63015` to `6:30:15` with `%-H%M%S`. I.e. support parsing one variable-length number followed a one or more fixed-length numbers.

djc:
Given the lack of response from the original submitter, going to just close this for now. IMO the original issue isn't a high priority.

jtmoon79:
> An alternative is to add new specifiers to allow parsing millisecond, microsecond, and nanosecond timestamps as a single integer. @djc any preferences?

This is a worthy improvement.

djc:
Worthy how? Have you personally had a use case for this? Given the relative lack of resources I would like to avoid adding features people just think of.

jszwedko:
The lack of support for parsing string millisecond timestamps does come up in https://vector.dev/ occasionally, where we use `chrono` for date/timestamp parsing. Recent example: https://github.com/vectordotdev/vector/discussions/19577

djc:
@jszwedko please open a new issue explaining your use case.

jszwedko:
> @jszwedko please open a new issue explaining your use case.

I'd actually opened the original issue 🙂 It's here: https://github.com/chronotope/chrono/issues/560

djc:
Okay, I'm open to reviewing a new PR that takes into account the feedback on this PR so far.

titaneric:
I would like to resolve this issue based on this implemetation!

Issue #553, opened 2021-04-14
Update the time dependency


stanislav-tkach:
It would be nice to update the `time` dependency to the `0.2` version before a new release (#552).

As far as I can see, the update isn't that straightforward because some functions are deprecated and a replacement for them returns `i128` instead of `i64` which requires some code tweaks.

dallaires:
I also think it would be a good idea to update the dependency. I'm doing a project using Rocket and Diesel and I have to separate both projects to manage cookies because Rocket depends on `time 0.2` and Diesel depends on this project which indirectly depends on `time 0.1` because of `chrono`. I can lend a hand if this issue can be resolved by a new rustacean. Let me know if you want me to do a pull request on this.

timvisee:
I'm looking into updating `time` to `0.2` now.

I think changing the `chrono` API to match the `time` type changes is the correct approach, so it'll definitely be a breaking change.

timvisee:
I did some initial work on this, but got stuck at a failing test. I'd be very happy if someone else could take a look at this as well.

See https://github.com/chronotope/chrono/pull/567

bootrecords:
I'd like to point out that the `time` crate is looking to be updated to version `0.3` with major changes soon. It might make sense to rather look towards bumping the `time` dependency to `0.3` directly, as `0.2` also has an unmaintained dependency (`stdweb`), which will be dropped with `0.3`.

DarthHater:
An FYI, `time` before `2.2.23` has a vulnerability, so updating this would be helpful. Info on the vulnerability here: https://github.com/RustSec/advisory-db/blob/main/crates/time/RUSTSEC-2020-0071.md

timvisee:
@DarthHater thanks for pointing this out!

The PR (https://github.com/chronotope/chrono/pull/567) has been updated to use `^0.2.23`, thanks to @blackghost1987.

All tasks are now resolved, so I marked it as ready for review. It's now up to the `chrono` maintainers to merge this.

bootrecords:
Not to nag, but as previously pointed out (though without reference), [`time v0.3` appears to be around the corner](https://github.com/time-rs/time/issues/248) - may be worth syncing up with those developments?

timvisee:
`time v0.3` seems to be scheduled for release tomorrow, with various breaking changes.

[...]

PR #567, opened 2021-06-04
Bump time to 0.2.23

timvisee:
Fixes https://github.com/chronotope/chrono/issues/553

This updates the `time` dependency to `0.2.23`. The crate has been refactored to follow `time` API changes.

~~This is mostly done, but still requires some work. I'm seeing a failing test ([this](https://github.com/timvisee/chrono/blob/b8bdf5a13d50ba15573ccadf0e566c6c23527c44/src/naive/datetime.rs#L2343) case fails) for which I don't clearly see the problem nor a solution.~~

~~@ quodlibetor Do you have some insight on this failing test?~~

### Tasks:
- [x] Bump `time` to `0.2.23`
- [x] Refactor codebase for `time` API changes
- [x] Fix test `naive::datetime::tests::test_datetime_add` ([failing case](https://github.com/timvisee/chrono/blob/b8bdf5a13d50ba15573ccadf0e566c6c23527c44/src/naive/datetime.rs#L2343)) (thanks @blackghost1987!)

[changelog]: ../CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Versions with only mechanical changes will be omitted from the following list.
 * Add support for microseconds timestamps serde serialization/deserialization (#304)
 * Fix `DurationRound` is not TZ aware (#495)
 * Implement `DurationRound` for `NaiveDateTime`
+* Bump `time` dependency to 0.2.23 (#567)
 
 ## 0.4.19
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ __doctest = []
 
 [dependencies]
 libc = { version = "0.2.69", optional = true }
-time = { version = "0.1.43", optional = true }
+time = { version = "0.2", optional = true }
 num-integer = { version = "0.1.36", default-features = false }
 num-traits = { version = "0.2", default-features = false }
 rustc-serialize = { version = "0.3.20", optional = true }
--- a/src/format/parsed.rs
+++ b/src/format/parsed.rs
@@ -1110,15 +1110,15 @@ mod tests {
         let min_days_from_year_1970 =
             MIN_DATE.signed_duration_since(NaiveDate::from_ymd(1970, 1, 1));
         assert_eq!(
-            parse!(timestamp: min_days_from_year_1970.num_seconds()),
+            parse!(timestamp: min_days_from_year_1970.whole_seconds()),
             ymdhms(MIN_DATE.year(), 1, 1, 0, 0, 0)
         );
         assert_eq!(
-            parse!(timestamp: year_0_from_year_1970.num_seconds()),
+            parse!(timestamp: year_0_from_year_1970.whole_seconds()),
             ymdhms(0, 1, 1, 0, 0, 0)
         );
         assert_eq!(
-            parse!(timestamp: max_days_from_year_1970.num_seconds() + 86399),
+            parse!(timestamp: max_days_from_year_1970.whole_seconds() + 86399),
             ymdhms(MAX_DATE.year(), 12, 31, 23, 59, 59)
         );
 
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -124,7 +124,7 @@ fn test_date_bounds() {
 
     // let's also check that the entire range do not exceed 2^44 seconds
     // (sometimes used for bounding `Duration` against overflow)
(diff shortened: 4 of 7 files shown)


blackghost1987:
I've tried to debug this, and solved the original panic, but the test case still fails with some overflow errors. Here are my results so far, maybe it will help somebody. I will investigate it further though and get back here if I've got anything.

So the failing test case is `test_datetime_add` and it fails in the last line, when testing the handling of the minimal duration [here](https://github.com/timvisee/chrono/blob/b8bdf5a13d50ba15573ccadf0e566c6c23527c44/src/naive/datetime.rs#L2343).
It will eventually call [NaiveTime::overflowing_add_signed](https://github.com/timvisee/chrono/blob/b8bdf5a13d50ba15573ccadf0e566c6c23527c44/src/naive/time.rs#L532), which causes the panic in [this line](https://github.com/timvisee/chrono/blob/b8bdf5a13d50ba15573ccadf0e566c6c23527c44/src/naive/time.rs#L558).

The problematic call is a simple subtraction of two `time::Duration` (OldDuration) instances. Duration subtraction is done by negating the input and using addition, but i64::min_value() cannot be negated, at least not inside i64. Fortunately that subtraction can be avoided, because the whole point of it is to get the nanoseconds part of the Duration. The nanoseconds part is stored in a field in v0.2 and can be accessed by the `subsec_nanoseconds` function, like so:
```
        let rhssecs = rhs.whole_seconds();
        let rhsfrac = rhs.subsec_nanoseconds();
        debug_assert_eq!(
            OldDuration::seconds(rhssecs) + OldDuration::nanoseconds(rhsfrac as i64),
            rhs
        );
```

Unfortunately now the code panics at the overflow calculations [here](https://github.com/timvisee/chrono/blob/b8bdf5a13d50ba15573ccadf0e566c6c23527c44/src/naive/time.rs#L588)

blackghost1987:
I think I figured this out. So the last part of `NaiveTime::overflowing_add_signed` is trying to make sure the result is a valid NaiveTime, so `0 < secs < 86400`. It manages this by adding or removing whole days while maintaining the `morerhssecs` which is the remainder value in seconds. The problem can be reproduced with the following code:
```
NaiveTime::from_hms(0, 0, 0).overflowing_add_signed(Duration::min_value())
```
The result is correctly calculated to be 08:29:51.000000001 but the remainder is so huge, it cannot be stored in an i64. We could detect the overflow on the subtraction, but the `overflowing_add_signed` function has no ways of reporting an overflow, it supposed to always work.

This could only be solved by changing the return type of the function, which would be a breaking change!

Maybe a better alternative would be to consider this as a "known limitation" of `NaiveTime`, and solve this in `NaiveDateTime`. The `NaiveDateTime::checked_add_signed` already returns None if there's an overflow, so maybe we could add some checks to it to prevent this. The date handling part already checks the remainder value received from `NaiveTime::overflowing_add_signed` against a MAX_SECS_BITS to see if the Duration::seconds function would overflow, and if so it returns early. Maybe we could add a check like that _before_ the overflowing_add_signed call, and return early if it would fail.

timvisee:
> (...)
> 
> This could only be solved by changing the return type of the function, which would be a breaking change!
> 
> Maybe a better alternative would be to consider this as a "known limitation" of `NaiveTime`, and solve this in `NaiveDateTime`. The `NaiveDateTime::checked_add_signed` already returns None if there's an overflow, so maybe we could add some checks to it to prevent this. The date handling part already checks the remainder value received from `NaiveTime::overflowing_add_signed` against a MAX_SECS_BITS to see if the Duration::seconds function would overflow, and if so it returns early. Maybe we could add a check like that _before_ the overflowing_add_signed call, and return early if it would fail.

Do you suggest to resolve this (here) before merging this PR?

blackghost1987:
> Do you suggest to resolve this (here) before merging this PR?
 I already did it, and you already merged it 🙂 I just copied the existing MAX_SECS_BITS check to check it _before and after_ the addition as well, and it did the trick. I'm not sure if it's required after the addition, but it doesn't hurt.

timvisee:
Superseded by https://github.com/chronotope/chrono/pull/578, which implements these changes with `time v0.3` instead.

djc:
I don't think we'll be updating the version of time used in chrono. Since #478, chrono has a minimal dependency on time, and in the next semver-incompatible version we'll remove the dependency entirely.

(Note that chrono is unaffected by https://rustsec.org/advisories/RUSTSEC-2020-0071.)

Issue #519, opened 2021-01-03
Build fails on target wasm32-unknown-emscripten with: target wasm32-unknown-emscripten


bobajeff:
Here's my shell output:
```bash
(base) [bobajeff@bobajeff-pc chrono]$ cargo build --release --target wasm32-unknown-emscripten
    Updating crates.io index
   Compiling autocfg v1.0.1
   Compiling libc v0.2.81
   Compiling num-traits v0.2.14
   Compiling num-integer v0.1.44
   Compiling time v0.1.44
   Compiling chrono v0.4.19 (/home/bobajeff/Rust/webassembly/chrono/chrono)
error[E0428]: the name `inner` is defined multiple times
  --> src/sys.rs:25:1
   |
21 | mod inner;
   | ---------- previous definition of the module `inner` here
...
25 | mod inner;
   | ^^^^^^^^^^ `inner` redefined here
   |
   = note: `inner` must be defined only once in the type namespace of this module

error: aborting due to previous error

For more information about this error, try `rustc --explain E0428`.
error: could not compile `chrono`

To learn more, run the command again with --verbose.
(base) [bobajeff@bobajeff-pc chrono]$ 
```

ch1seL:
Any plans for the release?

PR #568, opened 2021-06-05
Fix compiling for wasm32-unknown-emscripten target #519

orion78fr:
I encountered a bug while compiling for the wasm32-unknown-emscripten target (https://github.com/orion78fr/godot_keepass_rust_totp/runs/2742402148?check_suite_focus=true) and it seems I wasn't the only one :  #519 

I don't know if you really need (and how to do) a test for this as it seems that it haven't been caught by your CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Versions with only mechanical changes will be omitted from the following list.
 * Add support for microseconds timestamps serde serialization/deserialization (#304)
 * Fix `DurationRound` is not TZ aware (#495)
 * Implement `DurationRound` for `NaiveDateTime`
+* Correct build for wasm32-unknown-emscripten target
 
 ## 0.4.19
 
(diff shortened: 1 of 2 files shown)


jcaesar:
If wasm32-unknown-emscripten is `cfg(unix)`, wouldn't it be better to use the unix mod for it, instead of the stub one? It does compile, but I haven't tested whether it does anything meaningful.

(Nevermind the fact that I submitted a PR to fix the same problem without looking whether somebody already did. Sorry for that.)

hoodmane:
> wouldn't it be better to use the unix mod for it, instead of the stub one? It does compile, but I haven't tested whether it does anything meaningful.

Yes it is better. Emscripten has working implementations of `timegm`, `mktime`, and `localtime_r`.

hoodmane:
Both #574 and #607 make the correct choice to use the unix one on emscripten target but this PR doesn't. Hopefully one of those two PRs can be merged.

orion78fr:
@hoodmane #593 has been merged in the meantime, no need to comment on an old merged PR from 6 month ago

hoodmane:
Ah great, thanks!

djc:
Actually the fix was merged after @hoodmane made their comment and their comment helped me find #593. So thanks for that!

PR #574, opened 2021-07-15
Fix duplicate module import under wasm32-unknown-emscripten

jcaesar:
      about adding the PR number)
      
  No, feels way too small. Should I anyway?      
      
      we can't reintroduce it?
      
  No. Do you want me to add a new architecture to the CI setup?
      

[changelog]: ../CHANGELOG.md


jcaesar:
My apologies, this is a duplicate of #568.

PR #578, opened 2021-07-29
Bump time to 0.3

timvisee:
Fixes https://github.com/chronotope/chrono/issues/553, https://github.com/chronotope/chrono/pull/567

This updates the `time` dependency to `0.3`. The crate has been refactored to follow `time` API changes.

This includes not-yet-merged changes made in https://github.com/chronotope/chrono/pull/567.

### Tasks:
- [x] Bump `time` to `0.3` when released (see https://github.com/time-rs/time/issues/248)
- [x] Refactor codebase for `time` API changes
- [x] Fix test `naive::date::tests::test_date_add` ([failing case](https://github.com/timvisee/chrono/blob/eff91806cbe009caebc9bbf3b3879764d4d45271/src/naive/date.rs#L2224))
- [x] Fix test `naive::datetime::tests::test_datetime_add` ([failing case](https://github.com/timvisee/chrono/blob/eff91806cbe009caebc9bbf3b3879764d4d45271/src/naive/datetime.rs#L2348))

[changelog]: ../CHANGELOG.md

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ __doctest = []
 
 [dependencies]
 libc = { version = "0.2.69", optional = true }
-time = { version = "^0.2.23", optional = true }
+time = { version = "0.3", optional = true }
 num-integer = { version = "0.1.36", default-features = false }
 num-traits = { version = "0.2", default-features = false }
 rustc-serialize = { version = "0.3.20", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ __doctest = []
 
 [dependencies]
 libc = { version = "0.2.69", optional = true }
-time = { version = "0.1.43", optional = true }
+time = { version = "0.2", optional = true }
 num-integer = { version = "0.1.36", default-features = false }
 num-traits = { version = "0.2", default-features = false }
 rustc-serialize = { version = "0.3.20", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ __doctest = []
 
 [dependencies]
 libc = { version = "0.2.69", optional = true }
-time = { version = "0.2", optional = true }
+time = { version = "^0.2.23", optional = true }
 num-integer = { version = "0.1.36", default-features = false }
 num-traits = { version = "0.2", default-features = false }
 rustc-serialize = { version = "0.3.20", optional = true }
--- a/.github/workflows/test-release.yml
+++ b/.github/workflows/test-release.yml
@@ -31,9 +31,9 @@ jobs:
         - os: ubuntu-latest
           rust_version: nightly
         - os: ubuntu-18.04
-          rust_version: 1.13.0
+          rust_version: 1.51.0
         - os: macos-latest
-          rust_version: 1.13.0
+          rust_version: 1.51.0
           # time doesn't work on windows with 1.13
 
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,10 +39,11 @@ jobs:
         - os: ubuntu-latest
           rust_version: nightly
         - os: ubuntu-18.04
-          rust_version: 1.48.0
+          rust_version: 1.51.0
         - os: macos-latest
-          rust_version: 1.48.0
-          # time doesn't work on windows with 1.48
+          rust_version: 1.51.0
+        - os: windows-latest
+          rust_version: 1.51.0
 
     runs-on: ${{ matrix.os }}
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,10 +39,10 @@ jobs:
         - os: ubuntu-latest
           rust_version: nightly
         - os: ubuntu-18.04
-          rust_version: 1.13.0
+          rust_version: 1.48.0
         - os: macos-latest
-          rust_version: 1.13.0
-          # time doesn't work on windows with 1.13
+          rust_version: 1.48.0
+          # time doesn't work on windows with 1.48
 
     runs-on: ${{ matrix.os }}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Versions with only mechanical changes will be omitted from the following list.
 * Add support for microseconds timestamps serde serialization/deserialization (#304)
 * Fix `DurationRound` is not TZ aware (#495)
 * Implement `DurationRound` for `NaiveDateTime`
+* Bump `time` dependency to 0.3 (#578)
 
 ## 0.4.19
 
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ You can get the current date and time in the UTC time zone
 or in the local time zone
 ([`Local::now()`](https://docs.rs/chrono/0.4/chrono/offset/struct.Local.html#method.now)).
 
-```rust
+```rust,ignore
 use chrono::prelude::*;
 
 let utc: DateTime<Utc> = Utc::now();       // e.g. `2014-11-28T12:45:59.324310806Z`
@@ -150,7 +150,7 @@ Alternatively, you can create your own date and time.
 This is a bit verbose due to Rust's lack of function and method overloading,
 but in turn we get a rich combination of initialization methods.
 
-```rust
+```rust,ignore
 use chrono::prelude::*;
 use chrono::offset::LocalResult;
 
@@ -334,7 +334,7 @@ from a [`DateTime`](https://docs.rs/chrono/0.4/chrono/struct.DateTime.html). Add
 [`DateTime.timestamp_subsec_nanos`](https://docs.rs/chrono/0.4/chrono/struct.DateTime.html#method.timestamp_subsec_nanos)
 to get the number of additional number of nanoseconds.
 
-```rust
+```rust,ignore
 // We need the trait in scope to use Utc::timestamp().
 use chrono::{DateTime, TimeZone, Utc};
 
@@ -353,7 +353,7 @@ Chrono also provides an individual date type ([**`Date`**](https://docs.rs/chron
 It also has time zones attached, and have to be constructed via time zones.
 Most operations available to `DateTime` are also available to `Date` whenever appropriate.
 
-```rust
+```rust,ignore
 use chrono::prelude::*;
 use chrono::offset::LocalResult;
 
--- a/ci/github.sh
+++ b/ci/github.sh
@@ -29,7 +29,7 @@ meaningful in the github actions feature matrix UI.
 
     runv cargo --version
 
-    if [[ ${RUST_VERSION:-} != 1.13.0 ]]; then
+    if [[ ${RUST_VERSION:-} != 1.48.0 ]]; then
         if [[ ${WASM:-} == yes_wasm ]]; then
             test_wasm
         elif [[ ${WASM:-} == wasm_simple ]]; then
@@ -43,7 +43,7 @@ meaningful in the github actions feature matrix UI.
         else
             test_regular UTC0
         fi
-    elif [[ ${RUST_VERSION:-} == 1.13.0 ]]; then
+    elif [[ ${RUST_VERSION:-} == 1.48.0 ]]; then
         test_113
     else
         echo "ERROR: didn't run any tests"
--- a/src/format/mod.rs
+++ b/src/format/mod.rs
@@ -16,7 +16,7 @@
 //! C's `strftime` format. The available options can be found [here](./strftime/index.html).
 //!
 //! # Example
-//! ```rust
+//! ```rust,ignore
 //! # use std::error::Error;
 //! #
 //! # fn main() -> Result<(), Box<dyn Error>> {
--- a/src/format/parsed.rs
+++ b/src/format/parsed.rs
@@ -1110,15 +1110,15 @@ mod tests {
         let min_days_from_year_1970 =
             MIN_DATE.signed_duration_since(NaiveDate::from_ymd(1970, 1, 1));
         assert_eq!(
-            parse!(timestamp: min_days_from_year_1970.num_seconds()),
+            parse!(timestamp: min_days_from_year_1970.whole_seconds()),
             ymdhms(MIN_DATE.year(), 1, 1, 0, 0, 0)
         );
         assert_eq!(
-            parse!(timestamp: year_0_from_year_1970.num_seconds()),
+            parse!(timestamp: year_0_from_year_1970.whole_seconds()),
             ymdhms(0, 1, 1, 0, 0, 0)
         );
         assert_eq!(
-            parse!(timestamp: max_days_from_year_1970.num_seconds() + 86399),
+            parse!(timestamp: max_days_from_year_1970.whole_seconds() + 86399),
             ymdhms(MAX_DATE.year(), 12, 31, 23, 59, 59)
         );
 
--- a/src/format/parsed.rs
+++ b/src/format/parsed.rs
@@ -1109,16 +1109,34 @@ mod tests {
             NaiveDate::from_ymd(0, 1, 1).signed_duration_since(NaiveDate::from_ymd(1970, 1, 1));
         let min_days_from_year_1970 =
             MIN_DATE.signed_duration_since(NaiveDate::from_ymd(1970, 1, 1));
+        #[cfg(feature = "oldtime")]
         assert_eq!(
             parse!(timestamp: min_days_from_year_1970.whole_seconds()),
             ymdhms(MIN_DATE.year(), 1, 1, 0, 0, 0)
         );
+        #[cfg(not(feature = "oldtime"))]
+        assert_eq!(
+            parse!(timestamp: min_days_from_year_1970.num_seconds()),
+            ymdhms(MIN_DATE.year(), 1, 1, 0, 0, 0)
+        );
+        #[cfg(feature = "oldtime")]
         assert_eq!(
             parse!(timestamp: year_0_from_year_1970.whole_seconds()),
             ymdhms(0, 1, 1, 0, 0, 0)
         );
+        #[cfg(not(feature = "oldtime"))]
+        assert_eq!(
+            parse!(timestamp: year_0_from_year_1970.num_seconds()),
+            ymdhms(0, 1, 1, 0, 0, 0)
+        );
+        #[cfg(feature = "oldtime")]
         assert_eq!(
             parse!(timestamp: max_days_from_year_1970.whole_seconds() + 86399),
+            ymdhms(MAX_DATE.year(), 12, 31, 23, 59, 59)
+        );
+        #[cfg(not(feature = "oldtime"))]
+        assert_eq!(
+            parse!(timestamp: max_days_from_year_1970.num_seconds() + 86399),
             ymdhms(MAX_DATE.year(), 12, 31, 23, 59, 59)
         );
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -126,7 +126,7 @@
 //! or in the local time zone
 //! ([`Local::now()`](./offset/struct.Local.html#method.now)).
 //!
-//! ```rust
+//! ```rust,ignore
 //! use chrono::prelude::*;
 //!
 //! let utc: DateTime<Utc> = Utc::now();       // e.g. `2014-11-28T12:45:59.324310806Z`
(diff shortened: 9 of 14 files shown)


timvisee:
The two failing tests are caused by a negating overflow panic for `Duration::MIN`. This fails:

```rust
-Duration::MIN
```

clux:
I understand that people are busy, but is there any chance that this PR in particular could be expedited? :pray:

Due to the linked CVE, `cargo-audit` is about to hit the entire ecosystem.

quodlibetor:
CI should run if you rebase on main.

Milo123459:
Well, I think this is reasonable to be 1.0, but you have logic so I think a semver minor version increment would be better. Even

quodlibetor:
## On bumping straight to 1.0.

Chrono's API is known-bad, unfortunately, as it was written before the rust API guidelines were a thing. Fixing the stupid weirdnesses in the API are the goal of 0.5.

## On design for this

Actually thinking through it we shouldn't add a `time03` feature: it would be best to leave the `time` feature alone and just change the implementation of the `clock` feature to use time 0.3 internally. The fact that `time` is part of the public API is a mistake and I'd rather not propagate it -- other crates can be responsible for converting between `time` and `chrono` types, and I'd be happy to adopt one that does something like that into the chronotope organization if anyone wants to write and maintain it.

quodlibetor:
Thanks very much @timvisee 

> This thread became somewhat noisy.

lol

quodlibetor:
Darn, I forgot that one of the reasons I didn't bump `time` was because it doesn't have the same rustc version compatibility as Chrono.

There's nothing for you to change, I'll take a look and... think about what to do. I don't want to reduce Chrono's support to just "last 6 months of rustc" but depending on how much would need to change to support the clock feature I might get forced into it because of this CVE.

quodlibetor:
Oh yeah time does indeed bump MSRV very eagerly: https://github.com/time-rs/time/commit/e4d09e20e5c44d20ae4c68aa0797c672dd0e713d

Cargo.toml
```diff
@@ -36,7 +36,7 @@ __doctest = []
 
 [dependencies]
 libc = { version = "0.2.69", optional = true }
-time = { version = "0.1.43", optional = true }
+time = { version = "0.3", optional = true }
```
Milo123459, line 39: 0.3.4 was recently released, perhaps use it here?

Milo123459:
```suggestion
time = { version = "0.3.4", optional = true }
```

CryZe: There's no reason to bump a minor version unless it is actually required for compiling.

Milo123459: Odd, my local rust was complaining about something. Shall look into it.

Milo123459: Oh jeez, my bad. A branch didn't get cut off properly, ignore this.

pymongo:
Please add `const fn` to chrono::Duration methods

djc:
@Milo123459 since you appear to be the most active maintainer, what's your plan/timeline for addressing the https://rustsec.org/advisories/RUSTSEC-2020-0159 issue(s)? I'd like to deal with the issues raised for that advisory, but in some of my projects time 0.3 by itself is not a viable replacement, so I'd like to understand what your plans are.

Milo123459:
> @Milo123459 since you appear to be the most active maintainer, what's your plan/timeline for addressing the [rustsec.org/advisories/RUSTSEC-2020-0159](https://rustsec.org/advisories/RUSTSEC-2020-0159) issue(s)? I'd like to deal with the issues raised for that advisory, but in some of my projects time 0.3 by itself is not a viable replacement, so I'd like to understand what your plans are.
> 
> * This PR has been open since Jul 29
> * The advisory was published on Oct 18
> * You last reviewed this on Nov 1

I would absolutely love to get this merged ASAP, but after speaking to Brandon, I don't think I can merge this. As much as  I would want to, I think Brandon has to.

krojew:
@Milo123459 given the importance of this issue, is there a plan for moving forward in the near future?

Milo123459:
> @Milo123459 given the importance of this issue, is there a plan for moving forward in the near future?

I emailed Brandon last week and haven't got a response, sadly. I'll probably just kind of merge it, but, I don't have publish perms to crates.io (I also emailed Brandon). I'm in a tough spot.

krojew:
If the lack of communication persists, would you consider publishing a new crate e.g. `chrono2`? For better or worse, that seems to be the pattern in such situations.

Milo123459:
> If the lack of communication persists, would you consider publishing a new crate e.g. `chrono2`? For better or worse, that seems to be the pattern in such situations.

Not a bad idea. I just sent a follow up email, but, we can only wait. This should be considered a last resort, as, who is going to migrate their crate to use `chrono2`. I'm definitely trying to get this done ASAP.

Milo123459:
> Please add `const fn` to chrono::Duration methods

Why? This seems unrelated.

Milo123459:
@timvisee Hey, am I all good to fix the CI checks, or, do you want to do that?

timvisee:
> @timvisee Hey, am I all good to fix the CI checks, or, do you want to do that?

Please do!

Do you have commit access to my branch?

Milo123459:
No, I don't! 
Please grant this and I'll fix it up :)

timvisee:
> No, I don't! Please grant this and I'll fix it up :)

I believe you do, as you've [committed](https://github.com/timvisee/chrono/commit/b15e2f2026c55aef264e5cf511587949bdf5bf9e) on on it 2 hours ago. In any case, I've granted perms explicitly.

Milo123459:
> > No, I don't! Please grant this and I'll fix it up :)
> 
> I believe you do, as you've [committed](https://github.com/timvisee/chrono/commit/b15e2f2026c55aef264e5cf511587949bdf5bf9e) on on it 2 hours ago. In any case, I've granted perms explicitly.

Thanks for granting the permissions!
I committed because that was resolving a conflict, and, I don't need explicit permissions for that.

Milo123459:
> I'll message the owner to see what we can do about this.
> 
> You may temporarily switch to this PR by putting the following in your `Cargo.toml` until this is resolved:
> 
> ```toml
> [patch.crates-io]
> chrono = { git = "https://github.com/timvisee/chrono", branch = "time-0.3" }
> ```
> 
> Edit: this doesn't fully fix it according to [this](https://github.com/chronotope/chrono/issues/602#issuecomment-940445390), as `chrono` itself has a similar problem.

This PR doesn't seem to fix the entire vulnerability as the implementation of the function doesn't change. Do you think you might be able to add that in this PR?

timvisee:
> Do you think you might be able to add that in this PR?

Do you mean, whether I can fix it in this branch? Or whether I can update the PR message?

Milo123459:
> > Do you think you might be able to add that in this PR?
> 
> Do you mean, whether I can fix it in this branch? Or whether I can update the PR message?

The former, fixing it. But, don't feel pressured into it, completely up to you.

timvisee:
@Milo123459 IIRC this isn't fixable without breakage.

Milo123459:
> @Milo123459 IIRC this isn't fixable without breakage.

I know.

I don't want apps to break - but at the end of the day, security should come first. This is something I'm willing to stand by, thus, me thinking it is a good change.

timvisee:
> > @Milo123459 IIRC this isn't fixable without breakage.
> 
> I know.
> 
> I don't want apps to break - but at the end of the day, security should come first. This is something I'm willing to stand by, thus, me thinking it is a good change.

I'm not sure when I'd have time to properly look into this. Feel free to give it a go in the mean time.

Milo123459:
> > > @Milo123459 IIRC this isn't fixable without breakage.
> > 
> > 
> > I know.
> > I don't want apps to break - but at the end of the day, security should come first. This is something I'm willing to stand by, thus, me thinking it is a good change.
> 
> I'm not sure when I'd have time to properly look into this. Feel free to give it a go in the mean time.

Ah, alrighty.

leo60228:
I don't think chrono should make a breaking change until it's decided whether the bug is even in chrono. From what I've seen, there isn't a consensus that the root issue isn't in libstd.

Milo123459:
> I don't think chrono should make a breaking change until it's decided whether the bug is even in chrono. From what I've seen, there isn't a consensus that the root issue isn't in libstd.

The audit page says it is coming from this function

leo60228:
Yes. From what I've seen, there isn't a consensus that the advisory is correct.

leo60228:
To clarify, if it didn't involve a breaking change I'd say chrono should be fixed either way, but I don't think a breaking change should be made if it isn't actually necessary.

Milo123459:
> To clarify, if it didn't involve a breaking change I'd say chrono should be fixed either way, but I don't think a breaking change should be made if it isn't actually necessary.

What are you suggesting the issue is then?

End goal here on my end is to fix the vulnerabilities with minimal breaking changes.

leo60228:
My personal belief is that the vulnerability is in libstd and consists of
`std::env::set_var` being safe.

Milo123459:
> My personal belief is that the vulnerability is in libstd and consists of `std::env::set_var` being safe.

This is out of Chrono reaches for fixing? If so, the advisory should be updated to state that it is in libstd. 

Never knew, thanks for sharing insight.

djc:
@Milo123459 probably good to at least skim this internals thread:

https://internals.rust-lang.org/t/synchronized-ffi-access-to-posix-environment-variable-functions/15475/97

Milo123459:
> @Milo123459 probably good to at least skim this internals thread:
> 
> [internals.rust-lang.org/t/synchronized-ffi-access-to-posix-environment-variable-functions/15475/97](https://internals.rust-lang.org/t/synchronized-ffi-access-to-posix-environment-variable-functions/15475/97)

I have, and well, haven't gotten anything useful off of it. I can't find where we invocate it, any functions that invocate it, and any dead-obvious invocations.

Dushistov:
> > @Milo123459 probably good to at least skim this internals thread:
> > [internals.rust-lang.org/t/synchronized-ffi-access-to-posix-environment-variable-functions/15475/97](https://internals.rust-lang.org/t/synchronized-ffi-access-to-posix-environment-variable-functions/15475/97)
> 
> I have, and well, haven't gotten anything useful off of it. I can't find where we invocate it, any functions that invocate it, and any dead-obvious invocations. I also found nothing in the time crate.

Here chrono `TZ` env variable:

https://github.com/chronotope/chrono/blob/3467172c31188006147585f6ed3727629d642fed/src/sys/unix.rs#L65

if other thread call `time_to_local_tm` that call `localtime_r` then this is problem.

So basically if you use `chrono` to get local time in different threads that this is the problem.
Because of it set/read `TZ` variable, which was read by `localtime_r`.

And because of glibc and rust stdlib uses different mutexes to protect access to environment variables this is would be bad.

Milo123459:
This is leading me to question, and I genuinely have no idea why that is being used: why are we using an environment variable? 

We never read it, nor use it. What is the point?

Dushistov:
> This is leading me to question, and I genuinely have no idea why that is being used: why are we using an environment variable?
> 
> We never read it, nor use it. What is the point?

chrono write to it, glibc read to it.
Here we pass argument to `tzset` via `TZ` variable, temporary changing global timezone to UTC:
```
set_var("TZ", "UTC");
tzset();
```
after that we call `mktime` that return UTC data,
then we restore current timezone back.

Milo123459:
Oh I see. I guess there isn't an immediate work around.

zseri:
Heavy-weight alternative might be forking, meddling with the env there, and transmitting the result via a pipe, then wait for the spawned process.
More lightweight alternative might be implementing/duplicating/replicating the `tzset`/`mktime` functionality as much as possible, and using that instead.

Milo123459:
> Heavy-weight alternative might be forking, meddling with the env there, and transmitting the result via a pipe, then wait for the spawned process. (maybe also provide an `unsafe` alternative which requires from the user that no other threads are running) More lightweight alternative might be implementing/duplicating/replicating the `tzset`/`mktime` functionality as much as possible, and using that instead.

I like the former, I'll give it a go.

Dushistov:
I suppose it is possible to use something like `lazy_static` plus `Mutex<Option<std::thread::ThreadId>>`,
and call `panic!` if it is not the first run and thread called function are not the same.

zseri:
@Dushistov I don't think that fixes the problem, because we simply don't have control over the standard library, and need to guard calls to `std::env::var` in other threads (according to the racy example).

But we use `std::env::set_env`, which according to the docs, should already be safe (because it guards access to `environ` with a lock), why is that a problem?

Dushistov:
@zseri 

> why is that a problem?

Because of glibc also access environment variables. As I see getenv has no locks inside: https://code.woboq.org/userspace/glibc/stdlib/getenv.c.html .

So if one thread call Rust `std::env::set_var` which uses lock, while other call any libc function that call `getenv`,
then no protection.

In chrono case glibc `localtime_r` call `tzset_internal` which call getenv ("TZ").

zseri:
I think the appropriate solution would be (1) for rust's `libstd` to expose its env lock, so all affected calls on `chrono`'s side can be effectively guarded.
C code calling `getenv` without protection is a corner case which we can't protect against if called outside from rust (if it's called from rust then we could use the env lock mechanism, although it could get difficult quickly and it might be necessary to use a lock implementation which supports recursive locks).
`setenv` from multithreaded programs on the C side is deemed unsound anyways (per POSIX at least).
And (2) I think chrono should avoid calling into the libc anyways, because appropriately shielding that is not possible until the rust libstd exposes the necessary lock access. This would also fix the `getenv` via C problem, at least in the chrono case.

nico-abram:
> I think the appropriate solution would be (1) for rust's libstd to expose its env lock, so all affected calls on chrono's side can be effectively guarded.

If I understand correctly, this would in the best case require bumping MSRV to at least a couple releases into the future (If/When std gets the new APIs exposing it), so I'm not sure this would be a valid solution

jhpratt:
Rust exposing the environment lock was previously ruled out.

zseri:
oof. ok.

quodlibetor:
I continue to be disinclined to merge this. I thought for awhile that we could migrate just that type and things would be okay, but I didn't realize that `time 0.3` changed the names of many of the methods on the Duration type. Merging this would mean that adding/removing the `oldtime` feature would _change_ -- not just enhance -- the set of methods available to users.

IMO that would be an absurd burden to put on users to work around a crate advisory that cannot even be affected by Chrono's use of `time 0.1`. It's trivially true that this PR cannot improve security for users, since time 0.3 removed the features that may lead to a segfault, and yet it's possible for Chrono to upgrade with no meaningful changes other than to method names.

I've crafted #626 to more strongly recommend to users that they disable the `oldtime` feature, and to clarify the docs aruond the set_env. I encourage folks to comment there if they have opinions on the wording of the docs warnings or have further ideas for encouraging users to do the right thing.

Milo123459:
At this rate, chrono having its own Duration type and not having to depend on time with the same API is probably the best idea. No major changes and fixes the vulnerability.

quodlibetor:
Right, just disabling the `oldtime` feature entirely removes `time` from the chrono dependency tree. Sadly there is no way out of the vulnerability, since Chrono just calls `localtime_r` itself, and as mentioned I don't know any way around that. 

I believe we can't even reimplement `localtime_r` in rust, because the only officially sanctioned way to access kernel APIs on BSD-derived unixen (notably macOS) is via libc. That means reimplementing `localtime_r` violates OS API guarantees and we're out of luck.

Milo123459:
Damn, I guess there isn't a good way around this.

saethlin:
> because the only officially sanctioned way to access kernel APIs on BSD-derived unixen (notably macOS) is via libc.

@quodlibetor Hung? Rust can and does call into libc all the time. So I don't understand what the problem is here, can you try explaining in more detail?

quodlibetor:
The bug is in libc (it has no way of checking if the environment is being modified), so to fix this problem we would need to _not_ call libc, just do some syscalls directly.

sdroege:
> Damn, I guess there isn't a good way around this.

The only way around `localtime_r()` is to implement it around `gettimeofday()` (and equivalent) and implementing your own timezone handling... which then also needs to look at environment variables (`TZ`) for getting the current local timezone, and you wouldn't have gained anything either. That's FWIW what [GLib](https://gitlab.gnome.org/GNOME/glib) does.

This really needs to be solved at a lower layer, or by considering changes to environment variables unsafe in Rust (which they should've been IMHO but that's too late now...).

saethlin:
> so to fix this problem we would need to not call libc, just do some syscalls directly.

This seems like quite a leap to me. Can't we just restrict ourselves to parts of libc that don't rely on the environment?

> which then also needs to look at environment variables (TZ) for getting the current local timezone, and you wouldn't have gained anything either.

Couldn't a Rust implementation soundly access TZ through the env lock?

jokeyrhyme:
There are a finite number of useful values for `TZ`, right? So, even without safe locking, could we not read it over and over until the value makes sense?

saethlin:
> could we not read it over and over until the value makes sense

No, because this is a data race which is UB. Concretely, a call to `getenv` could return a pointer to an allocation which between getting the pointer and attempting to read through it, is freed by a call to `std::env::set_var`.

sdroege:
> Couldn't a Rust implementation soundly access TZ through the env lock?

I guess so, that only leaves code changing the environment via other APIs as a problem. Which is probably acceptable.

Implementing the tzdata parsing etc is a bit of effort though, and I'm not sure if there's an easy portable and autodetectable way to locate it on the filesystem.

Might be a solution to investigate further, I guess

jokeyrhyme:
> > could we not read it over and over until the value makes sense
>
> No, because this is a data race which is UB. 

Got it, agreed, we can't have a library looping infinitely, that's unacceptable

It just seems to me like we have to give up on any approach that relies on locking, because not everything on the system is required to use the same lock

So, I guess I'm just trying to think about ways we could apply defensive-programming here

> Concretely, a call to getenv could return a pointer to an allocation which between getting the pointer and attempting to read through it, is freed by a call to std::env::set_var.

Got it, so it is possible to detect when a use-after-free has occurred at runtime? And from my earlier message, there are a finite number of useful values for TZ, right?

Rather than relying on an inconsistently-applied lock, is it feasible to attempt to read TZ, test what we get back in some way, and determine whether the read was good or bad?

saethlin:
@jokeyrhyme You misunderstand. The problem is that this is a data race. That's it, end of story. You cannot reason about the behavior of a program in C/C++/Rust which has produced a data race. That's what it means for an operation to be UB.

If you (or other readers passing by) are interested in being involved in this layer of the stack, learning about UB is really not optional. If you're new to this concept, I suggest starting with this talk: https://www.youtube.com/watch?v=v1COuU2vU_w

rbtcollins:
As I see it there are two choices: A rust-only implementation of localtime; or a new libc function to fix this defined-in dependency on unsafe behaviour.

Interestingly gmtime_r is nearly there: we can go time -> UTC, then apply the DST offset; however gmtime_r also depends on __tz_convert within glibc, so it would still be subject to the race on the first call with the current implementation. It should be fixable though, I think.

quodlibetor:
@sdroege 

> The only way around localtime_r() is to implement it around gettimeofday() (and equivalent) and implementing your own timezone handling.

If possible this actually seems doable. I have glanced through musl libc's [`localtime_r`](https://git.musl-libc.org/cgit/musl/tree/src/time/localtime_r.c) and most of the work is in [`__secs_to_tz`](https://git.musl-libc.org/cgit/musl/tree/src/time/__tz.c#n360) which seems likely to be where the actual unsafety is.

We have some decent TZ parsing code in rust in the chrono-tz ecosystem, I'd be happy to pull some of that into chrono to get around this.

sdroege:
> which seems likely to be where the actual unsafety is.

There's no access to the environment variables there, that's just applying the previously obtained timezone or not?

> We have some decent TZ parsing code in rust in the chrono-tz ecosystem, I'd be happy to pull some of that into chrono to get around this.

If that all exists already then that sounds like a viable approach, yes. That reduces the amount of work considerably :)
Just please make sure this uses the system's timezone database, at least by default. Shipping that with every application is wasteful and also make it more annoying to update the timezone database.

quodlibetor:
`__secs_to_zone` calls [`do_tzset()`](https://git.musl-libc.org/cgit/musl/tree/src/time/__tz.c#n126) which accesses the environment.

Milo123459:
Please can you try and resolve conflicts? Thanks.

djc:
I don't think we'll be updating the version of time used in chrono. Since #478, chrono has a minimal dependency on time, and in the next semver-incompatible version we'll remove the dependency entirely.

(Note that chrono is unaffected by https://rustsec.org/advisories/RUSTSEC-2020-0071.)

PR #580, opened 2021-08-07
Add a feature and methods to override `now`

### LINKED ISSUE #105: Better support for test cases: Overriding the current time
Apologies if there's already a good solution for this out there.  This is more a feature request than a bug.

In a test, it's nice to force certain dates to be returned.  Sure, you can do something like add an "overrideTime" variable, set it in tests, then do this every time you'd check the time:

let current_time : i64 ;
if !overrideTime {
  current_time = UTC::now().timestamp();
} else {
  current_time = forced_timestamp;
}

But meh.  It'd be nicer if instead there was a function similar to, say, JODA setCurrentMillisFixed : 
http://joda-time.sourceforge.net/api-release/org/joda/time/DateTimeUtils.html#setCurrentMillisFixed(long)

So then you'd have something like:

#[test]
fn test_that_requires_a_specific_time() {
    // This time is 2017-02-28T10:00Z
    UTC::setFixed(1488276000);
    // UTC::now() will now lie and always return the above value.
   ...
}

And then include a setCurrentMillisSystem() that undoes the above, for all that it'd be rarely called.

Alternatively: create a provider class, and then have it understood that classes should only ever ask the provider for the time rather than making raw now() calls.  e.g. Clock has a getTime() call, and it's implemented by SystemClock & FakeClock.  System Clock calls now() like normal, but if you pass in a FakeClock in a test, then that FakeClock can do whatever craziness you want.  This is a bit more powerful since now people can write FakeClocks that, say, actually run forward but from a set date or the like.

bryanburgers:
Support more testing use cases by adding a way to override the current
time that is returned by `Utc::now()` and `Local::now()`.

```toml
[package]
resolver = "2"

[dependencies]
chrono = "0.4"

[dev-dependencies]
chrono = { version = "0.4", features = ["test-override"] }
```

```rust
use chrono::{Utc, Datelike};

pub fn is_today_leap_day() -> bool {
    let today = Utc::today();
    today.month() == 2 && today.day() == 29
}

#[cfg(test)]
mod tests {
    use super::is_today_leap_day;
    use chrono::{Utc, DateTime};

    #[test]
    fn returns_true_on_leap_day() {
        Utc::test_override_now(Utc.ymd(2020, 2, 29).and_hms(0, 0, 0));
        assert!(is_today_leap_day());
    }

    #[test]
    fn returns_false_on_other_days() {
        Utc::test_override_now(Utc.ymd(2020, 2, 28).and_hms(0, 0, 0));
        assert!(!is_today_leap_day());
        Utc::test_override_now(Utc.ymd(2020, 3, 1).and_hms(0, 0, 0));
        assert!(!is_today_leap_day());
    }
}
```

Fixes #105

      about adding the PR number)

[changelog]: ../CHANGELOG.md

### DIFF
--- a/src/offset/local.rs
+++ b/src/offset/local.rs
@@ -212,6 +212,7 @@ thread_local!(static OVERRIDE: std::cell::RefCell<Option<DateTime<Local>>> = std
 
 #[cfg(all(feature = "test-override"))]
 impl Local {
+    #[cfg_attr(docsrs, doc(cfg(feature = "test-override")))]
     /// Override the value that will be returned by `Local::now()`, for use in local tests.
     ///
     /// ```
@@ -231,6 +232,7 @@ impl Local {
         });
     }
 
+    #[cfg_attr(docsrs, doc(cfg(feature = "test-override")))]
     /// Clear an override created by `Utc::test_override_now()`.
     pub fn test_clear_override() {
         OVERRIDE.with(|o| {
@@ -238,6 +240,7 @@ impl Local {
         });
     }
 
+    /// Get the overriden time to return for `Utc::now()`.
     #[inline]
     fn test_get_override() -> Option<DateTime<Local>> {
         OVERRIDE.with(|o| *o.borrow())
--- a/src/offset/local.rs
+++ b/src/offset/local.rs
@@ -99,12 +99,26 @@ impl Local {
     /// Returns a `DateTime` which corresponds to the current date.
     #[cfg(not(all(target_arch = "wasm32", not(target_os = "wasi"), feature = "wasmbind")))]
     pub fn now() -> DateTime<Local> {
+        #[cfg(feature = "test-override")]
+        {
+            if let Some(t) = Self::test_get_override() {
+                return t;
+            }
+        }
+
         tm_to_datetime(Timespec::now().local())
     }
 
     /// Returns a `DateTime` which corresponds to the current date.
     #[cfg(all(target_arch = "wasm32", not(target_os = "wasi"), feature = "wasmbind"))]
     pub fn now() -> DateTime<Local> {
+        #[cfg(feature = "test-override")]
+        {
+            if let Some(t) = Self::test_get_override() {
+                return t;
+            }
+        }
+
         use super::Utc;
         let now: DateTime<Utc> = super::Utc::now();
 
@@ -193,6 +207,43 @@ impl TimeZone for Local {
     }
 }
 
+#[cfg(all(feature = "test-override"))]
+thread_local!(static OVERRIDE: std::cell::RefCell<Option<DateTime<Local>>> = std::cell::RefCell::new(None));
+
+#[cfg(all(feature = "test-override"))]
+impl Local {
+    /// Override the value that will be returned by `Local::now()`, for use in local tests.
+    ///
+    /// ```
+    /// use chrono::{Local, TimeZone, Datelike};
+    /// fn is_today_leap_day() -> bool {
+    ///     let today = Local::today();
+    ///     today.month() == 2 && today.day() == 29
+    /// }
+    ///
+    /// Local::test_override_now(Local.ymd(2020, 2, 29).and_hms(0, 0, 0));
+    /// assert!(is_today_leap_day());
+    /// Local::test_clear_override();
+    /// ```
+    pub fn test_override_now(datetime: DateTime<Local>) {
+        OVERRIDE.with(|o| {
+            *o.borrow_mut() = Some(datetime);
+        });
+    }
+
+    /// Clear an override created by `Utc::test_override_now()`.
+    pub fn test_clear_override() {
+        OVERRIDE.with(|o| {
+            *o.borrow_mut() = None;
+        });
+    }
+
+    #[inline]
+    fn test_get_override() -> Option<DateTime<Local>> {
+        OVERRIDE.with(|o| *o.borrow())
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::Local;
@@ -223,5 +274,17 @@ mod tests {
             "unexpected timestr {:?}",
             timestr
         );
+    }
+
+    #[cfg(feature = "test-override")]
+    #[test]
+    fn test_override() {
+        assert!(Local::now() > Local.ymd(2021, 8, 7).and_hms(1, 0, 0));
+
+        Local::test_override_now(Local.ymd(2020, 2, 29).and_hms(0, 0, 0));
+        assert_eq!(Local::now(), Local.ymd(2020, 2, 29).and_hms(0, 0, 0));
+
+        Local::test_clear_override();
+        assert!(Local::now() > Local.ymd(2021, 8, 7).and_hms(1, 0, 0));
     }
 }
--- a/src/offset/utc.rs
+++ b/src/offset/utc.rs
@@ -173,6 +173,7 @@ mod tests {
     #[cfg(feature = "test-override")]
     #[test]
     fn test_override_multiple_threads() {
+        // stub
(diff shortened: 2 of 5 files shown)


bryanburgers:
I'm definitely open to thoughts on what the name of the feature should be and what the name of the methods should be.

I also considered returning another struct when overriding that will restore on drop. Something like

```rust
{
    let o = Utc::test_override();
    o.set(Utc.ymd(2020, 2, 29).and_hms(0, 0, 0));
    assert!(Utc::now() < Utc.ymd(2021, 1, 1).and_hms(0, 0, 0)); 
}

assert!(Utc::now() > Utc.ymd(2021, 1, 1).and_hms(0, 0, 0));
```

but I thought that might add too much complexity, especially if `Utc::test_override` was called multiple times.

bryanburgers:
The other thought on naming is to put the override functions in their own module instead of hanging them off of `Utc` and `Local`.

* `chrono::test_override::set_utc(_: DateTime<Utc>)`
* `chrono::test_override::clear_utc()`
* `chrono::test_override::set_local(_: DateTime<Local>)`
* `chrono::test_override::clear_local()`

```rust
use chrono::{Utc, Datelike};

pub fn is_today_leap_day() -> bool {
    let today = Utc::today();
    today.month() == 2 && today.day() == 29
}

#[cfg(test)]
mod tests {
    use super::is_today_leap_day;
    use chrono::{Utc, DateTime};

    #[test]
    fn returns_true_on_leap_day() {
        chrono::test_override::set_utc(Utc.ymd(2020, 2, 29).and_hms(0, 0, 0));
        assert!(is_today_leap_day());
    }

    #[test]
    fn returns_false_on_other_days() {
        chrono::test_override::set_utc(Utc.ymd(2020, 2, 29).and_hms(0, 0, 0));
        assert!(!is_today_leap_day());
        chrono::test_override::set_utc(Utc.ymd(2020, 2, 29).and_hms(0, 0, 0));
        assert!(!is_today_leap_day());
    }
}

**Comment by bbigras** [comment]
Any progress on this?

**Comment by slashrsm** [comment]
I am working on some tests where I would need this. Is there anything I can do to help push this forward?

**Comment by esheppa** [comment]
Hi @slashrsm and @bryanburgers - I can see where this is useful for testing, just a few thoughts:
* Could your use cases instead be solved by passing a `DateTime` to the relevant functions, and having the actual current time generated somewhere earlier in the code?
* I'm a bit concerned about this being enabled by a dependency and then causing unexpected effects, potentially one way to resolve this could be changing the cfg directive from `#[cfg(feature = "test-override")]` to `#[cfg(all(test, feature = "test-override"))]`. Another (or additional) option could be making the `set` and `clear` functions private and having an API that runs the fixed now time on a separate thread to avoid interference with other code

**Comment by djc** [comment]
I think this would have to come with a very strong rationale for why (something like) this is the best possible design. I don't think it fits very well with the scope of chrono as a fairly low-level library.

**Comment by pitdicker** [comment]
> I think this would have to come with a very strong rationale for why (something like) this is the best possible design. I don't think it fits very well with the scope of chrono as a fairly low-level library.

I personally don't have any strong rationale to offer. I can imagine there are integration tests and unfortunately strutured codebases where adjusting the current time is the easiest or even only way to do some tests.

However https://github.com/chronotope/chrono/issues/105#issuecomment-633260201 in the related issue has collected 28 thumb ups in three years. This may be our most popular feature request.

**Comment by pitdicker** [comment]
To comment on this PR:
I would add just one standalone public functions in the `offset` module, which can serve to set and clear the override:
```rust
pub fn override_now(datetime: Option<DateTime<FixedOffset>>);
```

And by making it take a `FixedOffset` you can also use it to test different timezone offsets.

I also like the suggestion by @esheppa in https://github.com/chronotope/chrono/pull/580#issuecomment-1173780170 to confine the feature to testing only with `#[cfg(all(test, feature = "test-override"))]`.

@bryanburgers would you still be available to work on this after two years?

penso:
> I personally don't have any strong rationale to offer. I can imagine there are integration tests and unfortunately strutured codebases where adjusting the current time is the easiest or even only way to do some tests.

Yes, I have that exact scenarios, and it's not that uncommon. You're writing a test looking at data fixtures coming originally from an outside service, and your code is time related to `now()`. For example checking if more than a time period has passed, or should that session be expired or not.

jtmoon79:
To chime in "as a Test Engineer", I think the overall idea here is sound.

Skimming the code changes, they seem okay to me. However, I can't comment if this is necessarily the cleanest way to do this; would require a long think.

Additionally, I couldn't determine if the override of `now` behavior is being used for all the test cases that call `now`, which is what this PR should also do (if it's not already).

penso:
The way I do it so far is having an app context `Context` and use `ctx.now()` everywhere I need:

```
#[derive(Clone)]
pub struct AppContext {
    pub now: Option<DateTime<Utc>>,
}

impl AppContext {
    pub fn now(&self) -> DateTime<Utc> {
        self.now.unwrap_or(chrono::Utc::now())
    }

    pub fn set_now(&mut self, now: &str) {
        self.now = Some(now.parse().expect("Can't parse now"));
    }
}
```

pitdicker:
@jtmoon79 I planned on taking over this PR next week, but maybe it is more your cup of tea?

src/offset/local.rs
```diff
@@ -193,6 +207,46 @@ impl TimeZone for Local {
     }
 }
 
+#[cfg(all(feature = "test-override"))]
+thread_local!(static OVERRIDE: std::cell::RefCell<Option<DateTime<Local>>> = std::cell::RefCell::new(None));
```
jtmoon79, line 211: nitpick: "Override" is a broad term, can this named `LOCAL_NOW_OVERRIDE` ? 

Also, could this somewhat complex global variable include a `///` code comment describing why it's necessary?

src/offset/utc.rs
```diff
@@ -98,3 +112,98 @@ impl fmt::Display for Utc {
         write!(f, "UTC")
     }
 }
+
+#[cfg(all(feature = "clock", feature = "test-override"))]
+thread_local!(static OVERRIDE: std::cell::RefCell<Option<DateTime<Utc>>> = std::cell::RefCell::new(None));
```
jtmoon79, line 117: Same request for a code comment.

Same suggestion for name change to `UTC_NOW_OVERRIDE`, which also will disambiguate from the other `OVERRIDE`.

src/offset/local.rs
```diff
@@ -193,6 +207,46 @@ impl TimeZone for Local {
...
+
+    #[cfg_attr(docsrs, doc(cfg(feature = "test-override")))]
+    /// Clear an override created by `Utc::test_override_now()`.
+    pub fn test_clear_override() {
+        OVERRIDE.with(|o| {
+            *o.borrow_mut() = None;
+        });
+    }
+
+    /// Get the overriden time to return for `Utc::now()`.
```
jtmoon79, line 243: Did you mean

```
/// ... return for `Local::now()`
```

? Same question for other docstrings in this scope.

jtmoon79 (changes requested):
I'm still not sure where the override value is set such that other tests will use the override value during `cargo tests`.  Can you point it out to me?

jtmoon79:
> @jtmoon79 I planned on taking over this PR next week, but maybe it is more your cup of tea?

Before doing any work, I'll wait until @djc approves of the concept.

djc:
I guess I'm open to having something like this. I think it should default to off and should only be turned on with a `cfg` flag (not a Cargo feature, which can be turned on by indirect dependencies). We should offer a way both to set the current time and the timezone/offset.

pitdicker:
https://github.com/chronotope/chrono/pull/1244 is the updated PR.

PR #593, opened 2021-09-17
use stub for anything not unix and not windows

yu-re-ka:
### Thanks for contributing to chrono!

Alternative to https://github.com/chronotope/chrono/pull/568
Fixes https://github.com/chronotope/chrono/issues/519

- [ ] Have you added yourself and the change to the [changelog]? (Don't worry
      about adding the PR number)
- [x] If this pull request fixes a bug, does it add a test that verifies that
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md

PR #607, opened 2021-10-25
Fix emscripten target

sinkingsugar:
Emscripten is effectively a unix system and so `inner` would be declared twice resulting in a compilation error.
This PR fixes such scenario.


djc:
Would you be able to rebase this on current main? I'm happy to merge something like this.

djc:
Never mind, I think #593 is a better fix.

PR #613, opened 2021-10-29
Change DateTime::timestamp_nanos() to return Option

### LINKED ISSUE #586: DateTime.timestamp_nanos() on valid DateTime panics instead of returning a result/option
While it is documented behaviour, a valid DateTime should not panic while getting a timestamp from it.

### Minimal repro

```rust
let naive = NaiveDateTime::from_timestamp_opt(456456456456, 0).unwrap(); //Does not panic
let datetime = DateTime::<Utc>::from_utc(naive, Utc)); //Still no panic
println!("{}", datetime.timestamp_nanos()); //This panics

```

---

### Potential low-effort fix

```rust
pub fn timestamp_nanos(&self) -> Option<i64> {
    let as_ns = self.timestamp().checked_mul(1_000_000_000);
    match as_ns {
        Some(ns) => Some(ns + i64::from(self.timestamp_subsec_nanos())),
        None => None,
    }
}
```

While testing this potential fix, I uncovered the following points where other unrelated operations on DateTimes call `timestamp_nanos()`, which suggests to me that this might of greater concern than something that can be chalked up as a low importance "formatting bug".

`round.rs: 157`  
`round.rs: 186`

An alternate fix would be to add checked variants of `timestamp_nanos()` and `timestamp_millis()`.

---

As for my use-case, I can easily work around the issue, but this is still a big problem for other use-cases.

Milo123459:

      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md

Fixes #586

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -1573,7 +1573,7 @@ pub mod serde {
             S: ser::Serializer,
         {
             match *opt {
-                Some(ref dt) => serializer.serialize_some(&dt.timestamp_nanos()),
+                Some(ref dt) => serializer.serialize_some(&dt.timestamp_nanos().unwrap()),
                 None => serializer.serialize_none(),
             }
         }
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -242,13 +242,13 @@ impl<Tz: TimeZone> DateTime<Tz> {
     /// use chrono::TimeZone;
     ///
     /// let dt = Utc.ymd(1970, 1, 1).and_hms_nano(0, 0, 1, 444);
-    /// assert_eq!(dt.timestamp_nanos(), 1_000_000_444);
+    /// assert_eq!(dt.timestamp_nanos().unwrap(), 1_000_000_444);
     ///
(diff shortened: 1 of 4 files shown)


Milo123459:
It's failing due to an unrelated change. After looking into it, it is something broken with `bincode`.

Milo123459:
This is breaking because of MSRV. Once #578 gets merged, this can be merged safely.

src/datetime.rs
```diff
@@ -1420,7 +1420,7 @@ pub mod serde {
         where
             S: ser::Serializer,
         {
-            serializer.serialize_i64(dt.timestamp_nanos())
+            serializer.serialize_i64(dt.timestamp_nanos().unwrap())
```
nickelc, line 1423: these unwraps move the panic from `DateTime::timestamp_nanos` to the serde modules.

Milo123459: Good catch

andy128k:
Why not `i128`?

Issue #575, opened 2021-07-18
chrono Date<Utc>.format("%s") panics with unclear message


doivosevic:
panics with `panicked at 'a Display implementation returned an error unexpectedly: Error', /build/rustc-1.51.0-src/library/alloc/src/string.rs:2213:14`

I feel like there should be a better panic for this case and I guess some other format values have similar panics. This is especially problematic since format doesn't return Result so I wouldn't expect this to panic nor can I handle it gracefully

    #[test]
    fn test_format_timestamp_chrono() {
        let now = Utc::now().date(); // just Date panics on %s
        now.format("%s").to_string();
    }

t1ooo:
Can we just throw panic with clear error message instead of returning fmt::Error?

https://github.com/chronotope/chrono/blob/3467172c31188006147585f6ed3727629d642fed/src/format/mod.rs#L537

hustcer:
I have run into this as well, can it be fixed in the next release? thx

retrhelo:
> I have run into this as well, can it be fixed in the next release? thx

I wonder if my PR solves your problem? It's made a few months ago, so some conflicts have to be solved before it gets merged :(.

hustcer:
Yes, it does, Thanks for your contribution

esheppa:
Hi @hustcer  - please see https://github.com/chronotope/chrono/pull/614#issuecomment-1173070176 for some alternative solutions available in the currently released version of `chrono`.

pitdicker:
We have multiple issues that track the issues out formatting code can panic. I'm closing this one since the `Date` type is deprecated. #1127 is the best issue to watch.

PR #614, opened 2021-10-31
fix format panic problem

retrhelo:
Add some modification to `format::inner_format()`, which will prevent the program from panic when an inproper format string is provided. This is the problem mentioned in issue #575 . 

The new strategy I'm using is to **skip those inproper format substrings**. You can find some of the examples in the test I add. I think this strategy is better than directly panic, especially when handling user input strings. Skipping the wrong parts will also indicate an invalid format string, and giving the programmer an opportunity to fix the problem while running, instead of panic and terminating the program.

### DIFF
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -3050,4 +3050,25 @@ mod tests {
         assert_eq!(datetime_east, datetime_utc.with_timezone(&timezone_east));
         assert_eq!(datetime_west, datetime_utc.with_timezone(&timezone_west));
     }
+
+    #[test]
+    fn test_format_to_string() {
+        // create a naive date
+        let naive_date = NaiveDate::from_ymd(2000, 1, 12);
+        assert_eq!(format!("{}", naive_date.format("%Y %b %d")), String::from("2000 Jan 12"));
+        assert_eq!(format!("{}", naive_date.format("%F")), String::from("2000-01-12"));
+        // try to format something that the instance doesn't have
+        assert_eq!(format!("{}", naive_date.format("%r")), String::from(":: "));
+        // pass a wrong format
+        assert_eq!(
+            format!("{}", naive_date.format("%Y %{whatever} %d")),
+            String::from("2000 whatever} 12")
+        );
+
+        // create a naive time
+        let naive_time = NaiveTime::from_hms(12, 43, 33);
+        assert_eq!(format!("{}", naive_time.format("%H:%M:%S %p")), String::from("12:43:33 PM"));
+        // try to format something that doesn't exist
+        assert_eq!(format!("{}", naive_time.format("%F %H:%M:%S")), String::from("-- 12:43:33"));
+    }
 }
--- a/src/datetime.rs
+++ b/src/datetime.rs
@@ -3050,25 +3050,4 @@ mod tests {
         assert_eq!(datetime_east, datetime_utc.with_timezone(&timezone_east));
         assert_eq!(datetime_west, datetime_utc.with_timezone(&timezone_west));
     }
-
-    #[test]
-    fn test_format_to_string() {
-        // create a naive date
-        let naive_date = NaiveDate::from_ymd(2000, 1, 12);
-        assert_eq!(format!("{}", naive_date.format("%Y %b %d")), String::from("2000 Jan 12"));
-        assert_eq!(format!("{}", naive_date.format("%F")), String::from("2000-01-12"));
-        // try to format something that the instance doesn't have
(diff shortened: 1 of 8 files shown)


quodlibetor:
This unfortunately causes errors to become silent.

A better solution would be to introduce a `format_opt` method that correctly returns a `Result<DelayedFormat, FormatError>`.

src/datetime.rs
```diff
@@ -3050,4 +3050,25 @@ mod tests {
...
+        // create a naive date
+        let naive_date = NaiveDate::from_ymd(2000, 1, 12);
+        assert_eq!(format!("{}", naive_date.format("%Y %b %d")), String::from("2000 Jan 12"));
+        assert_eq!(format!("{}", naive_date.format("%F")), String::from("2000-01-12"));
+        // try to format something that the instance doesn't have
+        assert_eq!(format!("{}", naive_date.format("%r")), String::from(":: "));
+        // pass a wrong format
+        assert_eq!(
+            format!("{}", naive_date.format("%Y %{whatever} %d")),
+            String::from("2000 whatever} 12")
```
Milo123459, line 3065: If invalid, I would imagine this would simply produce "%{whatever}", not "whatever}"

retrhelo:
> If invalid, I would imagine this would simply produce "%{whatever}", not 
> "whatever}"

That sounds reasonable. I'll work on it.

By the way, I think the idea to introduce a new method that returns
`Result<DelayedFormat, FormatError>`, which is mentioned before, is also an
attractive solution to this problem. I can also give a try on it.

Milo123459:
Maybe create another function that returns a Result?

retrhelo:
> Maybe create another function that returns a Result?

I roughly implement a method called `try_format` for this purpose. It adds some checks after a `DelayedFormat` is initialized. I wonder if `try_format` fits the project's naming convention?

retrhelo:
> Maybe create another function that returns a Result?

I roughly created a `try_format` function for this purpose. However, I find that there's many format functions for each structure like `DateTime`, `Date`, `Time`, etc. Adding such a function will get a lot of files involved, since we're adding it to every structure that implements `format` function. I think create a new PR for this task is better choice.

hustcer:
Any update on this PR?

djc:
What's the path here? Is there some outer API here that's not propagating the error from the inner API, but panicking instead?

retrhelo:
> Is there some outer API here that's not propagating the error from the inner API, but panicking instead?

Yes, it's the panic problem. The original `format` function panics when passed in an inproper format string, making it hard to "format" a user-input string that's possibly inproper.

djc:
Yeah, I don't think this is an improvement. If we're going to stop panicking, we should explicitly signal an error some other way.

esheppa:
Just to add another option here, and I recognize this isn't super user friendly or discoverable, (we can potentially fix that with a few helper functions or more docs) but the potentially invalid user input problem can be solved in the currently released `chrono`, using [`StrftimeItems::new()`](https://docs.rs/chrono/latest/chrono/format/strftime/struct.StrftimeItems.html#method.new) and the [`write!`](https://doc.rust-lang.org/std/fmt/fn.write.html) macro depending on the task.  

If you want to just check whether the format string is valid or not, using `StrftimeItems` suffices, however if you want to format without panicking, using `write!` instead of `to_string` or `format!` allows catching this [(playground link)](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c47eb8476c6780049b75185d08f738b3).

```rust
use chrono::{Utc, format::{Item, strftime::StrftimeItems}}; // 0.4.19
use std::fmt::Write;
const BAD_FMT_STR: &str = "%Y-%m-%";
const GOOD_FMT_STR: &str = "%Y-%m-%d";

fn main() {
    assert!(StrftimeItems::new(BAD_FMT_STR).any(|i| matches!(i, Item::Error)));
    assert!(StrftimeItems::new(GOOD_FMT_STR).all(|i| !matches!(i, Item::Error)));
    
    let mut out = String::new();
    assert!(write!(&mut out, "{}", Utc::now().date().naive_local().format(GOOD_FMT_STR)).is_ok());
    assert!(write!(&mut out, "{}", Utc::now().date().naive_local().format(BAD_FMT_STR)).is_err());
}
```

esheppa:
I've added a draft PR which shows some options here: https://github.com/chronotope/chrono/pull/735

pitdicker:
@retrhelo Thank you for your work here!
We are going in a different direction to fix this panic, see https://github.com/chronotope/chrono/issues/1127.

PR #619, opened 2021-11-06
Make NaiveTime::from_str support HH:MM as well as HH:MM:SS

jamespharaoh:
Times are commonly written without seconds, NaiveTime::from_str should recognise this format as well. This adds that functionality.

This makes parsing data structures which include times easier, since you would otherwise need to specify a format string or construct the NaiveTime yourself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Versions with only mechanical changes will be omitted from the following list.
 * Implement `DurationRound` for `NaiveDateTime`
 * Add `DateTime::from_local()` to construct from given local date and time (#572)
 * Correct build for wasm32-unknown-emscripten target (#568)
+* Add support for "HH:MM" format in `NaiveTime::from_str()`
 
 ## 0.4.19
 
--- a/src/format/mod.rs
+++ b/src/format/mod.rs
@@ -294,6 +294,8 @@ pub enum Item<'a> {
     Numeric(Numeric, Pad),
     /// Fixed-format item.
     Fixed(Fixed),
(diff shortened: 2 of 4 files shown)


esheppa:
Thanks for this @jamespharaoh - apologies for the long delay. I like where you are going with this as it would be interesting to make the various `FromStr` implementations more flexible, however there are some tradeoffs which I'd like to seek comment on before going ahead with this:

a) Parse/Display Roundtrip - if we allow multiple valid input formats, then it is impossible guarantee a identity when combining parse and display. (This may also be the case currently, which we could fix instead if desired)
b) How we view the `FromStr` and `Display` impls compared to the `format` and `parse_from_str` methods. For now the `FromStr` and `Display` impls have some sensible defaults, but it could be interesting to have `FromStr` try a couple of different common patterns in order (hopefully in an unambiguous way) to make parsing common date/time formats easier (especially for new users of the library) at the cost of some performance.

This PR falls within the latter option and if it is decided to go down that path, it might be interesting to have an array of arrays of `Item<'static>` and then try each one in order - this would also avoid adding the `Truncated` variant to `Item`.

esheppa:
Thanks for the feedback @djc. In this case, if possible @jamespharaoh - would you mind implementing this using an array or arrays of `Item<'static>`? My line of thinking here is opening the way to allow for example `1500` to parse as a time as well in future. (Or alternatively, if we can make a case of `Item::Truncated` being generally useful, then maybe keep it as is)

pitdicker:
Closing in favor of https://github.com/chronotope/chrono/pull/1181.

PR #626, opened 2021-11-19
Clarify localtime and oldtime's docs

quodlibetor:
In response to #578, there really isn't anything Chrono can do to actually prevent the invalid
memory accesses in libc -- all we can do is tell people not to set env vars. In my experience and
from conversations surrounding this issue, setting env vars is rare and definitely just shouldn't
be done once the process has more than one thread running. Hopefully isn't an an awful burden on
users, either in terms of feature reduction or of security scanning.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -38,15 +38,22 @@
 //!
 //! Default features:
 //!
-//! - `alloc`: Enable features that depend on allocation (primarily string formatting)
 //! - `std`: Enables functionality that depends on the standard library. This
 //!   is a superset of `alloc` and adds interoperation with standard library types
 //!   and traits.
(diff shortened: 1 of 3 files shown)


nico-abram:
`segfault or crash in other interesting ways` would it not be better to explicitly mention data races and undefined behaviour?

If this is going to be accepted I feel like at the very least these functions should be deprecated and have unsafe counterparts.

pitdicker:
I agree the docs should say something about the issue with reading the `TZ` environment variable and our history with RUSTSEC-2020-0159. But since it is fixed with https://github.com/chronotope/chrono/pull/677 we should say something different from the changes in this PR.

PR #633, opened 2021-12-14
Derive Default for Parsed instead of implementing it manually

rnestler:
Fixes a clippy warning.

- [na] Have you added yourself and the change to the [changelog]? (Don't worry
      about adding the PR number) (I don't think this is worth a changelog entry.)
- [na] If this pull request fixes a bug, does it add a test that verifies that
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md


djc:
Thanks! This was fixed as part of #652.

Issue #309, opened 2019-03-17
Make Duration::seconds (and siblings) const fn


axos88:
This would allow to specify things like:

```const THRESHOLD: Duration = Duration::seconds(3);```

bspeice:
I don't think this is possible in current Rust - `const fn` isn't allowed to use branches as of Rust 1.33.

quodlibetor:
Yes, that's what I'm currently holding off on. Even once const fn is stable enough for chrono we'll need to think about how best to support them in a way that doesn't break backcompat too terribly. Chrono currently supports rust 1.13, but const fn is the strongest reason to upgrade to a new version that is upcoming. 

Obviously we'll update, but I'd like to keep things as compatible as possible, if that means adding some features or something else so that we can still work on old rustcs, I'm not sure.

hjmallon:
`const fn` on NaiveDate and friends would also be useful.

quodlibetor:
If anyone wants to create a PR for this behind something like an `unstable-nightly-const` feature gate I would be happy to take it. It should just be a matter of duplicating some functions and adding some `cfg` and `cfg(not)` attributes.

csm123199:
For the googlers that came here like me, it looks like the current blocker on this issue are const-panics ([RFC](https://github.com/rust-lang/rfcs/blob/master/text/2345-const-panic.md)) ([issue](rust-lang/rust#51999)) that are triggered when doing the .expect/panic-based bounds checking on the various `Duration` constructors.

Perhaps we can introduce variants that follow stdlib functions like [checked_{mul,div,...}](https://doc.rust-lang.org/stable/std/primitive.i64.html#method.checked_mul), and return an `Option<Duration>`, with `None` meaning out-of-bounds, just like checked_mul.

Since `Duration` is re-exported from the [time](https://github.com/time-rs/time) crate, shouldn't this really be an issue there?

zertyz:
Still not good to go on with that change?
I'd like to vote for it, because Duration constants make much sense, imho.
I actually had to quit using Duration for that and resort to ints (with a particular unit attached to their names), having to create a Duration variable every time...
Do you, authors, tought of a better way to address this scenario than using constants?
Found it nowhere in the docs...

robjtede:
In the https://github.com/dtolnay/rustversion there's a special case for making functions const conditionally on compiler versions.

msdrigg:
We now have const panics in rust 1.57 would it be possible to move this issue forward?

[...]

PR #638, opened 2021-12-22
Patch 1

msdrigg:
This commit updates `Duration::seconds` and siblings to be const. This
requires the const-panic api that only stabilized in 1.57, so it may be
best to hide this behid a feature flag somehow so as to not require a
MSRV of 1.57. This is my first pr so I don't know the policy around
this.

This pr also adds a bit of mess to the functions because we can't use
`.expect("error")` or `Datetime::partial_cmp`. I think it is worth it
because const durations would be really nice to have.

- [ x ] Added the change to the [changelog]? 
- [ x ] Added a test

[changelog]: ../CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Versions with only mechanical changes will be omitted from the following list.
 * Implement `DurationRound` for `NaiveDateTime`
 * Add `DateTime::from_local()` to construct from given local date and time (#572)
 * Correct build for wasm32-unknown-emscripten target (#568)
+* Make `Duration.seconds()` and siblings const fn
 
 ## 0.4.19
 
--- a/src/oldtime.rs
+++ b/src/oldtime.rs
@@ -70,53 +70,74 @@ impl Duration {
     /// Equivalent to `Duration::seconds(weeks * 7 * 24 * 60 * 60)` with overflow checks.
     /// Panics when the duration is out of bounds.
     #[inline]
-    pub fn weeks(weeks: i64) -> Duration {
-        let secs = weeks.checked_mul(SECS_PER_WEEK).expect("Duration::weeks out of bounds");
-        Duration::seconds(secs)
+    pub const fn weeks(weeks: i64) -> Duration {
+        let secs = weeks.checked_mul(SECS_PER_WEEK);
+        if let Some(secs) = secs {
+            Duration::seconds(secs)
+        } else {
+            panic!("Duration::weeks out of bounds");
+        }
     }
 
     /// Makes a new `Duration` with given number of days.
     /// Equivalent to `Duration::seconds(days * 24 * 60 * 60)` with overflow checks.
     /// Panics when the duration is out of bounds.
     #[inline]
-    pub fn days(days: i64) -> Duration {
-        let secs = days.checked_mul(SECS_PER_DAY).expect("Duration::days out of bounds");
-        Duration::seconds(secs)
+    pub const fn days(days: i64) -> Duration {
+        let secs = days.checked_mul(SECS_PER_DAY);
+        if let Some(secs) = secs {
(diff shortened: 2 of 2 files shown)


hoon-prestolabs:
It seems there's an error with lint. It doesn't finish for months.

djc:
As mentioned in #309, this will need to be conditional on an opt-in feature flag, otherwise we have to raise the MSRV too fast. Once that change is made I'm open to merging something like this.

msdrigg:
I'm pretty busy this week. But this coming weekend I'll figure out how to do a opt-in feature flag and make the change.

GeorgeHahn:
@djc Would you consider a dependency on a crate like [rustversion](https://crates.io/crates/rustversion) or [const_fn](https://crates.io/crates/const_fn)?

djc:
I'd prefer not to take on a procedural macro dependency. Maybe a declarative macro will work for this, something that can be called with an optional token for the `const` and dump a whole `impl` block with/without it?

mattfbacon:
Has this stagnated? I'd love to have this feature.

djc:
There has been no progress as far as I know. @msdrigg did you want to move this forward? @mattfbacon if not, would you be willing to open a new PR based on this branch and incorporate the review feedback?

msdrigg:
I dont think I will have time for this for the foreseeable future. If @mattfbacon can take the reigns, it would be much appreciated

mattfbacon:
Sure, I could take a look. I'll try to figure out if a `const` distinction can be done with an MBE macro.

mattfbacon:
Const on debug, non-const on release. I couldn't figure out how to support `meta` and `vis` tokens due to parsing ambiguity so I think they need to be added individually, as I did for `pub`.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=14d3de0dcf41949dee89664e1333ea63

djc:
Yeah, not super excited about pulling in a dependency or a build script for this stuff.

esheppa:
As an extra note, I'm experimenting with ways to use `core::time::Duration`, which could potentially fully or partly replace the existing Duration in an 0.5 release. Then people using more recent rust versions would automatically get these functions as const and we essentially farm out working out the const/non-const to the core/std library! 

@mattfbacon / @msdrigg - would having a `Duration` type in chrono that was either `core::time::Duration` or implemented `From<core::time::Duration>` provide similar utility as the changes here? If so, perhaps we close this in favour implementing the use of `core::time::Duration` in 0.5

djc:
core::time::Duration doesn't support negative Durations, right? I thought there was a fairly fundamental difference.

esheppa:
Correct - I'm trying out two options:
a) use core::time::Duration directly and remove all uses of negative duration - cleaner but causes more breakage
b) create a new `SignedDuration` that wraps the core::time::Duration - at the cost of some stack space
 ```rust
 pub enum SignedDuration {
    Forwards(Duration),
    Backwards(Duration),
}
``` 
once I've got all the tests passing, I'll create some PR's to collect feedback on whether any of those might be feasible, or if we can potentially find a path forward from one of those options to something feasible.

pitdicker:
I played with this PR, hoping to help push it over the finish line now that our MSRV is 1.56 and 1.57 is maybe possible.

But I now think it is not good to do this in the 0.4.x branch. We can only make the methods const for our own implementation, which you get with the `--no-default-features` cargo flag. Otherwise the non-const methods from time 0.1 will be used.

Suppose you use the const methods in your crate. When you add a dependency that somewhere in its dependency graph depends on chrono and doesn't set the flag, your crate suddenly fails to compile. The situation where enabling a feature flag makes less code compile does not seem smart.

msdrigg:
Glad to see the progress on this!

PR #642, opened 2022-01-06
Updated oldtime from version 0.1.43 to version 0.1.44 which is compatible with WASI

john-sharratt:

      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md


djc:
Thanks for your contribution!

This is unnecessary. The way semver works with Rust crates is that dependency requirements like 0.1.43 act as if they had the ^ operator implicitly specified, so that your downstream crates will automatically pick up time 0.1.44.

PR #644, opened 2022-01-20
add optional rkyv support

### LINKED ISSUE #583: rkyv Support
https://rkyv.org/

A zero-copy deserialization framework for rust. A sort of alternative to serde that has higher performance. chrono should add a feature gate where specifying "rkyv" will enable support for rkyv.

Types that should be serialized/deserialized should be given `#[derive(Archive, Serialize, Deserialize, PartialEq, Debug)]`.

dovahcrow:

      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md

Fixes #583

### DIFF
--- a/src/date.rs
+++ b/src/date.rs
@@ -18,7 +18,8 @@ use crate::naive::{self, IsoWeek, NaiveDate, NaiveTime};
 use crate::offset::{TimeZone, Utc};
 use crate::DateTime;
 use crate::{Datelike, Weekday};
-
+#[cfg(feature = "rkyv")]
+use rkyv::{Archive, Deserialize, Serialize};
 /// ISO 8601 calendar date with time zone.
 ///
 /// You almost certainly want to be using a [`NaiveDate`] instead of this type.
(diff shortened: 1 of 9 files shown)


djc:
This seems to have a lot of unrelated of changes. I'm afraid the changes in #661 have caused quite a few conflicts, but if you would rebase this on current main and make only the minimal changes necessary to support rkyv I'd be happy to review this.

dovahcrow:
@djc Done.

src/date.rs
```diff
@@ -18,7 +18,8 @@ use crate::naive::{self, IsoWeek, NaiveDate, NaiveTime};
 use crate::offset::{TimeZone, Utc};
 use crate::DateTime;
 use crate::{Datelike, Weekday};
-
+#[cfg(feature = "rkyv")]
+use rkyv::{Archive, Deserialize, Serialize};
```
djc, line 22: Please put this in a separate block of imports, between the std/core block and the crate-internal imports. (Same for all the other imports.)

djc (commented):
Thanks for rebasing!

src/date.rs
djc: It's not necessary for me to merge this PR, but if you don't mind doing the work that would be great! Please do it as a separate commit.

dovahcrow: done

djc (approved):
LGTM, thanks!

PR #647, opened 2022-02-08
change UTC::now() documentation to "date and time"

ModProg:
At least I was confused for a moment reading that it returns the current date, looking at it in the documentation.

      about adding the PR number)

[changelog]: ../CHANGELOG.md


djc:
I don't think this needs a changelog entry. Can you rebase on current main? It should fix the lint issues.

ModProg:
> I don't think this needs a changelog entry. Can you rebase on current main? It should fix the lint issues.

I just did the changelog because the template prompted me to do so

djc:
Thanks!

PR #649, opened 2022-02-11
Fix localtime unsoundness using `rl_localtime`

Kixunil:
This replaces current implementation with `rl_localtime` which is sound
thanks to it being a fork that calls to Rust `std` to get env vars.

**Important:** this is meant mainly as a demonstration how the change could look like.
I don't recommend merging until people check soundness of `rl_localtime` crate.

      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md


Kixunil:
Better use `tz` crate.

PR #653, opened 2022-03-06
Update to 2018 edition (and MSRV 1.32)

djc:


Milo123459 (approved):
LGTM. A good step in the right direction!

PR #639, opened 2021-12-29
bump time

Milo123459:
- bump time
- bump changelog

### Thanks for contributing to chrono!

- [x] Have you added yourself and the change to the [changelog]? (Don't worry
      about adding the PR number)
- [x] If this pull request fixes a bug, does it add a test that verifies that
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md

PR #655, opened 2022-03-14
Libtzfile localtime

esheppa:
This is a potential way to solve the [CVE-2020-0159](https://github.com/rustsec/advisory-db/blob/main/crates/chrono/RUSTSEC-2020-0159.md). This uses the library `libtzfile` to read the time zone information file and calculate the local offset from the information contained within. This work is hopefully orthogonal with that of #639. 

      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md


djc:
libtzfile uses 2021 edition (which means it doesn't support our MSRV), seems to have only one non-maintainer contributor, apparently has no CI in place, and the code doesn't look very clean to me. I don't think it's a good fit.

PR #656, opened 2022-03-14
Localtime using tz-rs

esheppa:
This is a potential way to solve the [CVE-2020-0159](https://github.com/rustsec/advisory-db/blob/main/crates/chrono/RUSTSEC-2020-0159.md). This uses the library `tz-rs` to read the time zone information file and calculate the local offset from the information contained within.

`tz-rs` is quite comprehensive as leap second handling has been implemented, however there are some outstanding questions on how to handle this in `from_local_datetime` (should we assume the local datetime does or doesn't include leap seconds?).

      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md


djc:
Thanks for working on this! I'm not yet sure if this is the right direction, will have to dig into it more.

esheppa:
Thanks for enabling CI @djc - I'll try and get the CI passing - then I recognize there are a bunch more things to figure out, this is intended as a proof of concept. If this is later decided to be a good solution, I'll go through and add more documentation etc.

Potential issues:
* Minimum supported rust version: This or #655 would require some changes to not affect `chrono`s MSRV - but these would be solvable, for example via PR'ing to the timezone file libraries to reduce their MSRV to match that of `chrono` (or alternatively increasing `chrono`s MSRV, but this may be less preferable)
* Extra dependency: there isn't really any way around this unless the code of a TZif parsing library is vendored into `chrono` itself.

Other options:
* Providing this or #655 as a separate crate, potentially alongside removing or altering the `Local` funcionality for `unix` systems in the main `chrono` crate.

x-hgg-x:
For information, I have reduced the MSRV to `rustc 1.45` for the version `v0.6.6` of `tz-rs`.

esheppa:
Thanks @x-hgg-x - I've updated to the latest version of `tz-rs`

src/offset/local.rs
```diff
@@ -75,6 +77,263 @@ fn datetime_to_timespec(d: &NaiveDateTime, local: bool) -> sys::Timespec {
...
+    fn utc_now() -> DateTime<Local> {
+        DateTime::<Local>::from_utc(Utc::now().naive_local(), FixedOffset::east(0))
+    }
+
+    fn utc_from_local(local: NaiveDateTime) -> DateTime<Local> {
+        DateTime::<Local>::from_utc(local, FixedOffset::east(0))
+    }
+
+    fn try_now() -> Result<DateTime<Local>, error::TzError> {
+        if path::Path::new(LOCALTIME_LOCATION).exists() {
```
MiSawa, line 270: Wouldn't we also need a `TZ` environment variable lookup and parse here and other places that use `LOCALTIME_LOCATION`?

esheppa: Hi @MiSawa - thanks for pointing this out. I've added a check for the `TZ` environment variable, and if it exists, use that with preference to the `/etc/localtime` file.

MiSawa: Awesome, thank you!

esheppa:
Thanks @djc - I'll close this and leave some comments on #677.

PR #657, opened 2022-03-18
Make library compatible with rust 2021

erikh:
Signed-off-by: Erik Hollensbe <git@hollensbe.org>

      about adding the PR number)

[changelog]: ../CHANGELOG.md


djc:
We're not going to do this because this raises the minimum supported Rust version way too fast. If you want to help with getting #653 working, that would be great.

PR #661, opened 2022-03-22
Cleanups

djc:
My goal here is to make the codebase easier to navigate and take advantage of the MSRV upgrade from #653 to employ more modern Rust standards.


src/datetime/rustc_serialize.rs
```diff
@@ -0,0 +1,119 @@
+use super::DateTime;
+#[cfg(feature = "clock")]
+use crate::offset::Local;
+use crate::offset::{FixedOffset, LocalResult, TimeZone, Utc};
+use core::fmt;
+use core::ops::Deref;
+use rustc_serialize::{Decodable, Decoder, Encodable, Encoder};
```
Milo123459, line 7: Why use rustc_serialize? Serde?

Milo123459 (approved):
LGTM, although serde should be the default.

src/datetime/rustc_serialize.rs
djc: I'm not making any backwards incompatible changes here, so that means keeping rustc_serialize.

djc:
> LGTM, although serde should be the default.

Do you mean that we should enable the serde feature by default? I don't agree -- most well-maintained libraries keep it as an off-by-default feature, so that downstream users who don't need it don't have to compile the extra dependency.

PR #664, opened 2022-03-23
Add support for microseconds & optional timestamps serde serialization for `NaiveDateTime`

nickelc:
This PR brings the `naive::datetime::serde` module on par with the `datetime::serde` module.


djc:
The changes look okay to me, but I notice there isn't any test coverage. Could you write some basic tests that cover the code added here?

nickelc:
The code has similar doc examples as tests as the `datetime::serde` module.

Issue #654, opened 2022-03-06
Get week start and end days


sestrella:
Inspired by Rails [DateAndTime::Calculations](https://edgeapi.rubyonrails.org/classes/DateAndTime/Calculations.html) module, it would be useful to have a set of functions that return week boundaries based on a weekday. This proposal - if approved - intends to add the following functions to `IsoWeek` implementation:

```rs
fn start_day(&self, weekday: Weekday) -> NaiveDate {
  NaiveDate::from_isoywd(self.year(), self.week(), weekday)
}
```

Equivalent to Rails [beginning_of_week](https://edgeapi.rubyonrails.org/classes/DateAndTime/Calculations.html#method-i-beginning_of_week).

```rs
fn end_day(&self, weekday: Weekday) -> NaiveDate {
  self.start_day(weekday) + Duration::days(6)
}
```

Equivalent to Rails [end_of_week](https://edgeapi.rubyonrails.org/classes/DateAndTime/Calculations.html#method-i-end_of_week).

Additionally, a function that returns the start and end of the week at once could be added as well:

```rs
fn start_and_end_days(&self, weekday: Weekday) -> (NaiveDate, NaiveDate) {
  let start_day = self.start_day(weekday);
  (start_day, start_day + Duration::days(6))
}
```

**Usage**

A common use case would be getting the week boundaries based on a day:

```rs
let week = Local::today().iso_week();

let start_day = week.start_day(Weekday::Mon)
let end_day = week.end_day(Weekday::Mon)
// or
let (start_day, end_day) = week.start_and_end_days(Weekday::Mon)
```

djc:
Can you talk more about the use case, what you are trying to achieve that is harder with the current API than (you think) it should be? There are any number of ways we could go about this, but you're describing potential solutions more than the problem you started out with.

sestrella:
Sure let me provide more context, when it comes down to reports it is quite common to collect data within a date range such as weekly, monthly, quarterly, or yearly, therefore, getting week boundaries based on a specific date could be useful, in fact, that is the main reason why I came out with this proposal since right now I'm pulling data from an API that takes `start_day` and `end_day` as arguments. About

> what you are trying to achieve that is harder with the current API than

I would say the current API is very polished and coming up with a solution for the use case described above was not difficult, but at the same time, it feels like a common use case which is why I made this proposal. I hope this extra context provides more insights about the problem. Thank you for reviewing this proposal.

djc:
[...]

PR #666, opened 2022-03-29
[#654] Get week start and end days

sestrella:

      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -487,7 +487,7 @@ pub use format::{ParseError, ParseResult};
 
 pub mod naive;
 #[doc(no_inline)]
-pub use naive::{IsoWeek, NaiveDate, NaiveDateTime, NaiveTime};
+pub use naive::{IsoWeek, NaiveDate, NaiveDateTime, NaiveTime, NaiveWeek};
 
 pub mod offset;
 #[cfg(feature = "clock")]
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -2517,7 +2517,7 @@ mod tests {
     }
 
     #[test]
-    fn test_naiveweek_first_and_last_day() {
+    fn test_naiveweek() {
         let date = NaiveDate::from_ymd(2022, 5, 18);
         let asserts = vec![
             (Weekday::Mon, "2022-05-16", "2022-05-22"),
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -5,7 +5,7 @@
 
 #[cfg(any(feature = "alloc", feature = "std", test))]
 use core::borrow::Borrow;
-use core::ops::{Add, AddAssign, Sub, SubAssign};
+use core::ops::{Add, AddAssign, Sub, SubAssign, RangeInclusive};
 use core::{fmt, str};
 
 use num_integer::div_mod_floor;
@@ -19,7 +19,7 @@ use crate::format::{parse, ParseError, ParseResult, Parsed, StrftimeItems};
 use crate::format::{Item, Numeric, Pad};
 use crate::naive::{IsoWeek, NaiveDateTime, NaiveTime};
 use crate::oldtime::Duration as OldDuration;
-use crate::{Datelike, Weekday};
+use crate::{Datelike, Duration, Weekday};
 
 use super::internals::{self, DateImpl, Mdf, Of, YearFlags};
 use super::isoweek;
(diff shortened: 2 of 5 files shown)


djc:
Thanks for working on this!

I think I suggested to make these `NaiveDateTime` instead of `NaiveDate`, are you still planning to change that?

src/naive/isoweek.rs
```diff
@@ -160,4 +180,29 @@ mod tests {
...
+        assert_eq!(week.end_day(Weekday::Tue), NaiveDate::from_ymd(2022, 4, 4));
+    }
+
+    #[test]
+    fn test_bounds() {
+        let date = NaiveDate::from_ymd(2022, 3, 28);
+        let week = date.iso_week();
+
+        assert!(week.bounds(Weekday::Mon).contains(&date));
+        assert!(week.bounds(Weekday::Tue).contains(&date));
```
sestrella, line 206: @djc I'm submitting this PR mainly to start a conversation. After adding the `bounds` function I realized that `IsoWeek` might be the wrong place to add this feature. In the context of `test_bounds` the assertion `week.bounds(Weekday).contains(&date)` must be true disregarding which `Weekday` is passed as an argument, returning a week where the original date is out of range might be a little bit counterintuitive. Therefore, I think it would be better to move these 3 functions to the place where `iso_week` is defined since it is required to have access to the `Weekday` of the original date in order to compute the right week boundaries. What do you think?

djc: Yeah, that makes sense. Would be nice to add some comments pointing this out so people can find this functionality if they're looking for it in the other place.

sestrella: For any `NaiveDate` the following assertions must always be fulfilled:

```rs
let date = // Any random NaiveDate
for week in Weekday::Mon..Weekday::Sun {
  assert!(date.week_bounds(week).contains(&date));
}
```

sestrella:
Regarding:

> I think I suggested to make these NaiveDateTime instead of NaiveDate, are you still planning to change that?

Sorry, I was about to leave a message explaining why I choose a `NaiveDate` over a `NaiveDateTime`. `NaiveDateTime` works well for the `start_day` and `end_day` functions, unfortunately, the week coming from `bounds(Weekday)` cannot be represented as a range of `NaiveDateTime` since there is no `Step` trait to create a `Range<NaiveDateTime>` and the internal iterator `NaiveDateDaysIterator` only works for `NaiveDate`. From the consumer standpoint, I think it would be useful if `bounds(Weekday)` returns something that could be treated as `Range` or `Iterator (NaiveDateDaysIterator)` in that sense `NaiveDate` might be a good choice for the `start_day` and `end_day` functions since they defined the bounds of the week for `bounds(Weekday)`, I'm thinking about the following use cases:

- Looping over the days in a week
- Checking if a day is part of a week (this could be done via [contains](https://doc.rust-lang.org/std/ops/struct.Range.html#method.contains) which is part of `Range`)

What do you think?

djc:
That makes sense to me, thanks for explaining your position in more detail!

sestrella:
@djc I think this PR is finally ready for a full review. Would you mind taking a look at it, please?

sestrella:
The [issue](https://github.com/chronotope/chrono/runs/6295991590?check_suite_focus=true) reported by the CI was fixed.

src/lib.rs
```diff
@@ -450,13 +448,15 @@ pub mod prelude {
     #[doc(no_inline)]
     pub use crate::SubsecRound;
     #[doc(no_inline)]
+    pub use crate::{Date, Week};
```
djc, line 451: I don't think these need to be added to the prelude, since they're pretty niche.

src/traits.rs
```diff
@@ -1,4 +1,63 @@
...
+    /// # Examples
+    ///
+    /// ```
+    /// use chrono::{NaiveDate, Weekday, Weeklike};
+    ///
+    /// let date = NaiveDate::from_ymd(2022, 4, 18);
+    /// let week = date.week(Weekday::Mon);
+    /// assert!(week.start_day() <= date);
+    /// ```
+    fn start_day(&self) -> Self::Day;
```
djc, line 21: How do you feel about renaming `start_day()` and `end_day()` to `first_day()` and `last_day()`? My worry is that `end_day()` could be interpreted as either the last day of the week or the first day of next week, which could be confusing, whereas this is explicit for `last_day()`.

src/traits.rs
```diff
@@ -1,4 +1,63 @@
...
+    /// let week = date.week(Weekday::Mon);
+    /// let range = week.range();
+    /// assert_eq!(range.start(), &week.start_day());
+    /// assert_eq!(range.end(), &week.end_day());
+    /// ```
+    ///
+    /// [start_day]: ./trait.Weeklike.html#tymethod.start_day
+    /// [end_day]: ./trait.Weeklike.html#method.end_day
+    #[inline]
+    fn range(&self) -> RangeInclusive<Self::Day> {
```
djc, line 57: I think I'd prefer to call this `days()` to make it clear what the range's `Item` is. What do you think?

src/naive/date.rs
```diff
@@ -1095,6 +1115,13 @@ impl NaiveDate {
     pub fn iter_weeks(&self) -> NaiveDateWeeksIterator {
         NaiveDateWeeksIterator { value: *self }
     }
+
+    /// Returns the [`NaiveWeek`] that the date belongs, starting with the
```
djc, line 1119: "belongs to".

I think this could use some example to make to make the semantics clearer. Also should we rename the `weekday` argument to `start` or something here?

src/naive/date.rs
```diff
@@ -50,6 +50,26 @@ const MIN_DAYS_FROM_YEAR_0: i32 = (MIN_YEAR + 400_000) * 365 + (MIN_YEAR + 400_0
 #[cfg(test)] // only used for testing, but duplicated in naive::datetime
 const MAX_BITS: usize = 44;
 
+/// A week represented by a [`NaiveDate`] and a [`Weekday`] which is the first
+/// day of the week.
+#[derive(Debug)]
+pub struct NaiveWeek {
+    date: NaiveDate,
+    weekday: Weekday,
```
djc, line 58: I think `weekday` should be called `start`, as that is the intended role within the type (as far as I understand).

src/naive/date.rs
```diff
@@ -50,6 +50,26 @@ const MIN_DAYS_FROM_YEAR_0: i32 = (MIN_YEAR + 400_000) * 365 + (MIN_YEAR + 400_0
...
+}
+
+impl Weeklike for NaiveWeek {
+    type Day = NaiveDate;
+
+    #[inline]
+    fn start_day(&self) -> Self::Day {
+        let start = self.date.weekday().num_days_from_monday() as i64;
```
djc, line 66: We could calculate this in `u8`, right? Why are we casting here?

src/naive/date.rs
```diff
@@ -50,6 +50,26 @@ const MIN_DAYS_FROM_YEAR_0: i32 = (MIN_YEAR + 400_000) * 365 + (MIN_YEAR + 400_0
...
+        let end = self.weekday.num_days_from_monday() as i64;
+        let days = start - end;
+        self.date - Duration::days(if days >= 0 { days } else { 7 + days })
```
djc, line 69: I think `if days >= 0 { days } else { 7 + days }` could be more naturally expressed with some kind of modulo?

djc (commented):
Sorry for the delay in reviewing this -- I was on holiday yesterday.

I think it mostly looks good, but I have some feedback to consider.

sestrella:
@djc the latest commit seems to fix the CI issue, unfortunately the build is not getting trigger without an approval:

image

However, I tried it locally and it works:

image

src/date.rs
```diff
@@ -16,12 +16,29 @@ use rkyv::{Archive, Deserialize, Serialize};
...
+use crate::naive::{self, IsoWeek, NaiveDate, NaiveTime, NaiveWeek};
 use crate::offset::{TimeZone, Utc};
 use crate::oldtime::Duration as OldDuration;
+use crate::traits::Weeklike;
 use crate::DateTime;
 use crate::{Datelike, Weekday};
 
+/// A week represented by a [`NaiveWeek`] + [Offset](crate::offset::Offset).
+#[derive(Debug)]
+pub struct Week<Tz: TimeZone> {
```
djc, line 28: So, after reviewing some other PRs I'm on the fence about this concept of types that don't point to a specific time but do reference a timezone. Since the type represents a time span, the timezone offset could conceivably be different at the start of the span vs the end of the span, which we cannot represent accurately.

How important is having this type included to you? Could we make do with just the `NaiveWeek`?

sestrella: @djc  - It makes sense to me, starting with `NaiveWeek` sounds fair. I just updated the PR simplifying the implementation a little bit.

sestrella:
Pushed a PR addressing the error reported by the linter, sorry for not catching that before.

```sh
> cargo fmt -- --check --color=always && echo "OK"
OK
```

djc:
Still some doc link issues, can you check that out?

sestrella:
Hopefully, everything turns green this time 🤞🏼 

```
> cargo +nightly doc && echo "OK"
OK
```

djc:
Thanks for seeing this through!

sestrella:
Thank you for all the feedback and time spent on this PR

DavidMazarro:
Closes #654

PR #669, opened 2022-04-05
Move the `datetime::serde` timestamp visitors into the respective modules

nickelc:
The internal structs can be moved into their respective modules because `pub(super)` is now supported by the MSRV.

This is technically a breaking change but these structs were hidden from the docs.


nickelc:
> so I'd like to postpone this until at least after the 0.4.20 release and maybe to 0.5.

sure, I'm fine with it.
The move is just to make the datetime serde modules more comparable.

esheppa (approved):

esheppa:
This looks good to me - are you happy to have this merged @djc?

djc:
Yes! Thanks.

PR #672, opened 2022-04-08
Update absolute path to 2018 syntax for WASM/wasmbind feature

danielalvsaaker:
Fixes a compile error where a struct path is in the 2015 format. Affected wasm32 targets with features "wasmbind" and "clock" enabled.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -121,6 +121,34 @@ jobs:
           RUST_VERSION: stable
           WASM: wasm_simple
 
+  wasm_emscripten:
+    strategy:
+      matrix:
+        os: [macos-latest]
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install rust
+        uses: actions-rs/toolchain@v1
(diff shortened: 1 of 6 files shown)


djc:
Thanks! Do you have a suggestion how we could test this in CI?

djc (approved):

djc:
The clippy failures are unrelated, but I don't have permissions to merge without the checks passing. If you have the chance, please fix them up in a separate commit, or I'll get to it soon.

danielalvsaaker:
> Thanks! Do you have a suggestion how we could test this in CI?

Actually, this error only occurs when both the "wasmbind" and "clock" features are enabled. CI would need to compile for this feature combinaton.

As for the lint errors, let me handle that shortly.

danielalvsaaker:
> The clippy failures are unrelated, but I don't have permissions to merge without the checks passing. If you have the chance, please fix them up in a separate commit, or I'll get to it soon.

Should we just silence clippy for the four trait methods it complains about? Renaming these would be breaking.

djc:
Lints have been fixed on main, please rebase.

PR #675, opened 2022-04-11
Duplicate vaguely platform-specific code

djc:


djc:
Let's handle this as part of #675 which includes the final steps to fix the vulnerability.

Issue #674, opened 2022-04-09
0.4.20 release planning


hoodmane:
Currently to build downstream packages for targets emscripten, I clone a fork of chrono with #593 backported onto v0.4.19 and patch the downstream package's `Cargo.toml` by adding
```patch
+[patch.crates-io]
+chrono = { path =  "path/to/forked/chrono" }
```
I am looking forward to not having to do this.

djc:
We can do a beta release for now, see #676. In order to get to a 0.4.20 final release I want to merge #677 (or some other solution for #499) and ideally we'll then do some release candidate testing.

cramhead:
I notice that version 0.4.19 is using version 0.1.43 of the time crate. I noticed https://rustsec.org/advisories/RUSTSEC-2020-0071 today via `cargo audit`. After looking into time a little I see that time is now on version 0.3.9, https://github.com/time-rs/time/blob/main/Cargo.toml. 
Could we add an upgrade to time to the next release?
I'd be happy to contribute, but I'm pretty new to rust.

Stargateur:
@cramhead this CVE doesn't affect chrono

krojew:
Any info when the new version will be released?

cgorski-airtime:
Is there a preferred way to track and provide feedback for the beta release? I am interested in helping to get the beta into an official release.

djc:
I've published 0.4.20-beta.1 to crates.io, would be happy to get everyone to test it!

djc:
I've published an rc.1 and written up some release notes for it: https://github.com/chronotope/chrono/releases/tag/v0.4.20-rc.1. Please let me know if anything is missing/unclear (@esheppa would especially value your feedback on the release notes).

esheppa:
@djc - the release notes look great! (However they made me realize that I probably need to document the behavior of the caching of the tzfile data)

Potentially we could add the following: 
* Rkyv support: https://github.com/chronotope/chrono/pull/644 and https://github.com/chronotope/chrono/pull/701/e920210e2f94ad85ff25c30ae902442d4e2872c2
* https://github.com/chronotope/chrono/pull/522

djc:
rkyv support was already there, near the top of the additions. I missed #522 on my previous run, and also added #572, #588 and #609.

djc:
0.4.20 has been released!

Xuanwo:
[...]

PR #676, opened 2022-04-12
Bump version to 0.4.20-beta.1

djc:
In the context of #674, it would be nice to make it easier for people to test the current state of the main branch.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "chrono"
-version = "0.4.19"
+version = "0.4.20-beta.1"
 description = "Date and time library for Rust"
 homepage = "https://github.com/chronotope/chrono"
 documentation = "https://docs.rs/chrono/"


djc:
@esheppa how about we get this out now/soon?

esheppa:
@djc - sounds good to me. It would probably be worth us revisiting https://github.com/chronotope/chrono/pull/683#discussion_r862423621 prior to the full release, but it doesn't block a beta

djc:
Can you create an issue about that? I was thinking we should write up some tentative release notes and publish them with the beta, this could go in there with a link to the feedback.

badboy:
This is now actually making it _harder_ for me to test out the `main` branch because just using this in the `patch` section of my `Cargo.toml` will complain that `0.4.20-beta.1` is unused in my build, because everything still uses `0.4.19`.

Can the beta release be published to crates.io now? Then I can upgrade to that for testing.

djc:
Sorry about that -- I've published beta.1 now.

PR #682, opened 2022-04-30
Test Local functions against date command

esheppa:
This is split from #678 and just adds the extra tests

### DIFF
--- a/src/offset/local/mod.rs
+++ b/src/offset/local/mod.rs
@@ -167,7 +167,7 @@ mod tests {
         //     assert_eq!("", date_command_str);
         // }
 
-        // This is used while a decision is made wheter the `date` output needs to 
+        // This is used while a decision is made wheter the `date` output needs to
         // be exactly matched, or whether LocalResult::Ambigious should be handled
         // differently
         match Local.from_local_datetime(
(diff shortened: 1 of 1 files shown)


src/offset/local/mod.rs
```diff
@@ -137,7 +137,149 @@ impl TimeZone for Local {
...
+    }
+
+    #[test]
+    #[cfg(unix)]
+    fn try_verify_against_date_command() {
+        // #TODO: investigate /bin/date command behaviour on macOS
+        // avoid running this on macOS, temporarily
+        // for date_path in ["/usr/bin/date", "/bin/date"] {
+        for date_path in ["/usr/bin/date"] {
+            if path::Path::new(date_path).exists() {
```
djc, line 200: Can you make this use an early continue? That would help reduce rightward drift.

djc (commented):
Looks great!

src/offset/local/mod.rs
esheppa: good call, the code was a bit gnarly, I've made some improvements by using NaiveDateTime instead

esheppa: I removed the loop across date commands as I tried this out on macOS, but I couldn't find the same command line arguments and so wasn't able the replicate the behavior desired to find the local timezone

PR #683, opened 2022-05-01
Implement `find_local_time_type_from_local`

esheppa:
* Implement `find_local_time_type_from_local` which uses a local instead of UTC timestamp to compare to the transitions and extra rules
* All transitions are converted to a local timestamp by adding the relevant offset, prior to comparison with the desired local timestamp
* Pass a `LocalResult` as far up as possible to ensure that ambiguous and skipped local timestamps are recognized

Potential additions:
* Having an extra function `try_from_local_datetime` in the `TimeZone` trait which returns a `LocalResult` allowing users of the library to handle ambiguous or skipped timestamps and avoid panics.


src/offset/local/unix.rs
```diff
@@ -12,33 +12,59 @@ use std::sync::Once;
...
 }
 
-fn offset(unix: i64) -> FixedOffset {
+fn offset(d: NaiveDateTime, local: bool) -> LocalResult<FixedOffset> {
     let info = unsafe {
         INIT.call_once(|| {
             INFO = Some(TimeZone::local().expect("unable to parse localtime info"));
         });
         INFO.as_ref().unwrap()
     };
```
esheppa, line 45: Potentially this caching behavior should be configurable. In a long running application, it might be unexpected that even when the system timezone changes, the application still uses the old timezone

esheppa:
Will try to recreate this to get CI running

PR #685, opened 2022-05-02
docs: add `deps.rs` badge

afonsojramos:
Adds the `deps.rs` badge to give some more awareness of the dependency status of the repository.

<hr>

> - [ ] Have you added yourself and the change to the [changelog]? (Don't worry
      about adding the PR number)

Does it make sense to add myself to the changelog in this case?

[changelog]: ../CHANGELOG.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 
 [![Chrono GitHub Actions][gh-image]][gh-checks]
 [![Chrono on crates.io][cratesio-image]][cratesio]
+[![Chrono Dependencies][depsrs-image]][depsrs]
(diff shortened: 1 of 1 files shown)


djc:
If this is about the time 0.1 dependency, that should be fixed soon. I don't think this is worth doing.

afonsojramos:
Well, it was actually because I noticed the `rustc-serialize` dependency, which maybe could be replaced by `serde` as per the recommendation of the Rust Library Team. This also helps to identify possible vulnerabilities, but I understand that this is not necessary _per se_.

djc:
We already have serde support -- the rustc-serialize support is just there to avoid breaking backwards compatibility.

PR #686, opened 2022-05-05
Fix parse panic

### LINKED ISSUE #645: Panic in `parse_from_rfc2822` with overflowing date
```rust
#[test]
fn panic() {
    chrono::DateTime::parse_from_rfc2822("31 DEC 262143 23:59 -2359");
}
```

botahamec:
This is to resolve #645. The test was previously panicking when subtracting the `FixedOffset` from the `NaiveDateTime`. There may be other places that use that subtraction that we want to look at in the future.


src/format/parsed.rs
```diff
@@ -626,6 +626,12 @@ impl Parsed {
         let offset = self.offset.ok_or(NOT_ENOUGH)?;
         let datetime = self.to_naive_datetime_with_offset(offset)?;
         let offset = FixedOffset::east_opt(offset).ok_or(OUT_OF_RANGE)?;
+
+        // this is used to prevent an overflow when calling FixedOffset::from_local_datetime
+        datetime
+            .checked_sub_signed(OldDuration::seconds(i64::from(offset.local_minus_utc())))
```
djc, line 632: Sorry, now that I'm looking at this in context I'm having second thoughts. What would it take to move this check into `TimeZone::from_local_datetime()`? Can you check what happens to that method in #677?

djc:

src/format/parsed.rs
botahamec: The problem I see is that `TimeZone::from_local_datetime()` returns a `LocalResult`, and I think it would be counter-intuitive if it returned `LocalResult::None` for that.

botahamec:
I had trouble fixing the merge conflict somehow, but it should be good now

djc:
Thanks!

PR #703, opened 2022-06-07
add second fraction normalisation

### LINKED ISSUE #702: `TimeLike` does not correct account for nanoseconds exceeding 1e9
I'm been using chrono as a reference implementation while fuzzing my own code, and I've found an error in chrono.

`.to_string()` correctly rounds seconds up to account for nanoseconds exceeding 1e9, but `.second()` and `.nanosecond()` from `TimeLike` do not.

Example:

```rs
use chrono::{NaiveDate, Timelike};

#[test]
fn chrono_error() {
    let chrono_dt = NaiveDate::from_ymd(2000, 1, 1).and_hms_nano(1, 0, 0, 1_010_000_000);
    assert_eq!(chrono_dt.to_string(), "2000-01-01 01:00:01.010");
    assert_eq!((chrono_dt.second(), chrono_dt.nanosecond()), (1, 10_000_000));
}
```

gives:

```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `(0, 1010000000)`,
 right: `(1, 10000000)`', src/main.rs:6:5
```

[rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fa333040c21e4cc1ab406911254390fb) showing the error.

samuelcolvin:
fix #702

      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md

### DIFF
--- a/src/naive/time/tests.rs
+++ b/src/naive/time/tests.rs
@@ -277,7 +277,6 @@ fn test_time_format() {
     assert_eq!(NaiveTime::from_hms_milli(23, 59, 59, 1_000).format("%X").to_string(), "23:59:60");
 }
 
-
 #[test]
 fn test_timelike_normalisation() {
     let t = NaiveTime::from_hms_nano(0, 0, 0, 1_123_456_789);
--- a/src/naive/time/tests.rs
+++ b/src/naive/time/tests.rs
@@ -276,3 +276,17 @@ fn test_time_format() {
     assert_eq!(NaiveTime::from_hms(13, 57, 9).format("%r").to_string(), "01:57:09 PM");
     assert_eq!(NaiveTime::from_hms_milli(23, 59, 59, 1_000).format("%X").to_string(), "23:59:60");
 }
+
+
+#[test]
+fn test_timelike_normalisation() {
+    let t = NaiveTime::from_hms_nano(0, 0, 0, 1_123_456_789);
+    assert_eq!(t.second(), 1);
+    assert_eq!(t.nanosecond(), 123_456_789);
+
+    let t = NaiveTime::from_hms_nano(23, 59, 59, 1_999_999_999);
+    assert_eq!(t.hour(), 24);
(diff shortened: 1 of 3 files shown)


src/naive/time/tests.rs
```diff
@@ -276,3 +276,17 @@ fn test_time_format() {
...
+
+
+#[test]
+fn test_timelike_normalisation() {
+    let t = NaiveTime::from_hms_nano(0, 0, 0, 1_123_456_789);
+    assert_eq!(t.second(), 1);
+    assert_eq!(t.nanosecond(), 123_456_789);
+
+    let t = NaiveTime::from_hms_nano(23, 59, 59, 1_999_999_999);
+    assert_eq!(t.hour(), 24);
```
samuelcolvin, line 288: I don't know if this is correct, the alternative is to alter `from_hms_nano_opt` and `with_nanosecond` to error if the total time exceeds 24hours which seems like a bigger change.

If we want to keep the change this way, I think better to be explicit about this with a test.

samuelcolvin:
Humm, I'm not sure what to do about these tests - it seems like to support leap seconds, you explicitly want the functionality that this PR alters?

djc:
Ah yeah, I don't have enough context on that -- I only became a maintainer of this crate recently.

src/naive/time/tests.rs
esheppa: In this case `.second()` should return 60 due to the nanosecond representation being equal to or greater than 1e9 which represents a leap second in `chrono`, but `.hour()` should still return 23

esheppa:
Hi @samuelcolvin - I've made some comments over at https://github.com/chronotope/chrono/issues/702 - please let me know your thoughts

pitdicker:
Closing due to inactivity and the concerns raised above. We still have https://github.com/chronotope/chrono/issues/702 to track this issue.

Issue #704, opened 2022-06-07
Update `time` packages version to 0.2.x because of security vulnerability


John15321:
Update `time` packages version to 0.2.x because of security vulnerability

By running `cargo audit` on your package you can clearly see that it's using a very old version of the `time` package which is not secure

```bash
$ cargo audit
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 417 security advisories (from /Users/john/.cargo/advisory-db)
    Updating crates.io index
    Scanning Cargo.lock for vulnerabilities (93 crate dependencies)
Crate:     rustc-serialize
Version:   0.3.24
Title:     Stack overflow in rustc_serialize when parsing deeply nested JSON
Date:      2022-01-01
ID:        RUSTSEC-2022-0004
URL:       https://rustsec.org/advisories/RUSTSEC-2022-0004
Solution:  No fixed upgrade is available!
Dependency tree:
rustc-serialize 0.3.24
└── chrono 0.4.19

Crate:     time
Version:   0.1.44
Title:     Potential segfault in the time crate
Date:      2020-11-18
ID:        RUSTSEC-2020-0071
URL:       https://rustsec.org/advisories/RUSTSEC-2020-0071
Solution:  Upgrade to >=0.2.23
Dependency tree:
time 0.1.44
└── chrono 0.4.19

Crate:     serde_cbor
Version:   0.11.2
Warning:   unmaintained
Title:     serde_cbor is unmaintained
Date:      2021-08-15
ID:        RUSTSEC-2021-0127
URL:       https://rustsec.org/advisories/RUSTSEC-2021-0127
Dependency tree:
serde_cbor 0.11.2
└── criterion 0.3.5
    └── chrono 0.4.19

error: 2 vulnerabilities found!
warning: 1 allowed warning found
```

You can also see the other ones giving a WARNING but the time on is the most important here

John15321:
Here is a linked PR if you could review it that would be very nice:
https://github.com/chronotope/chrono/pull/705

djc:
See https://github.com/chronotope/chrono/issues/602#issuecomment-1075915577.

Issue #602, opened 2021-10-11
CVE-2020-26235


acim:
Please update time dependency.

 Inverse Dependency graph
time 0.1.44 (registry+https://github.com/rust-lang/crates.io-index)
└── chrono 0.4.19 (registry+https://github.com/rust-lang/crates.io-index)

CryZe:
The CVE is wrong, 0.1 is not affected.

Arnavion:
Actually, this conversation is founded on a wrong assumption. chrono has the issue not because of its time 0.1 dependency, but because [it directly calls `localtime_r`]() in the same way that time 0.1 does. AFAICT the code from time 0.1 that calls `localtime_r` is never called by chrono, at least not in the `Local::now` call path.

So unless I'm missing something, while it's unfortunate that the PR to update to time 0.3 hasn't been merged for a long time, it also won't actually fix the crash reported here.

quodlibetor:
> After reading through the issues: would using a shared lock for set_var and the localtime_r invocation fix this issue?

`set_var` already takes the lock. Unfortunately last time I looked the rust stdlib does not expose the environment lock for clients (i.e., us). So we can't take the lock before calling `localtime_r`.

Even if we could, that wouldn't prevent other things that somehow affect the environment through libc from modifying it.

Arnavion:
cargo-audit already works fine in the case of optional dependencies that aren't enabled because the corresponding feature isn't enabled, because such dependencies are not in the lockfile, so cargo-audit doesn't know about them anyway.

tarcieri:
See https://github.com/rustsec/advisory-db/issues/288 and https://github.com/rustsec/rustsec/issues/21 for some potential options for false positive elimination

esheppa:
As a way of moving forward with this, are either of the following two solutions palatable:

1. Remove `chrono::offset::Local` from `chrono` and put it in a new library, this would mean that only people who explicity need to use `chrono::offset::Local` would be affected by the CVE (hopefully this is a much smaller group than the total users of `chrono`). This would require an `0.5` release of `chrono` which may or may not be acceptable.
[...]

PR #705, opened 2022-06-07
Update `time` packages version to a more secure one

John15321:

      about adding the PR number)
      we can't reintroduce it?
No, I don't really think I can write a test for `cargo audit` unless you want to wait a long time until it installs on your pipeline but we can discuss it I guess.

https://github.com/chronotope/chrono/issues/704

[changelog]: ../CHANGELOG.md


djc:
See https://github.com/chronotope/chrono/issues/602#issuecomment-1075915577.

John15321:
fair enough

PR #707, opened 2022-06-08
Morph elapsed_years() API into years_since()

djc:
@yozhgoor would you like to review? This:

* Is more flexible, since it allows you to provide both dates for the interval
* Is more precise, since it returns `None` if the `base` is in the future
* Doesn't have to be guarded with `[cfg(feature = "clock")]` since it doesn't rely on `Utc::today()`
* Does not expose imprecise new constants as part of the public API

### DIFF
--- a/src/date.rs
+++ b/src/date.rs
@@ -276,21 +276,16 @@ impl<Tz: TimeZone> Date<Tz> {
         self.date
     }
 
-    /// Retrieves the elapsed years from now to the given [`Date`].
-    #[cfg(feature = "clock")]
-    pub fn elapsed_years(&self) -> u32 {
-        let now = Utc::today().with_timezone(&self.timezone());
+    /// Returns the number of whole years from the given `base` until `self`.
+    pub fn years_since(&self, base: Self) -> Option<u32> {
+        let mut years = self.year() - base.year();
+        if (self.month(), self.day()) < (base.month(), base.day()) {
+            years -= 1;
+        }
 
-        let years = if (now.month(), now.day()) < (self.month(), self.day()) {
-            now.year() - self.year() - 1
-        } else {
-            now.year() - self.year()
-        };
-
-        if years.is_positive() {
-            years as u32
-        } else {
-            0
+        match years >= 0 {
+            true => Some(years as u32),
+            false => None,
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/date.rs
```diff
@@ -276,21 +276,16 @@ impl<Tz: TimeZone> Date<Tz> {
...
+    /// Returns the number of whole years from the given `base` until `self`.
+    pub fn years_since(&self, base: Self) -> Option<u32> {
+        let mut years = self.year() - base.year();
+        if (self.month(), self.day()) < (base.month(), base.day()) {
+            years -= 1;
+        }
+
+        match years >= 0 {
+            true => Some(years as u32),
+            false => None,
```
cecton, line 288: That's a sophisticated if-else xD

Is it intentional? It seems you did that everywhere. Maybe to avoid if-expressions?

src/lib.rs
```diff
@@ -508,20 +508,6 @@ pub use month::{Month, ParseMonthError};
...
-    pub mod f32 {
-        /// Number of weeks in a year.
-        pub const WEEKS_PER_YEAR: f32 = 52.1775;
-    }
-    /// Constants of type `f64`.
-    pub mod f64 {
-        /// Number of weeks in a year.
-        pub const WEEKS_PER_YEAR: f64 = 52.1775;
-    }
-}
```
cecton, line 523: This is not "imprecise". This actually very precise: https://en.wikipedia.org/wiki/Week#Definition_and_duration

src/naive/datetime/tests.rs
```diff
@@ -240,3 +240,22 @@ fn test_nanosecond_range() {
...
+#[test]
+#[cfg(feature = "clock")]
+fn test_years_elapsed() {
+    use crate::Utc;
+    const WEEKS_PER_YEAR: f32 = 52.1775;
+
+    // This is always at least one year because 1 year = 52.1775 weeks.
+    let one_year_ago = Utc::today() - Duration::weeks((WEEKS_PER_YEAR * 1.5).ceil() as i64);
+    // A bit more than 2 years.
+    let two_year_ago = Utc::today() - Duration::weeks((WEEKS_PER_YEAR * 2.5).ceil() as i64);
```
cecton, line 253: You should probably not use `today` here but give a specific date. This test is tricky to get it right. @yozhgoor did it right but it isn't easy to understand why.

If you give a specific date, it will be right all the time and it's easier to understand.

cecton (commented):
tbh I think it's complementary.

* `elapsed_years()` reminds me of `std::time::Instant::elapsed()` https://doc.rust-lang.org/std/time/struct.Instant.html#method.elapsed its API is more concise and fit a particular use case: how many years has been spent since... birthday, subscription date, posted, etc...
* `years_since()` is probably useful. I don't have a use case in mind though.

It doesn't cost much to implement both because elapsed_years can call years_since.

src/naive/datetime/tests.rs
cecton: I mean: the test success depends on the exact date you are running the test. If you get it wrong you can easily get a flaky test that breaks sometimes every 4 years or so.

src/date.rs
djc: I find `match` expressions easier to read (in the case where the branch fits on one line): they're slightly more vertically compact and they lay out both branches right next to each other. I use these all the time, so I wouldn't describe this as particularly sophisticated.

src/lib.rs
djc: Well, if you define a `YEAR` to be a "mean year" computed across 400 years, then yes. For the conventional meaning of a year (of either 365 or 366 days) it isn't very precise.

src/date.rs
cecton: Yes as I never do that it looks pretty strange to me but I think you have a good point

src/lib.rs
cecton: So I guess it's just the naming that is incorrect

src/date.rs
yozhgoor: Personally I prefer to use functions like `is_positive()` instead of `>= 0`.

djc: It's not obvious to me without actually testing it what `is_positive()` returns for `0`. The direct comparison is also more compact. Why do you prefer the method call here?

src/lib.rs
djc: All other things being equal I'd prefer not to expose relatively niche API surface.

src/naive/datetime/tests.rs
djc: The tests I had here (which I had moved around due to some confusion) are direct ports of the original tests to the new API), so I don't think this is different from before?

src/date.rs
yozhgoor: For the readability (but it's a personal habits). `is_positive()` returns `false` if the number is either zero or negative.

djc:
I agree to some extent that they're complementary. At the same time, I'd argue that both are relatively niche (if only for the fact that we've managed to get by without them for many years) and, given that I'm here as a replacement maintainer trying to get a handle on maintenance I'd prefer to have less public API, so I'm inclined to stick with only the more powerful API for now.

(TBH if someone were to submit a PR adding back the other one with some test coverage, I'd probably merge it.)

src/naive/datetime/tests.rs
cecton: I mean the test is right and will succeed but it depends on an external variable that is the current date and it is hard to read and understand why it will always succeed.

It's safer and easier to use a specific date (take the current date for instance) so it doesn't vary each time the test is ran.

So my point is not about the test being wrong or right but the maintainability.

cecton:
> (TBH if someone were to submit a PR adding back the other one with some test coverage, I'd probably merge it.)

Since @yozhgoor originally made the API for elapsed_years instead of years_since despite it would have been easier, I expect that he would probably benefit from it.

src/lib.rs
cecton: It's not like the number of weeks in a mean year is going to change though

djc:
Thanks for the feedback!

PR #710, opened 2022-06-15
Remove unused libc dependency from Cargo.toml

metent:

      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md

libc dependency should be removed as it is no longer directly used by chrono. I'm not sure whether the changelog should document this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ Versions with only mechanical changes will be omitted from the following list.
 * Implement `DoubleEndedIterator` for `NaiveDateDaysIterator` and `NaiveDateWeeksIterator`
 * Fix panicking when parsing a `DateTime` (@botahamec)
 * Add support for getting week bounds based on a specific `NaiveDate` and a `Weekday` (#666)
+* Remove libc dependency from Cargo.toml.
 
 ## 0.4.19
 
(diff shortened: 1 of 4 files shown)


esheppa:
Hi @metent - thanks for this, it looks good to me. If possible could you update the documentation about this feature [here](https://github.com/chronotope/chrono/blob/8dd7245f85d230aeb5216f934d3385376f27bbe1/src/lib.rs#L46). I think its worth putting in the changelog as this could be "potentially breaking" in a weird case - because libc is an optional crate, people could add it to their chrono features. We could resolve this by adding a feature called "libc" that does nothing as specified [here](https://doc.rust-lang.org/cargo/reference/semver.html#cargo-remove-opt-dep)

esheppa:
Hi @metent - thanks for this, it looks good to me. If possible could you update the documentation about this feature [here](https://github.com/chronotope/chrono/blob/8dd7245f85d230aeb5216f934d3385376f27bbe1/src/lib.rs#L46). I think its worth putting in the changelog as this could be "potentially breaking" in a weird case - because libc is an optional crate, people could add it to their chrono features. We could resolve this by adding a feature called "libc" that does nothing as specified [here](https://doc.rust-lang.org/cargo/reference/semver.html#cargo-remove-opt-dep)

metent:
Done

esheppa:
Thanks @metent

PR #711, opened 2022-06-15
`NaiveDateTime` to `DateTime` helper method

### LINKED ISSUE #687: Make conversion from NaiveDateTime to DateTime easier
Hello,

I noticed that the conversion from a NaiveDateTime to a DateTime by supplying a TimeZone is not really as easy as it could be (or IDE-friendly).

Mostly, I think the problem is that you have the conversion method on the `TimeZone` trait, instead of on the `NaiveDateTime` object. At least speaking for myself, if I want to convert a NaiveDateTime, I already have one (e.g. because I manually parsed it from somewhere without Time information) and then just want to supply the TimeZone object. As a normal developer, I would just press `.` on the object and hope that my IDE will display me a suitable method.

Therefore, what about something like this?

```rust
impl NaiveDateTime {
    pub fn to_datetime_local<Tz: TimeZone>(&self, tz: Tz) -> LocalResult<DateTime<Tz>> {
        tz.from_local_datetime(self)
    }
}
```

This makes it really obvious how to construct a DateTime<TimeZone> from a NaiveDateTime:

```rust
#[test]
fn test_naivedatetime_to_local() {
    let offset = FixedOffset::east(3600);
    let dt = NaiveDate::from_ymd(2022,05,07).and_hms(14,51,00)
        .to_datetime_local(offset);
    assert_eq!("2022-05-07T14:51:00+01:00", dt.single().unwrap().to_rfc3339());
}
```

Furthermore, while the `LocalResult` is of course correct, I think it would be nice to have some kind of "lossy" output. I would assume that most of the time there is only one result (as the DST transitions are mostly in the night anyway, where they have the least impact). Currently the user has to handle this situation every time, while most of the time the conversion will be correct. Anyway, I think it would be good for a time library to have some sensible default how to work with this, as not every user might know what to do in this case.

I checked on the Java time API, and they handle it like this (https://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html#atZone-java.time.ZoneId-):

> In most cases, there is only one valid offset for a local date-time. In the case of an overlap, where clocks are set back, there are two valid offsets. This method uses the earlier offset typically corresponding to "summer".
>
> In the case of a gap, where clocks jump forward, there is no valid offset. Instead, the local date-time is adjusted to be later by the length of the gap. For a typical one hour daylight savings change, the local date-time will be moved one hour later into the offset typically corresponding to "summer".
>
> To obtain the later offset during an overlap, call [ZonedDateTime.withLaterOffsetAtOverlap()](https://docs.oracle.com/javase/8/docs/api/java/time/ZonedDateTime.html#withLaterOffsetAtOverlap--) on the result of this method. To throw an exception when there is a gap or overlap, use [ZonedDateTime.ofStrict(LocalDateTime, ZoneOffset, ZoneId)](https://docs.oracle.com/javase/8/docs/api/java/time/ZonedDateTime.html#ofStrict-java.time.LocalDateTime-java.time.ZoneOffset-java.time.ZoneId-).

Why not create an additional method on `LocalResult` which returns something like the Java handling (not quite sure how other time libraries handle this situation)?

botahamec:
fixes #687

I chose `and_timezone` as the method name. The issue proposes using `to_datetime_local`. I can see pros and cons with both, so let me know if we want to use a different name

### DIFF
--- a/src/naive/datetime/mod.rs
+++ b/src/naive/datetime/mod.rs
@@ -732,7 +732,7 @@ impl NaiveDateTime {
     /// use chrono::{NaiveDate, Utc};
     /// let dt = NaiveDate::from_ymd(2015, 9, 5).and_hms(23, 56, 4).and_timezone(Utc).unwrap();
     /// assert_eq!(dt.timezone(), Utc);
-    pub fn and_timezone<Tz: TimeZone>(&self, tz: Tz) -> LocalResult<DateTime<Tz>> {
+    pub fn and_local_timezone<Tz: TimeZone>(&self, tz: Tz) -> LocalResult<DateTime<Tz>> {
         tz.from_local_datetime(self)
     }
 }
--- a/src/naive/datetime/mod.rs
+++ b/src/naive/datetime/mod.rs
@@ -21,7 +21,7 @@ use crate::naive::date::{MAX_DATE, MIN_DATE};
 use crate::naive::time::{MAX_TIME, MIN_TIME};
 use crate::naive::{IsoWeek, NaiveDate, NaiveTime};
 use crate::oldtime::Duration as OldDuration;
-use crate::{Datelike, Timelike, Weekday};
+use crate::{Datelike, Timelike, Weekday, DateTime, LocalResult, TimeZone};
 
 #[cfg(feature = "rustc-serialize")]
 pub(super) mod rustc_serialize;
@@ -721,6 +721,19 @@ impl NaiveDateTime {
     #[inline]
     pub fn format<'a>(&self, fmt: &'a str) -> DelayedFormat<StrftimeItems<'a>> {
         self.format_with_items(StrftimeItems::new(fmt))
+    }
+
+    /// Converts the `NaiveDateTime` into the timezone-aware `DateTime<Tz>`
+    /// with the provided timezone, if possible.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use chrono::{NaiveDate, Utc};
+    /// let dt = NaiveDate::from_ymd(2015, 9, 5).and_hms(23, 56, 4).and_timezone(Utc).unwrap();
+    /// assert_eq!(dt.timezone(), Utc);
+    pub fn and_timezone<Tz: TimeZone>(&self, tz: Tz) -> LocalResult<DateTime<Tz>> {
+        tz.from_local_datetime(self)
     }
 }
 
--- a/src/naive/datetime/mod.rs
+++ b/src/naive/datetime/mod.rs
@@ -726,11 +726,14 @@ impl NaiveDateTime {
     /// Converts the `NaiveDateTime` into the timezone-aware `DateTime<Tz>`
     /// with the provided timezone, if possible.
     ///
+    /// This is experimental and might be removed in the future. Feel free to
+    /// let us know what you think about this API.
+    ///
     /// # Example
     ///
     /// ```
     /// use chrono::{NaiveDate, Utc};
-    /// let dt = NaiveDate::from_ymd(2015, 9, 5).and_hms(23, 56, 4).and_timezone(Utc).unwrap();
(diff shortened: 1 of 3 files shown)


esheppa:
Thanks @botahamec. This is an interesting function however there is some history surrounding this, specifically [here](https://github.com/chronotope/chrono-rfcs/blob/master/notes/api-issues.md)

That said, there may be an argument for this way of generating a `DateTime<Tz>` using ymd/hms style functions, as the current way (`Utc.ymd().hms()`) requires the Date<Tz> which is also listed on the same page:
> `Date<Tz>` doesn't actually make sense.

Given all this, once the upcoming `0.4.20` release is out, there will likely be some scope for larger changes in the API for a `0.5`

The challenge here is to come up with an API that meets the following criteria:
* Encourages use of `DateTime` with `TimeZone` to avoid issues that come from using the `NaiveDateTime` style
* Relatively ergonomic
* One way to do things

Given the two options here of:
1.`tz.from_local_datetime(ndt)`
2. `ntd.and_timezone(tz)`

The first has the advantage of using the `TimeZone` trait, however I'm quite persuaded by the reasoning about the IDE friendliness of having inherent methods (in `0.5` we could also consider applying similar logic with `Datelike` and `Timelike`). 

After all this my conclusions are:
* Merge this in as it would be interesting to get feedback on whether people like using this API (perhaps we could document that it is experimental and might be removed/replaced in the future?)
* Potentially rename the function - it could be nice to include the word `local`, for example `local_with_timezone` 

Please let me know your thoughts

esheppa:
Hi @botahamec - potentially we could also call it `assert_timezone` - this is useful as this function has to either be potentially panic if returning a `DateTime<Tz>` or instead return `LocalResult<DateTime<Tz>>`. Given it currently returns the former, the `assert` in the name could suffice for both the 'from_local" part and the "can panic" part.

botahamec:
> this function has to either be potentially panic if returning a `DateTime<Tz>` or instead return `LocalResult<DateTime<Tz>>`. Given it currently returns the former, the `assert` in the name could suffice for both the 'from_local" part and the "can panic" part.

That'd be incorrect. It currently returns a `LocalResult`.

esheppa:
My apologies @botahamec, I should have re-checked the code rather than relying on memory - happy to go with either of the options suggested, or a better one you can think of, if you update the name (or leave it as-is, if you think that it is the best option), and add the documentation about stability/use then this should be good to merge

botahamec:
@esheppa I decided to go with `and_local_timezone`. I think this best matches the naming convention of the rest of the library. `with` is typically used to replace a part of the structure with something else, but `and` is used to change the type. I could go either way on whether it should be `and_local_timezone` or `local_and_timezone`.

src/naive/datetime/mod.rs
```diff
@@ -722,6 +722,22 @@ impl NaiveDateTime {
     pub fn format<'a>(&self, fmt: &'a str) -> DelayedFormat<StrftimeItems<'a>> {
         self.format_with_items(StrftimeItems::new(fmt))
     }
+
+    /// Converts the `NaiveDateTime` into the timezone-aware `DateTime<Tz>`
+    /// with the provided timezone, if possible.
+    ///
+    /// This is experimental and might be removed in the future. Feel free to
```
djc, line 729: Why are we calling this experimental again?

djc (approved):

src/naive/datetime/mod.rs
botahamec: It was proposed in this comment. Apparently this API gets reinvented a lot. https://github.com/chronotope/chrono/pull/711#issuecomment-1157647197

esheppa:
Thanks @botahamec

src/naive/datetime/mod.rs
djc: I don't think labeling this as experimental isn't helpful because it's not clear what downstream users should do with that information.

PR #712, opened 2022-06-16
Replace `winapi` with `windows-sys`

hotate29:

      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ default = ["clock", "std", "wasmbind"]
 alloc = []
 libc = []
 std = []
-clock = ["std", "winapi", "iana-time-zone"]
+clock = ["std", "windows-sys", "iana-time-zone"]
 wasmbind = ["wasm-bindgen", "js-sys"]
 unstable-locales = ["pure-rust-locales", "alloc"]
 __internal_bench = ["criterion"]
@@ -40,7 +40,7 @@ wasm-bindgen = { version = "0.2", optional = true }
 js-sys = { version = "0.3", optional = true } # contains FFI bindings for the JS Date API
 
 [target.'cfg(windows)'.dependencies]
-winapi = { version = "0.3.0", features = ["std", "minwinbase", "minwindef", "timezoneapi"], optional = true }
+windows-sys = { version = "0.42.0", features = ["Win32_System_Time", "Win32_Foundation"], optional = true }
 
 [dev-dependencies]
 serde_json = { version = "1" }
--- a/src/offset/local/windows.rs
+++ b/src/offset/local/windows.rs
@@ -13,9 +13,13 @@ use std::mem;
 use std::ptr;
 use std::time::{SystemTime, UNIX_EPOCH};
 
-use winapi::shared::minwindef::*;
-use winapi::um::minwinbase::SYSTEMTIME;
-use winapi::um::timezoneapi::*;
+use windows_sys::Win32::Foundation::FILETIME;
+use windows_sys::Win32::Foundation::SYSTEMTIME;
+use windows_sys::Win32::System::Time::FileTimeToSystemTime;
+use windows_sys::Win32::System::Time::GetTimeZoneInformation;
+use windows_sys::Win32::System::Time::SystemTimeToFileTime;
+use windows_sys::Win32::System::Time::SystemTimeToTzSpecificLocalTime;
+use windows_sys::Win32::System::Time::TzSpecificLocalTimeToSystemTime;
 
 use super::{FixedOffset, Local};
 use crate::{DateTime, Datelike, LocalResult, NaiveDate, NaiveDateTime, NaiveTime, Timelike};
@@ -177,7 +181,7 @@ const HECTONANOSEC_TO_UNIX_EPOCH: i64 = 11_644_473_600 * HECTONANOSECS_IN_SEC;
 
 fn time_to_file_time(sec: i64) -> FILETIME {
     let t = ((sec * HECTONANOSECS_IN_SEC) + HECTONANOSEC_TO_UNIX_EPOCH) as u64;
-    FILETIME { dwLowDateTime: t as DWORD, dwHighDateTime: (t >> 32) as DWORD }
+    FILETIME { dwLowDateTime: t as u32, dwHighDateTime: (t >> 32) as u32 }
 }
 
 fn file_time_as_u64(ft: &FILETIME) -> u64 {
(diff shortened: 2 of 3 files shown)


hotate29:
This change brings MSRV to (maybe) 1.35.0. In the meantime, I will wait for CI.

esheppa:
Hi @hotate29 - thanks for this PR. If possible could you do the following:
* Update the CI yaml to use 1.46 so we can ensure the CI does pass
* Add an empty feature called "winapi" to avoid potential breakage (as per [here](https://doc.rust-lang.org/cargo/reference/semver.html#cargo-remove-opt-dep))

As far as the increase in MSRV we may be better off waiting for an `0.5` release. It looks like [debian stretch](https://packages.debian.org/stretch/rustc) is at version 1.41 but I don't have a good understanding of where might be currently using version 1.32 and also packaging `chrono` or packages that depend on `chrono`. Also from a quick investigation `windows-rs` lists the [rust-version](https://github.com/microsoft/windows-rs/blob/d3eee4a51e8c5975f2398c679525a040a61ecb35/crates/libs/sys/Cargo.toml#L11) as 1.46 which I think will likely be too much of an increase in the short term.

A solution in the meantime could be instead having a feature like "use-windows-rs" which would use "windows-rs" instead of "winapi" - however as per [this issue](https://github.com/rust-lang/cargo/issues/1839) I don't think this is currently possible.

Please let me know your thoughts?

djc:
Would like to keep this open -- I tagged with the API-incompatible tag so we can easily find such changes in the future.

hotate29:
> - Update the CI yaml to use 1.46 so we can ensure the CI does pass
@esheppa Do you mean to replace the part using 1.32 with 1.46?

esheppa:
@hotate29 - yes in the `.github/workflows/test.yml`. It's up to you whether you wan't to do this as it might be a while before it gets merged, but its nice to leave it in a green state for when it gets picked up at the point that the MSRV can increase to >= 1.46

hotate29:
@esheppa I understand. thanks!

kennykerr:
Curious, what's holding this change up? Let me know if I can help.

djc:
@kennykerr what's the MSRV for the windows crates these days? We recently bumped to 1.38 but are being conservative about bumping more (probably the upper bound we'd consider is 1.48 at this point).

kennykerr:
The MSRV for `windows-sys` 0.36 is 1.46.

`parking_lot` has already moved to 1.49 so I would recommend that version if you're planning to switch. The next version of `windows-sys` will likely require 1.49 as well.

djc:
I think the limitation here is stable Debian, which IMO still affects a sizable contingent of downstream users. Anyway -- I'm generally interested in moving to windows-sys but this is why it's stalled. If you want to be lower in the stack than chrono, I think you should maybe be more conservative about MSRV bumps as well? 🙂

From my perspective (admittedly as a user of macOS on the desktop and Linux on servers -- and for other people I work with either at my day job or in OSS, it seems like Windows is a minority) I don't see a big benefit to switching away from winapi (and it's just a few small API calls, too) whereas I feel a substantial amount of responsibility towards downstream projects who want to serve users which have older compilers.

kennykerr:
Sorry for the delay - I lost track of this issue. 

> probably the upper bound we'd consider is 1.48 at this point

I just tested 1.48 and that works great: https://github.com/microsoft/windows-rs/pull/2173

I'm happy to enforce that. As a practical matter, the current version of [windows-sys](https://crates.io/crates/windows-sys) works fine with 1.48 if you'd like to go ahead with that.

esheppa:
Thanks for this @kennykerr.

@djc - what are your thoughts on using this in `0.5` with a MSRV of `1.48`?

djc:
Yes, that seems fine. @hotate29 are you interested in updating this to bump the MSRV to 1.48?

hotate29:
Yes, I am interested. I will work on that.

hotate29:
@djc Yes, I am interested. I will work on that.

hotate29:
I have added an empty feature for compatibility ([e65acd1](https://github.com/chronotope/chrono/pull/712/commits/e65acd14d170f519e91ef78b23f9e6794af55a40)), but if this change is included in the 0.5 release, I don't think it is needed.
Should I remove it?

djc:
Yes, please remove the empty feature.

Would you mind rebasing this so that there are two commits:

* One to bump the MSRV and deal with all the fallout (clippy fixes etc)
* Another separate one that actually replaces winapi

Can also do two separate PRs if that's easier for you (then we can squash while merging).

This would make it easier for me to review the changes.

hotate29:
I rebased.

hotate29:
Oops, found a mistake.

hotate29:
Fixed.

Cargo.toml
```diff
@@ -11,6 +11,7 @@ readme = "README.md"
 license = "MIT/Apache-2.0"
 exclude = ["/ci/*"]
 edition = "2018"
+rust = "1.48.0"
```
djc, line 14: This is supposed to be `rust-version` and that only starts working after 1.56, so I don't think we should add it here.

djc (approved):
Looks great, one small issue to address, please fix it in the originating commit?

Cargo.toml
hotate29: Deleted. thanks!

djc: So please squash this into the first commit?

esheppa (approved):
Thanks for this @hotate29 - apologies to ask you to rebase this again, but we have recently merged the past while of work from `0.4.x` to `main` and if you can rebase this on the HEAD of `main` that would clean up the diff in this PR. (Separately if you could use `git rebase -i` to move 288a5a6 into ef3fe078 that would be great)

Cargo.toml
hotate29: Squashed.

hotate29:
@esheppa Rebased.

src/offset/local/windows.rs
```diff
@@ -36,7 +37,7 @@ pub(super) fn naive_to_local(d: &NaiveDateTime, local: bool) -> LocalResult<Date
         tm_yday: 0,                // and this
         tm_isdst: -1,
         // This seems pretty fake?
-        tm_utcoff: if local { 1 } else { 0 },
+        tm_utcoff: local as i32,
```
esheppa, line 44: do you mind using `i32::from(local)` here as per line 223 below?

src/offset/local/windows.rs
```diff
@@ -67,12 +68,14 @@ fn tm_to_datetime(mut tm: Tm) -> DateTime<Local> {
...
-    let time = NaiveTime::from_hms_nano(
+
+    let time = NaiveTime::from_hms_nano_opt(
         tm.tm_hour as u32,
         tm.tm_min as u32,
         tm.tm_sec as u32,
         tm.tm_nsec as u32,
-    );
+    )
+    .expect("invalid time");
```
esheppa, line 82: this could be cast to a `LocalResult::None` which would avoid the `expect` call (this is semantically sub-optimal, but is consistent with usage elsewhere):

```rust
let time = NaiveTime::from_hms_nano_opt(...).ok_or(LocalResult::None)?;
```

esheppa (changes requested):
Thanks @hotate29 - there are just a few more changes I'd like you to take a look at if possible

src/offset/local/windows.rs
hotate29: Pushed changes (will squash later).

src/offset/local/windows.rs
```diff
@@ -25,7 +25,8 @@ use super::{FixedOffset, Local};
 use crate::{DateTime, Datelike, LocalResult, NaiveDate, NaiveDateTime, NaiveTime, Timelike};
 
 pub(super) fn now() -> DateTime<Local> {
-    tm_to_datetime(Timespec::now().local())
+    let datetime = tm_to_datetime(Timespec::now().local());
+    datetime.single().expect("invalid time")
```
hotate29, line 29: An expectation is that the system will not return an invalid time.

esheppa:
Thanks for those fixes @hotate29 - if possible can you squash them down to the original two commits? We are using "Rebase and Merge" rather than "Squash and Merge" which is why we have to be a bit more picky about keeping all the commits within a given PR clean (I'll set up some docs on this to make it an explicit policy)

hotate29:
@esheppa Squashed.

src/offset/local/windows.rs
esheppa: This looks reasonable, in the case of `now()`

esheppa (approved):
Thanks for this @hotate29, looks good to go!

hotate29:
Thanks to all who worked on this PR!

PR #713, opened 2022-06-19
Expose MIN_TIME and MAX_TIME of NaiveTime

rikvdkleij:
See https://github.com/chronotope/chrono/issues/386#issuecomment-1159030041

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ Versions with only mechanical changes will be omitted from the following list.
 * Add support for getting week bounds based on a specific `NaiveDate` and a `Weekday` (#666)
 * Remove libc dependency from Cargo.toml.
 * Add the `and_local_timezone` method to `NaiveDateTime`
+* Expose `MIN_TIME` and `MAX_TIME` of `NaiveTime`
 
 ## 0.4.19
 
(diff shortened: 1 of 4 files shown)


djc:
Hmm, would it make more sense to expose these as `NaiveTime::{MIN, MAX}` associated types? I think that would be more idiomatic in modern Rust, at the cost of some consistency with current constants. We could introduce `DateTime::{MIN_UTC, MAX_UTC}` to fix the consistency, deprecating the other constants. @esheppa thoughts?

esheppa:
@djc - I think making them associated is a great idea. Given the two options, either associated constants on the trait or associated constants in the inherent impl, I'd lean toward the latter, among other reasons, people are likely to have the type names already imported, but potentially not the traits, so its more ergonomic to get at the constants if they are associated constants on the inherent impl. (This is the same IDE friendliness reasoning that was persuasive in https://github.com/chronotope/chrono/pull/711)

djc:
Okay, so let's do the associated consts (with the `_UTC` suffix in the case of types that take a `TimeZone` argument) and deprecate all the old `MIN_` and `MAX_` constants, with a note in the deprecation pointing to the new API.

@rikvdkleij does that seem like something you can tackle?

rikvdkleij:
> @rikvdkleij does that seem like something you can tackle?

@djc No, sorry, it was just my intention to get access to the `MIN_TIME` constant and make it consistent with the other constants.

djc:
Fixed this with #726 instead.

djc:
I guess that PR doesn't yet actually expose those values, so feel free to submit a new PR for that.

Issue #497, opened 2020-11-06
Completely migrate to core::time::Duration


Plecra:
This would be ideal for usability of the `Duration` type, and promote more compatibility with other crates. It's been mentioned a few times over the years, and was a goal [back in 2016](https://github.com/chronotope/chrono/issues/58#issuecomment-244266035).

As back then, this would be a major breaking change and would need to be prepared for 0.5. `core::time::Duration` also doesn't have some of the utility methods like `oldtime::Duration::days`, which we might want to replace in some fashion.

The main issue is `time::Duration`'s lack of support for negative durations. This will force us to change a few APIs that return possibly-negative values. It'd be good to review which APIs will change before working on the implementation.

gerob311:
Hey, I'm pretty new to Rust and `chrono`, but so far, I'm not convinced that completely replacing `chrono::Duration` with `time::Duration` is a good idea, because of the lack of support for negative durations that you mentioned.

How would you represent negative durations under this proposal?

djc:
Note that #1229 already added a bunch of support for using `core::time::Duration` with chrono's types, and https://github.com/rust-lang/rust/issues/120301 proposed adding a few more utility methods to `core::time::Duration`.

pitdicker:
Interesting, I didn't know you were working on that in the standard library.

djc:
Well, this is motivated by my day job work sharing me and my notions of not wanting to use `chrono::Duration` for anything ideally.

So in my mind, our current direction is aligned with this issue.

djc:
To be clear, I think there are some places in the API where we might still have a need for something like chrono's `Duration` (called `TimeDelta` since #1406), but I think that use is probably limited to output types in subtraction/difference calculations, where an unsigned type can't represent the range of values required and `Option` or `Result`-wrapping the outcome might impair the UX. For input types, I think the unsigned `core::time::Duration` is sufficient.

Issue #154, opened 2017-05-25
Provide access to subsecond nanos in the public interface of Duration.


tylerreisinger:
The way the Duration struct is implemented, if there are more total nanoseconds represented than i64::MAX, it is impossible to see greater than millisecond precision on the outside despite that information being stored. It would be useful to have a method similar to [std::time::Duration::subsec_nanos](https://doc.rust-lang.org/std/time/struct.Duration.html#method.subsec_nanos) which can return only the number of nanoseconds past the nearest second.

Additionally, for interfacing with chrono, it would be useful to have a constructor to create a Duration from a number of whole seconds plus the additional nanoseconds similar to std::time::Duration::new to efficiently construct Duration objects with full precision.

lifthrasiir:
I don't intend to modify `time::Duration` by now, but I plan for the new duration-like type for Chrono 0.4.1 (originally intended for 0.4.0 but delayed due to the time constraint).

kosta:
@lifthrasiir is there a public branch that I can track for the `time::Duration` replacement?

kangalioo:
What is the state of this? I am working on implementing chrono support for the PyO3 Python bindings crate. Chrono's lacking Duration API artificially limits the size of Duration's that can be supported, which is unfortunate.

pickfire:
@kangalioo I think adding it directly does not requires us to change the public interface. https://github.com/chronotope/chrono/pull/542

pitdicker:
Added in https://github.com/chronotope/chrono/pull/1327.

kangalio:
Thanks!

I would vouch to re-open this issue though, because the second part of the issue is not part of #1327:

> Additionally, for interfacing with chrono, it would be useful to have a constructor to create a Duration from a number of whole seconds plus the additional nanoseconds similar to std::time::Duration::new to efficiently construct Duration objects with full precision.

This specifically is what would be needed in pyo3 to support sub-second precision in the Python<->Rust conversion ([code permalink](https://github.com/PyO3/pyo3/blob/2d3dc654285eb8c9b29bb65b1d93323f551a8314/src/conversions/chrono.rs#L262-L287))

djc:
Please just file a new issue so we can track that request separately.

pitdicker:
Sorry, I missed that part of the first post. https://github.com/chronotope/chrono/pull/1137 includes a `Duration::new` method.

PR #714, opened 2022-06-21
Use core::time::Duration via an enum

esheppa:
Setting this up so it can run through the CI for now, this shouldn't be merged until at least version 0.5, or possibly never.

This is one option to use `core::time::Duration` in chrono's duration. The main change here is to remove the `oldtime` modules, and rename the `Duration` type to `SignedDuration` which can now be infallibly converted both ways with `core::time::Duration` (albeit via an absolute value when going to `core::time::Duration`)

```rust
pub enum SignedDuration {
    /// A duration heading in forwards in time
    Forwards(Duration),
    /// A duration heading in backwards in time
    Backwards(Duration),
}
```


djc:
It's an interesting idea. Can you explain explicitly:

(a) what problems does this solve?
(b) which downsides do you see?
(c) are there alternative solutions to (a) that we should consider?

(Effectively, this is like a mini-RFC.)

esheppa:
Thanks @djc - good idea re. mini-RFC. Worth noting here that I'll (hopefully shortly) set up another branch with a different approach using `core::time::Duration` only.

(a) what problems does this solve?
1. https://github.com/chronotope/chrono/pull/638 - however I'd need to replicate all the Add, Sub, etc impls with `core::time::Duration` or have them work with `Into<SignedDuration>` to get this, so it's somewhat debatable whether this helps. 
2. Moves some code from `chrono` into `core`/`std` - less code to review/test/maintain
3. `core::time::Duration` => `chrono::SignedDuration` is now infallible, allowing APIs that work with `SignedDuration` to potentially be changed to `Into<SignedDuration>`
4. `chrono::SignedDuration` => `core::time::Duration` now has a lossy but infallible option
5. Disambiguates the name of the type from equivalents in `core`/`std` 
6. Somewhat resolves https://github.com/chronotope/chrono/issues/717 - although removing the `oldtime` feature also resolves this
7. Resolves some longstanding issues: https://github.com/chronotope/chrono/issues/58 and https://github.com/chronotope/chrono/issues/497 and https://github.com/chronotope/chrono/issues/117 and https://github.com/chronotope/chrono/issues/154

 (b) which downsides do you see?
- Risk of incorrect implementation or subtle semantic change that breaks existing code without a compile error (could be partially resolved by adding even more test coverage)
- Migration, when going from 0.4 to 0.5, all instances of `chrono::Duration` have to be renamed - however after a bulk rename is done code should compile and work as before, unless the implementation details of the original type were being relied upon.
- The total size of `SignedDuration` is larger than the previous `chrono::Duration` as it includes the enum tag as well
- Increase in MSRV

 (c) are there alternative solutions to (a) that we should consider?
- Keep the `chrono::Duration` as is
- Use only `core::time::Duration` (this means removing or changing some APIs, so its more of a breaking change)
- Potential better solutions that people comment with when they see this PR :D

esheppa:
Looks like this might also help with #613 / #586 - assuming we encourage users towards using `core::time::Duration` where possible, or otherwise change the return types on the `SignedDuration` methods

Liamolucko:
This is just an opinion, but I think `TimeOffset` might be a better name, since a negative offset makes a bit more semantic sense than a negative duration.

esheppa:
Thanks for the feedback @Liamolucko. There is also some discussion to be had whether we use `core::time::Duration` directly and don't allow for negative duration, however that may be too much of a breaking change. I'd originally planned to have two PRs but it is probably easier to just have the discussion on this one and only implement the other option if that looks to be better.

Having the add/sub methods take `Into<chrono::Duration>` means that at least cases where users wanted to use `core::time::Duration` (eg in constants or with `serde`) are solved by this

esheppa:
@Liamolucko - what are your thoughts on `TimeDifference` or `TemporalDifference`, my thinking here is to disambiguate from the concept of `Offset` in relation to a timezone - although perhaps it might make sense to investigate harmonizing these instead.

One difference between the current `Duration` and `FixedOffset` is that `FixedOffset`'s resolution is seconds, while `Duration` is nanoseconds.

Liamolucko:
Those work, although they're a bit long. Maybe `TimeDelta`?

Another difference between this and `FixedOffset` is that `FixedOffset` can only be +/- 24 hours.

esheppa:
`TimeDelta` is especially nice due to [precedent](https://docs.python.org/3/library/datetime.html#datetime.timedelta) so that sounds like a great choice to me!

> Another difference between this and FixedOffset is that FixedOffset can only be +/- 24 hours.

Excellent point

djc:
The `TimeDelta` name is great and makes it much clearer what this is for. I think I'm inclined to not make this an enum but just a simple struct `Duration` like we have today in `oldtime`.

esheppa:
The advantage of it being an enum is that it internally stores the `core::time::Duration` which means we can delegate to all its trait implementations for derives (`serde` being an important one here) - it could also be a struct with a boolean for the direction. 

However even if we were to move away from either of those two options, I'd argue that we should ensure that we have an `impl From<core::time::Duration> for TimeDelta` so that all the operations can use `Into<TimeDelta>` and most users can just use the `core::time::Duration` for their operations (which has const constructors)

djc:
Hmm, I was thinking we should use `TimeDelta` as an output type but allow both `TimeDelta` and `core::time::Duration` as input types? Having `impl Into<TimeDelta>` as an argument all over the place has the potential of incurring a decent amount of compile-time overhead because we potentially have to compile all the functions twice (and it adds a layer of abstraction that makes things harder to understand).

esheppa:
Makes sense, so we just duplicate the impls, once for `core::time::Duration` and once for `TimeDelta`?

djc:
Where it makes sense, yes.

esheppa:
ok - and do you have a preference between the below and using an enum?

```rust
struct TimeDelta {
    duration: core::time::Duration,
    forwards: bool,
}
```

djc:
I guess it doesn't really matter -- which one do you think will make the code most readable?

esheppa:
I think the `enum` is cleaner from a readability perspective - will get the type renamed to `TimeDelta` and the operations impls fixed, plus to a rebase

esheppa:
Will rebase this again once `0.4.22` is released

esheppa:
Hmm these test failures probably warrant more discussion. The fundamental issue stems from the `core::time::Duration` having seconds as `u64` and the functions needing to return `i64` for API compatibilty. One option is to change the functions to return `i128` which would make it fully compatible with the `MIN/MAX` values of `core::time::Duration`, but the resulting `i128` could be a bit of a pain to deal with

djc:
Where do we need to return `i64` for compatibility?

esheppa:
So my plan here was to make the migration process as painless as possible. To do this I've tried to keep as many as possible of the function signatures the same, while changing the name and underlying implementation of the type (I could potentially even add back the fallible conversions for improved compatibility). The existing [`Duration::num_seconds`](https://docs.rs/chrono/latest/chrono/struct.Duration.html#method.num_seconds) method returns `i64`, so I'm trying to avoid returning something else like `i128`. Another way to solve it could be `abs_seconds` which returns a `u64`.

However, IMO there is a better way to solve this - which is that, realistically even i64 is a larger range of seconds than would ever be useful (except perhaps for physics use-cases, which might be better served by [`hifitime`](https://docs.rs/hifitime/latest/hifitime/) or some other custom implementation, probably using floating point numbers). My proposal would be to have some lower, sensible limits on `TimeDelta` MAX/MIN, and potentially equivalents on `NaiveDate`/`NaiveDateTime`/etc. For example:
* Define TimeDelta::MAX as i64::MAX milli/microseconds, or even a number less than this, and when converting from `core::time::Duration` then panic / error / truncate when out of range, meaning only `num_nanoseconds` needs to deal with i128/u128. (microseconds is my preference here as this allows all but `num_nanoseconds` to infallibly return `i64` and still covers a very large range of years).
* Define `NaiveDate` or `NaiveDateTime` MIN/MAX, and then define `TimeDelta`'s MIN/MAX in the context of that

This potentially has an effect on doing zero-copy de/serialization however as there are byte patterns that aren't valid for the type itself (eg underlying storage might be i64/u64, but some of the valid i64/u64 values wouldn't be valid for `TimeDelta`)

djc:
> So my plan here was to make the migration process as painless as possible. To do this I've tried to keep as many as possible of the function signatures the same, while changing the name and underlying implementation of the type (I could potentially even add back the fallible conversions for improved compatibility).

Hmm, not sure I agree with this plan. Given the fact that we have such a broad user base, we won't get to break API compatibility very often. I think 0.5 should be more correct/robust (fewer, and well-documented, panics) and expose a minimal API, even if that means migration would be more work. Given that 0.5 will get rid of the time 0.1 compatibility, I think willingness for maintained downstream packages to migrate will be reasonably high even in the face of substantial changes.

esheppa:
>  I think 0.5 should be more correct/robust (fewer, and well-documented, panics) and expose a minimal API, even if that means migration would be more work.

Excellent, this sounds like a good approach!

Another thought - perhaps `TimeDelta` could look like the below, (and be renamed?) to be compatible with `Months` and `Days` (alternatively they could get their own `*Delta` types. This would allow it to share helper methods etc as well. We could potentially include `Backwards` and `Forwards` in the prelude as well. In this case it wouldn't have methods like `num_seconds()` etc, the user would need to get to the underlying `core::time::Duration` and ask that.

```rust
pub enum TimeDelta<D = core::time::Duration> {
    Forwards(D),
    Backwards(D),
}

impl<D: Copy> TimeDelta<D> {
    fn abs(self) -> D {
        todo!()   
    }
   fn forwards(self) -> Option<D> {
        todo!()  
   }
   fn backwards(self) -> Option<D> {
       todo!()    
  }
}

```

and use cases like:

```rust
// handle the direction
match my_datetime - my_other_datetime {
    TimeDelta::Forwards(d) => (),
    TimeDelta::Backwards(d) => (),
}

// just get the magnitude
(my_datetime - my_other_datetime).abs()

// handle the direction
// could be just `Delta` here?
match my_date - my_other_date {
    TimeDelta::Forwards(d) => (),
    TimeDelta::Backwards(d) => (),
}

// just get the magnitude
(my_date - my_other_date).abs()

// handle the direction
// could be just `Delta` here?
match my_date.signed_months_since(my_other_date) {
    TimeDelta::Forwards(d) => (),
    TimeDelta::Backwards(d) => (),
}

// just get the magnitude
my_date.signed_months_since(my_other_date).abs()
```

djc:
Let's first do a version without generics and see where we feel a need for generics. My spidey sense is that there's too much complexity to justify the benefit but I don't have a good feel for the potential benefit right now.

esheppa:
closing in favour of https://github.com/chronotope/chrono/pull/824

Issue #720, opened 2022-06-25
CI is broken on cross-targets


esheppa:
This looks to potentially be due to a change in the naming convention of github releases. 

It can be resolved for now by hard-coding the previous tag "v0.2.1".

PR #721, opened 2022-06-25
Removed a couple of zero-prefixed literals

botahamec:
Small change to remove a couple of clippy warnings.


botahamec:
I don't what that failed check is but I suspect it's not something I did wrong

djc:
Nope, the CI failure is #720.

djc (approved):

djc:
Thanks!

Issue #719, opened 2022-06-25
Decide on caching behaviour for parsed tzinfo data


esheppa:
Currently we parse this lazily the first time that a user interacts with a `Local` timezone. This could be problematic in cases where the symlink for /etc/localtime is updated. See linked comment for further discussion

https://github.com/chronotope/chrono/pull/683#discussion_r862423621

djc:
I don't think we really have natural API surface where this could be configured. Absent configuration API, I see 3 options:

1. Always cache. Upside: only have to parse once, downside: timezone might become wrong if changed while app is running.
2. Never cache: always in tune with system timezone, but getting the local timezone is relatively expensive.
3. Cache with expiration: reparse current timezone at most once in, say, an hour, or a few minutes? Cost of parsing is amortized over many calls and app will be only be wrong for a little bit after timezone changes, some synchronization overhead and increased implementation complexity.

I wonder what libc does? Arguably we will attain maximum compatibility if we do the same thing as, say, glibc.

esheppa:
From a brief look at glibc it seems to do the following:
* each time that `mktime` is called, it calls `tzset`
* `tzset` internally, whether TZ env variable or a tzfile is used, checks if it has been changed from the last call. If no change is detected it uses a cached version, otherwise it reparses the variable/file

We could potentially check the symlink metadata each time and only re-parse if it has been modified (in the case of using a file) which is I think quite similar to what glic does. `chrono` currently always reparses if using a `TZ` variable

djc:
Yeah, checking the symlink mtime (not sure it's always a symlink?) seems like a decent improvement. As a further optimization, I suppose we could only check the mtime once per minute or so?

We could take the `Hasher::hash()` in the case of a `TZ` variable, I suppose?

djc:
@esheppa would you be able to take a stab at this? IMO getting a 0.4.20 out with what we've got so far is higher priority than triaging old pull issues/PRs.

esheppa:
Solved via #728

PR #728, opened 2022-07-05
Caching of parsed TimeZone information

esheppa:
This is a basic implementation of caching as described in https://github.com/chronotope/chrono/issues/719

### DIFF
--- a/src/offset/local/unix.rs
+++ b/src/offset/local/unix.rs
@@ -8,7 +8,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use std::sync::Once;
+use std::{cell::RefCell, env, fs, time::SystemTime};
 
 use super::tz_info::TimeZone;
 use super::{DateTime, FixedOffset, Local, NaiveDateTime};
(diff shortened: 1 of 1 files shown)


src/offset/local/unix.rs
```diff
@@ -37,35 +37,89 @@ pub(super) fn naive_to_local(d: &NaiveDateTime, local: bool) -> LocalResult<Date
...
+        }
+    })
+}
+
+thread_local! {
+    static TZINFO: RefCell<Option<Cache>> = RefCell::new(None);
+}
+
+enum Source {
+    Localtime(SystemTime),
```
djc, line 88: Style nit: let's call  this `LocalTime`.

src/offset/local/unix.rs
```diff
@@ -37,35 +37,89 @@ pub(super) fn naive_to_local(d: &NaiveDateTime, local: bool) -> LocalResult<Date
 }
 
 fn offset(d: NaiveDateTime, local: bool) -> LocalResult<FixedOffset> {
-    let info = unsafe {
-        INIT.call_once(|| {
-            INFO = Some(TimeZone::local().expect("unable to parse localtime info"));
-        });
-        INFO.as_ref().unwrap()
-    };
+    TZINFO.with(|cache| {
```
djc, line 40: Let's move all this code into a `&self` method on `Cache`?

src/offset/local/unix.rs
```diff
@@ -37,35 +37,89 @@ pub(super) fn naive_to_local(d: &NaiveDateTime, local: bool) -> LocalResult<Date
...
+                    .find_local_time_type(d.timestamp())
+                    .expect("unable to select local time type")
+                    .offset(),
+            ))
+        }
+    })
+}
+
+thread_local! {
+    static TZINFO: RefCell<Option<Cache>> = RefCell::new(None);
```
djc, line 84: Style nit: I think calling this `TZ_INFO` would be more idiomatic.

So we need the `Option` here because we can't initialize a real `Cache` in a `static` context? Might be worth a comment.

src/offset/local/unix.rs
```diff
@@ -37,35 +37,89 @@ pub(super) fn naive_to_local(d: &NaiveDateTime, local: bool) -> LocalResult<Date
...
-            LocalResult::Single(tt) => LocalResult::Single(FixedOffset::east(tt.offset())),
         }
-    } else {
-        LocalResult::Single(FixedOffset::east(
-            info.find_local_time_type(d.timestamp())
-                .expect("unable to select local time type")
-                .offset(),
-        ))
+    }
+    fn out_of_date(&self) -> bool {
```
djc, line 104: Style nit: empty line between methods, please.

This still feels like an expensive thing to do on every invocation to `offset()`. Could we amortize the cost somehow?

src/offset/local/unix.rs
```diff
@@ -37,35 +37,89 @@ pub(super) fn naive_to_local(d: &NaiveDateTime, local: bool) -> LocalResult<Date
...
     }
 }
 
-static mut INFO: Option<TimeZone> = None;
-static INIT: Once = Once::new();
+struct Cache {
+    zone: TimeZone,
+    source: Source,
+}
+impl Cache {
```
djc, line 118: Style nit: empty line here, too.

src/offset/local/unix.rs
```diff
@@ -37,35 +37,89 @@ pub(super) fn naive_to_local(d: &NaiveDateTime, local: bool) -> LocalResult<Date
...
-        // the entire time was skipped in which case we will return LocalResult::None anywa.
-        match info
-            .find_local_time_type_from_local(d.timestamp(), d.year())
-            .expect("unable to select local time type")
-        {
-            LocalResult::None => LocalResult::None,
-            LocalResult::Ambiguous(early, late) => LocalResult::Ambiguous(
-                FixedOffset::east(early.offset()),
-                FixedOffset::east(late.offset()),
+        if local {
```
djc, line 50: How about we invert the condition here and do an early return for `if !local`?

src/offset/local/unix.rs
```diff
@@ -37,35 +37,89 @@ pub(super) fn naive_to_local(d: &NaiveDateTime, local: bool) -> LocalResult<Date
...
+            // we pass through the year as the year of a local point in time must either be valid in that locale, or
+            // the entire time was skipped in which case we will return LocalResult::None anywa.
+            match cache
+                .borrow()
+                .as_ref()
+                .unwrap()
```
djc, line 56: This `unwrap()` should not be necessary, since we statically know above that it was initialized. Can we write the code in a way that the compiler understands this?

djc (commented):
Good start! A bunch of thoughts...

src/offset/local/unix.rs
esheppa: renamed

src/offset/local/unix.rs
esheppa: good call. Hopefully I've struck the right balance of what I moved and what I've left in `naive_to_local`

src/offset/local/unix.rs
esheppa: this is resolved by moving most of the logic to an `&mut self` method on `Cache`

src/offset/local/unix.rs
esheppa: Some thoughts here:
* for the env var variant, this should be very unlikely to change, given the issue surrounding `set_var`, so we could just never alter the cache if the `TZ` env var is set
* In the case it is not set, it could be confusing if there was a lag between symlinking a different timezone and then seeing the change happen - perhaps we could have some configurable behavior with features (eg `never-invalidate-cache`, but as a default always check the `mtime`? If we have a feature it would have to use the previous code with the `Once` rather than thread local storage, to ensure that users don't end up with a different timezone cached on each thread!
* This might interact unexpectedly with asynchronous code (by blocking longer than expected), perhaps adding a new API, say, `LocalCached` where the struct that implements `TimeZone` actually stores the parsed tzinfo file, then users can choose if/when to invalidate the cache

src/offset/local/unix.rs
```diff
@@ -8,64 +8,118 @@
...
-    if local {
-        match offset(*d, true) {
-            LocalResult::None => LocalResult::None,
-            LocalResult::Ambiguous(early, late) => LocalResult::Ambiguous(
-                DateTime::from_utc(*d - early, early),
-                DateTime::from_utc(*d - late, late),
+    TZ_INFO.with(|maybe_cache| {
+        // can later change `.replace(None)` with  `.take()` if MSRV goes >= 1.50
+        // can also remove the `ref` if/when MSRV increases to allow mixed owned/ref patterns
+        let mut cache = maybe_cache.replace(None).unwrap_or_default();
```
djc, line 26: Seems like a case for [`Option::get_or_insert()`](https://doc.rust-lang.org/std/option/enum.Option.html#method.get_or_insert)?

src/offset/local/unix.rs
```diff
@@ -8,64 +8,118 @@
...
+
+        if !local {
+            let offset = FixedOffset::east(
+                self.zone
+                    .find_local_time_type(d.timestamp())
+                    .expect("unable to select local time type")
+                    .offset(),
+            );
+
+            LocalResult::Single(DateTime::from_utc(d, offset))
```
djc, line 99: Can we just return here so we don't need the extra level of indentation for the rest of the function?

djc (commented):
Looking a lot better, some further suggestions/thoughts!

src/offset/local/unix.rs
esheppa: 1s seems like a reasonable choice in that, it would perhaps be imperceptible to a user manually changing the symlink and then looking at application output. I guess the caching time could be included in the documentation somewhere, such that it is clear to users.

djc: Okay, seems good.

djc (approved):
Approved with one minor issue addressed!

src/offset/local/unix.rs
```diff
@@ -8,64 +8,134 @@
...
-        match offset(*d, true) {
-            LocalResult::None => LocalResult::None,
-            LocalResult::Ambiguous(early, late) => LocalResult::Ambiguous(
-                DateTime::from_utc(*d - early, early),
-                DateTime::from_utc(*d - late, late),
-            ),
-            LocalResult::Single(offset) => {
-                LocalResult::Single(DateTime::from_utc(*d - offset, offset))
+    TZ_INFO.with(|maybe_cache| {
+        maybe_cache.borrow_mut().get_or_insert(Cache::default()).offset(*d, local)
```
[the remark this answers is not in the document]
esheppa: agreed, perfect case for `get_or_insert_default()`. `clippy` has a lot of `_with()` functions to keep up with!

djc:
@esheppa would you mind rebasing this? Probably also good to just squash the commits.

src/lib.rs
```diff
@@ -421,9 +421,6 @@
 #![warn(unreachable_pub)]
 #![deny(dead_code)]
 #![cfg_attr(not(any(feature = "std", test)), no_std)]
-// can remove this if/when rustc-serialize support is removed
```
djc, line 424: Don't think this was intended?

esheppa: yeah not sure what happened there! I've fixed and re-squashed

djc (approved):

jhpratt:
Is this PR not based on the assumption that the system's time zone will never change during the execution of a program? This is certainly not the case — the tzdb being updated by the system is the most obvious situation, but there are others that are far more subtle.

djc:
We invalidate the cache within one second if the mtime of the tzdb file changes. Would be interested to hear more about the subtle scenarios you've considered for your solution in time!

jhpratt:
Ah, invalidating the cache is good and should catch most cases.

The `TZ` environment variable could change. That is the core of the problem that led to the soundness issues, as I'm sure you are aware at this point.

When you say the "tzdb file", what are you referring to? The file containing the version? If so, it's possible for someone to write one of the other files directly. I don't blame you if you say "too bad" in that situation, but it's something to be mindful of.

djc:
The symlink metadata, see here: https://github.com/chronotope/chrono/pull/728/files#diff-568a8cdf0e0b7a4922b13d27771df35ac8e561ac9f25c5229263120edc171486R51.

We're explicitly leaving the `TZ` variable changing out of scope for now, because changing the environment seems so fraught with issues.

M1cha:
In a proprietary project we're currently locked to chrono 0.4.19 because the recent addition of caching the timezone broke our code. I have to admit that the way we propagate the change to chrono is quite hacky and that we're relying on implementation details of both glibc and chrono.

So first of: yes, the TZ variable points to a symlink which systemd changes after updating the timezone. This is even the case on desktop arch linux.

So basically we're using dbus to listen for timezone changes that come from systemd-timesyncd.
Once that happened we need to make sure that `localtime_r` returns the updated value. The issue is that glibc caches the value, too. The difference is that they update it when the value of `TZ` has changed. So we can simply change the value, call tzset, change the value back, call tzset again.

That hack basically works for C and worked for chrono too because it always called `localtime_r`. Since that's not the case anymore, the change doesn't propagate to chrono anymore. So basically we need a way to propagate a timezone change to chrono.

djc:
@M1cha please file a new issue, or better yet, propose some code changes (likely around https://github.com/chronotope/chrono/blob/main/src/offset/local/unix.rs#L41) that would fix your issue as a PR. I'm surprised that it doesn't work, we designed the caching approach to be responsive to approaches such as yours.

PR #729, opened 2022-07-07
Allow deprecated when rustc-serialize feature is enabled

esheppa:


djc:
BTW, technically this is not clippy, but just the rustc lints.

Issue #732, opened 2022-07-08
DateTime::parse_from_rfc2822 not compliant with RFC2822 comments


Finomnis:
# Problem
According to [RFC2822](https://datatracker.ietf.org/doc/html/rfc2822#section-3.3), an RFC2822 string can end with a `CFWS`.

However, the `DateTime::parse_from_rfc2822` cannot deal with comments:
```rust
use chrono::DateTime;

fn main() {
    let time = DateTime::parse_from_rfc2822("2 Nov 2021 14:26:12 +0000 (UTC)").unwrap();
    println!("{}", time);
}
```
```none
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ParseError(TooLong)', src/main.rs:4:80
```

# Implementation Details

The implementation [explicitly distances itself from comments](https://github.com/chronotope/chrono/blob/051e1170c41477ce162301c8711110a4577c1a23/src/format/parse.rs#L80) and calls itself *not an actual rfc2822 parser*:

```rust
    // - we do not recognize a folding white space (FWS) or comment (CFWS).
    //   for our purposes, instead, we accept any sequence of Unicode
    //   white space characters (denoted here to `S`). any actual RFC 2822
    //   parser is expected to parse FWS and/or CFWS themselves and replace
    //   it with a single SP (`%x20`); this is legitimate.
```

I'd argue that `chrono` is the defacto go-to library for anything date/time related in Rust, so this parser should be an *actual* parser.

# Rationale
The [documentation](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.parse_from_rfc2822) of `DateTime::parse_from_rfc2822` states:

> Parses an RFC 2822 date and time string such as Tue, 1 Jul 2003 10:52:37 +0200, then returns a new DateTime with a parsed FixedOffset.
> 
> RFC 2822 is the internet message standard that specifices the representation of times in HTTP and email headers.

This sounds like it should:
- accept the official RFC 2822 standard
- be compatible with email headers

Sadly, most email headers use a comment after the string that reflects the timezone, like `"2 Nov 2021 14:26:12 +0000 (UTC)"`. So I'd argue that the behaviour of the library is not consistent with the documentation; it should state that this is not an actual `RFC2822` parser, but instead only parses a subset and is explicitly not compatible with email headers.

# Solutions
I can see two possible solutions:
- adjust the documentation to reflect that this is not an actual RFC2822 parser
- adjust the code to make this an actual RFC2822 parser

[...]

PR #733, opened 2022-07-09
Add compatibility with rfc2822 comments

Finomnis:
Fixes https://github.com/chronotope/chrono/issues/732.

### DIFF
--- a/src/format/parse.rs
+++ b/src/format/parse.rs
@@ -53,7 +53,10 @@ fn parse_rfc2822<'a>(parsed: &mut Parsed, mut s: &'a str) -> ParseResult<(&'a st
 
     // an adapted RFC 2822 syntax from Section 3.3 and 4.3:
     //
-    // date-time   = [ day-of-week "," ] date 1*S time *S
+    // c-char      = <any char except '(', ')' and '\\'>
+    // c-escape    = "\" <any char>
+    // comment     = "(" *(comment / c-char / c-escape) ")" *S
+    // date-time   = [ day-of-week "," ] date 1*S time *S *comment
     // day-of-week = *S day-name *S
     // day-name    = "Mon" / "Tue" / "Wed" / "Thu" / "Fri" / "Sat" / "Sun"
     // date        = day month year
@@ -79,9 +82,10 @@ fn parse_rfc2822<'a>(parsed: &mut Parsed, mut s: &'a str) -> ParseResult<(&'a st
     //
     // - we do not recognize a folding white space (FWS) or comment (CFWS).
     //   for our purposes, instead, we accept any sequence of Unicode
-    //   white space characters (denoted here to `S`). any actual RFC 2822
-    //   parser is expected to parse FWS and/or CFWS themselves and replace
-    //   it with a single SP (`%x20`); this is legitimate.
+    //   white space characters (denoted here to `S`). For comments, we accept
+    //   any text within parentheses while respecting escaped parentheses.
+    //   Any actual RFC 2822 parser is expected to parse FWS and/or CFWS themselves
+    //   and replace it with a single SP (`%x20`); this is legitimate.
     //
     // - two-digit year < 50 should be interpreted by adding 2000.
     //   two-digit year >= 50 or three-digit year should be interpreted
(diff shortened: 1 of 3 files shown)


src/format/parse.rs
```diff
@@ -817,6 +829,12 @@ fn test_rfc2822() {
         ("Tue, 20 Jan 2015 17:35:20 -0800", Ok("Tue, 20 Jan 2015 17:35:20 -0800")), // normal case
         ("Fri,  2 Jan 2015 17:35:20 -0800", Ok("Fri, 02 Jan 2015 17:35:20 -0800")), // folding whitespace
         ("Fri, 02 Jan 2015 17:35:20 -0800", Ok("Fri, 02 Jan 2015 17:35:20 -0800")), // leading zero
+        ("Tue, 20 Jan 2015 17:35:20 -0800 (UTC)", Ok("Tue, 20 Jan 2015 17:35:20 -0800")), // trailing comment
+        (
+            "Tue, 20 Jan 2015 17:35:20 -0800 ( (UTC ) (\\( (a)\\(( \\t ) ) \\\\( \\) ))",
+            Ok("Tue, 20 Jan 2015 17:35:20 -0800"),
+        ), // complex trailing comment
+        ("Tue, 20 Jan 2015 17:35:20 -0800 (UTC\\)", Err(TOO_LONG)), // incorrect comment, not enough closing parentheses
```
esheppa, line 837: It seems like `TOO_SHORT` is more optimal in this case (admittedly obscure) - but it seems like actually getting `TOO_SHORT` at this stage might not be trivial, unless we early return from `scan::comment_2822`. Potentially the logic could be that if there is at least one `(` then it makes sense to early return the `TOO_SHORT` if it is never closed. Alternatively, the parser could be more permissive and just remove all the comment characters and return the parsed date if the rest is in the correct format - let me know what your thoughts are here?

esheppa (commented):
Looks good from my perspective - I've just left one comment for some discussion

src/format/parse.rs
Finomnis: I had the same thought, but for one, it's indeed not trivial to change that; and further, it's not completely clear. Technically `"Tue, 20 Jan 2015 17:35:20 -0800 ("` could be both too short or too long. `"Tue, 20 Jan 2015 17:35:20 -0800 x"`, for example, is definitely too long, so one could argue that the first one is also a valid string plus `" ("`. So I'd argue that "a comment" is everything that parses as a comment, and "too long" is everything attached to the timestamp that doesn't parse as a comment. Which includes `" ("`.

Finomnis: I don't think just accepting *everything* is the right choice, at least not with your current code structure. `parse.rs` just parses a single rfc2822 datetime string, and the way it is currently written means it could be used as a part of a more complex parser. So I wouldn't allow just everything, otherwise I think the entire `TOO_LONG` value would be pointless and we could simply accept all the strings that *start* with a valid rfc2822 datetime.

If you want me to change something, then what I would agree with the most is to short-circuit `TOO_SHORT`. But that also wouldn't be true the way the code is written right now, because the `parse_rfc2822` conceptually is supposed to *attempt* to parse a rfc2822 datetime string + return the leftover input that wasn't part of the rfc2822 datetime. And it does exactly that in the case of `"Tue, 20 Jan 2015 17:35:20 -0800 ("`, because `"Tue, 20 Jan 2015 17:35:20 -0800"` is a valid rfc2822 string and `" ("` is a valid rest. It just wasn't able to consume `" ("`, which is totally valid.

So if we *want* to change something, it must be somewhere in the method `parse_internal`. And this one currently is written so that it always returns `TOO_LONG` whenever any tokens are left over from parsing. So I'm not sure how `TOO_SHORT` would conceptually fit the current code structure.

esheppa: Thanks @Finomnis - really good point about the rfc2822 item being able to be reused within other formats, we definitely shouldn't break that so it looks like it is best to leave it as is.

Finomnis:
@esheppa Rebased to fix merge conflict in `CHANGELOG.md`

esheppa:
Thanks @Finomnis

PR #734, opened 2022-07-14
Fix Duration::abs() for negative durations with nanos

abreis:
The `abs()` method on `oldtime::Duration`, introduced in #418, was incorrect whenever a negative duration had non-zero `nanos`. This PR fixes it and adds a few simple tests.


djc (approved):

djc:
Thanks!

PR #735, opened 2022-07-24
Some experiments around formatting using constants

esheppa:
This can be done now using the `format::Item` `enum`, however it can be quite verbose.

These are some experiments using a macro_rules and some helper functions to make it a bit more convenient.


pitdicker:
It has been one and a half year... Closing for now.

PR #736, opened 2022-07-25
upgrade time package to version 0.3.11

sariug:

      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md

upgrading time package to get rid of the cargo audit problem.


djc:
See #602.

sariug:
my bad. when is the release?

djc:
See #674.

PR #737, opened 2022-07-25
Simplify RFC 2822 comment parser

djc:
Tweak the implementation from #733 to simplify the parser.

cc @Finomnis


esheppa:
Looks good - looks like this disallows training whitespace which is consistent with the behavior pre acd4ecf09fd0e5e35e2b5d5e074f6e1cc77172fc. Only thought I had was changing the test strings to raw strings as the multiple levels of escaping is a bit tricky - I added a PR to this branch to do that, however please feel free to ignore if you think it's not necessary

esheppa (approved):

src/format/parse.rs
```diff
@@ -150,11 +150,8 @@ fn parse_rfc2822<'a>(parsed: &mut Parsed, mut s: &'a str) -> ParseResult<(&'a st
     }
 
     // optional comments
-    s = s.trim_left();
     while let Ok((s_out, ())) = scan::comment_2822(s) {
-        // Trim left after every found comment, as comments are allowed to have whitespace
-        // between them
-        s = s_out.trim_left();
+        s = s_out;
```
Finomnis, line 154: Why was this changed? Multiple comments are allowed, and whitespace between/around as well.

src/format/parse.rs
```diff
@@ -831,10 +828,10 @@ fn test_rfc2822() {
         ("Fri, 02 Jan 2015 17:35:20 -0800", Ok("Fri, 02 Jan 2015 17:35:20 -0800")), // leading zero
         ("Tue, 20 Jan 2015 17:35:20 -0800 (UTC)", Ok("Tue, 20 Jan 2015 17:35:20 -0800")), // trailing comment
         (
-            "Tue, 20 Jan 2015 17:35:20 -0800 ( (UTC ) (\\( (a)\\(( \\t ) ) \\\\( \\) ))",
+            r"Tue, 20 Jan 2015 17:35:20 -0800 ( (UTC ) (\( (a)\(( \t ) ) \\( \) ))",
```
Finomnis, line 831: Based on my other comment, we should probably add a test here for multiple comments

Finomnis (changes requested):
The reworked parser is quite elegant. I don't understand why support for multiple comments was removed, though.

src/format/parse.rs
Finomnis: OK I saw now that you added the whitespace into the comment parser.

src/format/scan.rs
```diff
@@ -349,49 +349,33 @@ pub(super) fn timezone_name_skip(s: &str) -> ParseResult<(&str, ())> {
...
+/// Tries to consume an RFC2822 comment including preceding ` `.
+///
+/// Returns the remaining string after the closing parenthesis.
+pub(super) fn comment_2822(s: &str) -> ParseResult<(&str, ())> {
+    use CommentState::*;
+
+    let mut state = Start;
+    for (i, c) in s.bytes().enumerate() {
+        state = match (state, c) {
+            (Start, b' ') => Start,
```
Finomnis, line 361: According to the RFC, whitespace consists of more than just spaces, it can be space, tab, \r, \n, etc.

Finomnis (changes requested):

src/format/scan.rs
```diff
@@ -408,10 +392,10 @@ fn test_rfc2822_comments() {
...
         ("(x(x(x)x)x)", Ok("")),
         ("( x ( x ( x ) x ) x )", Ok("")),
-        ("(\\)", Err(TOO_SHORT)),
-        ("(\\()", Ok("")),
-        ("(\\))", Ok("")),
-        ("(\\\\)", Ok("")),
+        (r"(\)", Err(TOO_SHORT)),
+        (r"(\()", Ok("")),
+        (r"(\))", Ok("")),
+        (r"(\\)", Ok("")),
```
Finomnis, line 398: Missing additional tests for the newly introduced whitespace parsing.

Finomnis:
Created PR with fixes for my comments: https://github.com/chronotope/chrono/pull/739

Finomnis (approved):

esheppa (approved):

djc:
Thanks, great collaboration here!

Issue #740, opened 2022-07-26
wasm32/wasmbind: cannot find function tm_to_datetime in this scope


AmateurECE:
Compiling at 0903ab1a19426fd43a96ba7f2ed86b9fc08bf292 for target wasm32-unknown-unknown with feature `wasmbind` enabled results in a compilation error:

```
[edtwardy@Hackintosh trunk-test]$ cargo build --target wasm32-unknown-unknown
warning: URL fragment `#0903ab1a` in git URL is ignored for dependency (chrono). If you were trying to specify a specific git revision, use `rev = "0903ab1a"` in the dependency declaration.
    Updating git repository `https://github.com/chronotope/chrono#0903ab1a`
   Compiling chrono v0.4.20-beta.1 (https://github.com/chronotope/chrono#0903ab1a#0903ab1a)
error[E0425]: cannot find function `tm_to_datetime` in this scope
   --> /home/edtwardy/.cargo/git/checkouts/chrono-4fedc6b82b24d624/0903ab1/src/offset/local/stub.rs:17:5
    |
17  |     tm_to_datetime(Timespec::now().local())
    |     ^^^^^^^^^^^^^^ help: a function with a similar name exists: `tm_to_time`
...
199 | fn tm_to_time(tm: &Tm) -> i64 {
    | ----------------------------- similarly named function `tm_to_time` defined here

For more information about this error, try `rustc --explain E0425`.
```

[build-error.log](https://github.com/chronotope/chrono/files/9185552/build-error.log)

esheppa:
This should be solved by #746

PR #746, opened 2022-07-28
Update CI test for wasm32-unknown-unknown to enable wasmbind feature

esheppa:
Testing whether this triggers the CI

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,10 +4,10 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
     paths:
       - "**.rs"
       - .github/**
+      - .ci/**
       - Cargo.toml
 
 jobs:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -173,6 +173,30 @@ jobs:
           RUST_VERSION: stable
           WASM: wasm_unknown
 
+  wasm_wasi:
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          target: wasm32-wasi
+          override: true
+          default: true
+
+      - name: Build and Test
+        run: bash ci/github.sh
+        env:
+          RUST_VERSION: stable
+          WASM: wasm_wasi
+
   cross-targets:
     strategy:
       matrix:
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -47,7 +47,6 @@
 //!
 //! Optional features:
 //!
-//! - `wasmbind`: Enable integration with [wasm-bindgen][] and its `js-sys` project
 //! - [`serde`][]: Enable serialization/deserialization via serde.
 //! - `unstable-locales`: Enable localization. This adds various methods with a
 //!   `_localized` suffix. The implementation and API may change or even be
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ Versions with only mechanical changes will be omitted from the following list.
 * Add the `and_local_timezone` method to `NaiveDateTime`
 * Fix the behavior of `Duration::abs()` for negative durations with non-zero nanos
 * Add compatibility with rfc2822 comments (#733)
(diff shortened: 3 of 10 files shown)


src/lib.rs
```diff
@@ -474,6 +474,7 @@ macro_rules! try_opt {
     };
 }
 
+
```
djc, line 477: Spurious?

.github/workflows/test.yml
```diff
@@ -6,6 +6,7 @@ on:
   pull_request:
     branches: [main]
     paths:
+      - ".ci/**"
```
djc, line 9: Hah, good call.

src/lib.rs
esheppa: just introduced some whitespace to get the CI running.. then realised there was a better way :D

.github/workflows/test.yml
esheppa: I'm thinking of removing `branches: [main]` as well, that way PR's to PR's will also run CI which would be great

djc: Sounds good! In a separate commit?

esheppa: Yep - just squashing them

djc: I meant, would be good to have the `test.yml` changes in a separate commit from the `Cargo.toml` changes since they're conceptually independent.

esheppa: Good point, will push up two separate commits

djc:
Also I think we'll no longer need @AmateurECE's original commit as part of this PR, right?

esheppa:
I initially thought we still needed, but then I saw the inconsistency between the `#[cfg` flags in the Cargo.toml and in the code, so I just used `#[cfg(all(target_arch = "wasm32", not(any(target_os = "emscripten", target_os = "wasi"))))]` and its negation everywhere which avoids breaking emscripten while still removing use of the `wasmbind` feature. This should be safe as using `rustup target list` there are only three wasm targets, `wasm32-unknown-unknown`, `wasm32-unknown-emscripten` and `wasm32-wasi`, so there shouldn't be any other wasm targets getting unexpected `wasm-bindgen` or `js-sys`

morenol:
Hey, now that the  wasm-bindgen is in the dependency tree and it is not possible to remove that from wasm targets, it adds a bunch of wasm_bindgen imports that runtimes like wasmtime and wasmer don't provide. Is it ok if I create a PR adding a feature_flag that is off by default that we can use to disable wasm_bingen from being installed?

RIght now I am having difficulties in using chrono crate with wasmtime. In chrono = 0.4.19 I didnt have that issue

esheppa:
@morenol - this PR was to fix #740 where there is essentially a discoverability problem, however it is not ideal that we have broken your use case. If I'm understanding you correctly, you are proposing to have a feature which disables wasm-bindgen, but this might break the additive nature of features. Instead, it may be possible to make those two dependencies optional and add wasmbind to the default features (and include wasm-bindgen and js-sys in wasmbind feature)?

Also to note, because the features would be additive, if any of your dependencies used the wasmbind feature, then you would get those crates in your tree as well. Is this preventing you from using chrono with wasmtime, eg compile error or runtime error - or is this more of an efficiency concern?

morenol:
This is preventing me from using chrono with wasmtime. The solution that you propose would also work, the thing that I need is to disable wasm-bindgen from the dependency tree.

The issue that I am having is that when wasm-bindgen is in the dependency tree, it injects some imports that are [intended](https://github.com/rustwasm/wasm-bindgen/issues/3013#issuecomment-1201915572) for JavaScripts hosts.

Just as a reference, this is the specific line in wasmtime that is returing an error: 

https://github.com/bytecodealliance/wasmtime/blob/c3e31c99461f7f109637528d805dbaa349cf4a22/crates/wasmtime/src/instance.rs#L830

djc:
> Instead, it may be possible to make those two dependencies optional and add wasmbind to the default features (and include wasm-bindgen and js-sys in wasmbind feature)?

This sounds good to me.

esheppa:
@morenol  - are you using the target `wasm32-wasi` or `wasm32-unknown-unknown`?

morenol:
wasm32-unknown-unknown

esheppa:
I've set up https://github.com/chronotope/chrono/pull/771 allow this. Do you mind checking whether it works for you? I might need to make a few more changes. We can continue the conversation over there

PR #747, opened 2022-07-29
Clarify docs for `DateTime::with_timezone`

kevincox:
The previous docs don't really make sense, why am I calling a method if it doesn't change the value? The new wording clarifies what is held the same and what changes.

      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md


djc:
Thanks!

PR #748, opened 2022-07-29
Update time dependency

TWALL9:

- [X ] Have you added yourself and the change to the [changelog]? (Don't worry
      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md


djc:
See #602.

PR #750, opened 2022-07-31
add LocalOffset type that stores the timezone name

### LINKED ISSUE #749: Wrong output from `%Z` specifier when using `Local::now().format()`
See below for an example:
```
use chrono::Local;

fn main() {
    println!("{}",Local::now().format("%z"));
    println!("{}",Local::now().format("%Z"));
}
```
Output:
```
+0200
+02:00
```
Expected output:
```
+0200
CEST
```
Interestingly, everything works fine when you substitute `Local` for `Utc`

esheppa:
This could also be implemented as an extra field on `FixedOffset`, which would mean it would no longer be a breaking change, but in that case it would make the `FixedOffset` type more than three times larger.

This should fix #749 on unix platforms. We would need to capture timezone name on Windows as well, potentially from [here](https://docs.microsoft.com/en-us/windows/win32/api/timezoneapi/ns-timezoneapi-time_zone_information) hence why this is in draft status for now

### DIFF
--- a/src/datetime/tests.rs
+++ b/src/datetime/tests.rs
@@ -233,12 +233,13 @@ fn test_to_string_round_trip() {
     let _dt: DateTime<FixedOffset> = ndt_fixed.to_string().parse().unwrap();
 }
 
-#[test]
-#[cfg(feature = "clock")]
-fn test_to_string_round_trip_with_local() {
-    let ndt = Local::now();
-    let _dt: DateTime<FixedOffset> = ndt.to_string().parse().unwrap();
-}
+// TEMPORARY: disabled due to changing behaivour of Local (with LocalOffset)
+// #[test]
+// #[cfg(feature = "clock")]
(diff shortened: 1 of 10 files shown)


pamburus:
Hi, what stops this pull request from being fixed and merged?

pitdicker:
> We would need to capture timezone name on Windows as well, potentially from [here](https://docs.microsoft.com/en-us/windows/win32/api/timezoneapi/ns-timezoneapi-time_zone_information) hence why this is in draft status for now

The `TIME_ZONE_INFORMATION` structure doesn't seem to play along. The documentation says it returns names such as "EST" and "PDT", but in reality it returns full, localized names instead. We would need some other source.

pitdicker:
Closing this PR for now because realistically there have been enough changes that it is easier to start fresh. #1563 tracks adding a `LocalOffset` type.

PR #752, opened 2022-08-03
Revise Months API

djc:
* Provide `checked_add_months()` and `checked_sub_months()` for
  callers that would like to avoid panics
* Document panic potential in `Add` and `Sub` implementations
* Implement additional traits for `Months` per API guidelines
* Hide inner type for `Months` and add constructor
* Use lower-level APIs to clamp day

Since I didn't get another chance to review #731 before it was merged (and the suggestion to make panicking explicit doesn't seem to have been followed up), here are some proposed improvements.

cc @avantgardnerio

### DIFF
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -597,31 +597,99 @@ impl NaiveDate {
         parsed.to_naive_date()
     }
 
-    /// Private function to calculate necessary primitives for `Add<Month>`
+    /// Add a duration in [`Months`] to the date
     ///
-    /// # Arguments
+    /// If the day would be out of range for the resulting month, use the last day for that month.
     ///
-    /// * `delta` - Number of months (+/-) to add
+    /// Returns `None` if the resulting date would be out of range.
     ///
-    /// # Returns
+    /// ```
+    /// # use chrono::{NaiveDate, Months};
+    /// assert_eq!(
+    ///     NaiveDate::from_ymd(2022, 2, 20).checked_add_months(Months::new(6)),
+    ///     Some(NaiveDate::from_ymd(2022, 8, 20))
+    /// );
+    /// assert_eq!(
+    ///     NaiveDate::from_ymd(2022, 7, 31).checked_add_months(Months::new(2)),
+    ///     Some(NaiveDate::from_ymd(2022, 9, 30))
+    /// );
+    /// ```
+    pub fn checked_add_months(self, months: Months) -> Option<Self> {
+        if months.0 == 0 {
+            return Some(self);
+        }
+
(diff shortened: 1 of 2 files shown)


src/naive/date.rs
```diff
@@ -597,31 +597,99 @@ impl NaiveDate {
...
+    ///     NaiveDate::from_ymd(2022, 7, 31).checked_add_months(Months::new(2)),
+    ///     Some(NaiveDate::from_ymd(2022, 9, 30))
+    /// );
+    /// ```
+    pub fn checked_add_months(self, months: Months) -> Option<Self> {
+        if months.0 == 0 {
+            return Some(self);
+        }
+
+        match months.0 <= core::i32::MAX as u32 {
```
esheppa, line 622: Could we avoid this by having `try_new` and `new` constructors on the `Months` which only accept valid u32 values?

esheppa (commented):
This looks very clean and consistent with other `chrono` APIs with the `checked_add_*` functions. I've set up #753 with a potential change moving the adding months to a month logic to `overflowing_add/sub` functions in the `month` module, however feel free to ignore that if you think it doesn't make sense to have the `overflowing_add/sub` functions.

src/naive/date.rs
```diff
@@ -597,31 +597,99 @@ impl NaiveDate {
...
         };
-        let res_years = self.year() + delta_years;
-        let res_months = res_months % 12;
-        let res_months = if res_months < 0 { res_months + 12 } else { res_months };
-        NaiveDate::from_ymd(res_years, res_months as u32 + 1, 1)
+
+        // Clamp original day in case new month is shorter
+
+        let flags = YearFlags::from_year(year);
+        let feb_days = if flags.ndays() == 366 { 29 } else { 28 };
```
avantgardnerio, line 688: This is a much better implementation, thank you! Had I known about `ndays()` I'd have done things differently. Either way, very happy to have support for adding Months in chrono!

src/naive/date.rs
```diff
@@ -597,31 +597,99 @@ impl NaiveDate {
...
+    ///     Some(NaiveDate::from_ymd(2021, 8, 20))
+    /// );
+    /// ```
+    pub fn checked_sub_months(self, months: Months) -> Option<Self> {
+        if months.0 == 0 {
+            return Some(self);
+        }
+
+        // Copy `i32::MIN` here so we don't have to do a complicated cast
+        match months.0 <= 2_147_483_648 {
```
avantgardnerio, line 647: :+1:

avantgardnerio (approved):
Very nice, thank you!

esheppa (approved):
Looks good!

Issue #727, opened 2022-07-05
`.month()` should return `chrono::Month`


Chris00:
Thanks for your useful library.  I was wondering why doesn't [`.month()`](https://docs.rs/chrono/0.4.19/chrono/trait.Datelike.html#tymethod.month) return [`chrono::Month`](https://docs.rs/chrono/0.4.19/chrono/enum.Month.html) which can be cast to an integer if desired.  To give the same result as now, `Month` would need to be declared as
```rust
pub enum Month {
    January = 1,
    February = 2,
    March = 3,
 ...
```

Chris00:
> It would of course change the API in an incompatible way. 

Sure.  That would be for version 0.5.0. 😀

> we might chose to expose the number as a method instead of relying on the integer cast?

Yes but you would need one method per integer type (`.as_u8()`, `.as_u32()`,...) while the cast takes care of that for you.  The possibility to cast may be mentioned in the documentation to ease the transition.

djc:
I think we would only provide `as_u8()` and leave casting it up to the user, that seems fine.

PR #753, opened 2022-08-03
Month::overflowing_add/sub

esheppa:
Add an overflowing_add/sub similar to the equivalents for `NaiveTime`, which are useful for `NaiveDate` + `Months`
Also change `Months` constructor to only accept `u32 <= i32::MAX` and add a `try_new`

### DIFF
--- a/src/month.rs
+++ b/src/month.rs
@@ -81,6 +81,52 @@ impl Month {
         }
     }
 
+    /// Add `Months` to the current `Month` returning the new `Month`
+    /// and any years that have been crossed.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use chrono::{Month, Months};
+    ///
+    /// assert_eq!(Month::January.overflowing_add(Months::new(3)), (0, Month::April));
+    /// assert_eq!(Month::January.overflowing_add(Months::new(15)), (1, Month::April));
+    /// ```
+    pub fn overflowing_add(&self, months: Months) -> (i32, Month) {
+        let month = self.number_from_month() + months.0 % 12;
+
+        let (years, month) = if month > 12 {
+            (months.as_i32() / 12 + 1, month % 12)
+        } else {
+            (months.as_i32() / 12, month)
+        };
+
+        (years, num_traits::FromPrimitive::from_u32(month).unwrap())
+    }
+
+    /// Subtract `Months` from the current `Month` returning the new `Month`
+    /// and any years that have been crossed
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use chrono::{Month, Months};
+    ///
+    /// assert_eq!(Month::April.overflowing_sub(Months::new(3)), (0, Month::January));
+    /// assert_eq!(Month::April.overflowing_sub(Months::new(15)), (-1, Month::January));
+    /// ```
+    pub fn overflowing_sub(&self, months: Months) -> (i32, Month) {
+        let rem_mths = months.0 % 12;
+
+        let (years, month) = if self.number_from_month() <= rem_mths {
+            (-months.as_i32() / 12 - 1, (12 + self.number_from_month()) - rem_mths)
+        } else {
+            (-months.as_i32() / 12, self.number_from_month() - rem_mths)
+        };
+
+        (years, num_traits::FromPrimitive::from_u32(month).unwrap())
+    }
+
     /// The previous month.
     ///
     /// `m`:        | `January`  | `February` | `...` | `December`
@@ -196,7 +242,11 @@ pub struct Months(pub(crate) u32);
 impl Months {
     /// Construct a new `Months` from a number of months
     pub fn new(num: u32) -> Self {
-        Self(num)
+        Months(num)
+    }
+
+    fn as_i32(&self) -> i32 {
+        self.0 as i32
     }
 }
 
@@ -318,7 +368,7 @@ mod month_serde {
 
 #[cfg(test)]
 mod tests {
-    use super::Month;
+    use super::{Month, Months};
     use crate::{Datelike, TimeZone, Utc};
 
     #[test]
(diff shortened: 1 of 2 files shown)


esheppa:
If this is looking like a good API, I'll add some more documentation prior to this being merged

esheppa:
I still need to do the `MIN_YEAR` / `MAX_YEAR` checks in `diff_months` as well

src/naive/date.rs
```diff
@@ -643,53 +640,28 @@ impl NaiveDate {
...
+        let month: crate::Month = num_traits::FromPrimitive::from_u32(self.month()).unwrap();
 
-            (year + 1, month - 12)
-        } else {
-            (year, month)
+        let (years, month) = match forwards {
+            true => month.overflowing_add(months),
+            false => month.overflowing_sub(months),
         };
+        let year = self.year() + years;
```
djc, line 655: No more overflow/underflow checking for the years?

src/naive/date.rs
```diff
@@ -643,53 +640,28 @@ impl NaiveDate {
...
-
-            (year - 1, month + 12)
-        } else if month > 12 {
-            if year == MAX_YEAR {
-                return None;
-            }
+    fn diff_months(self, months: Months, forwards: bool) -> Option<Self> {
+        // safe as .month() returns [1..12]
+        // can be further improved when .month() returns a Month
+        let month: crate::Month = num_traits::FromPrimitive::from_u32(self.month()).unwrap();
```
djc, line 649: I don't see the point of using `num_traits::FromPrimitive` and using `unwrap()` if we can just use `as` here.

src/month.rs
```diff
@@ -195,8 +223,23 @@ pub struct Months(pub(crate) u32);
 
 impl Months {
     /// Construct a new `Months` from a number of months
+    pub fn new_opt(num: u32) -> Option<Self> {
+        if num <= core::i32::MAX as u32 {
```
djc, line 227: I'm not really a fan of this as the constraint ends up not being expressed in the type system, so we have to lean on convention/remembering/understanding. IMO it's better just to it as `u32` (which also makes the initialization API simple) and check the range in usage context.

djc (commented):
In principle I'm open to the modularization of moving some of `diff_months()`'s internals into public API on `Month`, but I'm not sure this API is really natural/otherwise useful.

src/naive/date.rs
```diff
@@ -2153,26 +2125,13 @@ mod tests {
     use std::{i32, u32};
 
     #[test]
-    fn diff_months() {
+    fn test_diff_months() {
```
djc, line 2128: What does the `test_` prefix buy us? This seems like an unrelated change.

src/naive/date.rs
```diff
@@ -643,53 +640,28 @@ impl NaiveDate {
...
-            (year, month)
+        let (years, month) = match forwards {
+            true => month.overflowing_add(months),
+            false => month.overflowing_sub(months),
         };
+        let year = self.year() + years;
+        let month = month.number_from_month();
 
         // Clamp original day in case new month is shorter
-
```
djc, line 686: I'd prefer to keep the empty line here

src/naive/date.rs
esheppa: I need the actual `Month` type here as that's what `overflowing_add/sub` take as the `&self` parameter. This will be able to be cleaned up later if/when we fix #727

src/naive/date.rs
esheppa: Accidentally removed too much from the function. I've added it back as:

```rust
        if year < MIN_YEAR || year > MAX_YEAR {
            return None;
        }
```

Or do you mean the `self.year() + years` overflowing/underflowing the `i32`?

src/month.rs
esheppa: My thinking with this is that if the `Months` is invalid, its better for this to show up when creating the months rather than adding it, but admittedly most `Months` will be made and added as part of the same expression. Perhaps another option is taking `i32` here but returning `None` on negative values? Or else removing `new_opt` and having the check be done inside `overflowing_add/sub`. This also forces months to have one less possible value on the sub side due to `i32::MAX != abs(i32::MIN)`, but potentially this is acceptable as it is erring on the side of allowing a slightly smaller maximum subtraction?

src/naive/date.rs
djc: Both...

src/month.rs
djc: But that's just the point: a `Months::new(u32::MAX)` is not invalid by itself -- it's conceptually perfectly valid. It's only invalid in the context of what `checked_add_months()`/`checked_sub_months()` are trying to do, as such IMO the burden of validation should fall on the context, not the type itself.

esheppa:
With regard to the API, I agree that it is debatable if it is generally useful. A couple of potential use cases are:
* Users wanting to implement `add_months` but with their own clamping behavior on invalid days
* Users wanting to know the month n months after another, irrespective of the year

Alternatively if you like this but don't think think the public API is useful, it could be made private?

src/naive/date.rs
esheppa: will add the check on `i32` as well

src/naive/date.rs
esheppa: I don't think it is possible to go from `u32` to an enumeration thought?

src/month.rs
esheppa: Makes sense, will move the checks there

djc:
I think this is only good if the public API is useful. Otherwise the indirection makes the code harder to understand and I'd prefer to keep it as I had it in my original PR. It sounds to me like we don't yet have a very convincing use case in mind for exposing this publicly, so in that case I think we should leave this for now and keep this in mind as a potential direction for future evolution.

djc:
(Sorry, closed it accidentally.)

PR #754, opened 2022-08-04
Bump version to 0.4.20

djc:
I think it's time to unleash this thing!

Added #731, #747, #744 and #746 to the release notes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "chrono"
-version = "0.4.20-rc.1"
+version = "0.4.20"
 description = "Date and time library for Rust"
 homepage = "https://github.com/chronotope/chrono"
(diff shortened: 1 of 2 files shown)


esheppa (approved):
Sounds good to me! All the remaining (recent) issues are API-incompatible so it is great to get this out first

tdkt:
Is there a typo in the changelog about minimal compatible rust version? It mentions both 1.32 and 1.52.

djc:
> The minimum supported Rust version for 0.4.20 is 1.32.0, which is intentionally still quite conservative. If you are using chrono 0.4 with a Rust version older than 1.52, we'd like to hear from you since we'd like to further modernize the code base to ease maintenance.

What part of this is unclear?

tdkt:
Sorry, my bad after re-reading it attentively it makes sense.

Issue #755, opened 2022-08-05
Panic with 0.4.20 on latest ArchLinux and Alpine


jcamiel:
With `chrono 0.4.20`, using `let now: DateTime<Local> = Local::now();` results in a panic.

```
thread 'main' panicked at 'unable to parse localtime info: Io(Os { code: 2, kind: NotFound, message: "No such file or directory" })', /root/.cargo/registry/src/[github.com](http://github.com/)-1ecc6299db9ec823/chrono-0.4.20/src/offset/local/[unix.rs:95](http://unix.rs:95/):37
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Repro step:

1. use latest ArchLinux docker image `$ docker run -it archlinux`
2. install rust `$ pacman -Sy rust`
3. create a sample
```
$ cargo init foo
$ cd cargo
$ cargo add chrono
$ cargo run
```
4. Edit the main and rerun
```rust
use chrono::{DateTime, Local};

fn main() {
    let _now: DateTime<Local> = Local::now();
}
```

djc:
Does Alpine store the tzdb somewhere else or does it not have it installed at all by default?

djc:
It looks like this is happening while we're trying to read `/etc/localtime`, maybe Alpine/Arch use something different for that?

djc:
https://wiki.archlinux.org/title/System_time#Time_zone seems to suggest that Arch at least does put that in the usual place.

https://stackoverflow.com/questions/68996420/how-to-set-timezone-inside-alpine-base-docker-image suggests Alpine also follows the usual process.

jcamiel:
Ah sorry, I've forgot to tell that I was in a Docker context.
Effectively, with the latest `alpine` image, if set the timezone "by hand"

```
$ apk add -U tzdata
$ cp /usr/share/zoneinfo/Europe/Paris /etc/localtime
$ cargo run
```
Everything is ok now ! Sorry for the trouble!

djc:
I'm curious, what did you previously get as the local offset in that container?

jcamiel:
I don't know how to test it, just running the latest `archlinux` or `alpine`:

# On Alpine:

```
$ docker run -it alpine
$ date
Fri Aug  5 11:22:04 UTC 2022
$ ls -l /etc/localtime
ls: /etc/localtime: No such file or directory
$ find /usr/share/zoneinfo/
find: /usr/share/zoneinfo/: No such file or directory
```

[...]

PR #756, opened 2022-08-05
default to UTC when /etc/localtime is missing

esheppa:
This is one potential solution to #755. 

I've temporarily commented out some tests in src/offset/local/tz_info/timezone.rs, awaiting some discussion on where our "default to UTC" behavior should be (in unix.rs or in timezone.rs):
* If it belongs in `unix.rs` then the assumption of the test is invalidated as they will all panic
* If we move the logic to `timezone.rs` then the question is which errors should we recover from and which should we propogate (eg should filesystem errors cause a default to UTC as well?)
* When we lookup in the current offset we also use an `expect` - should this default to UTC instead of panicing:
```rust
                self.zone
                    .find_local_time_type(d.timestamp())
                    .expect("unable to select local time type")
                    .offset(),
```


esheppa:
Ideally in `0.5` we could consider whether `now()` should be fallible which would then allow (and encourage by the type system) users to decide how to fallback

djc:
Nice, thanks for working on this!

My inclination is to keep this scoped to Unix for now, which is where we broke compatibility with 0.4.20. I think for now we should stick to only defaulting to UTC if we get a NotFound error, and keep panicking for other types of errors. I'd like to hear about other types of failures before we tackle them.

Maybe we can also improve on the error message people see in this case, to make it explicit that it's about `/etc/localtime`, thereby making it easier to fix a problem if it comes up?

djc:
Oh, and we could add a `try_now()`  already and expose it as fallible?

djc:
I guess putting the logic in timezones.rs might work? I'm not sure, did you have a chance at how/when libc falls back?

esheppa:
I tested this locally against date command which also seems to default to UTC in the case of no `/etc/localtime` symlink.

With regards to putting the logic in `timezone.rs` -  I've looked into this again and the caching logic also needs to handle fallback so it's nice to keep it in one file (albeit, there are probably further optimizations of the logic between `timezone.rs` and `unix.rs`.

However this may actually not be the right path - these cases where we are falling back to UTC due to a missing `/etc/localtime` make sense on docker, but not in android. We may need something like an `android.rs` and others, to handle platform specific logic/APIs. 

I'm also somewhat tempted by the suggestion of using `iana_time_zone` - however this would have to be in the form of a recommendation to use it with `chrono-tz` as it would need the compiled in time zone database. Potentially a long term solution is refactoring so that `chrono-tz` is merged into `chrono` and feature gated, then can be turned on for specific platforms local time implementations.

Alternatively to make this work in `chrono` for [android](https://android.googlesource.com/platform/ndk/+/android-4.2.2_r1.2/docs/system/libc/OVERVIEW.html):
* We have to first check `TZ` (which we do) and then check 'persist.sys.timezone' system property
* This only gives us the name of the selected timezone, which we then need to find in `/system/usr/share/zoneinfo`, then we can parse the correct one and continue as normal

djc:
I think we should go the other way and rely on the system timezone database, I was thinking this morning about exposing other timezones in some way through the type system (where available).

You aren't quite explicit about why this solution doesn't work on Android, is that because we need to check this extra property? How would that work?

djc:
Okay, after reviewing the code in iana-time-zone I can see that it would make sense to pull that in as a dependency.

esheppa:
> I think we should go the other way and rely on the system timezone database, I was thinking this morning about exposing other timezones in some way through the type system (where available). 

This is an excellent idea IMO! As its better to have the automatically updated system tzdb (plus, a "static" feature could be possible for deployments to an environment lacking tzdb). 

> You aren't quite explicit about why this solution doesn't work on Android, is that because we need to check this extra property? How would that work?

We have to use the android SDK or equivalent to find the name of the currently chosen timezone

> Okay, after reviewing the code in iana-time-zone I can see that it would make sense to pull that in as a dependency. But not sure why that should be used with chrono-tz?

It only returns the name of the timezone, like "Asia/Kuala_Lumpur" or "America/Los_Angeles". This works if the system has the tzdb somewhere (eg Android has it, but in a different location), but `chrono-tz` is needed on systems that don't have a tzdb on the filesystem (potentially iOS)

esheppa:
Ideally the fallback here would only be applied if all the following conditions are met:
* TZ environment variable is not set, or the file it points to is invalid
* Nothing at `/etc/localtime`
* iana-time-zone returns an error

djc:
Yes, that sounds good.

I think we should prioritize shipping 0.4.21 that fixes this for systems that have a tzdb on the file system, we can take some more time to deal with iOS, I suppose (what did your "potentially" refer to?).

esheppa:
One thing to note is that there are two main classes of failures we've seen currently:
* Alpine docker images -> no tzdb, no /etc/localtime -> solution: default to UTC (will use iana-time-zone and verify it returns an error before defaulting to UTC)
* Android -> has tzdb in different location, no /etc/localtime -> solution: use iana-time-zone to find zone name and then read the file from /system/usr/share/zoneinfo

Fixing these two gets us to a much better state than now, and then further work can be done for iOS etc (I'll raise separate issues for these)

esheppa:
> For iOS I guess we could pull in the core_foundation crate and call CFTimeZone::system().seconds_from_gmt().
This is ok for `Local::now` but we also need `from_utc_datetime` and `from_local_datetime`

djc:
I guess we can see about getting support for [`daylightSavingTimeOffset`](https://developer.apple.com/documentation/foundation/timezone/2293268-daylightsavingtimeoffset), but yeah, let's defer this until after 0.4.21.

Cargo.toml
```diff
@@ -40,6 +40,7 @@ serde = { version = "1.0.99", default-features = false, optional = true }
 pure-rust-locales = { version = "0.5.2", optional = true }
 criterion = { version = "0.3", optional = true }
 rkyv = {version = "0.7", optional = true}
+iana-time-zone = { version = "0.1.41", optional = true }
```
djc, line 43: By the way, this requires an MSRV bump IIRC.

src/offset/local/unix.rs
```diff
@@ -95,11 +96,27 @@ struct Cache {
     source: Source,
 }
 
+fn get_fallback_timezone() -> Option<TimeZone> {
+    let tz_name = try_opt!(iana_time_zone::get_timezone().ok());
+
+    #[cfg(target_os = "android")]
+    let tzdb_location = " /system/usr/share/zoneinfo";
```
djc, line 103: Let's make these `const`?

Cargo.toml
esheppa: hmm true. Potentially we could resolve this by vendoring parts but that is also not ideal

src/offset/local/unix.rs
```diff
@@ -95,11 +96,27 @@ struct Cache {
     source: Source,
 }
 
+fn get_fallback_timezone() -> Option<TimeZone> {
+    let tz_name = try_opt!(iana_time_zone::get_timezone().ok());
```
djc, line 100: I think we can actually just use `?` instead of `try_opt!()`?

Cargo.toml
djc: Yeah, let's just go with 1.38 for now which is still fairly conservative. Actually might be a good middle step towards 1.48, giving downstream a chance to complain?

esheppa: sounds good to me! This gets us `TryFrom` as well which is very valuable IMO

esheppa:
I've kept the MSRV change as a separate commit so it is clear when it happened

Cargo.toml
djc: Agreed!

Cargo.toml
```diff
@@ -40,12 +40,14 @@ serde = { version = "1.0.99", default-features = false, optional = true }
 pure-rust-locales = { version = "0.5.2", optional = true }
 criterion = { version = "0.3", optional = true }
 rkyv = {version = "0.7", optional = true}
-iana-time-zone = { version = "0.1.41", optional = true }
 
 [target.'cfg(all(target_arch = "wasm32", not(any(target_os = "emscripten", target_os = "wasi"))))'.dependencies]
 wasm-bindgen = { version = "0.2" }
 js-sys = { version = "0.3" } # contains FFI bindings for the JS Date API
 
+[target.'cfg(not(any(target_os = "emscripten", target_os = "wasi", target_os = "solaris")))'.dependencies]
```
djc, line 48: I think this belongs in the other commit?

src/offset/local/unix.rs
```diff
@@ -99,18 +99,19 @@ struct Cache {
 #[cfg(target_os = "android")]
 const TZDB_LOCATION: &str = " /system/usr/share/zoneinfo";
 
+#[allow(dead_code)] // keeps the cfg simpler
```
djc, line 102: The changes here also belong in the other commit, right?

src/offset/local/unix.rs
```diff
@@ -95,11 +96,32 @@ struct Cache {
...
 }
 
+#[cfg(target_os = "android")]
+const TZDB_LOCATION: &str = " /system/usr/share/zoneinfo";
+
+#[cfg(not(target_os = "android"))]
+const TZDB_LOCATION: &str = "/usr/share/zoneinfo";
+
+#[cfg(any(target_os = "emscripten", target_os = "wasi", target_os = "solaris"))]
+fn get_fallback_timezone() -> Option<TimeZone> {
```
djc, line 107: Let's strip the `get_()` prefix here in accordance with the API guidelines?

src/offset/local/tz_info/timezone.rs
```diff
@@ -817,14 +817,16 @@ mod tests {
...
             } else {
-                let time_zone_local = TimeZone::local()?;
-                let time_zone_local_1 = TimeZone::from_posix_tz("localtime")?;
-                let time_zone_local_2 = TimeZone::from_posix_tz("/etc/localtime")?;
-                let time_zone_local_3 = TimeZone::from_posix_tz(":/etc/localtime")?;
-
-                assert_eq!(time_zone_local, time_zone_local_1);
-                assert_eq!(time_zone_local, time_zone_local_2);
-                assert_eq!(time_zone_local, time_zone_local_3);
+                // To be discussed - does this make sense in the context
```
djc, line 820: What should we do with these tests now?

src/offset/local/unix.rs
```diff
@@ -91,8 +97,9 @@ struct Cache {
 
 impl Default for Cache {
     fn default() -> Cache {
+        // default to UTC if no local timezone can be found
         Cache {
-            zone: TimeZone::local().expect("unable to parse localtime info"),
+            zone: TimeZone::local().unwrap_or_else(|_| TimeZone::utc()),
```
djc: I think this can be written as `unwrap_or_else(TimeZone::utc)`?

djc (commented):
Looking good!

Cargo.toml
esheppa: perhaps the best solution is to squash the last two commits?

src/offset/local/tz_info/timezone.rs
esheppa: I think removing is the best option, they are just re-implementing the logic of `TimeZone::local()` anyway, and to make it work we'd also have to re-implement the UTC fallback here. The first case with TZ env var is useful as that verifies that the env var overrides other settings

djc: Sounds good.

Cargo.toml
djc: No, this is from the MSRV commit, which we'd like to keep separate. I'm fine with squashing the other two though.

src/offset/local/unix.rs
djc: Forgot to rename this one?

esheppa: yep, sorted now

src/offset/local/unix.rs
esheppa: this should be sorted but the changes don't show up here

djc (approved):
Great!

esheppa:
would be great to add android and iOS to the CI, will investigate in a seperate PR

PR #759, opened 2022-08-06
GNU coreutils `date`-like time zone formatting

jarkonik:
Pull request adds additional string formatting options for time zones using multiple colons as syntax, which are identical to formatting options in GNU coreutils `date` ( https://man7.org/linux/man-pages/man1/date.1.html )

This would fix https://github.com/uutils/coreutils/issues/3780 

The pull request adds additional format specifiers for `chrono::format::strftime`.

In addition to currently implemented `%z` and `%:z` PR adds 2 new time zone formatting identifiers:

1. `%::z` -Offset from the local time to UTC with seconds. 
Example output: `+09:30:00` 
2. `%:::z`  - Offset from the local time to UTC without minutes. 
Example output: `+09`

### DIFF
--- a/src/format/mod.rs
+++ b/src/format/mod.rs
@@ -224,13 +224,13 @@ pub enum Fixed {
     /// The offset is limited from `-24:00` to `+24:00`,
     /// which is the same as [`FixedOffset`](../offset/struct.FixedOffset.html)'s range.
     TimezoneOffsetColon,
-    /// Offset from the local time to UTC (`+09:00` or `-04:00` or `+00:00`).
+    /// Offset from the local time to UTC with seconds (`+09:00:00` or `-04:00:00` or `+00:00:00`).
     ///
     /// In the parser, the colon can be omitted and/or surrounded with any amount of whitespace.
     /// The offset is limited from `-24:00` to `+24:00`,
     /// which is the same as [`FixedOffset`](../offset/struct.FixedOffset.html)'s range.
     TimezoneOffsetDoubleColon,
-    /// Offset from the local time to UTC with seconds (`+09:00:00` or `-04:00:00` or `+00:00:00`).
+    /// Offset from the local time to UTC without minutes (`+09` or `-04` or `+00`).
     ///
     /// In the parser, the colon can be omitted and/or surrounded with any amount of whitespace.
     /// The offset is limited from `-24:00` to `+24:00`,
--- a/src/format/mod.rs
+++ b/src/format/mod.rs
@@ -236,7 +236,7 @@ pub enum Fixed {
     /// The offset is limited from `-24:00` to `+24:00`,
     /// which is the same as [`FixedOffset`](../offset/struct.FixedOffset.html)'s range.
     TimezoneOffsetTripleColon,
-    /// Offset from the local time to UTC without minutes (`+09` or `-04` or `+00`).
+    /// Offset from the local time to UTC (`+09:00` or `-04:00` or `Z`).
     ///
     /// In the parser, the colon can be omitted and/or surrounded with any amount of whitespace,
     /// and `Z` can be either in upper case or in lower case.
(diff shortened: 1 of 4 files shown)


djc:
Would you mind documenting in the PR description the intended behavior?

sylvestre:
@jarkonik ping ? :)

jarkonik:
I've expanded the PR description a bit. If more information is needed please let me know what should I change/add.

src/format/mod.rs
```diff
@@ -557,15 +578,32 @@ fn format_inner<'a>(
                 result: &mut String,
                 off: FixedOffset,
                 allow_zulu: bool,
-                use_colon: bool,
+                colon_type: Option<ColonType>,
```
djc, line 581: I think this should not take an `Option`, just make it `Colons` directly (requiring the callers to pass `Colons::None`).

src/format/mod.rs
```diff
@@ -274,6 +286,15 @@ enum InternalInternal {
     Nanosecond9NoDot,
 }
 
+#[cfg(any(feature = "alloc", feature = "std", test))]
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum ColonType {
```
djc, line 291: I think I'd like to name this `Colons`.

djc (commented):
Seems reasonable. Some minor suggestions below.

djc (approved):

djc:
Thanks, this looks good to me.

djc:
@esheppa any opinions?

esheppa:
Looks good, just one thing I'd noticed is that potentially it could loop forever if the input was like `%:::::::::: ...etc`. I've set up a PR to @jarkonik which proposes two alternative implementations (one is commented out) - https://github.com/jarkonik/chrono/pull/1

My concern with the looping forever is that the strftime items can be used to validate user input, so it could be a DOS attack

src/format/strftime.rs
```diff
@@ -71,6 +71,8 @@ The following specifiers are available both to formatting and parsing.
 | `%Z`  | `ACST`   | Local time zone name. Skips all non-whitespace characters during parsing. [^9] |
 | `%z`  | `+0930`  | Offset from the local time to UTC (with UTC being `+0000`).                |
 | `%:z` | `+09:30` | Same as `%z` but with a colon.                                             |
+| `%::z` | `+09:30:00` | Offset from the local time to UTC with seconds.
```
esheppa, line 74: Looks like this is missing the final `|`

esheppa: And ideally would be aligned with the other columns

src/format/mod.rs
```diff
@@ -224,6 +224,18 @@ pub enum Fixed {
     /// The offset is limited from `-24:00` to `+24:00`,
     /// which is the same as [`FixedOffset`](../offset/struct.FixedOffset.html)'s range.
     TimezoneOffsetColon,
+    /// Offset from the local time to UTC with seconds (`+09:00:00` or `-04:00:00` or `+00:00:00`).
+    ///
+    /// In the parser, the colon can be omitted and/or surrounded with any amount of whitespace.
+    /// The offset is limited from `-24:00` to `+24:00`,
```
esheppa, line 230: Potentially we should format this to match the above, eg, `-24:00:00`

src/format/mod.rs
```diff
@@ -224,6 +224,18 @@ pub enum Fixed {
...
+    /// which is the same as [`FixedOffset`](../offset/struct.FixedOffset.html)'s range.
+    TimezoneOffsetDoubleColon,
+    /// Offset from the local time to UTC without minutes (`+09` or `-04` or `+00`).
+    ///
+    /// In the parser, the colon can be omitted and/or surrounded with any amount of whitespace.
+    /// The offset is limited from `-24:00` to `+24:00`,
```
esheppa, line 236: Potentially we should format this to match the above, eg, `-24`

esheppa (dismissed):
Thanks @jarkonik, if possible, do you mind making a couple of fixes in the docs?

djc:
Thanks!

sylvestre:
Well done!
Sorry for the question but how often do you release ? :)

PR #760, opened 2022-08-06
fixes #758 -- correctly document that %Y parses negative years

### LINKED ISSUE #758: DateTime::parse_from_str accepts negative year with %Y
This contradicts the claims at: https://docs.rs/chrono/latest/chrono/format/strftime/index.html#fn1

[Minimal example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b9dd201a264bf779ef33d4972c5f9935):

```rs
use chrono;

fn main() {
    dbg!(chrono::DateTime::parse_from_str("-1\n110723459+1002", "%Y%m%d%H%M%S%z"));
}
```

Reproduced with 0.4.20

alex:


djc:
Thanks. I wonder if we should more thoroughly the edge cases here -- like how many characters this needs for NaiveDate::MIN.

alex:
I think you dropped a word there :-) What should we do more thoroughly?

djc:
Oops! "document"

alex:
That's probably a good idea. Here I'm just trying to do the minimum to address the issue I ran into :-)

djc (approved):
LGTM. @esheppa what do you think?

I think it's better hear to keep the range for `%Y` consistent with the range we allow internally (as much as possible), such that we should allow negative years here -- and change the documentation -- vs following the documentation and clamping the year to some number (`0000` or `0001`, I suppose) to avoid negative years.

The documentation originates in https://github.com/chronotope/chrono/commit/43ee68b5223ee4deb3e5e5cb8eb8b8cda741eaf4. The implementation started allowing negative years in https://github.com/chronotope/chrono/commit/0399ba1849766177f5ebb435989452f70c6844a3.

esheppa (approved):
Looks good to me, and there are some tests for this in [`src/format/parse.rs`](https://github.com/chronotope/chrono/blob/a185d3bdfac0fca5fc0de4f392d7f22bfee8d486/src/format/parse.rs#L598) so no need to add any more for this PR.

PR #761, opened 2022-08-08
Remove unnecessary try_opt!() macro

djc:


esheppa:
I'd just assumed this was an MSRV thing

djc:
It probably was, but from before the 1.32 bump?

PR #767, opened 2022-08-10
Do not include iana-time-zone for sgx

trevor-crypto:
In previous versions chrono with default features built successfully on the sgx platform. Since `iana-time-zone` is not implemented for this platform, do not include it when `target_env = "sgx"`.

      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -86,6 +86,28 @@ jobs:
         run: cargo build --target thumbv6m-none-eabi --color=always
         working-directory: ./ci/core-test
 
+  intel_sgx:
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install rust with fortanix sgx toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          target: x86_64-fortanix-unknown-sgx
+          override: true
+      - uses: Swatinem/rust-cache@v1
+
+      - name: Build sgx lib
+        run: cargo build --target x86_64-fortanix-unknown-sgx --color=always
+        working-directory: ./ci/core-test
+
   wasm:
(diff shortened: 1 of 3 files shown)


djc:
Thanks! Would you be able to contribute something that would let us test this in CI?

trevor-crypto:
> Thanks! Would you be able to contribute something that would let us test this in CI?

Yes, sure. Let me add that

esheppa:
Maybe we should change this to be inclusive instead of exclusive, and only include iana-time-zone for the [implemented](https://github.com/strawlab/iana-time-zone/blob/50a12c5257688cfc937d18c09b44622c69f5ee65/src/lib.rs#L21-L37) OS's

esheppa:
I've set up https://github.com/trevor-crypto/chrono/pull/1 for the latter option using the fallback feature. @trevor-crypto  and @djc  let me know what you think. This should prevent us breaking other platforms/OS's

djc:
It looks like your https://github.com/trevor-crypto/chrono/pull/1 targeted the main branch in the fork, whereas this PR comes from the iana-sgx branch. So your commit doesn't seem to show up?

esheppa:
yep I'd targeted the wrong branch at first, should be fixed now

trevor-crypto:
> It looks like your [trevor-crypto#1](https://github.com/trevor-crypto/chrono/pull/1) targeted the main branch in the fork

@djc we resolved that :smile:

esheppa:
Just set up https://github.com/trevor-crypto/chrono/pull/3 to avoid this having cyclic features on macos/ios. Happy to retarget that to chrono instead if that is more suitable

djc:
It seems unrelated to SGX, so I don't think it should become part of this PR.

I'm also inclined to wait on https://github.com/strawlab/iana-time-zone/pull/50 instead. (And if we do this, it should probably just iOS since the original method works okay on macOS IIRC.)

djc (approved):

PR #773, opened 2022-08-12
Prepare 0.4.22

djc:
Bump the iana-time-zone dependency to 0.1.44, which breaks the dependency cycle on iOS/macOS and fixes some issues on RedHat-based Linux distributions.

### DIFF
--- a/src/month.rs
+++ b/src/month.rs
@@ -201,7 +201,7 @@ impl Months {
 }
 
 /// An error resulting from reading `<Month>` value with `FromStr`.
-#[derive(Clone, PartialEq)]
+#[derive(Clone, PartialEq, Eq)]
 pub struct ParseMonthError {
     pub(crate) _dummy: (),
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,19 @@ breakage if you use `no-default-features`.
 There were/are numerous minor versions before 1.0 due to the language changes.
 Versions with only mechanical changes will be omitted from the following list.
 
-## 0.4.20 (unreleased)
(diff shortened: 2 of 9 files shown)


src/month.rs
```diff
@@ -201,7 +201,7 @@ impl Months {
 }
 
 /// An error resulting from reading `<Month>` value with `FromStr`.
-#[derive(Clone, PartialEq)]
+#[derive(Clone, PartialEq, Eq)]
```
Kijewski, line 204: +`Copy` while you're at it? Or don't you want to make this promise in case of possible future changes?

djc: I'm just fixing up clippy stuff here, don't want to muddle it with unrelated changes.

esheppa:
Just set up a PR (#775) to update the changelog as it had gotten a little behind, aside from that looks good to me

djc:
I kinda want to get rid of change log in favor of release notes.

esheppa:
Sounds good to me. Only thing I'd like to consider in relation to that is that I somewhat like the idea that the changelog is in the repository rather than as metadata - perhaps there is a way we can automatically generate the release notes - and we could have a folder with a .md file for each release?

esheppa:
Just had a try of this to see if the content of a tag could be used but it doesn't seem to render as markdown, and I guess would have to be manually copied to the release creation form. We could still write them manually as markdown then copy that to the release creation form but that doesn't really meet the requirement of being less work than now. 

Should we remove the changelog for this release?

djc:
I think we probably want to keep it around for historical purposes? Anyway, this release is done, let me just get it out.

Issue #770, opened 2022-08-10
v0.4.21 patch upgrade breaks SGX builds


nhynes:
`chrono v0.4.21` introduces a new default feature `clock` which transitively pulls in `iana-time-zone`, which [doesn't build on SGX](https://github.com/strawlab/iana-time-zone/blob/392c9dee3d844d897bbc5c276855aa5bceaf8fb6/src/platform.rs#L7). This PATCH version upgrade has caused a sever-incompatible change on the SGX target (and others for which iana-time-zone doesn't work).

I would suggest:
1. yanking v0.4.21
2. publishing v0.5.0 with the new feature+transitive dep

Until then, essentially all users of the SGX target via core crates such as [mbedtls](https://github.com/fortanix/rust-mbedtls/blob/master/mbedtls/Cargo.toml#L36) will need to carefully manage lockfiles, which isn't super fun.

https://github.com/chronotope/chrono/pull/767 thankfully fixes the problem. Hopefully the new patch version gets through okay.

esheppa:
Thanks for the report @nhynes and @rdeaton - and apologies for the breakage. It looks like we have a few options here:
* To avoid the cyclic dependency we could temporarily turn off `iana-time-zone` for macOS/iOS (at least until https://github.com/strawlab/iana-time-zone/pull/50 is merged)
* Yanking v0.4.21 sounds like a good idea to me, but potentially we could release the fixes as v0.4.22  - essentially the reason for yanking would be the semver incompatibility
* In the specific case of `mbedtls` it doesn't look like this is using `Local`, and in fact, it looks like the usage of `chrono` could be replaced by `SystemTime`

djc:
Sorry for the regression!

It seems like 0.4.20 would also be broken on SGX, and I don't think we're going to yank both -- and I'm guessing the audience for whom 0.4.21 improves things is much larger than the audience for whom it makes things worse. In this case I think we should just get a 0.4.22 out as soon as possible.

PR #774, opened 2022-08-12
Update iana-time-zone to 0.1.44

Kijewski:
The circular dependency problems mentioned in <https://github.com/chronotope/chrono/issues/770#issuecomment-1211570887> where fixed in `iana-time-zone` 0.1.44. `iana-time-zone` now depends on `core-foundation-sys` instead of `core-foundation`.

- [ ] If this pull request fixes a bug, does it add a test that verifies that we can't reintroduce it?

[changelog]: ../CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ breakage if you use `no-default-features`.
 There were/are numerous minor versions before 1.0 due to the language changes.
 Versions with only mechanical changes will be omitted from the following list.
 
+## 0.4.22 (unreleased)
+
+* Fix circular dependency problems between iana-time-zone, core-foundation, and chrono
+
 ## 0.4.20 (unreleased)
 
 * Add more formatting documentation and examples.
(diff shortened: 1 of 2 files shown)


djc:
See #773, though I guess maybe I got the version wrong?

Kijewski:
I had this as open PR against [trevor-crypto:iana-sgx](https://github.com/trevor-crypto/chrono/tree/iana-sgx), but it was not picked in time. :-/ I guess I did not press F5 before updating my commit, and didn't see https://github.com/chronotope/chrono/pull/773 before opening this PR.

#773 uses the correct version.

djc:
Well, like I [said](https://github.com/chronotope/chrono/pull/767#issuecomment-1211938107) in that PR, it doesn't belong there anyway.

I fixed #773 to use the right version. Thanks for helping out with all this!

Kijewski:
You're welcome! Fun fact: Even before this problem, I added SGX as the prime example for a target, that is not implemented in iana-time-zone. :) https://github.com/strawlab/iana-time-zone/commit/5edaf34ffa31495b7275be4fb24d54935d49c795

PR #775, opened 2022-08-13
update changelog

esheppa:


esheppa:
would be interesting to see if we can harmonize the work for the changelog and release notes

PR #777, opened 2022-08-13
Clarify serde module contains defaults too in docs

### LINKED ISSUE #776: Deserialization / Serialization with Serde docs
It seems to do serializations with `serde` chrono requires you use the `serde` feature, also for `DateTime`, `NaiveDateTime`, and `NaiveDate`, is that correct?

```rs
/// documented at re-export site
#[cfg(feature = "serde")]
pub(super) mod serde;
```

Yet the docs still specify here and there that the `serde` feature is for "alternative" serialization options. Would maintainers be open to a PR trying to clarify this in the docs. I think the main line that could use updating is:

```
Serialize into a rfc3339 time string

See [the serde module](https://docs.rs/chrono/latest/chrono/serde/index.html) for alternate serialization.
```

Thanks for a great library! 😄

alextes:
Makes it more clear the serde module isn't *only* for alternatives it
is also required for the default implementations.

Adds a small hint in italic that the serde flag needs to be enabled to consume mentioned defaults.

Did not feel the change was notable so left it out of the changelog.

Closes #776 

      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -506,14 +506,13 @@ pub use traits::{Datelike, Timelike};
 #[doc(hidden)]
 pub use naive::__BenchYearFlags;
 
-/// Serialization/Deserialization in alternate formats
+/// Serialization/Deserialization with serde.
 ///
-/// The various modules in here are intended to be used with serde's [`with`
-/// annotation][1] to serialize as something other than the default [RFC
-/// 3339][2] format.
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Thanks!

PR #778, opened 2022-08-15
Impl Add<Month> + Sub<Month> for NaiveDateTime

avantgardnerio:

      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md

### DIFF
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -629,18 +629,15 @@ impl NaiveDate {
     ///
     /// If the day would be out of range for the resulting month, use the last day for that month.
     ///
+    /// Returns `None` if the resulting date would be out of range.
+    ///
     /// ```
     /// # use chrono::{NaiveDate, Months};
     /// assert_eq!(
     ///     NaiveDate::from_ymd(2022, 2, 20).checked_sub_months(Months::new(6)),
     ///     Some(NaiveDate::from_ymd(2021, 8, 20))
     /// );
-    /// ```
     ///
-    /// Returns `None` if the resulting date would be out of range.
-    ///
-    /// ```
-    /// use chrono::{Months, NaiveDate};
     /// let m = Months::new(core::i32::MAX as u32 + 1);
     /// assert_eq!(NaiveDate::from_ymd(2014, 1, 1).checked_sub_months(m), None);
     /// ```
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -642,7 +642,7 @@ impl NaiveDate {
     /// ```
     /// use chrono::{Months, NaiveDate};
     /// let m = Months::new(core::i32::MAX as u32 + 1);
-    /// assert_eq!(NaiveDate::from_ymd(2014, 1, 0).checked_sub_months(m), None);
+    /// assert_eq!(NaiveDate::from_ymd(2014, 1, 1).checked_sub_months(m), None);
     /// ```
     pub fn checked_sub_months(self, months: Months) -> Option<Self> {
         if months.0 == 0 {
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -649,8 +649,8 @@ impl NaiveDate {
             return Some(self);
         }
 
-        // Copy `i32::MIN` here so we don't have to do a complicated cast
-        match months.0 <= 2_147_483_648 {
+        // Copy `i32::MAX` here so we don't have to do a complicated cast
+        match months.0 <= 2_147_483_647 {
             true => self.diff_months(-(months.0 as i32)),
             false => None,
         }
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -629,14 +629,20 @@ impl NaiveDate {
     ///
     /// If the day would be out of range for the resulting month, use the last day for that month.
     ///
-    /// Returns `None` if the resulting date would be out of range.
-    ///
     /// ```
     /// # use chrono::{NaiveDate, Months};
     /// assert_eq!(
     ///     NaiveDate::from_ymd(2022, 2, 20).checked_sub_months(Months::new(6)),
     ///     Some(NaiveDate::from_ymd(2021, 8, 20))
     /// );
+    /// ```
+    ///
+    /// Returns `None` if the resulting date would be out of range.
+    ///
+    /// ```
+    /// use chrono::{Months, NaiveDate};
+    /// let m = Months::new(core::i32::MAX as u32 + 1);
+    /// assert_eq!(NaiveDate::from_ymd(2014, 1, 0).checked_sub_months(m), None);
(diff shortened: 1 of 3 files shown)


djc:
This seems reasonable. I think we should probably also add `checked_add_months()` and `checked_sub_months()`, and I'd like to express the new `Add` and `Sub` implementation in terms of `NaiveDate::checked_add_months()` and `checked_sub_months()` as well, to reduce the amount of layering and make the panicking behavior more explicit.

avantgardnerio:
@djc I added these functions, and I think they meet the requirements listed above. I inadvertently found an edge-case that was panicking, and added a test and fixed that too (see build logs for details).

src/naive/date.rs
```diff
@@ -629,22 +629,28 @@ impl NaiveDate {
...
-    ///
     /// ```
     /// # use chrono::{NaiveDate, Months};
     /// assert_eq!(
     ///     NaiveDate::from_ymd(2022, 2, 20).checked_sub_months(Months::new(6)),
     ///     Some(NaiveDate::from_ymd(2021, 8, 20))
     /// );
     /// ```
+    ///
+    /// Returns `None` if the resulting date would be out of range.
```
djc, line 640: I'd prefer to keep this explanation before the examples (and I think that's common style for rustdoc).

src/naive/datetime/mod.rs
```diff
@@ -533,6 +533,34 @@ impl NaiveDateTime {
...
+    ///
+    /// ```
+    /// use std::str::FromStr;
+    /// use chrono::{Months, NaiveDateTime};
+    /// let dt = |s| NaiveDateTime::from_str(s).unwrap();
+    /// let m = Months::new(core::i32::MAX as u32 + 1);
+    /// assert_eq!(dt("2014-01-01T01:00:00").checked_add_months(m), None);
+    /// ```
+    pub fn checked_add_months(self, rhs: Months) -> Option<NaiveDateTime> {
+        self.date().checked_add_months(rhs).map(|dt| dt.and_time(self.time()))
```
djc, line 561: I think this can be simplified to

```rust
self.date = self.date.checked_add_months(rhs)?;
```

I definitely think we should not be using the `date()` and `time()` accessors here.

src/naive/datetime/mod.rs
```diff
@@ -1289,6 +1345,37 @@ impl AddAssign<OldDuration> for NaiveDateTime {
...
+    /// assert_eq!(dt("2014-01-01T01:00:00").unwrap() + m(1), dt("2014-02-01T01:00:00").unwrap());
+    /// assert_eq!(dt("2014-01-01T00:02:00").unwrap() + m(11), dt("2014-12-01T00:02:00").unwrap());
+    /// assert_eq!(dt("2014-01-01T00:00:03").unwrap() + m(12), dt("2015-01-01T00:00:03").unwrap());
+    /// assert_eq!(dt("2014-01-01T00:00:04").unwrap() + m(13), dt("2015-02-01T00:00:04").unwrap());
+    /// assert_eq!(dt("2014-01-31T00:05:00").unwrap() + m(1), dt("2014-02-28T00:05:00").unwrap());
+    /// assert_eq!(dt("2020-01-31T06:00:00").unwrap() + m(1), dt("2020-02-29T06:00:00").unwrap());
+    /// ```
+    fn add(self, rhs: Months) -> Self::Output {
+        let new_date = self.date() + rhs;
+        new_date.and_time(self.time())
```
djc, line 1375:
```rust
Self {
    date: self.date.checked_add_months(rhs).unwrap(),
    time: self.time,
}
```

src/naive/date.rs
```diff
@@ -629,22 +629,28 @@ impl NaiveDate {
...
+    /// ```
     pub fn checked_sub_months(self, months: Months) -> Option<Self> {
         if months.0 == 0 {
             return Some(self);
         }
 
-        // Copy `i32::MIN` here so we don't have to do a complicated cast
-        match months.0 <= 2_147_483_648 {
+        // Copy `i32::MAX` here so we don't have to do a complicated cast
+        match months.0 <= 2_147_483_647 {
```
djc, line 653: Ah, missed that we're only making this negative after the cast to `i32`.

src/naive/datetime/mod.rs
djc: This doesn't seem to be resolved?

avantgardnerio: Sorry, I fixed it in the other two places and forgot to fix it here. It is resolved now.

djc (approved):

src/naive/datetime/mod.rs
```diff
@@ -606,6 +634,34 @@ impl NaiveDateTime {
         Some(NaiveDateTime { date, time })
     }
 
+    /// Subtracts given `Months` to the current date and time.
```
djc, line 637: Sorry, minor nit I forgot to mention: "Subtracts .. to" should use have "from" instead of "to".

djc (approved):

src/naive/datetime/mod.rs
```diff
@@ -533,6 +533,34 @@ impl NaiveDateTime {
...
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use std::str::FromStr;
+    /// use chrono::{Months, NaiveDateTime};
+    /// let dt = |s| NaiveDateTime::from_str(s).unwrap();
+    /// let ca = |dt: NaiveDateTime, m: Months| dt.checked_add_months(m).unwrap();
+    /// let m = Months::new;
+    /// assert_eq!(ca(dt("2014-01-01T01:00:00"), m(1)), dt("2014-02-01T01:00:00"));
```
esheppa, line 548: I'm a little uncomfortable with this style of code (albeit similar to what is used in the private tests) being in the public docs. IMO this would look better if we just used the full functions directly, with more intermediate variables if needed for readability. 

I've set up https://github.com/avantgardnerio/chrono/pull/1 as one possibility here, but I'm not set on that option, and keen to have a discussion on the merits of each approach.

esheppa (changes requested):
Thanks @avantgardnerio - good idea to add this to `NaiveDateTime` (I'm keen to set up a PR to add it to `DateTime<Tz>` shortly after this :smile:) - I've set up a PR to your branch with a proposal for a change to the documentation, but apart from that this looks great!

src/naive/datetime/mod.rs
djc: I agree that this makes the examples harder to understand (and somewhat unidiomatic).

avantgardnerio: I think they are all fixed now. Please ping me if I missed any!

avantgardnerio:
Thanks all! :tada:

PR #779, opened 2022-08-16
 Add feature documentation for docs.rs 

Kijewski:
This way the documentation will tell user which feature(s) are needed to use a function, trait, etc. docs.rs displays the information like

> Available on crate feature `serde` only.

Use this command to preview the changes:

```sh
RUSTDOCFLAGS="-D warnings --cfg docsrs" cargo +nightly doc --features=serde
```

- [x] If this pull request fixes a bug, does it add a test that verifies that we can't reintroduce it?

[changelog]: ../CHANGELOG.md

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -415,6 +415,7 @@ impl Default for DateTime<Utc> {
 }
 
 #[cfg(feature = "clock")]
+#[cfg_attr(docsrs, doc(cfg(feature = "clock")))]
 impl Default for DateTime<Local> {
     fn default() -> Self {
         Local.from_utc_datetime(&NaiveDateTime::default())
@@ -440,6 +441,7 @@ impl From<DateTime<Utc>> for DateTime<FixedOffset> {
 
 /// Convert a `DateTime<Utc>` instance into a `DateTime<Local>` instance.
 #[cfg(feature = "clock")]
+#[cfg_attr(docsrs, doc(cfg(feature = "clock")))]
 impl From<DateTime<Utc>> for DateTime<Local> {
     /// Convert this `DateTime<Utc>` instance into a `DateTime<Local>` instance.
     ///
@@ -462,6 +464,7 @@ impl From<DateTime<FixedOffset>> for DateTime<Utc> {
 
 /// Convert a `DateTime<FixedOffset>` instance into a `DateTime<Local>` instance.
 #[cfg(feature = "clock")]
+#[cfg_attr(docsrs, doc(cfg(feature = "clock")))]
 impl From<DateTime<FixedOffset>> for DateTime<Local> {
     /// Convert this `DateTime<FixedOffset>` instance into a `DateTime<Local>` instance.
     ///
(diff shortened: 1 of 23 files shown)


djc (approved):
Great, thanks!

src/lib.rs
```diff
@@ -485,6 +491,7 @@ pub use naive::{IsoWeek, NaiveDate, NaiveDateTime, NaiveTime, NaiveWeek};
 
 pub mod offset;
 #[cfg(feature = "clock")]
+#[cfg_attr(docsrs, doc(cfg(feature = "clock")))]
```
esheppa, line 494: It looks like this doesn't work with Re-exports

esheppa: or perhaps `no_inline` is causing the issue, as the requirement for `clock` feature doesn't show up in the prelude either

esheppa: I've tried toggling no_inline but doesn't seem to help, so I'm assuming it's a re-export thing

esheppa (changes requested):
Thanks @Kijewski this is a great addition! I've just noted a small issue with the labels on re-exports, also perhaps we can instead label the `local` module as needing the clock feature?

src/datetime/mod.rs
```diff
@@ -36,10 +36,12 @@ use crate::{Datelike, Timelike, Weekday};
 use rkyv::{Archive, Deserialize, Serialize};
 
 #[cfg(feature = "rustc-serialize")]
+#[cfg_attr(docsrs, doc(cfg(feature = "rustc-serialize")))]
```
esheppa, line 39: can be removed as this is documented at the top level instead

src/datetime/mod.rs
```diff
@@ -36,10 +36,12 @@ use crate::{Datelike, Timelike, Weekday};
...
 pub(super) mod rustc_serialize;
 
 /// documented at re-export site
 #[cfg(feature = "serde")]
+#[cfg_attr(docsrs, doc(cfg(feature = "serde")))]
```
esheppa, line 44: can be removed as this is documented at the top level instead

src/lib.rs
esheppa: I'm going to resolve this as it seems it's a rustdoc bug

esheppa (approved):
So I looked through the tracking issue and found: https://github.com/rust-lang/rust/issues/83428 and https://github.com/rust-lang/rust/issues/85043 

Specifically https://github.com/rust-lang/rust/issues/83428#issuecomment-966687079 I think shows the exact issue we are having here - so I think this overrides all my requests and we can just merge as is (however if you are happy to do so it could be worth removing the two `cfg_attr`s on the `pub(super)` items

djc:
@Kijewski if you can just rebase this once more, I'm happy to merge it.

src/datetime/mod.rs
Kijewski: Done.

src/datetime/mod.rs
Kijewski: Done.

Kijewski:
Rebased.

djc:
Thanks!

PR #780, opened 2022-08-17
Bump optional dependency iana-time-zone for RUSTSEC-2022-0049

tnballo:
Per [RUSTSEC-2022-0049](https://rustsec.org/advisories/RUSTSEC-2022-0049.html): `src/offset/local/unix.rs:107` calls vulnerable function `iana_time_zone::get_timezone`. Affects MacOS/iOS targets. This PR bumps to non-vulnerable version of the optional dependency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ breakage if you use `no-default-features`.
 There were/are numerous minor versions before 1.0 due to the language changes.
 Versions with only mechanical changes will be omitted from the following list.
 
+## 0.4.23
+
+* Update `iana-time-zone` version to 1.45 for [RUSTSEC-2022-0049](https://rustsec.org/advisories/RUSTSEC-2022-0049.html)
+
 ## 0.4.22
 
 * Allow wasmbindgen to be optional on `wasm32-unknown-unknown` target [(#771)](https://github.com/chronotope/chrono/pull/771)
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ serde = { version = "1.0.99", default-features = false, optional = true }
 pure-rust-locales = { version = "0.5.2", optional = true }
 criterion = { version = "0.3", optional = true }
 rkyv = {version = "0.7", optional = true}
-iana-time-zone = { version = "0.1.44", optional = true, features = ["fallback"] }
+iana-time-zone = { version = "0.1.45", optional = true, features = ["fallback"] }
 
 [target.'cfg(all(target_arch = "wasm32", not(any(target_os = "emscripten", target_os = "wasi"))))'.dependencies]
 wasm-bindgen = { version = "0.2", optional = true }


esheppa:
Thanks for this @tnballo. Looks like we probably need to yank `0.4.22` as well - let me know your thoughts @djc

esheppa (approved):

djc:
Why would we need to yank? We only have a lower boundary so 0.4.22 can be used with 0.1.45 just fine. I'm fine with merging this but it doesn't seem like a big deal on our side.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ breakage if you use `no-default-features`.
 There were/are numerous minor versions before 1.0 due to the language changes.
 Versions with only mechanical changes will be omitted from the following list.
 
+## 0.4.23
+
+* Update `iana-time-zone` version to 1.45 for [RUSTSEC-2022-0049](https://rustsec.org/advisories/RUSTSEC-2022-0049.html)
```
djc, line 16: The version number is wrong here.

Kijewski:
The two vulnerable versions of iana-time-zone were yanked, so the next time you invoke `cargo update`, a safe version is chosen anyway, even with `-Z minimal-versions`. And without updating, the user wouldn't get a new chrono version either, so I don't think that chrono needs to release a new version right now.

esheppa:
> We only have a lower boundary so 0.4.22
good point, didn't think through the interaction here, looks like the yanking of the earlier `iana-time`zone` versions solves it

tnballo:
Ah, that's true - this PR isn't necessary due to the yank. Closing. Thanks for double-checking!

giraffate:
IMHO, It would be helpful to update iana-time-zone to 0.1.45 because it doesn't update iana-time-zone version when doing `cargo update -p chrono`. I didn't know that it's a conservative update without `--agressive`.

djc:
This argument doesn't make much sense to me. The problematic dependency is not Chrono, it's not surprising that only updating Chrono doesn't fix it, and it's fixed quite simply by updating iana-time-zone directly (which is actually the cause of the issue).

Kijewski:
Also, this would crate a avalanche where the thousands crates that depend on `chrono` would need to release a new version, too. Then the hundreds of crates that depend on `humantime`. Then the thousands of crates that depend on `clap`. And soon every single crate would have released a new version.

giraffate:
> This argument doesn't make much sense to me. The problematic dependency is not Chrono, it's not surprising that only updating Chrono doesn't fix it, and it's fixed quite simply by updating iana-time-zone directly (which is actually the cause of the issue).

Oh, yes, you're right. Sorry, my comment was way off the mark.

Thanks for your response!

esheppa:
Could this could be included as part of the `0.4.23` release? However potentially I've missed something

djc:
> Could this could be included as part of the `0.4.23` release? However potentially I've missed something

I think we just established that it wouldn't help much? We can but I don't think it matters that much.

PR #783, opened 2022-08-17
functions from core-foundation

esheppa:
This is a proof of concept PR to go along with https://github.com/servo/core-foundation-rs/pull/521 at `core-foundation`. If/when that PR is accepted, this can be fixed (change the git dependency to a proper dependency) and properly documented

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ alloc = []
 libc = []
 std = []
 clock = ["std", "winapi", "iana-time-zone"]
+with-core-foundation = ["core-foundation"]
 oldtime = ["time"]
 wasmbind = ["wasm-bindgen", "js-sys"]
 unstable-locales = ["pure-rust-locales", "alloc"]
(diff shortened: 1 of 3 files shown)


djc:
I think it definitely make sense to invert the dependency here -- core-foundation seems like it ought to be lower-level than chrono. Would like to better understand who's using this code (if anyone still is) before we move this code into chrono.

djc:
> Ideally we could merge this ASAP after https://github.com/servo/core-foundation-rs/pull/521 so that any users have a migration path.

Hmm, I think I'd prefer to let it linger a little bit to find out what those users actually need, instead of moving code around that was introduced 5 years ago and might, for all we know, no longer have an actual user.

esheppa:
Fair point, no need for a migration path if no one is using the code anyway :smile:

Kijewski:
Maybe this https://twitter.com/sourcegraph/status/1547991489149423620 can come in handy? 

> Starting today, Sourcegraph can search and navigate through 87k+ Rust Crates and all their versions.
>
> For example, see all the crates and versions starting with "tokio" here: [https://sourcegraph.com/search?q=repo%](https://t.co/DW61GvIKQH)

Kijewski:
I found nothing outside of core-foundation itself: <https://sourcegraph.com/search?q=context:global+CFDate.*naive_utc&patternType=regexp>. Maybe @ratake can tell for what they use(d) the feature?

esheppa:
Great find @Kijewski!

pitdicker:
@djc Do you have an opinion on how to proceed with this PR?

pitdicker:
Personally I think it is fine to leave the conversion between a `CFDate` and `NaiveDateTime` to the `core-foundation` crate.

djc:
I pinged the upstream PR again, now that the Servo project is a bit more alive.

pitdicker:
I see you removed the chrono dependency in core-foundation in https://github.com/servo/core-foundation-rs/pull/666. Still it seems no-one was using this code?

djc:
Yeah, we're only adding this back if some people come asking for it.

Issue #339, opened 2019-09-04
There is no way to add a "day" to a DateTime


sparky8251:
Just like in #318, this sample code can be used to explain the problem.

```rust
use chrono::NaiveDateTime;
use chrono::TimeZone;
use chrono::{DateTime, Duration};
use chrono_tz::US::Eastern;
use chrono_tz::Tz;

fn main() {
    let timestamp = 1572580800;

    let mut dt: DateTime<Tz> = Eastern.from_utc_datetime(&NaiveDateTime::from_timestamp(timestamp, 0));

    println!("Written datetime: {}", dt);

    //increment past the daylight savings change

    dt = dt + Duration::days(3);

    println!("Written datetime: {}", dt);
}
```

When it's run, the output is:
```
Written datetime: 2019-11-01 00:00:00 EDT
Written datetime: 2019-11-03 23:00:00 EST
```
What I would expect adding 3 days is that I would get `2019-11-4 00:00:00 EST`. I didn't add `86400 * 3` to the time, I added `Duration::days(3)` so this feels inconsistent as a consumer of this crate.

I don't know if this belongs here or if it belongs in something like a `chrono-utils` crate, but I do feel it needs addressing or at least some sort of call out in the docs that it's not actually days. It's non-obvious that it's just an abstraction for "number of seconds in a day * X".

frondeus:
Hi, 

I noticed this problem appears only in newest versions of `chrono`:

I wrote this test:
```
#[cfg(test)]
mod tests {
    extern crate chrono_tz;

    use self::chrono_tz::Europe::Warsaw;

    #[test]
    fn test_chrono_add() { // https://github.com/chronotope/chrono/issues/339
        use std::ops::Add;
        use chrono::*;

        let timezone = Warsaw;
        let mut date = timezone.from_utc_datetime(&"2017-10-28T22:00:00Z".parse::<DateTime<Utc>>().unwrap().naive_utc());

        dbg!(&date);
        assert_eq!("2017-10-29", &date.format("%Y-%m-%d").to_string());
        date = date.add(Duration::days(1));
        assert_eq!("2017-10-30", &date.format("%Y-%m-%d").to_string());
        date = date.add(Duration::days(1));
        assert_eq!("2017-10-31", &date.format("%Y-%m-%d").to_string());
    }
}
```

With `chrono = "=0.4.10"` it's failing - "Expected 2017-10-30, found 2017-10-29"
While with `chrono = "=0.4.7"` it's working correctly

[...]

Issue #290, opened 2018-11-14
`Datelike` should have methods to add/sub a number of calendar day/month/year


vincentdephily:
`Datelike` currently offers a set of `with_*()` methods to set the day/month/year, but the resulting date must be valid (for example calling `with_month(13)` doesn't translate to january of the next year). You can also add a `Duration` to a `Datelike`, but there's no good `Duration` value for year/month/day because of leap years, various month lengths, and DST. 

I'd like to have methods `add_days/month/years(i32) -> Self` to easily add/subtract the corresponding number of time units. Might want to have `add_hour/min/sec/etc()` for consistency's sake, although adding a `Duration` works well in this case (unless we want to handle DST ?).

My usecase is deceptively simple: I write a report and want to know when the next day/week/month/year starts, after a given timestamp. For example I have `2018-11-14T12:34:56z` as input and want `2018-12-14T00:00:00z` as output (beginning of next month).

This is ugly but doable for "next year":
```rust
fn next_year<T>(d: T) -> T where T: Datelike {
  d.with_year(d.year()+1)
    .unwrap()
   .with_month(1)
   .unwrap()
   .with_day(1)
   .unwrap()
   .with_hour(0)
   .unwrap()
   .with_minute(0)
   .unwrap()
   .with_second(0)
   .unwrap()
   // Thankfully I know that my input had a second granularity so I don't care about nanoseconds.
}
```
But the implementation for `next_month/week/day()` is not going to be fun.

Incidentally, it with also be nice to have a `with_year/month/day/etc()` that panics instead of returning a result, for cases like this where the input is hardcoded and known to be correct. Or a `with_ymd()` and `with_hmsn()` ;)

felipenoris:
I have a similar issue. I just want to get the next/last day out of a `Datelike` date. For now I'm doing this:

```rust
fn next_date<T: Datelike + Copy>(date: T, fwd: bool) -> T {
    let inc: i32 = {
        if fwd {
            1
        } else {
            -1
        }
    };

    match date.with_ordinal((date.ordinal() as i32 + inc) as u32) {
        Some(dt) => dt,
        None =>  {
            if fwd {
                date
                    .with_year(date.year() + 1).unwrap()
                    .with_ordinal(1).unwrap()
            } else {
                date
                    .with_year(date.year() - 1).unwrap()
                    .with_month(12).unwrap()
                    .with_day(31).unwrap()
            }
        }
    }
}
```

nifker:
[...]

PR #784, opened 2022-08-17
add/sub Days

esheppa:
This matches the work done by @avantgardnerio and @djc with the `Months` datatype but for calendar `Days`.

This should resolve https://github.com/chronotope/chrono/issues/339 and https://github.com/chronotope/chrono/issues/290 and is relevant to the clarification in https://github.com/chronotope/chrono/issues/100.

Potential future work could:
* Deprecate the `days` and `weeks` helper functions on `Duration` as they are somewhat misleading, directing users to use `Days` instead
* Potentially deprecate `succ` and `pred` on `NaiveDate` - although sometimes it's nice to have multiple ways to do something

This is a bit light on documentation as I'm not sure whether this is wanted or not. If the idea here is looking good, I'll add more docs.


src/naive/date.rs
```diff
@@ -692,6 +680,83 @@ impl NaiveDate {
...
+    ///     Some(NaiveDate::from_ymd(2021, 8, 20))
+    /// );
+    /// ```
+    pub fn checked_sub_months(self, months: Months) -> Option<Self> {
+        if months.0 == 0 {
+            return Some(self);
+        }
+
+        // Copy `i32::MIN` here so we don't have to do a complicated cast
+        match months.0 <= 2_147_483_648 {
```
avantgardnerio, line 702: There's a bug here, best make it `2_147_483_647` https://github.com/chronotope/chrono/pull/778#discussion_r947706844

avantgardnerio (changes requested):

src/naive/date.rs
```diff
@@ -625,31 +638,6 @@ impl NaiveDate {
...
-    /// Returns `None` if the resulting date would be out of range.
-    ///
-    /// ```
-    /// # use chrono::{NaiveDate, Months};
-    /// assert_eq!(
-    ///     NaiveDate::from_ymd(2022, 2, 20).checked_sub_months(Months::new(6)),
-    ///     Some(NaiveDate::from_ymd(2021, 8, 20))
-    /// );
-    /// ```
-    pub fn checked_sub_months(self, months: Months) -> Option<Self> {
```
djc, line 641: Please don't move unrelated code around in the same commit.

src/naive/date.rs
```diff
@@ -692,6 +680,83 @@ impl NaiveDate {
...
+        }
+
+        // Copy `i32::MIN` here so we don't have to do a complicated cast
+        match months.0 <= 2_147_483_648 {
+            true => self.diff_months(-(months.0 as i32)),
+            false => None,
+        }
+    }
+
+    /// Add a duration in [`Days`] to the date\
```
djc, line 708: Spurious `\`?

src/naive/datetime/mod.rs
```diff
@@ -606,6 +606,24 @@ impl NaiveDateTime {
         Some(NaiveDateTime { date, time })
     }
 
+    /// Add a duration in [`Days`] to the date part of the `NaiveDateTime`
+    ///
+    /// Returns `None` if the resulting date would be out of range.
+    pub fn checked_add_days(self, days: Days) -> Option<Self> {
+        let new_date = self.date().checked_add_days(days)?;
```
djc, line 613: Simplifying:

```rust
Some(Self {
    date: self.date.checked_add_days(days)?,
    time: self.time,
})
```

src/naive/date.rs
```diff
@@ -115,6 +115,19 @@ impl NaiveWeek {
     }
 }
 
+/// A duration in calendar days. This is useful becuase when using `TimeDelta` it is possible
```
djc, line 118: I think only the first sentence should be on the first line. Also "becuase", should backtick-quote `TimeDelta::days(1)`. And it's weird to be talking about `TimeDelta` here because that doesn't actually exist in this branch.

If we're going with `TimeDelta` as an output type only, we probably shouldn't have constructors like `TimeDelta::days()`?

esheppa: Yep looks like I got ahead of myself with the docs here. I'll leave the comparison to `Duration::days` for now so that users can figure out when to use each. 

`TimeDelta` would still be an output type when using the `signed_duration_since` and equivalent operations Ideally though, we would make `Days` be the output type for `NaiveDate::signed_duration_since`, however `Days` doesn't currently allow negative, so this is something to think about. Overall we probably want to be consistent in allowing signed duration everywhere or nowhere, but perhaps it is also possible to have the calendar durations (`Months` and `Days`) not support negative, and the precise durations (now `Duration`, later potentially `TimeDelta`) support fowards and backwards.

If we don't want to support signed calendar duration (my preference), then we could change `NaiveDate::signed_duraiton_since` to `NaiveDate::days_since` or `NaiveDate::absolute_days_since` or equivalent (and also have `NaiveDate::months_since`).

src/naive/date.rs
esheppa: this code was accidentally moved, I've now rebased to include your recently merged fixes

src/naive/datetime/mod.rs
esheppa: good call - I've also used the `..self` syntax, although I can use the above if you prefer for consistency

src/naive/date.rs
```diff
@@ -698,6 +715,58 @@ impl NaiveDate {
...
+    ///     NaiveDate::from_ymd(2022, 7, 31).checked_add_days(Days::new(2)),
+    ///     Some(NaiveDate::from_ymd(2022, 8, 2))
+    /// );
+    /// ```
+    pub fn checked_add_days(self, days: Days) -> Option<Self> {
+        if days.0 == 0 {
+            return Some(self);
+        }
+
+        match i64::try_from(days.0) {
```
djc, line 738: How do you feel about turning this into `.ok()?.map(|d| self.diff_days(d))`?

src/naive/date.rs
```diff
@@ -698,6 +715,58 @@ impl NaiveDate {
...
+    ///     NaiveDate::from_ymd(2022, 2, 20).checked_sub_days(Days::new(6)),
+    ///     Some(NaiveDate::from_ymd(2022, 2, 14))
+    /// );
+    /// ```
+    pub fn checked_sub_days(self, days: Days) -> Option<Self> {
+        if days.0 == 0 {
+            return Some(self);
+        }
+
+        match i64::try_from(days.0) {
```
djc, line 760: Same here.

djc (approved):
LGTM with two style suggestions.

src/naive/date.rs
esheppa: >  but I think that's maybe a feature more than a bug, as it's actually the tricky case?

This sounds excellent to me. IMO using a negative duration as an input would usually provide unexpected results. This also allows unfallible versions returning `TimeDelta` and equivalent fallible ones returning `Duration`

src/datetime/mod.rs
```diff
@@ -339,6 +339,26 @@ impl<Tz: TimeZone> DateTime<Tz> {
...
+    ///
+    /// Returns `None` if the resulting date would be out of range.
+    pub fn checked_add_days(self, days: Days) -> Option<Self> {
+        self.datetime
+            .checked_add_days(days)?
+            .and_local_timezone(TimeZone::from_offset(&self.offset))
+            .single()
+    }
+
+    /// Subtract a duration in [`Days`] to the date part of the `DateTime`
```
djc, line 352: Subtract from.

djc (approved):
One more minor nit I missed the first time around.

avantgardnerio (approved):
Looks great!

djc (approved):

esheppa:
Having thought about this a bit, I think I need to add some tests before this gets merged :smile:

esheppa:
I've reused the tests from `test_date_add/sub` - this should be useful as we want to keep these even if the impl `Add/Sub<Duration> for NaiveDate` is removed

esheppa:
@djc - what is the best way to get these commits into main as well - should we periodically PR `0.4.x` into `main`?

djc:
Yes, I think so.

PR #785, opened 2022-08-21
checked+regular add/sub months impls

esheppa:
This is similar to https://github.com/chronotope/chrono/pull/778 but adds the functionality to `DateTime<Tz>`. This has one extra failure case, aside from overflow, which occurs when the existing local time is not valid on the new date once the months are added.

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -30,6 +30,7 @@ use crate::offset::Local;
 use crate::offset::{FixedOffset, Offset, TimeZone, Utc};
 use crate::oldtime::Duration as OldDuration;
 use crate::Date;
+use crate::Months;
 use crate::{Datelike, Timelike, Weekday};
 
 #[cfg(feature = "rkyv")]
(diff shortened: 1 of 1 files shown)


src/datetime/mod.rs
```diff
@@ -329,6 +330,19 @@ impl<Tz: TimeZone> DateTime<Tz> {
         Some(tz.from_utc_datetime(&datetime))
     }
 
+    /// Adds given `Months` to the current date and time.
+    ///
+    /// Returns `None` when it will result in overflow, or if the existing
+    /// local time is not valid on the newly calculated date.
+    ///
+    /// See [`NaiveDate::checked_add_months`] for more details on behaivour
```
djc, line 338:
```suggestion
    /// See [`NaiveDate::checked_add_months`] for more details on behavior.
```

src/datetime/mod.rs
```diff
@@ -339,6 +353,19 @@ impl<Tz: TimeZone> DateTime<Tz> {
         Some(tz.from_utc_datetime(&datetime))
     }
 
+    /// Subtracts given `Months` from the current date and time.
+    ///
+    /// Returns `None` when it will result in overflow, or if the existing
+    /// local time is not valid on the newly calculated date.
+    ///
+    /// See [`NaiveDate::checked_sub_months`] for more details on behaivour
```
djc, line 361:
```suggestion
    /// See [`NaiveDate::checked_sub_months`] for more details on behavior.
```

src/datetime/mod.rs
```diff
@@ -329,6 +330,19 @@ impl<Tz: TimeZone> DateTime<Tz> {
         Some(tz.from_utc_datetime(&datetime))
     }
 
+    /// Adds given `Months` to the current date and time.
+    ///
+    /// Returns `None` when it will result in overflow, or if the existing
```
djc, line 335: I feel like the "existing" word here doesn't add much value and distracts from the explanation. And maybe use "not valid for" instead of "not valid on"? Not a native speaker, so you may have a better idea than me.

djc (commented):
Nice!

src/datetime/mod.rs
esheppa: Removed the "existing" - agreed that didn't add much.

With the on/for, I'd lean toward "on" for a specific date, and "for" when talking about a longer period like a week. Did a quick search found only this: https://english.stackexchange.com/questions/182660/on-vs-at-with-date-and-time

djc (approved):

PR #787, opened 2022-08-22
Make the changelog historical from now on

djc:
The changelog causes a lot of conflicts in PRs, which makes maintenance harder. For release notes, it's better to use GitHub's releases feature, which gets nice per-release pages you can easily link to as well as being easier to find for automated tools such as Dependabot. This means more of the onus on producing release notes is on the maintainer, but I usually find it useful to edit the changelog a bit anyway.


esheppa (approved):
Looks good! (Good call on the template change as well :smile:)

esheppa:
Potentially these lines should be altered/removed from the README.md: https://github.com/chronotope/chrono/blob/69f1770df85dd8e9b3fbecdf7b4e7a974469bb48/README.md?plain=1#L34-L35

esheppa:
However I think it is better to merge this first. I've got some plans for a separate cleanup of the readme including potentially using a `#![doc = include_str!("README.md")]` to simplify the CI checks and avoid the issue where the `lint` stage fails due to the readme not being updated.

djc:
Oh, good call on the sentence in the README, I've just cut it out for now.

djc:
I've now scrapped the README/top-level docs sync because it seems like a crappy implementation and IMO the README was rather too detailed. I'd rather have a more curated README that sometimes falls slightly out of date...

esheppa (approved):
Looks good! I like the succinct readme

PR #789, opened 2022-08-24
Add additional checks for AlternateTime

x-hgg-x:

      about adding the PR number)
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md

This PR adds additional checks for the `AlternateTime` struct in the `tz-info` module, not existing in the initial implementation (see https://github.com/chronotope/chrono/pull/677#issuecomment-1096927057).


esheppa:
Thanks for this @x-hgg-x - I'll need a bit of time to understand this better, however one thing I noticed was:

> `/// Check DST transition rules consistency, which ensures that the DST start and end time are always in the same order.`

Is this empirically always the case in the IANA tzdb? From memory while testing I'd come across cases where there were regular transitions that switched orders, although potentially these checks only apply to the rules that are used once all the hard-coded transition times are exhausted?

x-hgg-x:
Yes all the rules in the IANA database are correct. In the case that a new version of the database adds an invalid rule, the [tzdb](https://github.com/Kijewski/tzdb) crate will not compile since the check is done at compile time, so I will be notified.
I think this is unlikely to happen since in this case, the resulting rule is ill-defined and not covered by the POSIX specification.

x-hgg-x:
> these checks only apply to the rules that are used once all the hard-coded transition times are exhausted

Yes the check only apply to the extra rule after the last transition, defined by a [POSIX `TZ` string](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html).

esheppa:
> Yes all the rules in the IANA database pass the check. In the case that a new version of the database adds an invalid rule, the [tzdb](https://github.com/Kijewski/tzdb) crate will not compile since the check is done at compile time in tz-rs, so I will be notified.

With regards to this - would you mind adding a test that parses all the available files in the systems timezone database? That way it runs in the CI here, meaning we would be notified here if the validation is causing an issue. (Separately, I'll schedule the CI to run at least once a day on `main` which would probably be useful as well)

x-hgg-x:
> would you mind adding a test that parses all the available files in the systems timezone database

Done in 19a27f6.

djc:
If we're "statically" validating that the tzdb zones pass some particular test, do we still need run-time verification? Seems like overkill to me.

esheppa:
One potential benefit of the runtime validation is that we could then make more assumptions about the structure of the files and therefore change fallible functions to be non-fallible, provided the validation has already passed, which might simplify some of the downstream code.

x-hgg-x:
> If we're "statically" validating that the tzdb zones pass some particular test, do we still need run-time verification? Seems like overkill to me.

I don't think there is any compile-time validation of the tzdb zones at the moment in the `chrono` repository.

x-hgg-x:
A quick benchmark on my PC for the `test_read_all_tz_files()` test shows that the `check_dst_transition_rules_consistency()` validation check represents less than 1% of the runtime of the `TimeZone::from_tz_data()` function, so I think it is ok even if don't put it behind a feature.

PR #790, opened 2022-08-24
Start work on 0.5

djc:
This breaks compatibility on the main branch by removing the dependency on time 0.1, rustc-serialize and num-traits. Before we merge this, I propose we create a 0.4.x branch to continue development of compatible features and fixes. In the pull request template, we could suggest targeting the PR at 0.4.x if it makes semver-compatible changes, since it is easier to merge forward than back.

This also renames the internal `oldtime::Duration` to `TimeDelta` (without making other changes to the functionality). That seemed like an interesting small change to get us started on an incremental path toward a better design.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,6 @@ __doctest = []
 
 [dependencies]
 num-integer = { version = "0.1.36", default-features = false }
-num-traits = { version = "0.2", default-features = false }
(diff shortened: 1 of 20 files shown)


esheppa:
https://github.com/chronotope/chrono/pull/792 should cover all the remaining CI checks

djc:
Closing this in favor of #794.

PR #792, opened 2022-08-26
Fix some doc tests etc

esheppa:


djc:
Thanks for helping out with this! Can you split this along the lines of my commits and just fixup the commits from my branch in place? It looks like there are other changes as well, would be good to add a separate commit for those.

esheppa:
I've split this out into commits here and also set up https://github.com/chronotope/chrono/pull/794 as I wasn't quite sure whether you wanted me to split these commits, or split the single one here but then correlate the bits with your other commits

djc:
Either is fine, really. I think working from your new PR is easiest at this point. Maybe you could consider working on branches in this repo, too, so that we may more easily collaborate on a single branch in cases like this?

esheppa:
That is a great point RE. working on branches in this repo, I didn't realise I was able to but it makes sense that I can.

PR #793, opened 2022-08-26
Use xtask for CI

esheppa:
Just trying this out to see if it can make the CI setup more user friendly


djc:
I definitely would like to get rid of ~all of the shell script stuff here, but I think I'd prefer moving most of it into the workflow config itself rather than nesting things into something else. If it turns out we can achieve substantial simplification from doing it this way I guess I'm open to that, but I think we should first try to do stuff in a way that feels more native to GitHub Actions.

esheppa:
Fair call. I've tried to copy across the shell script logic as closely as possible with this, with a plan to simplify/change it in another PR - as I don't want to accidentally stop doing some testing that we are currently doing. As part of that subsequent PR, if it is decided that the `xtask` method is a good choice, I'd aim in that case to make the workflow `yml` file as simple as possible, moving logic into the `xtask`. The main reasoning behind this is that it enables as much as possible to be tested locally (for example it could potentially call rustup to change the toolchain rather than manually commenting/uncommenting a `rust-toolchain.toml` file).

I think the most optimal comparison is between:
* Most logic in github actions config, with ideally none in shell scripts/xtask
* Most logic in shell scripts/xtask with minimal (just the matrix stuff to get parallelism) in github actions config

esheppa:
Looks like the wasm_simple test has been failing quietly for a while, will investigate further. The features powerset checking has also picked up a few issues (albeit, unlikely anyone would have them in practice)

.github/workflows/lint.yml
```diff
@@ -2,25 +2,22 @@ name: lint
 
 on:
   push:
-    branches: [main, master]
+    branches: [main]
```
djc, line 5: This seems unrelated to xtask?

.github/workflows/lint.yml
```diff
@@ -2,25 +2,22 @@ name: lint
...
   pull_request:
-    branches: [main, master]
+    paths:
```
djc, line 7: So is this, right?

.github/workflows/lint.yml
```diff
@@ -2,25 +2,22 @@ name: lint
...
     runs-on: ubuntu-latest
 
     steps:
       - uses: actions/checkout@v2
-      - name: Install rust
-        uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-          override: true
+      - uses: dtolnay/rust-toolchain@stable
```
djc, line 20: And this?

djc (commented):
So because this moves a bunch of stuff around in an area that's pretty sensitive (affecting our code coverage), I'd really like to review this as minimally small commits, whether in a single PR or multiple PRs. As it is, this has a lot of changes packed into a relatively large PR, which I find harder to review. So I'd like this to be split up more.

.github/workflows/lint.yml
esheppa: I've now moved this to a seperate commit which cleans up `lint.yml` and switches to a different rust toolchain manager

esheppa:
Let me know if this is the right balance for commit splitting. In the first commit, the code changes are necessary for the CI to pass due to the newly introduced features powerset test. The changes to `lint.yml` and `test.yml` are primarily to show how switching to xtask can somewhat simplify those files. 

I could potentially split that further to just adding the xtask and code changes and then simplifying the CI in a subsequent commit

djc:
Okay, I really just want to get this done without xtask, just moving everything into the YAML world. My problem here is that writing Rust code is a pretty shitty way to then again just shell out to Cargo processes, especially with all `::group::` stuff which just seems like dark magic. I'd rather have a lot of boilerplate YAML than clever Rust code in an xtask crate, because it's easier to navigate and I don't have to learn a bunch of stuff that I'd mostly get for free with the YAML.

I understand the benefit would be that we get to run it locally, but to me that benefit is too limited to trade off against the additional layer of complexity here (at least as formulated in this PR so far). IIRC there is also a Cargo extension command that can do the powerset stuff, but IMO the first priority here is to get rid of all the shell script craziness and to limit all the duplicate jobs. I think OS / Rust version / Cargo features / timezones are all orthogonal axes, and we don't need an extensive matrix for each of these.

esheppa:
Sounds reasonable to me. We can use [`cargo-hack`](https://github.com/taiki-e/cargo-hack) to do the powerset testing.

djc:
Yeah -- sorry for the time you put in on this.

esheppa:
Given its a delicate change, its worth experimenting with a few different options. We will likely want to still fix the `wasm_simple` test and add the other attribute changes so it was at least useful to find those issues.

PR #794, opened 2022-08-27
0.5 alpha compressed commits

esheppa:


src/lib.rs
```diff
@@ -425,13 +417,7 @@
 #![cfg_attr(feature = "rustc-serialize", allow(deprecated))]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
-#[cfg(feature = "oldtime")]
-#[cfg_attr(docsrs, doc(cfg(feature = "oldtime")))]
-extern crate time as oldtime;
-#[cfg(not(feature = "oldtime"))]
-mod oldtime;
 // this reexport is to aid the transition and should not be in the prelude!
```
djc, line 414: Should remove this line, too, right?

src/lib.rs
```diff
@@ -425,13 +417,7 @@
 #![cfg_attr(feature = "rustc-serialize", allow(deprecated))]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
-#[cfg(feature = "oldtime")]
-#[cfg_attr(docsrs, doc(cfg(feature = "oldtime")))]
-extern crate time as oldtime;
-#[cfg(not(feature = "oldtime"))]
-mod oldtime;
```
djc, line 432: Without this module declaration (and the import below, I suppose), I'm pretty sure this commit does not compile (and test). Would be nice to fix that properly.

esheppa: ah yikes I must have grabbed this in the wrong commit

esheppa: Hmm looks like I wiped out some commits when redoing them, will separate them again (I'm learning a bunch of `git` features I never knew existed!)

djc: If you did this using rebase -i, watch out for amending commits vs continuing the rebase, that's usually how I accidentally squash commits.

esheppa: Classic, that is exactly what I did! Should be right now, I've checked that it can compile at each commit

djc (approved):
Awesome! Just needs a rebase.

djc:
I've pushed a 0.4.x branch from the HEAD of main before merging this.

PR #800, opened 2022-08-29
Add sub days from 0.4.x v2

esheppa:


djc:
Hmm, so I think we want to keep explicit merge commits for this stuff? It looks like you rebased here, but that won't give us merge tracking (that is, it won't help us figure out which part we've already merged).

djc:
My original point is, if we backport from main to 0.4.x we have to keep track of what we've backported. If we use git merges to merge forward, then Git will track that for us, and each update is just `git merge 0.4.x`.

esheppa:
Perfect - I've set up a new PR with the proper merge

PR #802, opened 2022-08-30
Add sub days from 0.4.x v3

esheppa:
Trying this out using `git merge 0.4.x` as per https://github.com/chronotope/chrono/pull/800#issuecomment-1230345807

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -27,7 +27,7 @@ use crate::format::DelayedFormat;
 use crate::format::Locale;
 use crate::format::{parse, ParseError, ParseResult, Parsed, StrftimeItems};
 use crate::format::{Fixed, Item};
-use crate::naive::{IsoWeek, NaiveDate, NaiveDateTime, NaiveTime};
+use crate::naive::{Days, IsoWeek, NaiveDate, NaiveDateTime, NaiveTime};
 #[cfg(feature = "clock")]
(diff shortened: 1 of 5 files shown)


esheppa:
The diff now looks identical to that of #800 which is ideal

djc:
So we don't want to rebase or squash, which leaves the option of adding a merge commit... which I guess means we end up with two merge commits? I wonder if we can prevent that somehow...

esheppa:
how would "Rebase and merge" look for this? It would change the commit hashes, but potentially would preserve the details of the merge commit pointing back to the original?

djc:
Hmm, maybe. But actually we can't rebase because GitHub can't deal with the conflicts?

esheppa:
Maybe we can try targeting another branch first to see what happens?

djc:
Sounds good!

esheppa:
Hmm I'd not clicked through the options, as you said, unable to do the rebase (presumably it would work manually, but the main branch is probably protected). I went with the merge - it doesn't actually look too bad, however I should probably avoid using the word "merge" in the message of the first commit, which would make it look better

esheppa:
Maybe just squashing would be better?

esheppa:
squash doesn't look great (test-main-v2). I think the double-merge-commit is the least bad option, and the two merges is a reasonable price to pay for keeping the history linked

PR #804, opened 2022-08-30
Add sub days from 0.4.x v3

esheppa:
Will redo the commit message for this to remove the word "merge"

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -27,7 +27,7 @@ use crate::format::DelayedFormat;
 use crate::format::Locale;
 use crate::format::{parse, ParseError, ParseResult, Parsed, StrftimeItems};
 use crate::format::{Fixed, Item};
-use crate::naive::{IsoWeek, NaiveDate, NaiveDateTime, NaiveTime};
+use crate::naive::{Days, IsoWeek, NaiveDate, NaiveDateTime, NaiveTime};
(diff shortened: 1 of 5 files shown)


djc:
I guess the double merge option is fine -- just seems crappy that GitHub doesn't have a better way to do this.

esheppa:
lets hold off for a moment as there are a few other options I can try

esheppa:
Actually I guess the only way to meet both requirements:
* avoid double merge commit
* maintain some link to the history

Is to not have a protected main branch, and just do one merge commit. But that seems worse than just having double merge commits - so then the choices become:

* Double merge
* Double rebase (I've done this as `test-main-v3` just to make sure I've tried all the options)

Comparing `test-main`, `test-main-v2` and `test-main-v3` I think `test-main` is the clear winner, as the commits can be tracked back. Alternatively if we want only linear history then `test-main-v3` is better, however the only way to track the commits is by message.

djc:
I think I agree. So is this good to merge?

esheppa:
Yes good to go unless we think its worth rewording the `merge 0.4.x - Add/Sub days` commit message

esheppa:
This is really nice - the `0.4.x` branch shows up as exactly 8 commits behind, and 0 ahead, so that makes the double merge definitely the best choice

Issue #263, opened 2018-07-13
Improvements to parsing into Utc, converting from other timezones to Utc


mqudsi:
Thanks for making dealing with dates and times in rust easier and for providing what has become the defacto standard for date-based operations.

I do believe there is room for improvement however when it comes to dealing with `DateTime<Utc>` instances; unless I'm mistaken and there's actually an easier way, this is what my code currently looks like to parse a timestamp from a string into a `DateTime<Utc>` instance:

```rust
        let created: DateTime<Utc> = DateTime::from_utc(
            DateTime::<FixedOffset>::parse_from_rfc3339("1996-12-19T16:39:57-08:00")
                .unwrap()
                .naive_utc(),
            Utc,
        );
```

Aside from the fact that it isn't very pretty or terse, I think it's also extremely undiscoverable. In particular, the usage of DateTime::**from_utc** to convert _to_ a `DateTime<Utc>` object is unexpected (I understand why it works).

It would be extremely helpful if `DateTime<FixedOffset>` had a `to_utc()` helper method or some other means of converting more intuitively between `DateTime` flavors.

Is there a fundamental reason why there are no `std::convert` impls to facilitate conversion between these types? Since in the code above there is only the first `.unwrap()` and all other operations are not expected to ever produce an error, I imagine it would be possible to use `std::convert` in its current form (vs waiting for `std::convert::TryFrom` to finally stabilize).

If `std::convert::From<chrono::DateTime<chrono::FixedOffset>>` were implemented for `chrono::DateTime<chrono::Utc>` (or `std::convert::Into<chrono::DateTime<chrono::Utc>>` implemented for `chrono::DateTime<chrono::FixedOffset>` if one type isn't aware of the other) the code above could be simplified considerably:

```rust
        let created: DateTime<Utc> = DateTime::<FixedOffset>::parse_from_rfc3339("1996-12-19T16:39:57-08:00").unwrap().into();
```

(Although I would still suggest that `parse_from_rfcxxxx` be implemented for `DateTime<Utc>` in all cases.)

dpc:
I came here, looking for help, found this issue so I'll post my experiences.

I was trying to convert date in `Local` timezone to a `FixedOffset` timezone.

The timezone/offset conversions are very confusing and very hard to discover. Some of the traits involved don't even make sense to me:

Like offset vs timezone - what is `Local` if not a `FixedOffset` timezone of system-configured value? Donno. Maybe I'm missing something.

[...]

PR #806, opened 2022-08-31
Add `parse_from_xxx` to `DateTime<Utc>`

mqudsi:
As discussed in #263. Note that this breaks existing code that did not
explicitly specify the offset in calls to `DateTime::parse_from_*`, as
they are now ambiguous.

Relies on #271 for conversion back into `DateTime<Utc>` from
`DateTime<FixedOffset>` as the latter is used under the hood.

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -524,11 +524,13 @@ where
 }
 
 impl DateTime<FixedOffset> {
-    /// Parses an RFC 2822 date and time string such as `Tue, 1 Jul 2003 10:52:37 +0200`,
-    /// then returns a new [`DateTime`] with a parsed [`FixedOffset`].
+    /// Parses an RFC 2822 date-and-time string into a `DateTime<FixedOffset>` value.
     ///
-    /// RFC 2822 is the internet message standard that specifies the
-    /// representation of times in HTTP and email headers.
+    /// This parses valid RFC 2822 datetime strings (such as `Tue, 1 Jul 2003 10:52:37 +0200`)
+    /// and returns a new [`DateTime`] instance with the parsed timezone as the [`FixedOffset`].
+    ///
+    /// RFC 2822 is the internet message standard that specifies the representation of times in HTTP
+    /// and email headers.
     ///
     /// ```
     /// # use chrono::{DateTime, FixedOffset, TimeZone};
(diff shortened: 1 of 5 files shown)


src/datetime/mod.rs
```diff
@@ -585,6 +585,49 @@ impl DateTime<FixedOffset> {
     }
 }
 
+impl DateTime<Utc> {
+    /// Parses an RFC 2822 date and time string such as `Tue, 1 Jul 2003 10:52:37 +0200`,
```
djc, line 589: Nit: would like these doc strings to start with a single concise line describing the functionality, then an empty line, then a longer/more detailed description.

djc (approved):
Makes sense to me, I think this is helpful new API. @esheppa any thoughts?

esheppa:
My only thought is whether this should assert that the timezone is UTC, either via a trailing `Z` or a `+00:00` offset? It is arguable either way, but it might be confusing if a time is parsed that has an offset, only for the timestamp to change as it is converted to UTC.

mqudsi:
@esheppa I respectfully disagree. The strongly typed type declared in the code is how one wants the date and time to be stored. The kind of date that it comes from is specified by the input format and can diverge without a problem. 

eg in my code someone wants all datetime variables to be in UTC because that’s what they want  to deal with and enforce is stored in the db, on the disk, displayed (unless converted), etc. But as long as I can convert from a non-UTC datetime losslessly, I would absolutely hate to be blocked from using that. 

In this specific instance, the function names indicate what input formats are allowed. In this case,  the input should be any valid RFC 2822 or RFC 3339 string, neither of which are restricted to UTC-only. 

If this were a “deserialize” impl perhaps you could get away with requiring (for no measurable benefit) that datetime strings are all in UTC but that’s not the point of this question - and even in that case, I don’t see the point in not being flexible and forgiving in what we allow as input. 

The function signatures tell you what to reasonably expect: any valid RFC 2822/3339 text input and a `DateTime<Utc>` output. It would make no sense to arbitrarily restrict the inputs to a subset of what the RFC supports and I would argue that if we had code that did that, sooner or later someone would rightly file a bug complaining about it.

src/datetime/mod.rs
mqudsi: I agree. These were copy and pasted from the original impl for FixedOffset at the time the patch was first authored - I don’t know if those impls have had their docs updated but I agree that it would be the correct way given the way rust docs are generated and displayed and given that it’s the defacto approach the community has adopted. 

I’ll update all the docs in this PR accordingly and squash the changes.

esheppa:
Thanks for your reply @mqudsi - that sounds reasonable to me - and from a second look at the docs it looks like there are a few examples that show this

src/datetime/mod.rs
djc: Cool, yeah, if you update the docs, let's merge this!

Note: this PR targets main but main is now intended for 0.5.0 development. Since this is additive, we could merge it into 0.4.x to make it available sooner, so maybe rebase and change the PR's target branch?

mqudsi:
I updated the branch with the requested changes to the documentation. Note that I didn't squash and force push because it turns out the docs for the `DateTime<FixedOffset>` parsing functions were also incorrectly format, so I decided to just tackle both in one go.

I'm happy to open a separate PR targeting 0.4.x with these commits rebased on top of the current 0.4 branch but I want to confirm that the API breakage is ok for a 0.4.x release? While the changes are purely additive, any existing usage of `DateTime::parse_xxx(...)` would break as that is now an ambiguous method call that may return either a `DateTime<Utc>` or a `DateTime<FixedOffset>` value (I honestly have no idea why these were only implemented for `DateTime<FixedOffset>` in the first place).

djc:
Ah yeah, in that case it's probably better to stick with 0.5.x only.

PR #786, opened 2022-08-22
datetime_from_str is exact about whitespace

jtmoon79:
_THIS IS A PROOF-OF-CONCEPT PULL REQUEST, NOT FINAL, CODE IS POOR QUALITY_

`format/parse.rs:datetime_from_str` is exact about whitespace in
the passed data `s` and passed strftime format `fmt`

Issue #660

See https://github.com/chronotope/chrono/issues/660#issuecomment-1221812907

### Thanks for contributing to chrono!

- [ ] Have you added yourself and the change to the [changelog]? (Don't worry
      about adding the PR number)
- [ ] If this pull request fixes a bug, does it add a test that verifies that
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md

Issue #660, opened 2022-03-21
Local.datetime_from_str errantly ignoring format whitespace


jtmoon79:
Function [`chrono::Local.datetime_from_str`](https://docs.rs/chrono/0.4.9/chrono/offset/trait.TimeZone.html#method.datetime_from_str) ignores trailing whitespace in the passed variable `fmt`.

[Given code](https://play.rust-lang.org/?version=beta&mode=debug&edition=2021&gist=9b8fd4039378dafb7f3c2bf42b443ae2)
```rust
extern crate chrono;
use chrono::{Local, TimeZone};

fn main() {
    let a = Local.datetime_from_str("2000-01-01 02:03:01", "%Y-%m-%d %H:%M:%S ");
    println!("{:?}", a);
    let a = Local.datetime_from_str(" 2000-01-01 02:03:02\n", "%Y-%m-%d %H:%M:%S ");
    println!("{:?}", a);
}

```
results in unexpected successful parsing.
```text
Ok(2000-01-01T02:03:01+00:00)
Ok(2000-01-01T02:03:02+00:00)
```

The passed variable `fmt` has a trailing space that is not present in the passed `s`,  yet `datetime_from_str` returns `Ok`. These are errant matches.

---

Using `chrono` version `0.4.19`.

Similar to Issue #548

jtmoon79:
In case anyone needs a hack workaround. ([full code with tests](https://play.rust-lang.org/?version=beta&mode=debug&edition=2021&gist=3f879edb1dd50d0080f498be6ac4f861))

```rust
/// workaround for chrono Issue #660 https://github.com/chronotope/chrono/issues/660
/// match spaces at beginning and ending of inputs
/// TODO: handle all Unicode whitespace.
///       This fn is essentially counteracting an errant call to `std::string:trim`
///       within `Local.datetime_from_str`.
///       `trim` removes "Unicode Derived Core Property White_Space".
///       This implementation handles three whitespace chars. There are twenty-five whitespace
///       chars according to
///       https://en.wikipedia.org/wiki/Unicode_character_property#Whitespace .
pub fn datetime_from_str_workaround_Issue660(value: &str, pattern: &str) -> bool {
    let spaces = " ";
    let tabs = "\t";
    let lineends = "\n\r";
    
    // match whitespace forwards from beginning
    let mut v_sc: u32 = 0;  // `value` spaces count
    let mut v_tc: u32 = 0;  // `value` tabs count
[...]

PR #807, opened 2022-08-31
Issue #660 whitespace exact

jtmoon79:
be exact about whitespace Issue #660

Be exact about whitespace in parsing. This changes
pattern matching in `format::parse::parse` as it does not allow
arbitrary whitespace before, after, or between the datetime specifiers.

`format/parse.rs:datetime_from_str` is exact about whitespace in
the passed data `s` and passed strftime format `fmt`.

Also be more exacting about colons and whitespace around timezones.
Instead of unlimited colons and whitespace, only match a more limited
possible set of leading colons and whitespace.
(See `fn colon_or_space` https://github.com/chronotope/chrono/pull/807/files#diff-679cb1eea69fd36c51536ec228e61d6b0e40ade95997877f6f853e6ad31de5f1R224-R227).

Also rename some test functions to better describe what is tested.

---

~To help reviewers understand how test results changed _before_ and _after_, I created a branch `Issue660_whitespace_exact_try6_prechange_testsonly` that has added same set of test cases  as this branch `Issue660_whitespace_exact_try6`, but does not have the underlying changes for whitespace and colons (i.e. `Issue660_whitespace_exact_try6_prechange_testsonly`  is the "before significant changes" branch, and `Issue660_whitespace_exact_try6` is the "after significant changes" branch)~ https://github.com/jtmoon79/chrono/compare/Issue660_whitespace_exact_try6_prechange_testsonly..Issue660_whitespace_exact_try6
~The tests that result in different values are marked with `// TESTDIFF`.~

---

This is a lot of changes in this PR.  IMO, it covers one idea so it suitable for one PR; colons and whitespace should be handled more exactly. The changes for both of these tended to be the same code lines, so I thought it was sensible to make the changes all in one commit.  Hopefully the additional test cases provide enough confidence for accepting this significant change.

---

      about adding the PR number)
- [X] If this pull request fixes a bug, does it add a test that verifies that
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md

<br/>

I did not add to the `CHANGELOG.md` because there was no section `0.5` to add a new line item. I can add such a section if the reviewers would like.
(In a prior review, it was recommended that these changes should wait until a significant version change number (like from `0.4...` to `0.5`). https://github.com/chronotope/chrono/issues/660#issuecomment-1224052050 )

---

Also see initial change proposal explanation at https://github.com/chronotope/chrono/issues/660#issuecomment-1221812907

This PR is a further draft of #786 .


jtmoon79:
I converted this back to a Draft. After reviewing here I realized a few more tests were needed.

Sorry for the noise.

jtmoon79:
To understand how test cases have changed I created a branch `Issue660_whitespace_exact_try6_prechange_testsonly` that has added test cases https://github.com/jtmoon79/chrono/compare/Issue660_whitespace_exact_try6_prechange_testsonly..Issue660_whitespace_exact_try6
This comparison can help reviewers understand how behavior changed.

Admittedly, there is some feature creep in this PR (changed test function names). Hopefully that isn't an issue.

jtmoon79:
@djc @esheppa any opinion on this PR?

djc:
@jtmoon79 both @esheppa and I are working on maintaining this project as volunteers, so when you start pinging us after 1 day it feels entitled. We will review this when we have time, usually that doesn't take more than a few days.

jtmoon79:
> @jtmoon79 both @esheppa and I are working on maintaining this project as volunteers, so when you start pinging us after 1 day it feels entitled. We will review this when we have time, usually that doesn't take more than a few days.

Pardon me. Yes, per your convenience.

djc:
Okay, this is way too big to review in one go. It can be all in one PR but then I'd like the big commit split up in more, smaller commits. Multiple PRs would also be a fine approach. Here's what I'm thinking:

* One commit to add test cases that verify the current behavior (that will still pass after your changes)
* One commit to make the actual code changes and relevant document changes, and changes any tests as needed
* One commit to add further tests that verify the new behavior

Some further notes: please remove type annotations unless necessary for the compiler, don't use variable names like `s_c` (although `c` is fine if it is for a `char`) -- it seemed like you were using `_c` as a sort of Hungarian notation? Also for tests that test invalid input, please add a comment saying why the particular line of input is invalid.

src/datetime/tests.rs
```diff
@@ -205,19 +253,430 @@ fn test_datetime_from_str() {
...
+        "2015-2-18T13:16:9.15-10:00".parse::<DateTime<Utc>>(),
+        Ok(Utc.ymd(2015, 2, 18).and_hms_milli(23, 16, 9, 150))
+    );
+    assert!("2015-2-18T23:16:9.15".parse::<DateTime<Utc>>().is_err());
+
+    // no test for `DateTime<Local>`, we cannot verify that much.
+}
+
+#[test]
+fn test_utc_datetime_from_str_with_spaces() {
```
jtmoon79, line 381: test function devoted to just spacing variations

src/format/mod.rs
```diff
@@ -218,27 +218,27 @@ pub enum Fixed {
...
     TimezoneOffsetDoubleColon,
     /// Offset from the local time to UTC without minutes (`+09` or `-04` or `+00`).
     ///
-    /// In the parser, the colon can be omitted and/or surrounded with any amount of whitespace.
+    /// In the parser, the colon may be omitted,
     /// The offset is limited from `-24` to `+24`,
     /// which is the same as [`FixedOffset`](../offset/struct.FixedOffset.html)'s range.
     TimezoneOffsetTripleColon,
-    /// Offset from the local time to UTC (`+09:00` or `-04:00` or `Z`).
+    /// Offset from the local time to UTC (`+09:00` or `-0400` or `Z`).
```
jtmoon79, line 239: Update docstring to show allowed format variations.

jtmoon79:
> Okay, this is way too big to review in one go. It can be all in one PR but then I'd like the big commit split up in more, smaller commits.

This PR push is slightly different than your ask. It's currently 2 commits but I can revise into 3 commits if needed. Hopefully my reasoning is agreeable. 🙂

> I'd like the big commit split up in more, smaller commits.

I made two large commits for easier comparison:
- one commit that adds many test cases and no behavior changes
- one commit that adds behavior changes, tweaks test cases, adds few test cases

Having this in two commits makes "before/after behavior comparison" easier for humans to grok, IMO.

<br/>

I reduced timezone colon delimiters from arbitrary whitespace+colon to one or zero of
patterns `":"`, `" "`, `" :"`, `": "`, or `" : "`.
This seemed like a reasonable compromise of previous matching flexibility and desired matching exactness.

I dropped PR creep around function name changes and a few other things.

<br/>

> One commit to add test cases that verify the current behavior (that will still pass after your changes)

Commit fd946e8923feecfab5d2d4f1e64ca010d0be9988

This adds a substantial number of tests as well. Some may seem redundant. But they are to help clarify behavior changes that will occur in the next commit.

Since there were significant changes to `parse_internal`, I also tests for `Item`s like `Nanosecond3NoDot`.
e.g. for testing of `Nanosecond3NoDot` make sure test number lengths of 2 and 4 ("edges") and 6, 9 (other signifcant fraction lengths).
These seemingly unrelated test cases gave me a little more confidence that nothing broke.

<br/>

> One commit to make the actual code changes and relevant document changes, and changes any tests as needed

Commit 742fb955a935e2faf3ccf30e751bfb28e33cf839

Some tests that appear "added" were already existiing tests that were moved several lines.,
e.g. the test case `" 4 : 3 : 2.1 "` here:
```
--- a/src/naive/time/tests.rs
+++ b/src/naive/time/tests.rs
@@ -196,9 +196,6 @@ fn test_date_from_str() {
         "0:0:0",
         "0:0:0.0000000",
         "0:0:0.0000003",
-        " 4 : 3 : 2.1 ",
-        " 09:08:07 ",
-        " 9:8:07 ",
         "01:02:03",
         "4:3:2.1",
         "9:8:7",
@@ -266,6 +263,18 @@ fn test_date_from_str() {
         "1441497364.649",    // valid datetime, not a NaiveTime
         "+1441497364.649",   // valid datetime, not a NaiveTime
         "+1441497364",       // valid datetime, not a NaiveTime
+        "01 :02:03",         // space after hour
+        "01: 02:03",         // space before minute
+        "01 : 02:03",        // space around hour-minute divider
+        "01:02 :03",         // space after minute
+        "01:02: 03",         // space before second
+        "01:02 : 03",        // space around minute-second divider
+        "01:02:03 .456",     // space after second
+        "01:02:03. 456",     // space before fraction
+        "01:02:03 ",         // trailing space
+        "01:02:03.456 ",     // trailing space
+        " 01:02:03",         // leading space
+        " 4 : 3 : 2.1 ",     // spaces intermixed throughout
```

<br/>

> One commit to add further tests that verify the new behavior

I thought this to be unnecessary for my "A/B comparison" approach.

<br/>

> please remove type annotations unless necessary for the compiler

Removed.

<br/>

> don't use variable names like s_c (although c is fine if it is for a char) -- it seemed like you were using _c as a sort of Hungarian notation?\

Renamed unused variable `_c` to `_`.

There are some places using `c_` (a `char`). The `c_` is to help humans distinguish from outer scope variable `c` (a `char`).

<br/>

> Also for tests that test invalid input, please add a comment saying why the particular line of input is invalid.

I added comment explanations for many invalid input tests. However, I did not add a comment for all invalid input tests.
I was concerned that in some areas those comments would decrease grokability.
Happy to add more comments to the invalid test cases if that is needed.

<br/>

I ran `cargo fmt` at the project top-level directory.
This produced some unexpected alignments of trailing comments, e.g.
```
+        "2012 -12-12T12:12:12Z",                                     // space after year
+        "2012  -12-12T12:12:12Z",                                    // multi space after year
+        "2012- 12-12T12:12:12Z",                                     // space after year divider
+        "2012-  12-12T12:12:12Z", // multi space after year divider
+        "2012-12-12T 12:12:12Z",  // space after date-time divider
```
At first glance, this looks like a bug in `rustfmt`.

src/format/scan.rs
```diff
@@ -198,9 +198,67 @@ pub(super) fn space(s: &str) -> ParseResult<&str> {
...
+        Some(_) | None => s,
+    }
+}
+
+/// Allow a colon with possible one-character whitespace padding.
+/// Consumes zero or one of these leading patterns:
+/// `":"`, `" "`, `" :"`, `": "`, or `" : "`.
 pub(super) fn colon_or_space(s: &str) -> ParseResult<&str> {
-    Ok(s.trim_left_matches(|c: char| c == ':' || c.is_whitespace()))
+    let c0_ = match s.chars().next() {
```
djc, line 226: Okay, can we just get rid of the underscore suffixes? I don't see their value.

src/format/scan.rs
```diff
@@ -238,6 +296,16 @@ where
     };
     s = &s[1..];
 
+    // special check for `Z` to return more accurate error `INVALID`.
```
djc, line 299: This should be a separate commit.

src/format/scan.rs
```diff
@@ -198,9 +198,67 @@ pub(super) fn space(s: &str) -> ParseResult<&str> {
...
+/// If the first `char` is whitespace then consume it and return `s`.
+/// Else return `s`.
+pub(super) fn space1(s: &str) -> &str {
+    match s.chars().next() {
+        Some(c) if c.is_whitespace() => s_next(s),
+        Some(_) | None => s,
+    }
+}
+
+/// Allow a colon with possible one-character whitespace padding.
```
djc, line 222: I think this should be a separate commit.

src/format/strftime.rs
```diff
@@ -466,28 +466,45 @@ impl<'a> Iterator for StrftimeItems<'a> {
...
             Some(c) if c.is_whitespace() => {
-                // `%` is not a whitespace, so `c != '%'` is redundant
-                let nextspec = self
-                    .remainder
-                    .find(|c: char| !c.is_whitespace())
-                    .unwrap_or(self.remainder.len());
-                assert!(nextspec > 0);
-                let item = sp!(&self.remainder[..nextspec]);
-                self.remainder = &self.remainder[nextspec..];
+                let ws = self.remainder;
```
djc, line 471: Could this be a different commit? Please get rid of the `_` suffix.

djc (commented):
I think this is an improvement (keeping the first commit separate definitely helps) but I'd still like to see the changes that seem to be more or less independent split out into separate commits.

src/format/strftime.rs
jtmoon79: > Could this be a different commit?

I re-ran the old code and I could not elicit a test case that justified the new code. I'm pretty sure I added the code early-on in this PR endeavor to fix a multi-byte unicode character parsing problem. But later, I fixed something else such that this change was unnecessary.  So I reverted my changes.

I left the added test cases to have some assurance about my claim; multi-byte characters are correctly processed.

src/format/scan.rs
jtmoon79: In commit [62d5791](https://github.com/chronotope/chrono/pull/807/commits/62d57919d99d8d01750b8768ab4702468506b3c7)

src/format/scan.rs
jtmoon79: I decided to drop this. After adding more test cases around this, I realized it was somewhat tunnel-vision change (it was not as helpful as I thought).  The benefit is middling in exchange for PR feature creep.

src/format/scan.rs
jtmoon79: Fixed.

src/format/parse.rs
```diff
@@ -292,13 +296,37 @@ where
                 s = &s[prefix.len()..];
             }
 
-            Item::Space(_) => {
-                s = s.trim_left();
+            Item::Space(item_space) => {
+                for item_c in item_space.chars() {
```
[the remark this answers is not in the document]
jtmoon79: Changed to `expect` and `actual`.

jtmoon79:
Regarding [83e9ba0](https://github.com/chronotope/chrono/pull/807/commits/83e9ba0d3f8576651aa0db25d27ce7b6d2178a56), see failed job _rust_msrv (ubuntu-latest)_ https://github.com/chronotope/chrono/actions/runs/3278766675/jobs/5397533682

djc (approved):
@esheppa I think this is good to go, any concerns?

src/format/scan.rs
```diff
@@ -198,6 +198,27 @@ pub(super) fn space(s: &str) -> ParseResult<&str> {
...
+        None => {
+            // one or zero chars in `s`, return empty string
+            &s[s.len()..]
+        }
+    }
+}
+
+/// If the first `char` is whitespace then consume it and return `s`.
+/// Else return `s`.
+pub(super) fn space1(s: &str) -> &str {
```
esheppa, line 215: potentially this could be better named `trim1` - to contrast from the existing `space` function which is fallible and asserts the existence of at least one space

src/format/parse.rs
```diff
@@ -424,27 +451,24 @@ where
                     | &TimezoneOffsetDoubleColon
                     | &TimezoneOffsetTripleColon
                     | &TimezoneOffset => {
-                        let offset = try_consume!(scan::timezone_offset(
-                            s.trim_left(),
-                            scan::colon_or_space
-                        ));
+                        s = scan::space1(s);
```
esheppa, line 454: why do we allow one optional whitespace here?

src/format/parse.rs
```diff
@@ -292,13 +296,37 @@ where
                 s = &s[prefix.len()..];
             }
 
-            Item::Space(_) => {
-                s = s.trim_left();
+            Item::Space(item_space) => {
```
esheppa, line 299: The implementations here now look very similar to the `Item::*Literal` ones - what are your thoughts on the following code? We could perhaps even deprecate `Item::*Space` and just always use literal? Maybe we need a different item to allow for optional whitespace/literals.

```rust
            Item::Literal(prefix) | Item::Space(prefix) => {
                if s.len() < prefix.len() {
                    return Err((s, TOO_SHORT));
                }
                if !s.starts_with(prefix) {
                    return Err((s, INVALID));
                }
                s = &s[prefix.len()..];
            }

            #[cfg(any(feature = "alloc", feature = "std", test))]
            Item::OwnedLiteral(ref prefix) | Item::OwnedSpace(ref prefix) => {
                if s.len() < prefix.len() {
                    return Err((s, TOO_SHORT));
                }
                if !s.starts_with(&prefix[..]) {
                    return Err((s, INVALID));
                }
                s = &s[prefix.len()..];
            }
```

src/format/parse.rs
```diff
@@ -424,27 +451,24 @@ where
...
                     }
 
                     &TimezoneOffsetColonZ | &TimezoneOffsetZ => {
-                        let offset = try_consume!(scan::timezone_offset_zulu(
-                            s.trim_left(),
-                            scan::colon_or_space
-                        ));
+                        s = scan::space1(s);
+                        let offset =
+                            try_consume!(scan::timezone_offset_zulu(s, scan::colon_or_space));
```
esheppa, line 462: should we (here or in another PR) consider being explicit about all the parts of the offset (eg offset hours, offset minutes)? As far as I can see in the strftime docs we don't mention being permissive with space and/or colons in `%z`, `%:z` or `%#z`

src/format/scan.rs
```diff
@@ -219,9 +219,48 @@ pub(super) fn space1(s: &str) -> &str {
     }
 }
 
-/// Consumes any number (including zero) of colon or spaces.
+/// Allow a colon with possible one-character whitespace padding.
+/// Consumes zero or one of these leading patterns:
+/// `":"`, `" "`, `" :"`, `": "`, or `" : "`.
 pub(super) fn colon_or_space(s: &str) -> ParseResult<&str> {
```
esheppa, line 225: This functions return type is `ParseResult` but as far as I can see no code paths can return `Err`. We should either implement some `Err` code paths or have the function return just `&str` instead

src/format/scan.rs
```diff
@@ -219,9 +219,48 @@ pub(super) fn space1(s: &str) -> &str {
...
-    Ok(s.trim_left_matches(|c: char| c == ':' || c.is_whitespace()))
+    let c0s = match s.chars().next() {
```
esheppa, line 226: Potentially the following could be done a bit more simply? I've included one possibility below. Also I've included some commented error paths where we could potentially be a bit more specific, eg allowing only the specified patterns. If we prefer to keep it always returning ok, the function could potentially be renamed to something like `maybe_colon_or_space`

```rust
if s.is_empty() {
        // return Err(TOO_SHORT)
        return Ok(s)
    }

    fn next_char_is_whitespace(input: &str) -> bool {
        input.chars().next().map(|c| c.is_whitespace()).unwrap_or_default()
    }

    if s.starts_with(':') {
        s = s_next(s);
        if next_char_is_whitespace(s) {
            s = s_next(s);
        }
        return Ok(s)
    }
    
    
    if next_char_is_whitespace(s) {
        s = s_next(s);
        if s.starts_with(':') {
            s = s_next(s);
        } else {
            return Ok(s)
        }
        if next_char_is_whitespace(s) {
            s = s_next(s);
        }
        return Ok(s)
    }

    // Err(INVALID) 
    Ok(s)
```

esheppa (changes requested):
Thanks for this @jtmoon79 - I've included a few comments for discussion, please let me know your thoughts

src/format/scan.rs
jtmoon79: I'll rename this.

src/format/parse.rs
jtmoon79: Previously, infinite whitespace was allowed. So strftime specifier `%Y/%m/%d %H:%M:%S%z` would match values `"2022/01/31 00:01:59-0700"`, and  `"2022/01/31 00:01:59 -0700"`, and `"2022/01/31 00:01:59          \n\t\n -0700"`.

I guessed a significant portion of chrono users would be affected. So I erred away from too many breaking changes.
Allowing one whitespace seemed like a reasonable compromise , e.g. `"2022/01/31 00:01:59-0700"` and `"2022/01/31 00:01:59 -0700"` will still match the specifier string.

If you would prefer to be strict about this, e.g. only `"2022/01/31 00:01:59-0700"` would match the specifier string, then that would be fine with me. My personal preference is to be strict because that simplifies reasoning about correctness.

src/format/parse.rs
jtmoon79: I noticed the same. Seems like a good change to me. However, I decided that this should be put into a different PR so this PR is not overloaded.
If you'd like, I could open a _Code Improvement_ Issue for this.

src/format/parse.rs
jtmoon79: Yes, definitely. It would help users avoid unwanted surprises.
How about I first create a tracking Issue around this?

src/format/scan.rs
jtmoon79: The `colon_or_space` is passed to `src/format/scan.rs:timezone_offset` functions.

https://github.com/chronotope/chrono/blob/v0.4.23/src/format/scan.rs#L219
Those functions require passing an object with trait
`FnMut(&str) -> ParseResult<&str>`.
It would touch too much code for this PR.

After a brief code review, I think `colon_or_space` should remain as is. Not quite as perfect as it could be but good enough IMO

src/format/scan.rs
jtmoon79: 🤔

...

Yes, this fits with this PR, and would eliminate a few extra code paths. I will include this in the next update.

(I'm somewhat contradicting my prior comment https://github.com/chronotope/chrono/pull/807#discussion_r1021114078)

jtmoon79: This is in the latest PR push. https://github.com/chronotope/chrono/pull/807/commits/1e9f90bdf827699cef7dd6da41c37f3b95f9cdcd

jtmoon79:
Latest PR push also includes a few extra tests in `test_rfc3339`
https://github.com/chronotope/chrono/pull/807/files#diff-abd7a03adfeaa1c9ca4623371f789cf48c5d794d82457a4068ad5be00fe7c784R1458-R1492

jtmoon79:
FYI the last PR update was a rebase to `main`.

jtmoon79:
@esheppa don't mean to pester, just wondering if you saw this updated PR?  I know you both maintain this project on your own time, so whenever.

djc:
@esheppa ping, would you mind doing another round of review on this?

esheppa:
Will get this reviewed tomorrow

src/format/parse.rs
esheppa: I think that given this will be a breaking change anyway, lets be explicit about all white space rather than being relaxed in some areas and explicit in others. 

Even in the case that we do decide to leave this optional white space in, we could probably leave a comment justifying why

src/format/parse.rs
esheppa: Excellent, I like the idea of this being in a separate PR

esheppa (approved):

esheppa:
Thanks for your patience @jtmoon79. This is looking great. I think it makes sense to have any further changes done in subsequent PR's rather than this one, so I'm happy for this to go ahead and be merged. It looks like this will need a rebase before being able to be merged

jtmoon79:
This rebase was tedious and it looks like I messed up some stuff along the way.  I'll be working on this some more when I can.

jtmoon79:
I think this is ready to merge.

I'm a little concerned about my decision to allow some flexibility in the timezone offset matching, e.g. patterns `00:00`, `00 :00`, `00: 00`, `00 : 00` match (https://github.com/chronotope/chrono/commit/5bd127812226a51f3430b982cee54e8bb6eecfd2). Perhaps adding this little corner of wiggle room contradicts the purpose of this large change, and so could be a little confusing. I can rework this PR to simply be strict about the pattern, e.g. `00:00` only.

On the other hand, maybe just getting this PR merged would be better than putting off this big change any longer?

jtmoon79:
On a different topic, would you like help reviewing Issues and PRs for this project?  It's an important project and it seems like some assistance would help.  Since I have some familiarity now, I might be of some use.
I could check-in on occasion and give (non-authoritative) feedback to PRs and Issues.

djc:
I agree it probably makes sense to leave the time whitespace as is for this PR, maybe revisit that in another one?

Would it be feasible to squash the current commit history down to something more or less equivalent to the original three commits, or some other sequence that doesn't have as much back and forth?

Your help with maintaining this project would be much appreciated!

jtmoon79:
> Would it be feasible to squash the current commit history down

Squashed.

> Your help with maintaining this project would be much appreciated!

I'll subscribe to the github alerts. I'll comment on PRs where I can.

PR #809, opened 2022-09-01
move shell commands to yml

esheppa:


djc:
Would you mind splitting the first commit into pieces, maybe one part per job or something? I think the move of the date command tests should be in a separate commit as well. Otherwise this gets too hard to review.

esheppa:
There are still a few things I need to do here (some of which may be better in a separate PR)
* Consider whether we need ubuntu-20.04 - I've currently removed this but it was probably there for a reason. 
* Consider whether we need to run the `no_std`/`alternative-targets`/`cross-targets` elsewhere than linux (previous was running also on macOS)
* Re-implement the wasm simple test (using wasm-pack, or perhaps trunk).
* Run on android (I've added a build for this but could be nice to run the tests as well)
* Potentially consolidate the `cross-targets` and `no-std` blocks, 
* Potentially move android and iOS to the `cross-targets` block
* Potentially add some more common targets

djc:
This is looking great! So the one thing I would probably do different is to remove stuff from the shell scripts in the same commit as it is added/moved in the CI config, so it's easier to review how coverage changes. Do you think that would be hard to do?

esheppa:
Yep that sounds like a good move, and makes it one less commit as well.

esheppa:
potentially this should target `0.4.x` instead of `main`, and then it can later be merged to `main` - this would be nice as we can avoid maintaining two CI setups

djc:
Yes, merging this into 0.4.x first makes sense to me.

esheppa:
This has been merged via #814 instead

PR #811, opened 2022-09-06
Add niche optimization for `NaiveTime`

Kijewski:
`NaiveTime` contains two integer fields, both with a restricted range:

- `secs`, the second of the day, is less than 86,400.
- `frac` is used for two things. To store the nanosecond of the second,
  and a leap second, so its highest value is 1,999,999,999.

So, in order to allow niche optimization, one or both values could be
implemented in a type that does not allow higher values. Because `frac`
is stored later in the struct, it has the better option to yield
optimized code, e.g. when used with `Result<NaiveTime, E>`. The compiler
has a longer consecutive sequence of bytes to store non-NaiveTime its
data.

### DIFF
--- a/src/naive/time/mod.rs
+++ b/src/naive/time/mod.rs
@@ -12,6 +12,7 @@ use num_integer::div_mod_floor;
 #[cfg(feature = "rkyv")]
 use rkyv::{Archive, Deserialize, Serialize};
 
+use self::u31::U31;
 #[cfg(any(feature = "alloc", feature = "std", test))]
 use crate::format::DelayedFormat;
 use crate::format::{parse, ParseError, ParseResult, Parsed, StrftimeItems};
@@ -23,6 +24,8 @@ mod serde;
 
 #[cfg(test)]
 mod tests;
+
+mod u31;
 
 /// ISO 8601 time without timezone.
 /// Allows for the nanosecond precision and optional leap second representation.
@@ -187,7 +190,7 @@ mod tests;
 #[cfg_attr(feature = "rkyv", derive(Archive, Deserialize, Serialize))]
 pub struct NaiveTime {
     secs: u32,
-    frac: u32,
+    frac: U31,
 }
 
 impl NaiveTime {
@@ -392,7 +395,8 @@ impl NaiveTime {
             return None;
         }
         let secs = hour * 3600 + min * 60 + sec;
-        Some(NaiveTime { secs, frac: nano })
+        let frac = unsafe { U31::new_unchecked(nano) };
+        Some(NaiveTime { secs, frac })
     }
 
     /// Makes a new `NaiveTime` from the number of seconds since midnight and nanosecond.
@@ -443,7 +447,8 @@ impl NaiveTime {
         if secs >= 86_400 || nano >= 2_000_000_000 {
             return None;
         }
-        Some(NaiveTime { secs, frac: nano })
+        let frac = unsafe { U31::new_unchecked(nano) };
+        Some(NaiveTime { secs, frac })
     }
 
     /// Parses a string with the specified format string and returns a new `NaiveTime`.
@@ -534,7 +539,7 @@ impl NaiveTime {
     /// ```
     pub fn overflowing_add_signed(&self, mut rhs: TimeDelta) -> (NaiveTime, i64) {
         let mut secs = self.secs;
-        let mut frac = self.frac;
+        let mut frac = self.frac.get();
 
         // check if `self` is a leap second and adding `rhs` would escape that leap second.
         // if it's the case, update `self` and `rhs` to involve no leap second;
@@ -551,6 +556,7 @@ impl NaiveTime {
             } else {
                 frac = (i64::from(frac) + rhs.num_nanoseconds().unwrap()) as u32;
                 debug_assert!(frac < 2_000_000_000);
+                let frac = unsafe { U31::new_unchecked(frac) };
                 return (NaiveTime { secs, frac }, 0);
             }
         }
@@ -592,7 +598,8 @@ impl NaiveTime {
         }
         debug_assert!((0..86_400).contains(&secs));
 
-        (NaiveTime { secs: secs as u32, frac: frac as u32 }, morerhssecs)
+        let frac = unsafe { U31::new_unchecked(frac as u32) };
+        (NaiveTime { secs: secs as u32, frac }, morerhssecs)
     }
 
     /// Subtracts given `TimeDelta` from the current time,
@@ -688,12 +695,12 @@ impl NaiveTime {
         use core::cmp::Ordering;
 
         let secs = i64::from(self.secs) - i64::from(rhs.secs);
-        let frac = i64::from(self.frac) - i64::from(rhs.frac);
+        let frac = i64::from(self.frac.get()) - i64::from(rhs.frac.get());
 
         // `secs` may contain a leap second yet to be counted
         let adjust = match self.secs.cmp(&rhs.secs) {
             Ordering::Greater => {
-                if rhs.frac >= 1_000_000_000 {
+                if rhs.frac.get() >= 1_000_000_000 {
                     1
                 } else {
                     0
@@ -701,7 +708,7 @@ impl NaiveTime {
             }
             Ordering::Equal => 0,
             Ordering::Less => {
-                if self.frac >= 1_000_000_000 {
+                if self.frac.get() >= 1_000_000_000 {
                     -1
                 } else {
                     0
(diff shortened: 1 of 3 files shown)


src/naive/time/u31.rs
```diff
@@ -0,0 +1,287 @@
...
+    fn default() -> Self {
+        Self::ZERO
+    }
+}
+
+#[repr(C)]
+#[cfg(target_endian = "little")]
+#[derive(Copy, Clone)]
+struct Buf {
+    align: [u32; 0],
```
djc, line 83: Is this trick documented anywhere? I'd like this to be commented with a (more or less) authoritative citation on how this behaves.

src/naive/time/u31.rs
```diff
@@ -0,0 +1,287 @@
+use core::{cmp, fmt, hash, mem};
+
+/// A 31-bit unsigned integer
+#[allow(unreachable_pub)] // public through `rkyv::Archive<NaiveTime>`
+#[repr(transparent)]
```
djc, line 5: I suppose this is necessary in order for the compiler to leverage the niche? What does it mean for semver compatibility going forward?

src/naive/time/mod.rs
```diff
@@ -551,6 +556,7 @@ impl NaiveTime {
             } else {
                 frac = (i64::from(frac) + rhs.num_nanoseconds().unwrap()) as u32;
                 debug_assert!(frac < 2_000_000_000);
+                let frac = unsafe { U31::new_unchecked(frac) };
```
djc, line 559: I'd like some comments here that clearly document how/where the invariant is upheld in release builds.

djc (commented):
Thank you for working on this, interesting approach!

> So, in order to allow niche optimization, one or both values could be implemented in a type that does not allow higher values. Because `frac` is stored later in the struct, it has the better option to yield optimized code, e.g. when used with `Result<NaiveTime, E>`. The compiler has a longer consecutive sequence of bytes to store non-NaiveTime its data.

If we're going to do this, why not take the 15-bit niche from `secs` instead of 1 bit from `frac`? How have you tested this? Do you have an actual use case where this optimization has a substantial impact?

djc:
Also, why not go with the ux crate or the arbitrary-int crate?

(IMO the awesome version of this is a crate that we can use at test time to generate code into our `src` directory specifically for the types that we need, without adding a compile-time dependency for chrono. This would allow the upstream maintainers to improve the code over time without us having to take responsibility for a particular implementation.)

src/naive/time/u31.rs
Kijewski: It's not a public type, so the user should not need to care how U31 is implemented. Unless a user does very odd stuff like accessing an `Option<NaiveTime>` in C, they should not be bothered by this change.

Actually, I think I don't need `struct U31(Buf)` anyhow. I could just let U31 be the endianness-guarded struct directly.

Kijewski:
Neither crate uses niche optimization. I have an open PR for ux to implement it there: https://github.com/kjetilkjeka/uX/pull/49.

My system is far from crappy, but it takes me 17 seconds to compile ux in release mode. I don't think this additional compile time would appropriate to add to chrono.

src/naive/time/mod.rs
Kijewski: I assumed chrono is well enough tested, so I can assume the invariant to hold. :dagger: 

If I read the code correctly it is tested that `frac < 1_000_000_000` and `rhs.num_nanoseconds()` cannot return a value >= 1e9. Is the second assumption true? Then I this as a `// SAFETY` comment.

src/naive/time/u31.rs
Kijewski: > Slices have the same layout as the section of the array they slice. -- [Type layout # Slice layout](https://doc.rust-lang.org/reference/type-layout.html#slice-layout)

This is the same for 0 sized slices.

Kijewski:
> If we're going to do this, why not take the 15-bit niche from secs instead of 1 bit from frac?

In my current implementation, if the last bit of the struct is set, then the whole thing it not a `NaiveTime`. So sec can be every value anyway.

NaiveTime is used as the last member in NaiveDateTime, and DateTime, so it better to have the one bit at the very end of the structs, then somewhere in the middle. Otherwise you make the usable byte sequence a little bit smaller.

djc:
> My system is far from crappy, but it takes me 17 seconds to compile ux in release mode. I don't think this additional compile time would appropriate to add to chrono.

Fair point!

> > If we're going to do this, why not take the 15-bit niche from secs instead of 1 bit from frac?
> 
> In my current implementation, if the last bit of the struct is set, then the whole thing it not a `NaiveTime`. So sec can be every value anyway.
> 
> NaiveTime is used as the last member in NaiveDateTime, and DateTime, so it better to have the one bit at the very end of the structs, then somewhere in the middle. Otherwise you make the usable byte sequence a little bit smaller.

The compiler is allowed to reorder fields at will, right? Why wouldn't it do so here? And if not, we can trivially reorder `secs` and `frac` within `NaiveTime`.

esheppa:
Sorry to jump in kind of late here, but could we use something like: [nonmax](https://docs.rs/nonmax/latest/nonmax/index.html) to hold the nanos?

Kijewski:
nonmax generates terrible byte code, because it has to negate the value on every access. Most likely twice: once for reading, once for writing. This is enough to make the optimizer not understand what is going on.

Otherwise, how often will you access the data? It is obviously great that nonmax does not need unsafe operations to access the data. Nonmax could be "good enough".

esheppa:
Is it worth us benchmarking both methods so that we have some empirical evidence to inform the direction we choose?

djc:
In general, these questions haven't been answered yet:

> How have you tested this? Do you have an actual use case where this optimization has a substantial impact?

Kijewski:
> Is it worth us benchmarking both methods so that we have some empirical evidence to inform the direction we choose?

Old values use nonmax, new values use U31:

```text
bench_datetime_parse_from_rfc2822                                                                            
                        time:   [196.87 ns 197.23 ns 197.60 ns]
                        change: [+5.3629% +5.6315% +5.8895%] (p = 0.00 < 0.05)
                        Performance has regressed.

bench_datetime_parse_from_rfc3339                                                                            
                        time:   [163.58 ns 164.23 ns 164.98 ns]
                        change: [+6.6111% +7.8651% +9.1036%] (p = 0.00 < 0.05)
                        Performance has regressed.

bench_datetime_from_str time:   [245.06 ns 246.12 ns 247.37 ns]                                    
                        change: [-4.6067% -3.6377% -2.7998%] (p = 0.00 < 0.05)
                        Performance has improved.

bench_datetime_to_rfc2822                                                                             
                        time:   [565.76 ns 568.01 ns 570.94 ns]
                        change: [-2.2735% -1.8146% -1.3809%] (p = 0.00 < 0.05)
                        Performance has improved.

bench_ser_naivedatetime_writer                                                                            
                        time:   [285.23 ns 287.71 ns 290.49 ns]
                        change: [-4.5103% -3.4291% -2.3967%] (p = 0.00 < 0.05)
                        Performance has improved.

bench_ser_naivedatetime_string                                                                            
                        time:   [301.43 ns 302.13 ns 302.88 ns]
                        change: [-6.1510% -5.6113% -5.0359%] (p = 0.00 < 0.05)
                        Performance has improved.
```

The other benches were within noise threshold. So, `bench_datetime_parse_from_rfc2822` & `bench_datetime_parse_from_rfc3339` are significantly faster with nonmax, and `bench_ser_naivedatetime_writer` & `bench_ser_naivedatetime_string` are faster with U31. I ran both benchmarks twice, so there's a chance that this results are actually meaningful.

This is a strong indication to throw this PR away and go with nonmax instead. Otherwise, there seem to be no benchmarks for arithmetic operations or am I overlooking them?

Kijewski:
> How have you tested this?

Whether the optimization works at all you mean? https://github.com/chronotope/chrono/pull/811/commits/fb79395dfbba15f1d51bb47b4934fab5f9613643

> Do you have an actual use case where this optimization has a substantial impact?

Any API that returns a fallible `DateTime`, or stores an `Option<DateTime>`. In my case I have the latter a lot.

esheppa:
Interestingly if this works nicely with `nonmax`, potentially it could be used to replace the `i32` of the `DateImpl` as well, allowing `Option<NaiveDate>` to be the same size as `NaiveDate`

src/naive/time/u31.rs
x-hgg-x: For information, you can also use [`#[repr(align(4))]`](https://doc.rust-lang.org/reference/type-layout.html#representations) to do the same thing (from [RFC 1358](https://github.com/rust-lang/rfcs/blob/master/text/1358-repr-align.md)).

Kijewski: For some platforms the alignment of u32 might be 16bits. msp430-none-elf is a 16bit, [tier 3](https://doc.rust-lang.org/nightly/rustc/platform-support.html) platform.

PR #812, opened 2022-09-06
Expose functions needed for pyo3

pickfire:
Companion pull request for https://github.com/PyO3/pyo3/pull/2604

Changes to API
- Make some functions public but doc hidden
  - `Mdf`
  - `TimeDelta` fields (`secs`, `nanos`) - standard library `Duration` already provides something similar as `as_secs` and `as_nanos`
- Move duration min max into `TimeDelta::MIN` and `TimeDelta::MAX` as public function similar to standard library `Duration`

Thanks to @Psykopear for helping out with the pull request and help pushing this forward.

cc @Psykopear @davidhewitt @djc

This is still a draft and can be discussed, @Psykopear made a version that works on stable chrono without exposing any internal details https://github.com/pickfire/pyo3/pull/1

### DIFF
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -995,7 +995,8 @@ impl NaiveDate {
 
     /// Returns the packed month-day-flags.
     #[inline]
-    fn mdf(&self) -> Mdf {
+    #[doc(hidden)]
+    pub fn mdf(&self) -> Mdf {
         self.of().to_mdf()
     }
 
--- a/src/naive/internals.rs
+++ b/src/naive/internals.rs
@@ -372,8 +372,9 @@ impl fmt::Debug for Of {
 /// The whole bits except for the least 3 bits are referred as `Mdl`
 /// (month, day of month and leap flag),
 /// which is an index to the `MDL_TO_OL` lookup table.
+#[doc(hidden)]
 #[derive(PartialEq, PartialOrd, Copy, Clone)]
-pub(super) struct Mdf(pub(super) u32);
+pub struct Mdf(pub(super) u32);
 
 impl Mdf {
     #[inline]
@@ -420,8 +421,10 @@ impl Mdf {
         }
     }
 
+    // Exported for pyo3.
+    #[doc(hidden)]
     #[inline]
-    pub(super) fn month(&self) -> u32 {
+    pub fn month(&self) -> u32 {
         let Mdf(mdf) = *self;
         mdf >> 9
     }
@@ -433,8 +436,10 @@ impl Mdf {
         Mdf((mdf & 0b1_1111_1111) | (month << 9))
     }
 
+    // Exported for chrono.
+    #[doc(hidden)]
     #[inline]
-    pub(super) fn day(&self) -> u32 {
+    pub fn day(&self) -> u32 {
         let Mdf(mdf) = *self;
         (mdf >> 4) & 0b1_1111
     }
--- a/src/naive/time/mod.rs
+++ b/src/naive/time/mod.rs
@@ -792,7 +792,8 @@ impl NaiveTime {
     }
 
     /// Returns a triple of the hour, minute and second numbers.
-    fn hms(&self) -> (u32, u32, u32) {
+    #[doc(hidden)]
+    pub fn hms(&self) -> (u32, u32, u32) {
         let (mins, sec) = div_mod_floor(self.secs, 60);
         let (hour, min) = div_mod_floor(mins, 60);
         (hour, min, sec)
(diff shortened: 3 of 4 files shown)


davidhewitt:
Please note that I would _not_ merge a PR to PyO3 which used `#[doc(hidden)]` APIs from `chrono`.

I think it is fair for library developers to use `#[doc(hidden)]` APIs as implementation detail exempt from semver, so coupling PyO3 to hidden APIs is unfair for the `chrono` maintainers as well as produces a poor UX if refactoring ever does break something.

src/naive/date.rs
```diff
@@ -995,7 +995,8 @@ impl NaiveDate {
 
     /// Returns the packed month-day-flags.
     #[inline]
-    fn mdf(&self) -> Mdf {
+    #[doc(hidden)]
+    pub fn mdf(&self) -> Mdf {
```
djc, line 999: What problem does this solve? What API do you need from the `Mdf`, and can we support that from `NaiveDate` directly?

src/naive/internals.rs
```diff
@@ -420,8 +421,10 @@ impl Mdf {
         }
     }
 
+    // Exported for pyo3.
+    #[doc(hidden)]
     #[inline]
-    pub(super) fn month(&self) -> u32 {
+    pub fn month(&self) -> u32 {
```
djc, line 427: For this, how about we add a method on `NaiveDate` that yields a `Month` and add a method on `Month` that gives you a `u8`?

src/naive/internals.rs
```diff
@@ -433,8 +436,10 @@ impl Mdf {
         Mdf((mdf & 0b1_1111_1111) | (month << 9))
     }
 
+    // Exported for chrono.
+    #[doc(hidden)]
     #[inline]
-    pub(super) fn day(&self) -> u32 {
+    pub fn day(&self) -> u32 {
```
djc, line 442: Similarly, we can just expose this on `NaiveDate` directly somehow.

src/naive/time/mod.rs
```diff
@@ -792,7 +792,8 @@ impl NaiveTime {
     }
 
     /// Returns a triple of the hour, minute and second numbers.
-    fn hms(&self) -> (u32, u32, u32) {
+    #[doc(hidden)]
+    pub fn hms(&self) -> (u32, u32, u32) {
```
djc, line 796: I'm open to making this a part of the public API, though I think we should probably yield `u8` instead of `u32` (even if this is different from other methods right now).

djc (commented):
I think this should target the 0.4.x branch, doing it in a semver-compatible way for now, and we can later see how the API evolves in 0.5. However, this constrains the options for `Duration` (`TimeDelta` on `main`) since this can be the time 0.1 version of the type which we don't control the public API for.

> Please note that I would _not_ merge a PR to PyO3 which used `#[doc(hidden)]` APIs from `chrono`.

I agree this probably isn't a good way to control an API that's public across project boundaries.

esheppa:
> I think this should target the 0.4.x branch, doing it in a semver-compatible way for now, and we can later see how the API evolves in 0.5

It might potentially be easier to just target `0.5` for this - as if we end up merging a switch to `core::time::Duration` then that will negate the need for much of the changes to`TimeDelta` in this PR.

src/naive/date.rs
pickfire: This is to allow getting the day, month and year directly, but if we run the day, month and year then internally chrono have to get it from mdf multiple times, maybe the compiler can optimize it?

src/naive/internals.rs
pickfire: `NaiveDate` already exposes a month function that returns u32.
```rust
    #[inline]
    fn year(&self) -> i32 {
        self.ymdf >> 13
    }
    #[inline]
    fn month(&self) -> u32 {
        self.mdf().month()
    }
    #[inline]
    fn day(&self) -> u32 {
        self.mdf().day()
    }
```

pickfire:
> It might potentially be easier to just target 0.5 for this - as if we end up merging a switch to core::time::Duration then that will negate the need for much of the changes toTimeDelta in this PR.

So `TimeDelta` is not going to stay in the 0.5 branch? Yes, the standard library `Duration` exposes quite a lot of stuff that is public so we can just use it.

src/naive/date.rs
djc: This seems to be a micro-optimization then? Until we have a benchmark showing that this has a substantial performance impact, let's skip it.

djc:
`TimeDelta` is going to stay on the 0.5 branch, but it doesn't exist on the 0.4.x branch. And, we'll likely start removing features from `TimeDelta` on the 0.5 branch, it's intended to become an output only type from chrono's perspective.

pickfire:
Thanks for the information. I will be closing this pull request for https://github.com/PyO3/pyo3/pull/2612

PR #813, opened 2022-09-08
Support AIX's TZDB location

ecnelises:
This is to support AIX operating system. Since it's not been official recognized target of rust, this PR is in draft status.

### DIFF
--- a/src/offset/local/unix.rs
+++ b/src/offset/local/unix.rs
@@ -99,8 +99,11 @@ struct Cache {
 #[cfg(target_os = "android")]
 const TZDB_LOCATION: &str = " /system/usr/share/zoneinfo";
 
+#[cfg(target_os = "aix")]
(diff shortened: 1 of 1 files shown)


djc (approved):
I'm happy to merge this independent of whether it's an official target given the minimal changes (and it seems unlikely the OS identifier would change).

esheppa (approved):

djc:
(BTW, you may also want to submit a PR for this to the 0.4.x branch.)

esheppa:
Thanks for this @ecnelises. By any chance do you know how the current timezone is specified on AIX? It looks like one option is the `TZ` environment variable, but there may be others, however I've not been able to find any documentation. This would be useful as we could PR this to the `iana-time-zone` library which we depend on

Kijewski:
@esheppa, OP is already working on a PR for iana-time-zone: https://github.com/strawlab/iana-time-zone/pull/57. I linked some man pages in my first comment.

PR #814, opened 2022-09-08
move shell commands to yml (v2)

esheppa:
This is a version of https://github.com/chronotope/chrono/pull/809 that targets `0.4.x`


.github/workflows/test.yml
```diff
@@ -11,58 +11,57 @@ on:
...
+        rust_version: ["1.38.0", "stable", "beta", "nightly"]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ matrix.rust_version }}
-          override: true
+      - uses: Swatinem/rust-cache@v1
+      - run: cargo test --lib --features serde --color=always -- --color=always
```
djc, line 51: Why does this separate `--lib` and `--doc`? Why do we include serde here, but not other features? I think we should ideally test `--all-features` across the Rust versions, or do we have features that don't work on 1.38?

djc (commented):
This is great, thanks again for working on this! Some minor nits:

.github/workflows/test.yml
esheppa: I've separated `--lib` and `--doc` to avoid the integration tests, specifically the long-running test against the `date` command (that is run in some other specific cases) - however it would probably be cleaner to use `--all-features` here.

It looks like the `rkyv` feature doesn't work on `1.38`, but this can be solved by not using the default features.

esheppa:
so it looks like rykv doesn't work on 1.38.0 even with default features set to false. Potentially this should be documented, but I'll do in a separate commit

djc (approved):

djc:
This is great, awesome work!

Issue #815, opened 2022-09-10
Consider refactoring chrono to avoid systematic misuse of panics


udoprog:
Many constructors in `chrono` has the behavior to panic on invalid inputs. This issue proposes to systematically change this behavior to instead make use of fallible constructors which in effect means the removal of the `*_opt` family of functions.

I'll be trying to make an argument for why below, at the end are the concrete changes I'm proposing.

#### Why should we do this?

The use of `chrono` has been the culprit of many unwanted panics in code that I've personally written in my career as a Rust programmer. The reason why this happens is sloppiness from my part. I use a method without realizing that it panics even though it says so in the documentation. But after consideration I've come to believe that the design of these methods could be improved.

The standard I abide by when it comes to panicking is broadly that [correct Rust programs should not panic](https://blog.burntsushi.net/unwrap/#so-one-should-never-panic). So what I'll try and determine next is if uses of these constructors contributes to writing correct Rust programs that do or do not panic.

There are a few tests I'll be applying here each time one of these methods are used:
1. Does the caller check the necessary preconditions before calling the function?
2. Does the caller control the input to the function so that it's runtime invariant?
3. Is the function call in a context where it's OK to panic. Such as a test case or a short lived CLI program.

#### Preconditions

If we look at the doc for most of these functions they contain wording such as [`TimeZone::ymd`](https://docs.rs/chrono/latest/chrono/offset/trait.TimeZone.html#method.ymd):
> Panics on the out-of-range date, invalid month and/or day.

This is a difficult precondition to check. An out-of-range date isn't a mechanical test such as "is this value smaller than 1000". It requires a deep understanding of the gregorian calendar system which `chrono` itself is responsible for providing. It's implausible that uses ensure that its input passes the required condition due to complexity. So the first test probably fails systematically across many projects.

#### Runtime invariant inputs or tests

It's [easy to find many uses of these functions](https://github.com/search?q=Utc%3A%3Atimestamp_millis&type=Code) where the input is runtime variant and not tests. I've found that many of these relate to ad-hoc use in combination with `serde`, which is a strong indication that its arguments truly are runtime variant. A bit of taxonomy lead me to believe that there might be a case of systematic misuse.

#### What to do?

I believe many of the examples one can find of `chrono` production use fails the tests I proposed. The way to correct these uses would be through the `*_opt` family of methods. But since the non-opt methods seems to be considered *the default* and probably leads to systematic problems when used I think it's worth considering changing them.

[...]

PR #817, opened 2022-09-11
Make most constructors fallible (relates #815)

udoprog:
*most of this message is copied from [this comment](https://github.com/chronotope/chrono/issues/815#issuecomment-1242821829)*

This is just a suggestion for changes to accomplish what's being proposed in #815. Consider it heavily WIP and any naming is preliminary.

I've opted for a single `ChronoError` type which has a private `kind` field that determines the message presented to the user (but is not public). I initially started with one error type per kind of error, but eventually noticed that it would make propagating errors overly hard like how constructing a DateTime either raises an error because the date or the time was invalid.

I've opted to make loading the local timezone fallible instead of panicking. I think this is motivated because technically the underlying platform [might perform checks which might fail](https://docs.microsoft.com/en-us/windows/win32/api/timezoneapi/nf-timezoneapi-systemtimetofiletime#return-value) all though what these look like exactly is a bit unclear. These kind of errors are represented through `ChronoErrorKind::SystemError(std::io::Error)`.
This has the side effect of making a number of TimeZone methods fallible - which is unfortunate because a number of timezones can easily be interacted with without erroring (`Utc` and `FixedOffset`). To mitigate this I've introduced a separate `FixedTimeZone` trait that extends the TimeZone trait which has a number of non-fallible methods. Note that this could also in theory have a suite of methods which do not produce `LocalResult` values, because the time zone used is never ambiguous.

`FixedOffset::{east, west}` now returns an error if the provided values is outside of a 24-hour offset. This might legitimately be a case that we are OK with panicking for instead.

Making things fallible has the effect of clearly marking the standard trait implementations which might panic, these are:
* The `From<T>` implementations which interacts with the Local timezone, because this doesn't implement FixedTimeZone. Where appropriate these have been moved to be `TryFrom<Error = ChronoError> implementations.
* As mentioned above Add and Sub.

The last big change is that `LocalResult::None` has been removed in favor of handling the absent case as an error in the `unix` implementation. This has the nice side effect that `LocalResult::earlest` and `LocalResult::latest` are neither fallible nor needs to return an `Option<T>`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,27 +33,32 @@ which Chrono builds upon and should acknowledge:
 
 ## Limitations
 
-Only proleptic Gregorian calendar (i.e. extended to support older dates) is supported.
-Be very careful if you really have to deal with pre-20C dates, they can be in Julian or others.
+Only proleptic Gregorian calendar (i.e. extended to support older dates) is
+supported. Be very careful if you really have to deal with pre-20C dates, they
+can be in Julian or others.
 
-Date types are limited in about +/- 262,000 years from the common epoch.
-Time types are limited in the nanosecond accuracy.
+Date types are limited in about +/- 262,000 years from the common epoch. Time
+types are limited in the nanosecond accuracy.
 
-[Leap seconds are supported in the representation but
-Chrono doesn't try to make use of them](https://docs.rs/chrono/0.5/chrono/naive/struct.NaiveTime.html#leap-second-handling).
-(The main reason is that leap seconds are not really predictable.)
-Almost *every* operation over the possible leap seconds will ignore them.
-Consider using `NaiveDateTime` with the implicit TAI (International Atomic Time) scale
-if you want.
+[Leap seconds] are supported in the representation but Chrono doesn't try to
+make use of them. (The main reason is that leap seconds are not really
+predictable.) Almost *every* operation over the possible leap seconds will
+ignore them. Consider using `NaiveDateTime` with the implicit TAI (International
+Atomic Time) scale if you want.
 
-Chrono inherently does not support an inaccurate or partial date and time representation.
-Any operation that can be ambiguous will return `None` in such cases.
-For example, "a month later" of 2014-01-30 is not well-defined
-and consequently `Utc.ymd(2014, 1, 30).with_month(2)` returns `None`.
+Chrono inherently does not support an inaccurate or partial date and time
+representation. Any operation that can be ambiguous will return
+`Err(ChronoError)` in such cases. For example, "a month later" of 2014-01-30 is
+not well-defined and consequently `Utc.ymd(2014, 1, 30)?.with_month(2)?` returns
+`Err(ChronoError)`.
 
-Non ISO week handling is not yet supported.
-For now you can use the [chrono_ext](https://crates.io/crates/chrono_ext)
-crate ([sources](https://github.com/bcourtine/chrono-ext/)).
+Non ISO week handling is not yet supported. For now you can use the [chrono_ext]
+crate ([sources]).
 
-Advanced time zone handling is not yet supported.
-For now you can try the [Chrono-tz](https://github.com/chronotope/chrono-tz/) crate instead.
+Advanced time zone handling is not yet supported. For now you can try the
+[Chrono-tz] crate instead.
+
+[chrono_ext]: https://crates.io/crates/chrono_ext
+[Chrono-tz]: https://github.com/chronotope/chrono-tz/
+[Leap seconds]: https://docs.rs/chrono/0.5/chrono/naive/struct.NaiveTime.html#leap-second-handling
+[sources]: https://github.com/bcourtine/chrono-ext/
--- a/README.md
+++ b/README.md
@@ -47,10 +47,9 @@ ignore them. Consider using `NaiveDateTime` with the implicit TAI (International
 Atomic Time) scale if you want.
 
 Chrono inherently does not support an inaccurate or partial date and time
-representation. Any operation that can be ambiguous will return
-`Err(ChronoError)` in such cases. For example, "a month later" of 2014-01-30 is
-not well-defined and consequently `Utc.ymd(2014, 1, 30)?.with_month(2)?` returns
-`Err(ChronoError)`.
+representation. Any operation that can be ambiguous will return `Err(Error)` in
+such cases. For example, "a month later" of 2014-01-30 is not well-defined and
+consequently `Utc.ymd(2014, 1, 30)?.with_month(2)?` returns `Err(Error)`.
 
 Non ISO week handling is not yet supported. For now you can use the [chrono_ext]
 crate ([sources]).
--- a/README.md
+++ b/README.md
@@ -33,31 +33,27 @@ which Chrono builds upon and should acknowledge:
 
 ## Limitations
 
-Only proleptic Gregorian calendar (i.e. extended to support older dates) is
-supported. Be very careful if you really have to deal with pre-20C dates, they
-can be in Julian or others.
+Only proleptic Gregorian calendar (i.e. extended to support older dates) is supported.
+Be very careful if you really have to deal with pre-20C dates, they can be in Julian or others.
 
-Date types are limited in about +/- 262,000 years from the common epoch. Time
-types are limited in the nanosecond accuracy.
+Date types are limited in about +/- 262,000 years from the common epoch.
+Time types are limited in the nanosecond accuracy.
 
-[Leap seconds] are supported in the representation but Chrono doesn't try to
-make use of them. (The main reason is that leap seconds are not really
-predictable.) Almost *every* operation over the possible leap seconds will
-ignore them. Consider using `NaiveDateTime` with the implicit TAI (International
-Atomic Time) scale if you want.
+[Leap seconds are supported in the representation but
+Chrono doesn't try to make use of them](https://docs.rs/chrono/0.5/chrono/naive/struct.NaiveTime.html#leap-second-handling).
+(The main reason is that leap seconds are not really predictable.)
+Almost *every* operation over the possible leap seconds will ignore them.
+Consider using `NaiveDateTime` with the implicit TAI (International Atomic Time) scale
+if you want.
 
-Chrono inherently does not support an inaccurate or partial date and time
-representation. Any operation that can be ambiguous will return `Err(Error)` in
-such cases. For example, "a month later" of 2014-01-30 is not well-defined and
-consequently `Utc.ymd(2014, 1, 30)?.with_month(2)?` returns `Err(Error)`.
+Chrono inherently does not support an inaccurate or partial date and time representation.
+Any operation that can be ambiguous will return `Err(chrono::Error)` in such cases.
+For example, "a month later" of 2014-01-30 is not well-defined
+and consequently `Utc.ymd(2014, 1, 30)?.with_month(2)?` returns `Err(chrono::Error)`.
 
-Non ISO week handling is not yet supported. For now you can use the [chrono_ext]
-crate ([sources]).
+Non ISO week handling is not yet supported.
+For now you can use the [chrono_ext](https://crates.io/crates/chrono_ext)
+crate ([sources](https://github.com/bcourtine/chrono-ext/)).
 
-Advanced time zone handling is not yet supported. For now you can try the
-[Chrono-tz] crate instead.
-
-[chrono_ext]: https://crates.io/crates/chrono_ext
-[Chrono-tz]: https://github.com/chronotope/chrono-tz/
-[Leap seconds]: https://docs.rs/chrono/0.5/chrono/naive/struct.NaiveTime.html#leap-second-handling
-[sources]: https://github.com/bcourtine/chrono-ext/
+Advanced time zone handling is not yet supported.
+For now you can try the [Chrono-tz](https://github.com/chronotope/chrono-tz/) crate instead.
--- a/benches/chrono.rs
+++ b/benches/chrono.rs
@@ -35,14 +35,14 @@ fn bench_datetime_from_str(c: &mut Criterion) {
 }
 
 fn bench_datetime_to_rfc2822(c: &mut Criterion) {
-    let pst = FixedOffset::east(8 * 60 * 60);
-    let dt = pst.ymd(2018, 1, 11).and_hms_nano(10, 5, 13, 84_660_000);
+    let pst = FixedOffset::east(8 * 60 * 60).unwrap();
+    let dt = pst.ymd(2018, 1, 11).unwrap().and_hms_nano(10, 5, 13, 84_660_000).unwrap();
     c.bench_function("bench_datetime_to_rfc2822", |b| b.iter(|| black_box(dt).to_rfc2822()));
 }
 
 fn bench_datetime_to_rfc3339(c: &mut Criterion) {
-    let pst = FixedOffset::east(8 * 60 * 60);
-    let dt = pst.ymd(2018, 1, 11).and_hms_nano(10, 5, 13, 84_660_000);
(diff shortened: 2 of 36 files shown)


udoprog:
Here's a brief guide to how the new API's are used:

Methods on `TimeZone` have been made fallible. This is necessary since the `Local` might need to *fallibly* talk to the system to figure things out.

```rust
pub trait TimeZone: Sized + Clone {
    type Offset: Offset;

    fn ymd(&self, year: i32, month: u32, day: u32) -> Result<LocalResult<Date<Self>>, ChronoError>;
    fn yo(&self, year: i32, ordinal: u32) -> Result<LocalResult<Date<Self>>, ChronoError>;
    fn isoywd(&self, year: i32, week: u32, weekday: Weekday) -> Result<LocalResult<Date<Self>>, ChronoError>;

    fn timestamp(&self, secs: i64, nsecs: u32) -> Result<DateTime<Self>, ChronoError>;
    fn timestamp_millis(&self, millis: i64) -> Result<DateTime<Self>, ChronoError>;
    fn timestamp_nanos(&self, nanos: i64) -> Result<DateTime<Self>, ChronoError>;

    fn datetime_from_str(&self, s: &str, fmt: &str) -> ParseResult<DateTime<Self>>;

    fn from_offset(offset: &Self::Offset) -> Self;

    fn offset_from_local_date(&self, local: &NaiveDate) -> Result<LocalResult<Self::Offset>, ChronoError>;
    fn offset_from_local_datetime(&self, local: &NaiveDateTime) -> Result<LocalResult<Self::Offset>, ChronoError>;

    fn from_local_date(&self, local: &NaiveDate) -> Result<LocalResult<Date<Self>>, ChronoError>;
    fn from_local_datetime(&self, local: &NaiveDateTime) -> Result<LocalResult<DateTime<Self>>, ChronoError>;

    fn offset_from_utc_date(&self, utc: &NaiveDate) -> Result<Self::Offset, ChronoError>;
    fn offset_from_utc_datetime(&self, utc: &NaiveDateTime) -> Result<Self::Offset, ChronoError>;

    fn from_utc_date(&self, utc: &NaiveDate) -> Result<Date<Self>, ChronoError>;
    fn from_utc_datetime(&self, utc: &NaiveDateTime) -> Result<DateTime<Self>, ChronoError>;
}
```

To cover some of the shortfall, the following trait is defined which is implemented for `Utc` and `FixedOffset`, the result of operating with it is unambiguous so `LocalResult` no longer needs to be used:

```rust
pub trait FixedTimeZone: TimeZone {
    fn offset_from_utc_date_fixed(&self, utc: &NaiveDate) -> Self::Offset;
    fn offset_from_utc_datetime_fixed(&self, utc: &NaiveDateTime) -> Self::Offset;

    fn from_utc_date_fixed(&self, utc: &NaiveDate) -> Date<Self>;
    fn from_utc_datetime_fixed(&self, utc: &NaiveDateTime) -> DateTime<Self>;
}
```

With this `Date<Tz>` and `DateTime<Tz>` gets a `with_fixed_timezone` function, since these are used in places where non-fallible operations would be useful like `From` impls:

```rust
impl<Tz: TimeZone> Date<Tz> {
    pub fn with_fixed_timezone<Tz2: FixedTimeZone>(&self, tz: &Tz2) -> Date<Tz2>;
}

impl<Tz: TimeZone> DateTime<Tz> {
    pub fn with_fixed_timezone<Tz2: FixedTimeZone>(&self, tz: &Tz2) -> DateTime<Tz2>;
}
```

Note that it would be possible to evolve `FixedTimeZone` to get more fallible methods.

The following changes have been done to `DateLike`:

```rust
pub trait Datelike: Sized {
    fn with_year(&self, year: i32) -> Result<Self, ChronoError>;
    fn with_month(&self, month: u32) -> Result<Self, ChronoError>;
    fn with_month0(&self, month0: u32) -> Result<Self, ChronoError>;
    fn with_day(&self, day: u32) -> Result<Self, ChronoError>;
    fn with_day0(&self, day0: u32) -> Result<Self, ChronoError>;
    fn with_ordinal(&self, ordinal: u32) -> Result<Self, ChronoError>;
    fn with_ordinal0(&self, ordinal0: u32) -> Result<Self, ChronoError>;
}
```

The following changes have been done to to `TimeLike`:

```rust
pub trait TimeLike: Sized {
    fn with_hour(&self, hour: u32) -> Result<Self, ChronoError>;
    fn with_minute(&self, min: u32) -> Result<Self, ChronoError>;
    fn with_second(&self, sec: u32) -> Result<Self, ChronoError>;
    fn with_nanosecond(&self, nano: u32) -> Result<Self, ChronoError>;
}
```

The following methods have been changed on `DateTime<Tz>`:
```rust
impl<Tz: TimeZone> DateTime<Tz> {
    pub fn with_timezone<Tz2: TimeZone>(&self, tz: &Tz2) -> Result<DateTime<Tz2>, ChronoError>;
    pub fn checked_add_signed(self, rhs: TimeDelta) -> Result<DateTime<Tz>, ChronoError>;
    pub fn checked_add_months(self, rhs: Months) -> Result<DateTime<Tz>, ChronoError>;
    pub fn checked_sub_signed(self, rhs: TimeDelta) -> Result<DateTime<Tz>, ChronoError>;
    pub fn checked_sub_months(self, rhs: Months) -> Result<DateTime<Tz>, ChronoError>;
    pub fn checked_add_days(self, days: Days) -> Result<Self, ChronoError>;
    pub fn checked_sub_days(self, days: Days) -> Result<Self, ChronoError>;
}
```

The following methods have been changed on `Date<Tz>`:

```rust
impl<Tz: TimeZone> Date<Tz> {
    pub fn and_time(&self, time: NaiveTime) -> Result<DateTime<Tz>, ChronoError>;
    pub fn and_hms(&self, hour: u32, min: u32, sec: u32) -> Result<DateTime<Tz>, ChronoError>;
    pub fn and_hms_milli(&self, hour: u32, min: u32, sec: u32, milli: u32) -> Result<DateTime<Tz>, ChronoError>;
    pub fn and_hms_micro(&self, hour: u32, min: u32, sec: u32, micro: u32) -> Result<DateTime<Tz>, ChronoError>;
    pub fn and_hms_nano(&self, hour: u32, min: u32, sec: u32, nano: u32) -> Result<DateTime<Tz>, ChronoError>;
    pub fn succ(&self) -> Result<Date<Tz>, ChronoError>;
    pub fn pred(&self) -> Result<Date<Tz>, ChronoError>;
    pub fn with_timezone<Tz2: TimeZone>(&self, tz: &Tz2) -> Result<Date<Tz2>, ChronoError>;

    pub fn checked_add_signed(self, rhs: TimeDelta) -> Result<Date<Tz>, ChronoError>;
    pub fn checked_sub_signed(self, rhs: TimeDelta) -> Result<Date<Tz>, ChronoError>;
}
```

The following methods have been changed/added on `LocalResult<T>`:

```rust
impl<T> LocalResult<T> {
     pub fn single(self) -> Result<T, ChronoError>;
     pub fn earliest(self) -> T;
     pub fn latest(self) -> T;
     pub fn try_map<U, E, F: FnMut(T) -> Result<U, E>>(self, mut f: F) -> Result<LocalResult<U>, E>;
}
```

esheppa:
> Methods on TimeZone have been made fallible. This is necessary since the Local might need to fallibly talk to the system to figure things out.

Potentially we could include the error as an associated type (potentially with default) in the trait, such that implementors can specify [`infallible`](https://doc.rust-lang.org/core/convert/enum.Infallible.html) for timezones like `FixedOffset` and `Utc`

djc:
Here are some thoughts/suggestions so far (haven't dug into the code yet):

* I feel like we should name the type `chrono::Error` and leave prefixing it to downstream users.
* Instead of adding a `_fixed` suffix for infallible methods, we could consider a `try_` prefix on fallible methods.
* For `checked_` methods I feel like the convention is to yield `Option` instead of `Result`, and we should stick with that.
* For cases where the only error case is out of range (assuming there's a decent number of these?), I wonder whether we should yield `OutOfRange` (which IIRC we already have in 0.5) as a separate error type, instead of yielding a more generic error type with variants that actually aren't applicable.

udoprog:
> Instead of adding a _fixed suffix for infallible methods, we could consider a try_ prefix on fallible methods.

Are you suggesting this applies to all methods that have been made fallible or just the ones associated with `TimeZone` / `FixedTimeZone`? I was considering inverting the relationship between the two so that `TimeZone` is the infallible one (`Utc` / `FixedOffset`) and some other trait yet to be named would be the fallible one only implemented by `Local` but that change seemed more involved and can wait.

@esheppa This also relates to your suggestion since you wanted a generic error type. All though I'm not super happy with `Infallible` since the `Result`-oriented methods for these [aren't stable yet](https://doc.rust-lang.org/std/result/enum.Result.html#method.into_ok) and somewhat unergonomic. But maybe less so then a separate family of methods distinguished by a suffix like `with_fixed_timezone`.

> For cases where the only error case is out of range (assuming there's a decent number of these?), I wonder whether we should yield `OutOfRange` (which IIRC we already have in 0.5) as a separate error type, instead of yielding a more generic error type with variants that actually aren't applicable.

Many things could be *considered* out of range, so what specific standard do you have in mind to apply here? I'm tentative towards considering invalid dates to be out of range (even though they are defined to be outside `MIN_UTC` / `MAX_UTC`) because I want the error to say the same thing as the panic currently does (`"invalid date"`) rather than something generic (`"out of range"`).

udoprog:
Also since this is a big and *involved* PR, if you feel like you can quickly just make the necessary changes to merge this don't feel obliged to block on me. So feel free to just make the changes and merge.

djc:
Sorry, I have way too much on my plate in terms of maintenance (as well as moving to a new place with my family) so I'm all too happy having you take care of the code changes here.

But, maybe to make this easier to digest, it might be worth splitting it up in chunks, roughly along the lines that you presented in your comment? I think many of these changes are pretty independent so it might be nice to iron out some of the non-controversial stuff first and work from there.

README.md
```diff
@@ -33,27 +33,32 @@ which Chrono builds upon and should acknowledge:
 
 ## Limitations
 
-Only proleptic Gregorian calendar (i.e. extended to support older dates) is supported.
```
djc, line 36: Please leave out unrelated changes to reflow text, which mostly make it harder to review this.

udoprog:
> But, maybe to make this easier to digest, it might be worth splitting it up in chunks, roughly along the lines that you presented in your comment? I think many of these changes are pretty independent so it might be nice to iron out some of the non-controversial stuff first and work from there.

Don't think this is easily achievable. Most of the PR essentially flowed "what if I do the changes to `TimeZone` and see what needs to be fixed". Doing it piecemeal would also leave the codebase peculiar states and nontrivially increase the work involved to deal with intermediate compatibility (e.g. use of fallible methods in infallible ones would need to `.unwrap()` which would then have to be removed in a subsequent PR).

An alternative approach could be that you don't necessarily have to reach the perfect end result in this PR, but can work on changing it to what you want for release in subsequent ones. Which is easy enough to apply to changes like naming and moving things around.

Anyway, I'll do any concrete changes you're requesting to the best of my ability.

udoprog:
Ok, unless I'm missing something I've made the concrete changes which are no longer being discussed. The remaining considerations are:

@esheppa [Proposes a change to `TimeZone`](https://github.com/chronotope/chrono/pull/817#issuecomment-1245312310) do add an `Error` associated type instead of introducing `FixedTimeZone`. This is something I'd personally like to punt to a future PR thinking over the design of the fallible / non-fallible methods because there's a lot of potential to get stuck here. I've made `FixedTimeZone` non-public for now since it's only used in private APIs anyways. This should also effectively punt the naming concern @djc had over them.

@djc [Proposed to consider using `OutOfRange`](https://github.com/chronotope/chrono/pull/817#issuecomment-1245342388). I'd personally want to punt this for the future. It echoes the consideration I had initially on introducing detailed error types, but from a broader public API perspective and I feel it's a more complex discussion which I'd be happy to commit to putting work into before a 0.5 release. I.e.: apply a coherent pattern towards error handling.

Some markdown reflow is still present, but hopefully it's only in the odd API documentation and shouldn't be too much of a bother.

djc (commented):
I reviewed the `date` module. It looks like several unrelated changes have been made which I'd like to have reverted in order to keep this PR as a focused change, making it easier to review only the changes necessary to making everything fallible. As such, any other changes to the code style/structure should be avoided.

udoprog:
> I reviewed the `date` module. It looks like several unrelated changes have been made which I'd like to have reverted in order to keep this PR as a focused change, making it easier to review only the changes necessary to making everything fallible. As such, any other changes to the code style/structure should be avoided.

So this might be why it's better for a maintainer do these kind of large changes. I'm sorry, but I don't have the time to put in work to undo stylistic changes that slipped through due to the size of this change, so unless there's a good reason (e.g. there's an obvious non-improvement) I can't do that for you. I understand if that doesn't work for you so feel free to close the PR and maybe this can be done at a later time. I'll hold off resolving the conflict for now until we've sorted this out.

djc:
I'm sorry that you feel this is an unreasonable request. I understand that it may take some time and it is not fun work to do (but then the same is true for making the changes in this PR in the first place, to large extent). Personally I feel that it is generally reasonable to reduce time spent by the maintainer (that is, by making changes optimally easy to review) at the cost of requiring contributors to put in more time, given that maintainers are required to spend much more time, and their time is (within this particular context) "more valuable". (I am aware that you also maintain a whole bunch of projects -- although maybe you aren't as strict with contributors, so there's a different balance?)

I don't maintain chrono because I find the job particularly fun -- I do it because it needs to be done, and at the time I was the only person willing to put in the time and energy.

I hope you're willing to reconsider so that I can spend my time elsewhere (for example, by making similar changes in 0.4.x to deprecate non-`_opt()` constructors).

WhyNotHugo:
Easiest way to get rid of the unwanted stylistic changes is is by restoring the index and adding only the wanted changes:

```
git fetch upstream pull/817/head  # Fetch the PR
git checkout -b 817-v2 pull/817/head  # Create a new branch based on it
git reset upstream/main  # Keep changes, but drop commits
git add -up  # Add individual chunks
# Hint: use `s` to split large chunks
git commit
```

This would take a while, but it's a matter of picking y/n for each change. Do remember to keep authorship information in the new commit.

udoprog:
> I'm sorry that you feel this is an unreasonable request.

The request to undo `.and_then` removal is largely what would cost a lot of time to implement for no benefit other than this review. If issues similar to that require reverting it would basically mean redoing everything (except maybe test cases). But I don't think it's an unreasonable request, I should've been more prudent.

djc:
> Easiest way to get rid of the unwanted stylistic changes is is by restoring the index and adding only the wanted changes:
> 
> ```
> git fetch upstream pull/817/head  # Fetch the PR
> git checkout -b 817-v2 pull/817/head  # Create a new branch based on it
> git reset upstream/main  # Keep changes, but drop commits
> git add -up  # Add individual chunks
> # Hint: use `s` to split large chunks
> git commit
> ```
> 
> This would take a while, but it's a matter of picking y/n for each change. Do remember to keep authorship information in the new commit.

Unfortunately many of the changes do not neatly align to diff hunk boundaries, so it's not as easy as this. That said, if you want to contribute, that would be great!

djc:
> > I'm sorry that you feel this is an unreasonable request.
> 
> The request to undo `.and_then` removal is largely what would cost a lot of time to implement for no benefit other than this review. If issues similar to that require reverting it would basically mean redoing everything (except maybe test cases). But I don't think it's an unreasonable request, I should've been more prudent.

FWIW, I would prefer `and_then()` over introducing single-use bindings even if this was new code.

udoprog:
All right. I'll close for now and re-open if I ever have the time to redo the PR so it conforms to expectation. If anyone else wants to pick it up, feel free.

PR #818, opened 2022-09-11
NaiveDateTime::from_timestamp_millis(_opt)

Pscheidl:
Construct NaiveDateTime from millis since epoch.


djc:
Despite the consistency argument I'm inclined to ditch the `_opt()` suffix from the fallible method and leave potential unwrapping to the caller -- this seems clearly better to me and the way of the future. @esheppa thoughts?

Pscheidl:
> Despite the consistency argument I'm inclined to ditch the `_opt()` suffix from the fallible method and leave potential unwrapping to the caller (that is -- not add the panicking variant at all). This seems clearly better to me and the way of the future. @esheppa thoughts?

Done (we can always revert to the original).

src/naive/datetime/mod.rs
```diff
@@ -127,6 +127,28 @@ impl NaiveDateTime {
         datetime.expect("invalid or out-of-range datetime")
     }
 
+    /// Makes a new `NaiveDateTime` corresponding to a UTC date and time,
+    /// from the number of milliseconds
+    /// since the midnight UTC on January 1, 1970 (aka "UNIX timestamp").
+    ///
+    /// Returns `None` on the out-of-range number of milliseconds and/or invalid nanosecond.
```
djc, line 134: In this case, there's no "invalid nanosecond", right? Also I think "the" -> "an" would read better?

src/naive/datetime/mod.rs
```diff
@@ -127,6 +127,28 @@ impl NaiveDateTime {
         datetime.expect("invalid or out-of-range datetime")
     }
 
+    /// Makes a new `NaiveDateTime` corresponding to a UTC date and time,
```
djc, line 130: You probably copy-pasted this from one of the other methods, but we're trying to make the first line of docstrings self-contained sentence fragments. After that we should have an empty line, then a more detailed description. Would you mind tweaking the docs for this new method at least?

djc (commented):
Thanks for working on this! LGTM modulo some documentation nits.

src/naive/datetime/mod.rs
Pscheidl: No, there are no invalid nanos in this case. Changed.

src/naive/datetime/mod.rs
Pscheidl: I think the whole doc can be simplified. Gave it a try. Please have a look.

src/naive/datetime/mod.rs
```diff
@@ -126,12 +126,11 @@ impl NaiveDateTime {
...
-
-    /// Makes a new `NaiveDateTime` corresponding to a UTC date and time,
-    /// from the number of milliseconds
-    /// since the midnight UTC on January 1, 1970 (aka "UNIX timestamp").
+    /// Creates a new [NaiveDateTime] from milliseconds since UNIX Epoch.
+    ///
+    /// UNIX epoch starts on midnight, January 1, 1970, UTC.
     ///
-    /// Returns `None` on the out-of-range number of milliseconds and/or invalid nanosecond.
+    ///  Returns `None` on an out-of-range number of milliseconds.
```
djc, line 133: Should ditch the prefix space here.

src/naive/datetime/mod.rs
```diff
@@ -126,12 +126,11 @@ impl NaiveDateTime {
         let datetime = NaiveDateTime::from_timestamp_opt(secs, nsecs);
         datetime.expect("invalid or out-of-range datetime")
     }
-
-    /// Makes a new `NaiveDateTime` corresponding to a UTC date and time,
-    /// from the number of milliseconds
-    /// since the midnight UTC on January 1, 1970 (aka "UNIX timestamp").
+    /// Creates a new [NaiveDateTime] from milliseconds since UNIX Epoch.
```
djc, line 129: Both here and in the next line, I think prefixing "UNIX" with "the" ("the UNIX epoch") would read better.

djc (commented):
Seems like an improvement, some remaining nits.

src/naive/datetime/mod.rs
Pscheidl: done

src/naive/datetime/mod.rs
Pscheidl: Sure, done,

djc (approved):

src/naive/datetime/mod.rs
```diff
@@ -126,6 +126,27 @@ impl NaiveDateTime {
...
+    /// use chrono::NaiveDateTime;
+    /// let timestamp_millis: i64 = 1662921288; //Sunday, September 11, 2022 6:34:48 PM
+    /// let naive_datetime = NaiveDateTime::from_timestamp_millis(timestamp_millis);
+    /// assert!(naive_datetime.is_some());
+    /// assert_eq!(timestamp_millis, naive_datetime.unwrap().timestamp_millis());
+    /// ```
+    #[inline]
+    pub fn from_timestamp_millis(millis: i64) -> Option<NaiveDateTime> {
+        let secs = millis / 1000;
+        let nsecs = (millis % 1000) as u32 * 1_000_000;
```
esheppa, line 147: this looks problematic as the input `millis` could be negative, and will still be negative after taking the modulo ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7e7ecdb063e7200a2970e92632ff1fbc))

One option would be to have two cases, depending on whether `millis` is positive or negative, similar to the logic used [here](https://github.com/chronotope/chrono/blob/e4911675b9c9cf6f9dc33015987494843b35fee8/src/naive/time/mod.rs#L535-L596)

The other option is just:
`NaiveDate::from_ymd(1970,1, 1).and_hms(0, 0, 0) + TimeDelta::from_millis(millis)`


Also it would be preferable to use `TryFrom` instead of `as`

src/naive/datetime/mod.rs
```diff
@@ -126,6 +126,27 @@ impl NaiveDateTime {
         let datetime = NaiveDateTime::from_timestamp_opt(secs, nsecs);
         datetime.expect("invalid or out-of-range datetime")
     }
+    /// Creates a new [NaiveDateTime] from milliseconds since the UNIX Epoch.
```
esheppa, line 129: potentially the capitalization should be aligned between this and the next line, I'd probably go for lowercase in both

esheppa:
Thanks for the PR @Pscheidl - this will be a useful helper function to have, and the docs look good - just noted a possible issue in the implementation when the input timestamp is negative

esheppa (changes requested):

src/naive/datetime/mod.rs
Pscheidl: Awesome, let's deal with that. I'm super busy ATM, please give me by the end of the week :)

djc: I don't think we should go through `TimeDelta` here, which adds a bunch of overhead and feels like a layering violation. Would be better to do a minimal calculation and still defer to `NaiveDateTime::from_timestamp_opt()`.

esheppa: Good call RE. avoiding `TimeDelta` especially as `from_millis` may well get deprecated anyway

src/naive/datetime/mod.rs
Pscheidl: Done

src/naive/datetime/mod.rs
Pscheidl: Done. The nanos calculation is split, as when the timestamp in millis is negative, an additional `abs()` op is done.

Pscheidl:
All comments addressed.

src/naive/datetime/mod.rs
```diff
@@ -127,6 +132,40 @@ impl NaiveDateTime {
...
+    /// assert!(naive_datetime.is_some());
+    /// assert_eq!(timestamp_millis, naive_datetime.unwrap().timestamp_millis());
+    /// ```
+    #[inline]
+    pub fn from_timestamp_millis(millis: i64) -> Option<NaiveDateTime> {
+        let secs = millis / 1000;
+
+        if millis < 0 {
+            let nsecs = (millis % 1000).abs() as u32 * NANOS_IN_MILLISECOND;
+            NaiveDateTime::from_timestamp_opt(secs - 1, NANOS_IN_SECOND - nsecs)
```
esheppa, line 162: Thanks for the impl that handles negative millis. If possible could you subtract with [`checked_sub`](https://doc.rust-lang.org/std/primitive.i64.html#method.checked_sub) here to avoid potential overflow

src/naive/datetime/mod.rs
```diff
@@ -127,6 +132,40 @@ impl NaiveDateTime {
...
+    /// let naive_datetime = NaiveDateTime::from_timestamp_millis(timestamp_millis);
+    /// assert!(naive_datetime.is_some());
+    /// assert_eq!(timestamp_millis, naive_datetime.unwrap().timestamp_millis());
+    /// ```
+    #[inline]
+    pub fn from_timestamp_millis(millis: i64) -> Option<NaiveDateTime> {
+        let secs = millis / 1000;
+
+        if millis < 0 {
+            let nsecs = (millis % 1000).abs() as u32 * NANOS_IN_MILLISECOND;
```
esheppa, line 161: Do you mind switching `as` for `u32::try_from(expr).ok()?`

src/naive/datetime/mod.rs
```diff
@@ -127,6 +132,40 @@ impl NaiveDateTime {
...
+            NaiveDateTime::from_timestamp_opt(secs - 1, NANOS_IN_SECOND - nsecs)
+        } else {
+            let nsecs = (millis % 1000) as u32 * NANOS_IN_MILLISECOND;
```
esheppa, line 164: Do you mind switching as for u32::try_from(expr).ok()?

esheppa (changes requested):

esheppa:
Thanks @Pscheidl, the new negative millis handling looks good- I've just requested a few more small changes

src/naive/datetime/mod.rs
Pscheidl: Done

src/naive/datetime/mod.rs
Pscheidl: Done

src/naive/datetime/mod.rs
Pscheidl: Done

Pscheidl:
Awesome @esheppa , thank you. All comments addressed.

esheppa (approved):
Thanks for those changes @Pscheidl

Pscheidl:
Learned a lot about the crate, thank you for the guidance. 

I can't hit the `merge` button, as I don't have write access.

djc:
Oops, should have squashed this instead of rebasing. :(

djc:
Backport to 0.4.x is in #823, which also includes some deduplication of the logic.

PR #819, opened 2022-09-12
Test no default features

mqudsi:
This patch configures CI to assert that `cargo test --no-default-features` also passes (in addition to the default combinatoric feature matrix). 

An additional commit is included that fixes the existing errors preventing `cargo test` from passing under `--no-default-features`. The commit message explains the approach taken.

This PR is target 0.5.x as some of the fixes were to 0.5.x-only impls; I'll file a separate PR for 0.4.x after this one lands in case any changes are needed or requested.

### DIFF
--- a/ci/github.sh
+++ b/ci/github.sh
@@ -69,6 +69,7 @@ test_regular() {
     tz="$1" && shift
 
     runt env TZ="$tz" cargo test --features __doctest,unstable-locales --color=always -- --color=always
+    runt env TZ="$tz" cargo test --no-default-features --color=always -- --color=always
     for feature in "${FEATURES[@]}"; do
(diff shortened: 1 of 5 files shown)


esheppa:
Thanks for this PR @mqudsi  - this looks like a good complement to some work in #814, as this PR has picked up some issues with the doc tests that I missed on that PR - what do you think @djc?

esheppa (approved):

djc:
Yes, I was thinking we should rebase it on that PR (or I should just get off my ass and review it).

djc:
@mqudsi would you mind rebasing this on current 0.4.x?

pitdicker:
Thank you! Merged in https://github.com/chronotope/chrono/pull/1059.

PR #823, opened 2022-09-19
Port from_timestamp_millis() (#818) to 0.4.x

djc:
Added some changes to deduplicate logic in the code.

### DIFF
--- a/src/naive/datetime/mod.rs
+++ b/src/naive/datetime/mod.rs
@@ -160,18 +160,18 @@ impl NaiveDateTime {
     /// ```
     #[inline]
     pub fn from_timestamp_millis(millis: i64) -> Option<NaiveDateTime> {
-        let secs = millis / 1000;
-
+        let mut secs = millis / 1000;
         if millis < 0 {
-            let nsecs = u32::try_from((millis % 1000).abs()).ok()? * NANOS_IN_MILLISECOND;
-            NaiveDateTime::from_timestamp_opt(
-                secs.checked_sub(1)?,
-                NANOS_IN_SECOND.checked_sub(nsecs)?,
-            )
-        } else {
-            let nsecs = u32::try_from(millis % 1000).ok()? * NANOS_IN_MILLISECOND;
-            NaiveDateTime::from_timestamp_opt(secs, nsecs)
+            secs = secs.checked_sub(1)?;
         }
+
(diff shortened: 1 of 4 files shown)


src/naive/datetime/mod.rs
```diff
@@ -131,6 +135,39 @@ impl NaiveDateTime {
...
+    /// let timestamp_millis: i64 = -2208936075; //Mon Jan 01 1900 14:38:45 GMT+0000
+    /// let naive_datetime = NaiveDateTime::from_timestamp_millis(timestamp_millis);
+    /// assert!(naive_datetime.is_some());
+    /// assert_eq!(timestamp_millis, naive_datetime.unwrap().timestamp_millis());
+    /// ```
+    #[inline]
+    pub fn from_timestamp_millis(millis: i64) -> Option<NaiveDateTime> {
+        let mut secs = millis / 1000;
+        if millis < 0 {
+            secs = secs.checked_sub(1)?;
```
esheppa, line 163: I think we still need to include this part as well: `NANOS_IN_SECOND.checked_sub(nsecs)?,` (in the `millis < 0` case only)

esheppa (changes requested):

src/naive/datetime/mod.rs
djc: Oh yeah, good catch. Added a commit that enables CI to run on the 0.4.x branch, too...

esheppa (approved):
Looks good! Maybe backporting to `0.4.x` is cleaner - it keeps the history of the `main` branch nicer as well

djc:
Hmm, let's try a bit more -- we should be a little more forceful about getting authors to target 0.4.x, I think.

Pscheidl:
This version is cleaner, but should be slightly less efficient - two ifs instead of one and the 'abs' operation is always performed. Not sure if you care about the difference, just explaining the original implementation.

djc:
Yeah -- unless someone wants to build benchmarks that show a substantial improvement, I'd rather opt for code that's easier to follow.

As for the `abs()` call -- since `abs()` will internally call `is_negative()` anyway, branching in order to prevent an `abs()` call doesn't actually help -- it will have the same number of branches in the case it's not needed, and an extra branch in the case it is needed.

PR #825, opened 2022-09-26
Remove separate CI workflows

djc:
I don't think there is anything in the `test-release.yml` workflow that we still need, and it looks like it didn't participate in #814, leading to errors.


esheppa (approved):
Good idea to remove this. Looks like it had gotten out of sync with `test.yml` for a while anyway

PR #827, opened 2022-09-28
0.4.x: deprecate methods that have an `_opt()` alternative

djc:
As a compatible step towards a future for 0.5 where we try to not panic internally, this deprecates all methods that have an `_opt()` alternative in order to require the caller to `unwrap()` or `expect()` instead of the library doing that.

This is a small step towards #815 and allows us to test with current users how this direction is perceived.

### DIFF
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -68,7 +68,7 @@ impl NaiveWeek {
     /// ```
     /// use chrono::{NaiveDate, Weekday};
     ///
-    /// let date = NaiveDate::from_ymd(2022, 4, 18);
+    /// let date = NaiveDate::from_ymd_opt(2022, 4, 18).unwrap();
     /// let week = date.week(Weekday::Mon);
     /// assert!(week.first_day() <= date);
     /// ```
@@ -87,7 +87,7 @@ impl NaiveWeek {
     /// ```
     /// use chrono::{NaiveDate, Weekday};
     ///
-    /// let date = NaiveDate::from_ymd(2022, 4, 18);
+    /// let date = NaiveDate::from_ymd_opt(2022, 4, 18).unwrap();
     /// let week = date.week(Weekday::Mon);
     /// assert!(week.last_day() >= date);
     /// ```
@@ -105,7 +105,7 @@ impl NaiveWeek {
     /// ```
     /// use chrono::{NaiveDate, Weekday};
     ///
-    /// let date = NaiveDate::from_ymd(2022, 4, 18);
+    /// let date = NaiveDate::from_ymd_opt(2022, 4, 18).unwrap();
     /// let week = date.week(Weekday::Mon);
     /// let days = week.days();
     /// assert!(days.contains(&date));
@@ -197,8 +197,8 @@ pub const MAX_DATE: NaiveDate = NaiveDate::MAX;
 // we use a separate run-time test.
 #[test]
 fn test_date_bounds() {
-    let calculated_min = NaiveDate::from_ymd(MIN_YEAR, 1, 1);
-    let calculated_max = NaiveDate::from_ymd(MAX_YEAR, 12, 31);
+    let calculated_min = NaiveDate::from_ymd_opt(MIN_YEAR, 1, 1).unwrap();
+    let calculated_max = NaiveDate::from_ymd_opt(MAX_YEAR, 12, 31).unwrap();
     assert!(
         NaiveDate::MIN == calculated_min,
         "`NaiveDate::MIN` should have a year flag {:?}",
@@ -463,7 +463,7 @@ impl NaiveDate {
         if n == 0 {
             return None;
         }
-        let first = NaiveDate::from_ymd(year, month, 1).weekday();
+        let first = NaiveDate::from_ymd_opt(year, month, 1)?.weekday();
         let first_to_dow = (7 + weekday.number_from_monday() - first.number_from_monday()) % 7;
         let day = (u32::from(n) - 1) * 7 + first_to_dow + 1;
         NaiveDate::from_ymd_opt(year, month, day)
@@ -481,9 +481,9 @@ impl NaiveDate {
     /// let parse_from_str = NaiveDate::parse_from_str;
     ///
     /// assert_eq!(parse_from_str("2015-09-05", "%Y-%m-%d"),
-    ///            Ok(NaiveDate::from_ymd(2015, 9, 5)));
+    ///            Ok(NaiveDate::from_ymd_opt(2015, 9, 5).unwrap()));
     /// assert_eq!(parse_from_str("5sep2015", "%d%b%Y"),
-    ///            Ok(NaiveDate::from_ymd(2015, 9, 5)));
+    ///            Ok(NaiveDate::from_ymd_opt(2015, 9, 5).unwrap()));
     /// ```
     ///
     /// Time and offset is ignored for the purpose of parsing.
@@ -492,7 +492,7 @@ impl NaiveDate {
     /// # use chrono::NaiveDate;
     /// # let parse_from_str = NaiveDate::parse_from_str;
     /// assert_eq!(parse_from_str("2014-5-17T12:34:56+09:30", "%Y-%m-%dT%H:%M:%S%z"),
-    ///            Ok(NaiveDate::from_ymd(2014, 5, 17)));
+    ///            Ok(NaiveDate::from_ymd_opt(2014, 5, 17).unwrap()));
     /// ```
     ///
     /// Out-of-bound dates or insufficient fields are errors.
@@ -526,12 +526,12 @@ impl NaiveDate {
     /// ```
     /// # use chrono::{NaiveDate, Months};
     /// assert_eq!(
-    ///     NaiveDate::from_ymd(2022, 2, 20).checked_add_months(Months::new(6)),
-    ///     Some(NaiveDate::from_ymd(2022, 8, 20))
+    ///     NaiveDate::from_ymd_opt(2022, 2, 20).unwrap().checked_add_months(Months::new(6)),
+    ///     Some(NaiveDate::from_ymd_opt(2022, 8, 20).unwrap())
     /// );
     /// assert_eq!(
-    ///     NaiveDate::from_ymd(2022, 7, 31).checked_add_months(Months::new(2)),
-    ///     Some(NaiveDate::from_ymd(2022, 9, 30))
+    ///     NaiveDate::from_ymd_opt(2022, 7, 31).unwrap().checked_add_months(Months::new(2)),
+    ///     Some(NaiveDate::from_ymd_opt(2022, 9, 30).unwrap())
     /// );
     /// ```
     pub fn checked_add_months(self, months: Months) -> Option<Self> {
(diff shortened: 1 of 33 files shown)


src/naive/date.rs
```diff
@@ -247,7 +247,7 @@ impl NaiveDate {
     /// ```
     /// use chrono::{NaiveDate, Datelike, Weekday};
     ///
-    /// let d = NaiveDate::from_ymd(2015, 3, 14);
+    /// let d = NaiveDate::from_ymd_opt(2015, 3, 14).unwrap();
```
esheppa, line 250: these docs should potentially be removed as has been done in other places

src/naive/date.rs
```diff
@@ -951,14 +932,15 @@ impl NaiveDate {
     /// ```
     /// use chrono::{NaiveDate, NaiveDateTime, Datelike, Timelike, Weekday};
     ///
-    /// let d = NaiveDate::from_ymd(2015, 6, 3);
+    /// let d = NaiveDate::from_ymd_opt(2015, 6, 3).unwrap();
     ///
-    /// let dt: NaiveDateTime = d.and_hms_nano(12, 34, 56, 789_012_345);
+    /// let dt: NaiveDateTime = d.and_hms_nano_opt(12, 34, 56, 789_012_345).unwrap();
```
esheppa, line 937: these docs should potentially be removed as has been done in other places

src/offset/fixed.rs
```diff
@@ -69,10 +70,11 @@ impl FixedOffset {
     /// ```
     /// use chrono::{FixedOffset, TimeZone};
     /// let hour = 3600;
-    /// let datetime = FixedOffset::west(5 * hour).ymd(2016, 11, 08)
-    ///                                           .and_hms(0, 0, 0);
+    /// let datetime = FixedOffset::west_opt(5 * hour).unwrap().ymd_opt(2016, 11, 08).unwrap()
```
esheppa, line 73: these docs should potentially be moved to west_opt

src/offset/fixed.rs
```diff
@@ -39,10 +39,11 @@ impl FixedOffset {
     /// ```
     /// use chrono::{FixedOffset, TimeZone};
     /// let hour = 3600;
-    /// let datetime = FixedOffset::east(5 * hour).ymd(2016, 11, 08)
-    ///                                           .and_hms(0, 0, 0);
+    /// let datetime = FixedOffset::east_opt(5 * hour).unwrap().ymd_opt(2016, 11, 08).unwrap()
```
esheppa, line 42: these docs should potentially be moved to east_opt

src/offset/local/unix.rs
```diff
@@ -126,12 +126,13 @@ impl Cache {
...
         if !local {
-            let offset = FixedOffset::east(
+            let offset = FixedOffset::east_opt(
                 self.zone
                     .find_local_time_type(d.timestamp())
                     .expect("unable to select local time type")
                     .offset(),
-            );
+            )
+            .unwrap();
```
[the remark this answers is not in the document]
esheppa: Interesting question - I don't think `LocalResult::None` is ideal here as it is meant to represent the specific situation where the local timestamp is invalid due to the local offset moving east, however it may be better than panicking, and so could be reasonable here, and then later we could allow more kinds of errors from this function to better signify what has gone wrong

src/offset/mod.rs
```diff
@@ -219,8 +219,9 @@ pub trait TimeZone: Sized + Clone {
     /// ```
     /// use chrono::{Utc, TimeZone};
     ///
-    /// assert_eq!(Utc.ymd(2015, 5, 15).to_string(), "2015-05-15UTC");
+    /// assert_eq!(Utc.ymd_opt(2015, 5, 15).unwrap().to_string(), "2015-05-15UTC");
```
esheppa, line 222: these docs should potentially be removed as has been done in other places

src/offset/mod.rs
```diff
@@ -342,6 +346,14 @@ pub trait TimeZone: Sized + Clone {
     ///
     /// Returns `LocalResult::None` on out-of-range number of seconds and/or
     /// invalid nanosecond, otherwise always returns `LocalResult::Single`.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use chrono::{Utc, TimeZone};
+    ///
+    /// assert_eq!(Utc.timestamp(1431648000, 0).to_string(), "2015-05-15 00:00:00 UTC");
```
esheppa, line 355: should potentially be `timestamp_opt` here

src/offset/mod.rs
```diff
@@ -360,8 +372,9 @@ pub trait TimeZone: Sized + Clone {
     /// ```
     /// use chrono::{Utc, TimeZone};
     ///
-    /// assert_eq!(Utc.timestamp_millis(1431648000).timestamp(), 1431648);
+    /// assert_eq!(Utc.timestamp_millis_opt(1431648000).unwrap().timestamp(), 1431648);
```
esheppa, line 375: these docs should potentially be removed as has been done in other places

src/naive/time/mod.rs
```diff
@@ -255,12 +256,13 @@ impl NaiveTime {
     /// ```
     /// use chrono::{NaiveTime, Timelike};
     ///
-    /// let t = NaiveTime::from_hms_milli(23, 56, 4, 12);
+    /// let t = NaiveTime::from_hms_milli_opt(23, 56, 4, 12).unwrap();
```
esheppa, line 259: these docs should potentially be removed as has been done in other places

src/naive/time/mod.rs
```diff
@@ -207,12 +207,13 @@ impl NaiveTime {
     /// ```
     /// use chrono::{NaiveTime, Timelike};
     ///
-    /// let t = NaiveTime::from_hms(23, 56, 4);
+    /// let t = NaiveTime::from_hms_opt(23, 56, 4).unwrap();
```
esheppa, line 210: these docs should potentially be removed as has been done in other places

src/naive/time/mod.rs
```diff
@@ -307,12 +309,13 @@ impl NaiveTime {
     /// ```
     /// use chrono::{NaiveTime, Timelike};
     ///
-    /// let t = NaiveTime::from_hms_micro(23, 56, 4, 12_345);
+    /// let t = NaiveTime::from_hms_micro_opt(23, 56, 4, 12_345).unwrap();
```
esheppa, line 312: these docs should potentially be removed as has been done in other places

src/naive/time/mod.rs
```diff
@@ -357,12 +360,13 @@ impl NaiveTime {
     /// ```
     /// use chrono::{NaiveTime, Timelike};
     ///
-    /// let t = NaiveTime::from_hms_nano(23, 56, 4, 12_345_678);
+    /// let t = NaiveTime::from_hms_nano_opt(23, 56, 4, 12_345_678).unwrap();
```
esheppa, line 363: these docs should potentially be removed as has been done in other places

esheppa (commented):
This looks excellent and I think this is the right incremental approach to get this done. Hopefully we will get some good feedback once `0.4.23` is released which can be incorporated into extensions on this in `0.5`.

djc:
Good feedback, thanks for catching the messy example situation -- I stumbled upon it late and then figured I'd ask for feedback before doing another round of cleaning that up. I think I've removed all the examples from the non-`opt` methods, and made sure that all `_opt()` methods have at least one example.

src/naive/date.rs
```diff
@@ -1885,7 +1733,7 @@ impl Iterator for NaiveDateDaysIterator {
         // current < NaiveDate::MAX from here on:
         let current = self.value;
         // This can't panic because current is < NaiveDate::MAX:
-        self.value = current.succ();
+        self.value = current.succ_opt().unwrap();
```
esheppa, line 1736: could we use `self.value = current.succ_opt()?;` and then remove the whole `if` block?

src/naive/date.rs
```diff
@@ -1903,7 +1751,7 @@ impl DoubleEndedIterator for NaiveDateDaysIterator {
             return None;
         }
         let current = self.value;
-        self.value = current.pred();
+        self.value = current.pred_opt().unwrap();
```
esheppa, line 1754: could we use `self.value = current.pred_opt()?;` and then remove the whole if block?

esheppa (approved):
Looks good! I've left a few comments in NaiveDate days iterator, also had one general thought:

where there are blocks like:

```rust
    /// assert_eq!(NaiveDate::from_ymd_opt(2015, 6, 3).unwrap().succ_opt(),
    ///            Some(NaiveDate::from_ymd_opt(2015, 6, 4).unwrap()));
```

potentially this style could be used instead:

```rust
    /// assert_eq!(NaiveDate::from_ymd_opt(2015, 6, 3).unwrap().succ_opt(),
    ///            NaiveDate::from_ymd_opt(2015, 6, 4)));
```

however, I don't think its worth doing those changes here as anyway, it may change further if the functions are changed to return result instead in some cases

src/naive/date.rs
djc: I played with something like that at the time, but I don't think so? In the case `succ_opt()` returns `None` but current is `Some` we still have to return that last value. Arguably the iterator contract should be changed but I didn't want to take that on as part of this PR.

esheppa: sounds good - I just remembered why this stood out to me, I've made some changes to it in https://github.com/chronotope/chrono/pull/824/commits/c7e1b9b8deee5cb378766d32d16cabed7184125a anyway, so lets merge this as-is

hoodie:
I have to admit that this makes for some rather verbose code.

djc:
> I have to admit that this makes for some rather verbose code. 

Can you give some examples?

hoodie:
the examples actually already contain that
```rs
let dt: NaiveDateTime = NaiveDate::from_ymd_opt(2016, 7, 8).unwrap().and_hms_opt(9, 10, 11).unwrap();
```

which used to be 
```rs
let dt: NaiveDateTime = NaiveDate::from_ymd(2016, 7, 8).and_hms(9, 10, 11);
```

I completely understand the motivation of having a fallible default that doesn't panic, but for some cases (in my case just example code) it would have been acceptable to have the panicking version as long as it is well documented that it panics.

djc:
Maybe that could make sense? Would you be able to submit a PR for something like that?

esheppa:
@hoodie - I'm keen to understand the use cases for constructing dates from literals vs parsing them. If possible could you give a bit more detail about how/why you are using functions like `from_ymd_opt` and how often that occurs in your code base?. 

In terms of continuing, the `Option` and `Result` helper methods mean you can do something like:

```rust
let dt: NaiveDateTime = NaiveDate::from_ymd_opt(2016, 7, 8).and_then(|d| d.and_hms_opt(9, 10, 11)).unwrap();
```

hoodie:
@esheppa you are actually right. I also only use this in examples, which is why I was a bit disappointed that my own example code started looking a bit cumbersome after fixing these deprecations ;) 

I don't think this needs to be fixed actually, it just makes example code a bit less sexy now

tafia:
Would it be possible to reactivate it under certain conditions?

```rs
// not sure it actually works
#[cfg_attr(not(all(test, feature = "opt_unwrap")), deprecate)]
```

The rationale is
- this crate is used a LOT and testing a particular date is extremely common. The deprecation add lot of noise either in the code or the tests outputs
- I believe dates created in tests are overwhelmingly using valid dates and thus the opt is most of the time useless in that scenario 
  - we're expecting a valid date
  - if this is not a valid date, test will panic which is actually ok!

djc:
@tafia there is some discussion in #815 about adding macros that could statically verify date/time validity.

I don't think doing a Cargo feature would make sense for this.

njaard:
I really don't like the deprecation warning. Now I have to go through all my code and add _opt to hundreds of locations, then when chrono 0.5 comes out, I'll have to do it again, removing the _opt and adding an error check.

Considering that the breaking change in 0.5 will be a compile-time error, could I suggest that the deprecation warning be removed, or made as a feature flag?

djc:
We'll see what we land on for 0.5 -- that will still take some time. For the time being, I don't think we're going to (conditionally) drop the deprecation warning, you can always introduce a wrapper in your own code if you're worried about churn.

Issue #624, opened 2021-11-17
How to get "now" with DateTime<Timezone>?


tobymurray:
Analogous to `Utc::now`, I'd like to get now as a `DateTime<Timezone>` for a given timezone, would this be the idiomatic way to do so?

```
Utc::now().with_timezone(timezone)
```

Similar to #169, I struggled a lot more than I think I ought to have to try and do some relatively straightforward seeming tasks. 

I'd love to:
- hear if there's a better way to do it
- add to the documentation (open to suggestions where would make a good fit, maybe right on [DateTime#now](https://docs.rs/chrono/0.4.19/chrono/offset/struct.Utc.html#method.now)?)
- commiserate with how hard it is to make an intuitive time library when it's so damn complicated! Thanks for all the work.

epipheus:
I'd like to mention that many of us use DateTimeWithTimezone as DB storage by default but optionally still log UTC based date times which is just +0. The current `Utc::now().with_timezone(&FixedOffset::east(0))` feels non-obvious and verbose and isn't terribly well documented. as mentioned in #169. `Utc::now.as_fixed_offset()` or `Utc::now.as_date_with_timezone()`. Even is as verbose, feels readable and obvious.  My 2 cents.

pitdicker:
> Analogous to `Utc::now`, I'd like to get now as a `DateTime<Timezone>` for a given timezone, would this be the idiomatic way to do so?
> 
> ```rust
> Utc::now().with_timezone(timezone)
> ```

Yes, this is the idiomatic way to do it.

https://github.com/chronotope/chrono/pull/1030 added `DateTime::fixed_offset` and https://github.com/chronotope/chrono/pull/1192 will add some examples to the documentation on using `Utc::now()` and `Local::now()`.

Issue #169, opened 2017-06-29
Convience method for converting from DateTime<FixedOffset> to UTC


glennpierce:
Hi

Currently I am doing something like

let dt : DateTime<FixedOffset> = DateTime::parse_from_rfc3339("1970-01-01T00:00:00+00:00")?;
let dt_utc : DateTime<UTC> = DateTime::<UTC>::from_utc(dt.naive_utc(), UTC);

Is there any more convenient way todo this and vice versa ?

Thanks

quodlibetor:
The fact that `Utc` is a type (and therefore can't depend on anything at run-time) means that it's impossible for methods on `DateTime<FixedOffset>` (which is where the `parse_from_rfc3339` is implemented) to parse to a `DateTime<Utc>`. 

However, we could probably implement either a `DateTime<Utc>::parse_from_rfc3339()` (or probably nicer a `Utc::parse_from_rfc3339()`) that does the parsing and auto-timezone conversion. 

Doing _that_ depends a bit on the philosophical outlook of this trait: whether we want to implement a whole bunch of kind-of-trivial ergonomic methods or if we want to keep the API surface small. We're currently thinking about the tradeoffs of the two philosophies.

glennpierce:
Thanks

nathanaeljones:
I've spent about 30 minutes trying to figure out how to correctly convert between DateTime<Utc> and DateTime<FixedOffset>.  I strongly support making that conversion more ergonomic.

Walther:
If a convenience method (e.g. `FixedOffset::now()`) would not be feasible, maybe the documentation could be improved? On a quick reading, the Readme seems to warmly suggest using `FixedOffset` as "it stores the most information and does not depend on the system environment". However, there's no instruction / example on how to properly use it.

djc:
I'm going to close this issue. The original request has been addressed with #806 in that you can now call `parse_from_rfc3339()` on `DateTime<Utc>`. Additionally, an `impl From<DateTime<FixedOffset>> for DateTime<Utc>` already exists.

However, I'm definitely open to the idea that ergonomic issues still exist here! In order to address those, please open a new issue with some example code of where you're running into trouble/a conversion that seems to require more complex code than what you feel is necessary.

PR #828, opened 2022-09-28
Adding now() functionality for FixedOffest and a Utc::fixed_now() for DateTime w/ TimeZone (set to 0)

epipheus:
I'm new to Rust so please review.  If this is an acceptable way of doing it I can clean it up and maybe add tests.

added a `now()` function to Offset trait and a `fixed_now()` function to Utc that calls the trait's now. FixedOffset's `now()` function get the duration from the system the same way Utc::now does but afterwards uses timestamp_opt to return DateTime<FixedOffset>. 

This was in response to related comment in #624 and #169 

(again I'm new to Rust so forgive any meandering commits, I can make a cleaner one if this is an acceptable set of changes.)

### DIFF
--- a/src/offset/fixed.rs
+++ b/src/offset/fixed.rs
@@ -129,7 +129,7 @@ impl Offset for FixedOffset {
         *self
     }
     fn now(&self) -> DateTime<FixedOffset> {
-        Utc::now()
+        self.from_utc_datetime(&Utc::now().naive_utc())
     }
 }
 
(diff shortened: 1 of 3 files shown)


epipheus:
on my end tests pass

```
test result: ok. 258 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 24.75s
```

Is there a preferred stable for PRs?

djc:
Try running the following commands:

* `cargo fmt -- --check`
* `cargo check --no-default-features`

djc:
> Is there a preferred stable for PRs?

Not sure what you mean by this. This PR should probably target the 0.4.x branch.

epipheus:
ok... let me get this in some shape and come back.

PR #829, opened 2022-09-29
now() functionality for FixedOffest and a Utc::fixed_now() for DateTime w/ TimeZone (set to 0)

epipheus:
Trying this again, forgive me if this is annoying, I'm new to Rust, so you'll have to review to make sure I didn't screw anything up. 

added a now() function to Offset trait and a fixed_now() function to Utc that calls the trait's now. FixedOffset's now() function get the duration from the system the same way Utc::now does but afterwards uses timestamp_opt to return DateTime.

This was in response to related comment in https://github.com/chronotope/chrono/issues/624 and https://github.com/chronotope/chrono/issues/169


src/offset/utc.rs
```diff
@@ -3,7 +3,14 @@
 
 //! The UTC (Coordinated Universal Time) time zone.
 
+use super::{FixedOffset, LocalResult, Offset, TimeZone};
```
djc, line 6: Please don't make unrelated changes (the imports are sorted and split this way on purpose).

src/offset/mod.rs
```diff
@@ -194,6 +194,9 @@ impl<T: fmt::Debug> LocalResult<T> {
 pub trait Offset: Sized + Clone + fmt::Debug {
     /// Returns the fixed offset from UTC to the local time stored.
     fn fix(&self) -> FixedOffset;
+    /// Returns a fixed offset DateTime at the current date and time in self's timezone
```
djc, line 197: I don't think we should add this to the trait.

src/offset/fixed.rs
```diff
@@ -6,15 +6,17 @@
 use core::fmt;
 use core::ops::{Add, Sub};
 
-use num_integer::div_mod_floor;
```
djc, line 9: Please revert the unrelated import moves.

src/offset/utc.rs
```diff
@@ -102,6 +107,10 @@ impl Offset for Utc {
     fn fix(&self) -> FixedOffset {
         FixedOffset::east(0)
     }
+    #[cfg(feature = "std")]
```
djc, line 110: I don't think adding this makes sense.

src/offset/fixed.rs
```diff
@@ -128,6 +130,12 @@ impl Offset for FixedOffset {
     fn fix(&self) -> FixedOffset {
         *self
     }
+    #[cfg(feature = "std")]
+    fn now(&self) -> DateTime<FixedOffset> {
+        let now =
```
djc, line 135: How about writing this as `Utc::now().with_timezone(self)`?

djc (commented):
This also needs to at least one test.

epipheus:
I'll get it right eventually. Apologies.

src/offset/utc.rs
epipheus: apologies, overbearing IDE.

src/offset/fixed.rs
epipheus: I started that way then eventually wondered if FixedOffset should have it's own now()

src/offset/utc.rs
epipheus: If FixedOffset had it's own now that didn't need Utc but required systemtime which is in std...

src/offset/fixed.rs
epipheus: oh. i understand. yes.

esheppa:
Hi @epipheus  - thanks for setting up a PR for this - one suggestion, you could squash all your commits together using `git rebase -i`. It looks like some extra commits have also been added in here due to the change of base branch, so if you squash your commits, you can then `cherry-pick` that resulting commit into a new branch and PR which can directly target `0.4.x`.

In terms of the code, another potential option is extending the `TimeZone` trait, instead of the `Offset` trait, with something like:

```rust
pub trait TimeZone: Sized + Clone {
    // other methods skipped for brevity
    #[cfg(feature = "clock")]
    /// Get the current time in the specified timezone
    fn now(&self) -> DateTime<Self> {
        Self::from_utc_datetime(self, &Utc::now().naive_utc())
    }
}
```

this should enable things like:

```rust
Utc.now() // (same as `Utc::now()`)
Local.now() // (same as `Local::now()`)
FixedOffset::east(3600).now()
chrono_tz::Pacific::Auckland.now()
```


In terms of converting from `Utc` to `FixedOffset` - this is a little tricky as `FixedOffset` is a bit special in that it maintains state (the offset), so this either needs to be one of:
* special case inherent method on `DateTime` (probably the best option as it seems like it might be a commonly used function?), or
* use an instance of `FixedOffset::east(0)` and then one of:
    *  `FixedOffset::east(0).from_utc_datetime(utc_datetime.naive_utc())`
    * `utc_datetime.with_timezone(&FixedOffset::east(0))`

However I think there is a reasonable solution that could solve this generically as a inherent impl on `DateTime`:
```rust
/// Converts the timezone to just the current offset throwing away any extra information
impl<Tz: TimeZone> DateTime<Tz> {
    pub fn with_fixed_offset(self) -> DateTime<FixedOffset> {
        self.with_timezone(&self.offset.fix())
    }
}
```

epipheus:
Thanks for much for the comments and thanks for bearing with me, I am far from awesome in Rust but doing something will hopefully get me there.  I never knew anything about cherry-pick so this is cool, I was going to just start over from a fresh fork especially since I disturbed import order, etc. 

Will attempt doing what you said without messing up after the weekend.

pitdicker:
@epipheus Is this something you still want to work on?

epipheus:
Apologies for massive delay. At the time had a death in the family and I definitely completely fell off for quite a while. I also thought I saw a recent commit that perhaps made this moot. Happy to take a look.

pitdicker:
I am sorry to hear to hear that.

To be honest I don't know how much we really need this feature with recent commits.
Some examples of what we can currently write:
```rust
let now_utc = Utc::now();
let local_now = Local::now();

// `DateTime<FixedOffset>` with offset = +00:00
let now_utc_fixed = Utc::now().fixed_offset();
// or
let now_utc_fixed: DateTime<FixedOffset> = Utc::now().into();

// `DateTime<FixedOffset>` with offset of system timezone
let local_now_fixed = Local::now();
// or
let local_now_fixed: DateTime<FixedOffset> = Local::now().into();

// Convert to another timezone:
let offset = FixedOffset::east_opt(5 * 60 * 60).unwrap();
let now_with_offset = Utc::now().with_timezone(&offset);
```

It don't think adding `now()` to the `TimeZone` trait as suggested by @esheppa would be a good idea. It would cause the available methods of the trait to change depending on whether the `clock` features is enabled.

If you want to add something, maybe adding it to `DateTime<Tz>` would be a nice place to make this functionality available for all timezones.

epipheus:
This feels much better. Maybe we can just close this with current commits really covering this well.

epipheus:
Sure. Adding to time zones would be nice.

pitdicker:
> If you want to add something, maybe adding it to `DateTime<Tz>` would be a nice place to make this functionality available for all timezones.

I thought it would be possible to add something like this:
```rust
impl<Tz: TimeZone> DateTime<Tz> {
    /// Returns a `DateTime` which corresponds to the current date and time.
    #[cfg(feature = "clock")]
    #[must_use]
    pub fn now() -> DateTime<Tz> {
        Utc::now().into()
    }
}
```

But that doesn't work. We have no way to generically construct a `DateTime` with some arbitrary type that implements `TimeZone`.

Okay. I'll close this PR, but will open one to add the examples from https://github.com/chronotope/chrono/pull/829#issuecomment-1642683903 to the documentation somewhere.

PR #837, opened 2022-10-08
fix benches - plug in datetime type parameter

esheppa:
The benches are currently not compiling due to this. Potentially they should be added to the tests, or just a `cargo check --benches`. Happy to add in this PR but will wait until #836 is merged first and rebase.


djc (approved):

djc:
Nice!

PR #842, opened 2022-10-11
Merge 0.4.x (deprecated non-`_opt()` methods)

djc:


esheppa (approved):
Looks good - good idea to keep this up to date frequently

src/datetime/rustc_serialize.rs
```diff
@@ -0,0 +1,123 @@
+#![cfg_attr(docsrs, doc(cfg(feature = "rustc-serialize")))]
```
esheppa, line 1: this file should be removed in 0.5

djc:
It's less about time-based frequency and more about not having too many conflicts collected.

PR #843, opened 2022-10-11
Fallible `Date` internals

djc:


esheppa:
This looks really good, I like the idea of applying the fallible functions everywhere - I've added https://github.com/chronotope/chrono/pull/845 as some potential extras to add in here, depending on your thoughts, but otherwise happy with this as it is

PR #844, opened 2022-10-15
optimise rfc3339 (and rfc2822)

conradludgate:
We use `to_rfc3339()` a lot in our observability libraries at TrueLayer. Upon profiling, I noticed that it took up a large portion of time, so I looked into optimising it.

Screenshot 2022-10-15 at 10 43 30

Screenshot 2022-10-15 at 10 43 39

I was able to do more optimisations but the requirement of the year formatting causes the code to be pretty tricky. I can attempt it in a later PR if this is accepted.

In our observability code I know that the year is 0..=9999 and the timezone is Utc so I had a bit more gains from cheating 😅

### DIFF
--- a/src/format/mod.rs
+++ b/src/format/mod.rs
@@ -700,19 +700,7 @@ fn format_inner<'a>(
                     // same as `%a, %d %b %Y %H:%M:%S %z`
                     {
                         if let (Some(d), Some(t), Some(&(_, off))) = (date, time, off) {
-                            let sec = t.second() + t.nanosecond() / 1_000_000_000;
-                            write!(
-                                result,
-                                "{}, {:02} {} {:04} {:02}:{:02}:{:02} ",
-                                locale.short_weekdays[d.weekday().num_days_from_sunday() as usize],
-                                d.day(),
-                                locale.short_months[d.month0() as usize],
-                                d.year(),
-                                t.hour(),
-                                t.minute(),
-                                sec
-                            )?;
-                            Some(write_local_minus_utc(result, off, false, Colons::None))
+                            Some(write_rfc2822_inner(result, d, t, off, locale))
                         } else {
                             None
                         }
@@ -783,6 +771,46 @@ pub(crate) fn write_rfc3339(
     // this is faster in this way.
     write!(result, "{:?}", dt)?;
     write_local_minus_utc(result, off, false, Colons::Single)
+}
+
+#[cfg(any(feature = "alloc", feature = "std", test))]
+/// write datetimes like `Tue, 1 Jul 2003 10:52:37 +0200`, same as `%a, %d %b %Y %H:%M:%S %z`
+pub(crate) fn write_rfc2822(
+    result: &mut String,
+    dt: crate::NaiveDateTime,
+    off: FixedOffset,
+) -> fmt::Result {
+    write_rfc2822_inner(result, &dt.date(), &dt.time(), off, Locales::new(None))
+}
+
+#[cfg(any(feature = "alloc", feature = "std", test))]
+/// write datetimes like `Tue, 1 Jul 2003 10:52:37 +0200`, same as `%a, %d %b %Y %H:%M:%S %z`
+fn write_rfc2822_inner(
+    result: &mut String,
+    d: &NaiveDate,
+    t: &NaiveTime,
+    off: FixedOffset,
+    locale: Locales,
+) -> fmt::Result {
+    let year = d.year();
+    // RFC2822 is only defined on years 0 through 9999
+    if !(0..=9999).contains(&year) {
+        return Err(fmt::Error);
+    }
+
+    let sec = t.second() + t.nanosecond() / 1_000_000_000;
+    write!(
+        result,
+        "{}, {:02} {} {:04} {:02}:{:02}:{:02} ",
+        locale.short_weekdays[d.weekday().num_days_from_sunday() as usize],
+        d.day(),
+        locale.short_months[d.month0() as usize],
+        year,
+        t.hour(),
+        t.minute(),
+        sec
+    )?;
+    write_local_minus_utc(result, off, false, Colons::None)
 }
 
 /// Tries to format given arguments with given formatting items.
--- a/src/format/mod.rs
+++ b/src/format/mod.rs
@@ -40,6 +40,8 @@ use alloc::string::{String, ToString};
 #[cfg(any(feature = "alloc", feature = "std", test))]
 use core::borrow::Borrow;
 use core::fmt;
+#[cfg(any(feature = "alloc", feature = "std", test))]
+use core::fmt::Write;
 use core::str::FromStr;
 #[cfg(any(feature = "std", test))]
 use std::error::Error;
@@ -501,7 +503,6 @@ fn format_inner<'a>(
         )
     };
 
-    use core::fmt::Write;
     use num_integer::{div_floor, mod_floor};
 
     match *item {
@@ -576,45 +577,6 @@ fn format_inner<'a>(
 
         Item::Fixed(ref spec) => {
             use self::Fixed::*;
-
-            /// Prints an offset from UTC in the format of `+HHMM` or `+HH:MM`.
-            /// `Z` instead of `+00[:]00` is allowed when `allow_zulu` is true.
-            fn write_local_minus_utc(
-                result: &mut String,
-                off: FixedOffset,
-                allow_zulu: bool,
-                colon_type: Colons,
-            ) -> fmt::Result {
-                let off = off.local_minus_utc();
-                if !allow_zulu || off != 0 {
-                    let (sign, off) = if off < 0 { ('-', -off) } else { ('+', off) };
-
-                    match colon_type {
-                        Colons::None => {
-                            write!(result, "{}{:02}{:02}", sign, off / 3600, off / 60 % 60)
-                        }
-                        Colons::Single => {
-                            write!(result, "{}{:02}:{:02}", sign, off / 3600, off / 60 % 60)
-                        }
-                        Colons::Double => {
-                            write!(
-                                result,
-                                "{}{:02}:{:02}:{:02}",
-                                sign,
-                                off / 3600,
-                                off / 60 % 60,
-                                off % 60
-                            )
-                        }
-                        Colons::Triple => {
-                            write!(result, "{}{:02}", sign, off / 3600)
-                        }
-                    }
-                } else {
-                    result.push('Z');
-                    Ok(())
-                }
-            }
 
             let ret =
                 match *spec {
@@ -735,10 +697,7 @@ fn format_inner<'a>(
                     // same as `%Y-%m-%dT%H:%M:%S%.f%:z`
                     {
                         if let (Some(d), Some(t), Some(&(_, off))) = (date, time, off) {
-                            // reuse `Debug` impls which already print ISO 8601 format.
-                            // this is faster in this way.
-                            write!(result, "{:?}", crate::NaiveDateTime::new(*d, *t))?;
-                            Some(write_local_minus_utc(result, off, false, Colons::Single))
+                            Some(write_rfc3339(result, crate::NaiveDateTime::new(*d, *t), off))
                         } else {
                             None
                         }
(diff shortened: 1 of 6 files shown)


djc:
Nice results!

Is the datetime_from_str regression stable? Why did you move all that code out of the Debug impls into an inherent method?

conradludgate:
> Why did you move all that code out of the Debug impls into an inherent method

I was just adding some comments to the code 😅 
This allows for better code-reuse while allowing more inlining to avoid dynamic dispatch overhead

> Is the datetime_from_str regression stable

Yeah I'm not sure on that one. I'm still looking into it

conradludgate:
Ok, I'm done tinkering 🙏

src/format/mod.rs
```diff
@@ -756,6 +707,159 @@ fn format_inner<'a>(
...
+    d: &NaiveDate,
+    t: &NaiveTime,
+    off: FixedOffset,
+) -> fmt::Result {
+    #[cfg(feature = "unstable-locales")]
+    let (short_months, short_weekdays) =
+        (locales::short_months(Locale::POSIX), locales::short_weekdays(Locale::POSIX));
+    #[cfg(not(feature = "unstable-locales"))]
+    let (short_months, short_weekdays) = (
+        &["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"],
```
esheppa, line 797: Could these potentially be broken out to a const, which could be shared with the ones at the top of `format_inner`

esheppa (commented):
Looks great, it's nice to specialize the `to_rfc*` implementations and that is an impressive performance uplift

src/format/mod.rs
```diff
@@ -756,6 +707,159 @@ fn format_inner<'a>(
...
+
+    result.push_str(short_months[d.month0() as usize]);
+
+    // year
+    {
+        // 4 digit year 1234
+        // is split into cent=12 and subcent = 34
+        let cent = d.year() / 100;
+        let subcent = d.year() % 100;
+        bytes[1] = b'0' + (cent / 10) as u8;
```
esheppa, line 833: potentially some of this can be shared with `NaiveDate::write_into`

conradludgate: Not exactly sure what you had in mind, but I added a `write_double_digit` method which makes this clearer

src/format/mod.rs
conradludgate: I've split it into a `Locales` struct for now, I'm guessing locale code still needs a lot more love so I don't think I should make much more changes to it

djc:
So can you explain what changes caused the majority of the performance improvements?

In particular I'm still unclear why it's good for performance to move the `Debug` code into `write_into()` methods; is that actually necessary for performance or is there a way of calling `Debug::fmt` without a performance penalty relative to calling `write_into()`? All other things being equal, I think leaving that code in the `Debug` impl would be more idiomatic.

conradludgate:
Using the format methods needs a Formatter type. We don't always have access to one in our implementations. Currently in standard there's no way to use fmt into a String without using dynamic dispatch. This is what the generic `write_into` methods provide by making it generic over `fmt::Write`. This then supports Formatter and String directly, and it inlines very well.

I'm no longer sure if the byte arrays provide a huge benefit of speed, but it is the only way to avoid the dyn overhead of integer writing, so it remains for now. Initially the idea was that writing into a static bugger would reduce the number of allocation checks and potential reallocs, although considering we specify the capacity upfront I believe this is negligible

conradludgate:
I've just stumbled upon ufmt, which is a dyn-free formatting machinery library and it shows similar gains <https://docs.rs/ufmt/latest/ufmt/#benchmarks>

Their `Debug` trait is basically what the `write_into` is trying to be, since its generic over the writer

src/format/mod.rs
```diff
@@ -577,77 +596,41 @@ fn format_inner<'a>(
         Item::Fixed(ref spec) => {
             use self::Fixed::*;
 
-            /// Prints an offset from UTC in the format of `+HHMM` or `+HH:MM`.
-            /// `Z` instead of `+00[:]00` is allowed when `allow_zulu` is true.
-            fn write_local_minus_utc(
```
djc, line 582: Would like a separate commit for extracting this function.

src/format/mod.rs
```diff
@@ -453,34 +453,34 @@ pub fn format_item<'a>(
...
-            &[
+struct Locales {
+    short_months: &'static [&'static str],
+    long_months: &'static [&'static str],
+    short_weekdays: &'static [&'static str],
+    long_weekdays: &'static [&'static str],
+    am_pm: &'static [&'static str],
+}
+#[cfg(any(feature = "alloc", feature = "std", test))]
+impl Locales {
```
djc, line 465: Please create a separate commit or PR for the structural changes here (and add an empty line between the `struct` definition and the `impl` block).

src/format/mod.rs
```diff
@@ -577,77 +596,41 @@ fn format_inner<'a>(
...
-                            });
+                        let ampm = if t.hour12().0 { locale.am_pm[1] } else { locale.am_pm[0] };
+                        for char in ampm.chars() {
+                            result.extend(char.to_lowercase())
                         }
                         Ok(())
                     }),
                     UpperAmPm => time.map(|t| {
-                        result.push_str(if t.hour12().0 { am_pm[1] } else { am_pm[0] });
+                        result.push_str(if t.hour12().0 {
```
djc, line 629: I think a `match` expression would be more readable here.

src/format/mod.rs
```diff
@@ -714,19 +697,7 @@ fn format_inner<'a>(
...
-                                short_weekdays[d.weekday().num_days_from_sunday() as usize],
-                                d.day(),
-                                short_months[d.month0() as usize],
-                                d.year(),
-                                t.hour(),
-                                t.minute(),
-                                sec
-                            )?;
-                            Some(write_local_minus_utc(result, off, false, Colons::None))
+                            Some(write_rfc2822_inner(result, d, t, off, locale))
```
djc, line 700: Would like a separate commit for this function extraction.

src/naive/time/mod.rs
```diff
@@ -1140,6 +1142,45 @@ impl Sub<NaiveTime> for NaiveTime {
...
+
+        let date_len = if nano == 0 {
+            8
+        } else if nano % 1_000_000 == 0 {
+            12
+        } else if nano % 1_000 == 0 {
+            15
+        } else {
+            18
+        };
```
djc, line 1165: This could be more compactly written with a `match` expression, I think.

src/format/mod.rs
```diff
@@ -756,6 +724,128 @@ fn format_inner<'a>(
...
+/// `Z` instead of `+00[:]00` is allowed when `allow_zulu` is true.
+#[cfg(any(feature = "alloc", feature = "std", test))]
+fn write_local_minus_utc(
+    result: &mut String,
+    off: FixedOffset,
+    allow_zulu: bool,
+    colon_type: Colons,
+) -> fmt::Result {
+    let mut off = off.local_minus_utc();
+    if !allow_zulu || off != 0 {
```
djc, line 752: Would prefer to invert the condition and do an early return here.

djc (commented):
Okay, I reviewed this in more detail and I think all the changes here are good, but I would like to have it split up in smaller commits in order to review the changes more carefully. Probably one commit for moving things into `write_into()`, also added a bunch of suggestions for things I think should be separate commits (separate PRs is also fine if you prefer).

I also think the docstrings you added need some work. Please write docstrings that describe the invariants/interface the function provides, instead of adding comments-as-docstrings. Preferably the first line of a docstring can stand alone and is no longer than a single line.

I would also like some benchmarks that compare writing into a byte array, then copying into a `String` with directly writing into the `String` (should probably reserve capacity up front). I'm definitely not a fan of the `write_utf8_bytes()` strategy...

src/format/mod.rs
```diff
@@ -790,21 +789,33 @@ fn write_rfc2822_inner(
...
+    result.push(' ');
     write_local_minus_utc(result, off, false, Colons::None)
 }
 
+/// Equivalent to `{:02}` formatting for n < 100.
+pub(crate) fn write_hundreds(w: &mut impl Write, n: u8) -> fmt::Result {
+    let tens = b'0' + n / 10;
+    let ones = b'0' + n % 10;
+    w.write_char(tens as char)?;
+    w.write_char(ones as char)
```
conradludgate, line 816: This ended up being faster than `str::from_utf8()`, although it's a touch slower than `from_utf8_unchecked`. But it seems acceptable to avoid unsafe here

esheppa:
> So can you explain what changes caused the majority of the performance improvements?

I did some experimentation around this, and found:
* Pre allocating in the `format` function, gave about 15% of speedup
* The new implementations of RFC3339 and RFC2822 alone, but with the original code in `to_rfc*` (eg with the `Item::Fixed(Fixed::Rfc3339)`, gives about half of the speedup
* My assumption is that the rest comes from the specialized impl avoiding the `format` function

conradludgate:
@djc the commits are all there now, in small logical chunks with the benchmark results in the commit message. Hope that helps

src/format/mod.rs
djc: This one has not been resolved.

djc: Ah, never mind, that was fixed in a later commit.

djc (approved):
This is looking great, thanks!

src/format/mod.rs
conradludgate: Do you still want me to make this change? I was skeptical since it wasn't related to this PR but I'm happy to do it

djc: Let's leave it like this.

src/format/mod.rs
```diff
@@ -756,6 +726,100 @@ fn format_inner<'a>(
...
+    write_hundreds(result, t.minute() as u8)?;
+    result.push(':');
+    let sec = t.second() + t.nanosecond() / 1_000_000_000;
+    write_hundreds(result, sec as u8)?;
+    result.push(' ');
+    write_local_minus_utc(result, off, false, Colons::None)
+}
+
+/// Equivalent to `{:02}` formatting for n < 100.
+pub(crate) fn write_hundreds(w: &mut impl Write, n: u8) -> fmt::Result {
```
esheppa, line 816: should we have any extra protection (eg debug_assert or returning an error) where n is >= 100?

src/format/mod.rs
```diff
@@ -756,6 +726,100 @@ fn format_inner<'a>(
...
+    off: FixedOffset,
+    locale: Locales,
+) -> fmt::Result {
+    result.push_str(locale.short_weekdays[d.weekday().num_days_from_sunday() as usize]);
+    result.push_str(", ");
+    write_hundreds(result, d.day() as u8)?;
+    result.push(' ');
+    result.push_str(locale.short_months[d.month0() as usize]);
+    result.push(' ');
+    write_hundreds(result, (d.year() / 100) as u8)?;
```
esheppa, line 802: do we need something like `((d.year() % 10_000) / 100)` or alternatively a check that the year isn't >= 10000?

src/naive/date.rs
```diff
@@ -1820,14 +1820,22 @@ impl DoubleEndedIterator for NaiveDateWeeksIterator {
...
+            crate::format::write_hundreds(f, (year / 100) as u8)?;
+            crate::format::write_hundreds(f, (year % 100) as u8)?;
         } else {
             // ISO 8601 requires the explicit sign for out-of-range years
-            write!(f, "{:+05}-{:02}-{:02}", year, mdf.month(), mdf.day())
+            write!(f, "{:+05}", year)?;
         }
+
+        f.write_char('-')?;
+        crate::format::write_hundreds(f, mdf.month() as u8)?;
```
esheppa, line 1836: this is more of a nit, but it would be nice to have just `format::write_hundreds` here

esheppa (approved):
Thanks for this @conradludgate - I'm happy with this as is but I've left a few comments to discuss prior to merging

greyblake:
@djc @conradludgate Any updates? (looking very much for the release https://github.com/chronotope/chrono/issues/850 :eyes: )

src/format/mod.rs
conradludgate: I suppose that's a valid concern

src/format/mod.rs
conradludgate: Hmm. Currently for years > 9999 we print the full year, but RFC2822 is limited to 0..=9999. Not sure what's the best approach here

src/format/mod.rs
esheppa: It's a bit unfortunate that we are restricted to `fmt::Result` - I wonder whether it is worth considering using `ufmt` here (and later potentially elsewhere as well). @djc what are your thoughts on the extra dependency?

An alternative thought is that we should avoid the `std::fmt` API and just use `&mut String` everywhere until the last moment, allowing better control over potential errors.

However, in the interest of getting this merged, those problems should probably be solved elsewhere. I wonder whether just checking if the `u8` is valid and otherwise returning `fmt::Error` would be sufficient here?

src/format/mod.rs
esheppa: As above, potentially we can return `fmt::Error` if the year is out of range?

danielhenrymantilla (approved):

src/format/mod.rs
djc: That seems like a reasonable suggestion to me.

src/format/mod.rs
djc: > I wonder whether just checking if the `u8` is valid and otherwise returning `fmt::Error` would be sufficient here?

Sounds good to me.

djc: I think chrono should have pretty much no dependencies in the default configuration, probably with the exception of things like iana-time-zone that help us paper over platform differences. We can add other optional dependencies, I guess, like stuff for locales and (de)serialization. Maybe ufmt could be an opt-in dependencies? Not sure how much complexity that would add.

djc:
Thanks for sticking with it!

PR #845, opened 2022-10-16
move more code into internals

esheppa:
My goals here are:
* Move any bitwise and other "internal" style operations into internals.rs (this could be split to modules later)
* Use only methods on the types from internals in naive/date.rs and naive/week.rs
* Given the Mdf::new and Of::new functions return an option, do the validation in the constructor

### DIFF
--- a/src/datetime/rustc_serialize.rs
+++ b/src/datetime/rustc_serialize.rs
@@ -58,7 +58,7 @@ pub struct TsSeconds<Tz: TimeZone>(DateTime<Tz>);
 
 #[allow(deprecated)]
 impl<Tz: TimeZone> From<TsSeconds<Tz>> for DateTime<Tz> {
-    /// Pull the inner DateTime<Tz> out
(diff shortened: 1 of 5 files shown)


djc:
Hah, I was also thinking about wrapping `DateImpl` in a newtype here. In this case, I don't think this needs to be merged into my PR before that gets merged. I'd also request that this is split out into smaller commits (for example, one per bullet point in the PR description?). It would be nice to split renames out from other work, for example.

esheppa:
Good call - I'm even keen to go ahead and split `internals` to modules to get a bit more safety around the constructors

djc:
Makes sense, although I'd prefer inline modules given the size of the internals module.

esheppa:
after putting this in modules the code was looking a bit tangled, so I'm trying out a different path, but will focus on the other PRs first

esheppa:
Closing in favour of https://github.com/chronotope/chrono/pull/882

PR #848, opened 2022-10-17
Impl Arbitrary for Naive{Date,Time,DateTime}

greyblake:
I'd love to see https://github.com/chronotope/chrono/pull/559 moving forward, but looks like @asayers has no time to address the code review right now. 
Since I am very interested in the support of `Arbitrary` I've decided to cherry pick his commit and address the code review.

## What is in this PR:
* Implementation of `Arbitrary` behind feature flag for `NaiveDate`, `NaiveTime`, `NaiveDateTime`.
* Unit tests for `NaiveDate` and `NaiveTime` (no unit test for `NaiveDateTime`, since it's implementation of `Arbitrary` is derived)
* The implementation and tests are placed into `arbitrary` modules (in the same fashion like it's done with `serde`)

## Considerations

Despite 
```rust
let ord = u.int_in_range(1..=366)?;
```
 does not make large difference when constructing `NaiveDate` as @djc mentioned, I've decided to use `YearFlags`, so the both implementations for `NaiveDate` and `NaiveTime` are implemented in similar infallible fashion.

### DIFF
--- a/src/naive/time/arbitrary.rs
+++ b/src/naive/time/arbitrary.rs
@@ -1,0 +1,44 @@
+#![cfg_attr(docsrs, doc(cfg(feature = "arbitrary")))]
+
+use super::NaiveTime;
+use arbitrary::{Arbitrary, Unstructured};
+
+impl Arbitrary<'_> for NaiveTime {
+    fn arbitrary(u: &mut Unstructured) -> arbitrary::Result<NaiveTime> {
+        let secs = u.int_in_range(0..=86_399)?;
+        let nano = u.int_in_range(0..=1_999_999_999)?;
+        let time = NaiveTime::from_num_seconds_from_midnight_opt(secs, nano)
+            .expect("Could not generate a valid chrono::NaiveTime. It looks like implementation of Arbitrary for NaiveTime is erroneous.");
+        Ok(time)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    const UNSTRCUTURED_BINARY1: [u8; 8] = [0x8f, 0xc2, 0x95, 0xdc, 0x3e, 0x45, 0xb2, 0x3e];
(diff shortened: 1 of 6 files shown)


djc:
What is the value of adding unit tests here? The code seems relatively simple.

src/naive/time/arbitrary.rs
```diff
@@ -0,0 +1,44 @@
+#![cfg_attr(docsrs, doc(cfg(feature = "arbitrary")))]
+
+use super::NaiveTime;
+use arbitrary::{Arbitrary, Unstructured};
+
+impl Arbitrary<'_> for NaiveTime {
+    fn arbitrary(u: &mut Unstructured) -> arbitrary::Result<NaiveTime> {
+        let secs = u.int_in_range(0..=86_399)?;
+        let nano = u.int_in_range(0..=1_999_999_999)?;
+        let time = NaiveTime::from_num_seconds_from_midnight_opt(secs, nano)
```
djc, line 10: What's the rationale for panicking here vs returning an error?

djc (commented):
Thanks for picking this up!

src/naive/time/arbitrary.rs
greyblake: My primary motivation was this comment https://github.com/chronotope/chrono/pull/559#discussion_r996506593, which you thumbed up.
I am OK with both approaches, both have their pros and cons. With this one (unwrap/expect) we'll detect the problem faster if somethings gets broken. (Arbitrary is implement that way, that if something returns an error, it retries to generate data until it succeeds).

greyblake:
> What is the value of adding unit tests here? The code seems relatively simple.

I have this habit of adding them to almost every change I do. If you think they are not necessary I can drop them. Please let me know.

djc:
Just retarget this PR to 0.4.x, changes that get merged there eventually are merged into main.

greyblake:
@djc Thanks for the feedback. So I am closing this one in favor of https://github.com/chronotope/chrono/pull/849

Issue #820, opened 2022-09-13
Rename `NaiveDate` and potentially remove `Date<Tz>`


esheppa:
First some past commentary:

Chrono RFC's: https://github.com/chronotope/chrono-rfcs/blob/master/notes/api-issues.md
https://github.com/chronotope/chrono-rfcs/issues/1
https://github.com/chronotope/chrono-rfcs/issues/3

https://github.com/chronotope/chrono/issues/204
> ...which will allow us to deprecate Date<Tz> and rename NaiveDate to Date, over a couple releases.

I think #204 has the right approach here, and this could be combined or kept separate from other renaming (eg potential renaming of `NaiveTime`):

* Remove `Date<Tz>`
* Rename `NaiveDate` to `Date`

We now have functions like [`and_local_timezone`](https://docs.rs/chrono/latest/chrono/naive/struct.NaiveDateTime.html#method.and_local_timezone) which offer an alternative way of constructing literals in a given timezone, however I think that making literals in a given timezone like `Utc.ymd(2018, 1, 1).and_hms(1, 2, 3)` is probably not the common case anyway, as they would often either be parsed, be the current time, or be some function of one of the previous.

We could also use `LocalDate` instead of `Date` as mentioned in https://github.com/chronotope/chrono-rfcs/issues/3#issue-388096479 which is what java.time / NodaTime use - however my preference is just `Date`.

djc:
I am very on board with ditching `Date<Tz>` and renaming `NaiveDate` to `Date`.

mqudsi:
Just because no one else has mentioned it, I like the `LocalDate` name instead of just `Date` if we're ever going to have any other  `XxxDate` or `Date<Xxx>` because it guides newcomers in the right direction. (I also think NodaTime is _generally_ a good library to imitate, type system differences excepted.)

Also, for more bikeshedding options, .NET 6 added a "naive date" type and named it `DateOnly`.

djc:
I like just `Date` much better than either `LocalDate` and `DateOnly`. I think there are probably substantial cultural differences in naming between Rust and the Java/.NET ecosystems even beyond the type system.

mqudsi:
Fair enough.

esheppa:
My thinking around the naming is as follows
* Some types are prefixed with `Naive` to both signify that they are missing information, but also gently encourage use of non-`Naive` types, however we don't want to discourage use of a date datatype where that would be most suitable from a correctness perspective.
[...]

PR #851, opened 2022-10-20
Deprecate usage of the Date<Tz> type

djc:
Recommend reviewing per commit.

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -30,6 +30,7 @@ use crate::naive::{Days, IsoWeek, NaiveDate, NaiveDateTime, NaiveTime};
 use crate::offset::Local;
 use crate::offset::{FixedOffset, Offset, TimeZone, Utc};
 use crate::oldtime::Duration as OldDuration;
+#[allow(deprecated)]
 use crate::Date;
 use crate::Months;
 use crate::{Datelike, Timelike, Weekday};
@@ -167,8 +168,8 @@ impl<Tz: TimeZone> DateTime<Tz> {
     /// ```
     /// use chrono::prelude::*;
     ///
-    /// let date: DateTime<Utc> = Utc.ymd_opt(2020, 1, 1).unwrap().and_hms_opt(0, 0, 0).unwrap();
-    /// let other: DateTime<FixedOffset> = FixedOffset::east_opt(23).unwrap().ymd_opt(2020, 1, 1).unwrap().and_hms_opt(0, 0, 0).unwrap();
+    /// let date: DateTime<Utc> = Utc.with_ymd_and_hms(2020, 1, 1, 0, 0, 0).unwrap();
+    /// let other: DateTime<FixedOffset> = FixedOffset::east_opt(23).unwrap().with_ymd_and_hms(2020, 1, 1, 0, 0, 0).unwrap();
     /// assert_eq!(date.date_naive(), other.date_naive());
     /// ```
     #[inline]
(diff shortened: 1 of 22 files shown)


src/naive/date.rs
```diff
@@ -232,6 +241,20 @@ fn test_date_bounds() {
 }
 
 impl NaiveDate {
+    /// Returns a `DateTime` which corresponds to the current date and time.
```
esheppa, line 244: should read `NaiveDate` rather than `DateTime`

src/offset/mod.rs
```diff
@@ -206,6 +206,27 @@ pub trait TimeZone: Sized + Clone {
     /// The original `TimeZone` value can be recovered via `TimeZone::from_offset`.
     type Offset: Offset;
 
+    /// Make a new `DateTime` from year, month, day, time components and current time zone.
+    ///
+    /// This assumes the proleptic Gregorian calendar, with the year 0 being 1 BCE.
+    ///
+    /// Returns `LocalResult::None` on invalid input data.
+    fn with_ymd_and_hms(
```
esheppa, line 216: This looks good, but its a reminder for me that we need to look into moving to a regular `Result<DateTime<Self>, Error` where the local result could be one variant in the error

djc: I agree, but IMO that should only happen on the main branch where we can also change all the `_opt()` methods to lose the suffix and then return a `Result`.

src/offset/mod.rs
```diff
@@ -269,7 +263,8 @@ pub trait TimeZone: Sized + Clone {
     /// but it will propagate to the `DateTime` values constructed via this date.
     ///
     /// Panics on the out-of-range date and/or invalid DOY.
-    #[deprecated(since = "0.4.23", note = "use `ymd_opt()` instead")]
+    #[deprecated(since = "0.4.23", note = "use `from_local_datetime()` with a `NaiveDateTime` instead")]
```
esheppa, line 268: we could reccomend `.and_timezone()` instead of `.from_local_datetime()` here and in similar places

esheppa: except on probably `from_local_date` as there `from_local_datetime` is the natural choice

src/offset/mod.rs
esheppa: yep it could also partially be covered by: https://github.com/chronotope/chrono/pull/830

src/offset/mod.rs
djc: Yup, good call.

src/naive/date.rs
```diff
@@ -232,6 +241,23 @@ fn test_date_bounds() {
...
+    /// Returns a `NaiveDate` which corresponds to the current date.
+    #[cfg(all(
+        feature = "clock",
+        not(all(
+            target_arch = "wasm32",
+            feature = "wasmbind",
+            not(any(target_os = "emscripten", target_os = "wasi"))
+        ))
+    ))]
+    pub fn today() -> Self {
```
esheppa, line 253: as this is new API surface, potentially it should be fallible?

esheppa: or potentially we don't need this and can just recommend users go `Utc::now().date_naive()` instead

src/datetime/mod.rs
```diff
@@ -200,13 +201,12 @@ impl<Tz: TimeZone> DateTime<Tz> {
     /// # Example
     ///
     /// ```
-    /// use chrono::Utc;
-    /// use chrono::TimeZone;
+    /// use chrono::{Utc, TimeZone, NaiveDate};
     ///
-    /// let dt = Utc.ymd_opt(1970, 1, 1).unwrap().and_hms_milli_opt(0, 0, 1, 444).unwrap();
+    /// let dt = Utc.from_local_datetime(&NaiveDate::from_ymd_opt(1970, 1, 1).unwrap().and_hms_milli_opt(0, 0, 1, 444).unwrap()).unwrap();
```
esheppa, line 206: potentially we could consider using `and_timezone(Utc)`, if we prefer to encourage the use of that API?

esheppa (approved):
Looks great - excited to get this merged!

src/naive/date.rs
djc: Good point, got rid of `NaiveDate::today()` in favor of recommending `Utc::now().date_naive()`.

src/datetime/mod.rs
djc: Yes, changed ~all the instances of this pattern.

esheppa (approved):

esheppa (approved):

Ten0:
no issue reference with rationale?

esheppa:
Hi @Ten0 - please see the discussion in https://github.com/chronotope/chrono/issues/820 - if you have some code examples of use of `Date<Tz>` that would be greatly appreciated

Ten0:
Hi! thanks a lot :) 
I'm reasonably happy with that change, I just wanted to read the thinking around it by curiosity 😊

PR #852, opened 2022-10-20
derive `Hash` for most pub types that also derive `PartialEq`

bruceg:
We needed to store `Item` alongside other data in a hash set, which requires all of the contained data be hashable. While I was adding the derive for that type, I looked for other pub types that could use it.

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -44,7 +44,7 @@ mod tests;
 /// future, so exhaustive matching in external code is not recommended.
 ///
 /// See the `TimeZone::to_rfc3339_opts` function for usage.
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
 pub enum SecondsFormat {
     /// Format whole seconds only, with no decimal point nor subseconds.
     Secs,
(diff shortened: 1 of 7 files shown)


djc (approved):

esheppa (approved):
Looks good. Nice to see the `format::Item` API being used.

@bruceg, please let us know if there are any documentation improvements needed in the `format::Item` and `format::StrftimeItems` API

djc:
Probably want to backport this to 0.4.x, @bruceg do you want to submit that?

bruceg:
A good example of how it can be used would be useful for documentation, ie feeding it in to `format_with_items` or what have you.

A bigger issue we had with this type was ownership. When parsing a string, we get items that reference the same lifetime as the string. When we later want to collect those items into a vec and store it for repeated use, we had to convert those into `Item<'static>` by boxing all the `Literal`s and `Space`s into `OwnedLiteral` and `OwnedSpace`. A better interface, that mimics other Rust types, might be to drop the owned variants from `Item`, but have a `fn to_owned(&self) -> OwnedItem` which has all boxed variants. I haven't looked at the underlying code to see if this makes sense for all your uses, though.

Funny you should mention 0.4.x. We use that for our releases, so we already did backport the change there. I will submit that after the weekend.

Issue #834, opened 2022-10-05
timezone caching prevents updating the timezone


M1cha:
In a proprietary project we're currently locked to chrono 0.4.19 because the recent addition of caching the timezone broke our code. I have to admit that the way we propagate the change to chrono is quite hacky and that we're relying on implementation details of both glibc and chrono.

So first of: yes, the TZ variable points to a symlink which systemd changes after updating the timezone. This is even the case on desktop arch linux.

So basically we're using dbus to listen for timezone changes that come from systemd-timesyncd.
Once that happened we need to make sure that `localtime_r` returns the updated value. The issue is that glibc caches the value, too. The difference is that they update it when the value of `TZ` has changed. So we can simply change the value, call tzset, change the value back, call tzset again.

That hack basically works for C and worked for chrono too because it always called `localtime_r`. Since that's not the case anymore, the change doesn't propagate to chrono anymore. So basically we need a way to propagate a timezone change to chrono.

_Originally posted by @M1cha in https://github.com/chronotope/chrono/issues/728#issuecomment-1267035270_

M1cha:
Replying to https://github.com/chronotope/chrono/pull/728#issuecomment-1268051610 :
The issue is that if the value comes from the evironment variable, it's never read again:
```
// we don't bother storing the contents of the environment variable in this case.
// changing the environment while the process is running is generally not reccomended
```

To propose a solution I'd first have to understand why we need a cache in first place. Is it really that slow to parse the timezone? If we want read and compare the TZ variable on every call, would that prevent the cache from speeding up anything?

djc:
So there's multiple things here:

* Reading the environment variable
* Reading data from disk
* Parsing a timezone definition

All of these combined I think warrant having a cache, but maybe if some of the parts aren't needed we wouldn't? Also if you want to create some benchmarks for this we might decide we don't need (as much) caching. I seem to remember maybe @esheppa did some benchmarking, but can't find it now...

esheppa:
I think there are a few competing issues here:
* It looks like @M1cha is using the `TZ` variable rather than a `/etc/localtime` symlink, so the existing caching may not apply for this
* The reason why I'd introduced the quoted comment and not allowed for `TZ` updating is that I'd thought (potentially incorrectly) that updating environment variables was problematic even when `chrono` itself wasn't doing it - however from a re-read of the issues it looks like it is more around rust code calling libc code which internally calls `setvar`. 

There are a couple of options here:
[...]

PR #853, opened 2022-10-22
Allow for changing TZ vairable and cache it for Local timezone

esheppa:
This should solve https://github.com/chronotope/chrono/issues/834.

As discussed, I've done some basic benchmarks, and it seems that this has minimal effect on the standard `/etc/localtime` case, while being around 3% more expensive in the TZ variable case. However this seems justified as this now allows for the TZ variable to change.

### empty TZ variable, timezone is a symlink at `/etc/localtime`

Before (e0b16e9c…):
```
bench_get_local_time    time:   [79.480 ns 79.692 ns 79.923 ns]
```

After (360478ce…):
```
bench_get_local_time    time:   [79.232 ns 79.383 ns 79.556 ns]
```

### with TZ variable

Before (e0b16e9c…):
```
bench_get_local_time    time:   [77.322 ns 77.664 ns 78.002 ns]
```

After (360478ce…):
```
bench_get_local_time    time:   [79.711 ns 79.921 ns 80.155 ns]
```

### DIFF
--- a/src/offset/local/tz_info/timezone.rs
+++ b/src/offset/local/tz_info/timezone.rs
@@ -26,11 +26,10 @@ impl TimeZone {
     ///
     /// This method in not supported on non-UNIX platforms, and returns the UTC time zone instead.
     ///
-    pub(crate) fn local() -> Result<Self, Error> {
-        if let Ok(tz) = std::env::var("TZ") {
-            Self::from_posix_tz(&tz)
-        } else {
-            Self::from_posix_tz("localtime")
+    pub(crate) fn local(env_tz: Option<&str>) -> Result<Self, Error> {
+        match env_tz {
+            Some(tz) => Self::from_posix_tz(tz),
+            None => Self::from_posix_tz("localtime"),
         }
     }
 
@@ -813,7 +812,7 @@ mod tests {
             // so just ensure that ::local() acts as expected
             // in this case
             if let Ok(tz) = std::env::var("TZ") {
-                let time_zone_local = TimeZone::local()?;
+                let time_zone_local = TimeZone::local(Some(tz.as_str()))?;
                 let time_zone_local_1 = TimeZone::from_posix_tz(&tz)?;
                 assert_eq!(time_zone_local, time_zone_local_1);
             }
--- a/src/offset/local/unix.rs
+++ b/src/offset/local/unix.rs
@@ -8,7 +8,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use std::{cell::RefCell, env, fs, time::SystemTime};
+use std::{cell::RefCell, collections::hash_map, env, fs, hash::Hasher, time::SystemTime};
 
 use super::tz_info::TimeZone;
 use super::{DateTime, FixedOffset, Local, NaiveDateTime};
@@ -33,20 +33,18 @@ thread_local! {
 
 enum Source {
     LocalTime { mtime: SystemTime },
-    // we don't bother storing the contents of the environment variable in this case.
-    // changing the environment while the process is running is generally not reccomended
-    Environment,
+    Environment { hash: u64 },
 }
 
 impl Source {
     fn new(env_tz: Option<&str>) -> Source {
-        // use of var_os avoids allocating, which is nice
-        // as we are only going to discard the string anyway
-        // but we must ensure the contents are valid unicode
-        // otherwise the behaivour here would be different
-        // to that in `naive_to_local`
         match env_tz {
-            Some(_) => Source::Environment,
+            Some(tz) => {
+                let mut hasher = hash_map::DefaultHasher::new();
+                hasher.write(tz.as_bytes());
+                let hash = hasher.finish();
+                Source::Environment { hash }
+            }
             None => match fs::symlink_metadata("/etc/localtime") {
                 Ok(data) => Source::LocalTime {
                     // we have to pick a sensible default when the mtime fails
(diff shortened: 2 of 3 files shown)


src/offset/local/tz_info/timezone.rs
```diff
@@ -26,9 +27,9 @@ impl TimeZone {
     ///
     /// This method in not supported on non-UNIX platforms, and returns the UTC time zone instead.
     ///
-    pub(crate) fn local() -> Result<Self, Error> {
-        if let Ok(tz) = std::env::var("TZ") {
-            Self::from_posix_tz(&tz)
+    pub(crate) fn local(env_result: &Result<String, VarError>) -> Result<Self, Error> {
```
esheppa, line 30: this may be better as `&Option<String>` or similar

src/offset/local/tz_info/timezone.rs
```diff
@@ -7,6 +7,7 @@ use std::{cmp::Ordering, fmt, str};
 
 use super::rule::{AlternateTime, TransitionRule};
 use super::{parser, Error, DAYS_PER_WEEK, SECONDS_PER_DAY};
+use std::env::VarError;
```
djc, line 10: This should go in the `std` block.

src/offset/local/tz_info/timezone.rs
djc: So this is in order to reuse the result for both `Source::new()` and `TimeZone::local()`, right? I think we should probably pass it as `Option<&str>`. Also this change might make sense to do as a separate commit?

src/offset/local/unix.rs
```diff
@@ -31,69 +31,70 @@ thread_local! {
     static TZ_INFO: RefCell<Option<Cache>> = Default::default();
 }
 
+// weird looking function signature
+// but allows sharing the results of env::var around
+fn get_tz_env_hash(env_result: &Result<String, env::VarError>) -> Option<u64> {
```
djc, line 36: Given that this is single-caller, not sure why extracting it as a separate function is useful.

src/offset/local/unix.rs
```diff
@@ -31,69 +31,70 @@ thread_local! {
...
-                Ok(_) | Err(_) => *mtime,
-            },
-            Source::Environment => return false,
-        };
-
-        match Source::default() {
-            Source::LocalTime { mtime, .. } => {
-                *self = Source::LocalTime { mtime, last_checked: now };
-                prev != mtime
+    fn out_of_date(&self, other_source: &Source) -> bool {
```
djc, line 71: Add an empty line before this? I'd also rename `other_source` as `prev` or something, since you're already using names like `old_mtime` and `old_hash` for some of the inner data it would make sense to use the more specific semantics here.

src/offset/local/unix.rs
```diff
@@ -115,17 +116,41 @@ fn fallback_timezone() -> Option<TimeZone> {
...
     }
 }
 
 impl Cache {
     fn offset(&mut self, d: NaiveDateTime, local: bool) -> LocalResult<DateTime<Local>> {
-        if self.source.out_of_date() {
-            *self = Cache::default();
+        let now = SystemTime::now();
+
+        match now.duration_since(self.last_checked) {
```
djc, line 135: So this block has been moved from `Source::out_of_date()` to here, can we do that as a separate commit?

djc (commented):
Good idea to add a benchmark! I think these are good changes, asking for a bit more structure in the commit history so it becomes easier to infer the rationale for the constituent parts.

src/offset/local/tz_info/timezone.rs
esheppa: excellent, I've gone for `Option<&str>` in [d2a32c8](https://github.com/chronotope/chrono/pull/853/commits/d2a32c8e50a44fdfd5f33bc4fcc11b12da5507eb)

src/offset/local/unix.rs
esheppa: removed

src/offset/local/unix.rs
esheppa: I've ended up in-lining this as above, which resulted in a nicer progression from the old code to the new code

src/offset/local/unix.rs
esheppa: this is now done in [e46e357](https://github.com/chronotope/chrono/pull/853/commits/e46e357ca5c623652d3b71fa7da865ac586ff184)

src/offset/local/unix.rs
```diff
@@ -32,68 +32,40 @@ thread_local! {
...
-    // we don't bother storing the contents of the environment variable in this case.
-    // changing the environment while the process is running is generally not reccomended
-    Environment,
+    LocalTime { mtime: SystemTime },
+    Environment { hash: u64 },
 }
 
-impl Default for Source {
-    fn default() -> Source {
-        // use of var_os avoids allocating, which is nice
```
esheppa, line 43: I benchmarked using `var_os` vs `var` and the difference was negligible, while the code was messier, hence the removal of this comment and use of `var_os`

src/offset/local/unix.rs
```diff
@@ -89,14 +89,20 @@ fn fallback_timezone() -> Option<TimeZone> {
...
-            zone: TimeZone::local().ok().or_else(fallback_timezone).unwrap_or_else(TimeZone::utc),
-            source: Source::default(),
             last_checked: SystemTime::now(),
+            source: Source::new(env_ref),
+            zone: new_zone(env_ref),
         }
     }
 }
 
+fn new_zone(var: Option<&str>) -> TimeZone {
```
djc, line 100: I find the `new_` prefix a little confusing here. In the end this is just the current zone, right? Maybe just call this `current()` or `zone()` or `current_zone()`?

djc (approved):

src/offset/local/unix.rs
esheppa: good call, I've renamed to `current_zone`

djc (approved):

PR #859, opened 2022-10-31
add MonthsDelta

esheppa:
This has been separated out from #824

### DIFF
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -18,7 +18,7 @@ use rkyv::{Archive, Deserialize, Serialize};
 use crate::format::DelayedFormat;
 use crate::format::{parse, write_hundreds, ParseError, ParseResult, Parsed, StrftimeItems};
 use crate::format::{Item, Numeric, Pad};
-use crate::month::Months;
+use crate::month::{Months, MonthsDelta};
 use crate::naive::{IsoWeek, NaiveDateTime, NaiveTime};
 use crate::oldtime::Duration as OldDuration;
 use crate::{Datelike, Duration, Weekday};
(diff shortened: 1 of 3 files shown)


src/naive/date.rs
```diff
@@ -1054,6 +1054,49 @@ impl NaiveDate {
...
+    /// assert_eq!(
+    ///     NaiveDate::from_ymd(2022, 4, 30).months_since(NaiveDate::from_ymd(2022, 5, 1)),
+    ///     MonthsDelta::Backwards(Months::new(1))
+    /// );
+    /// assert_eq!(
+    ///     NaiveDate::from_ymd(2022, 4, 30).months_since(NaiveDate::from_ymd(2023, 5, 1)),
+    ///     MonthsDelta::Backwards(Months::new(13))
+    /// );
+    /// ```
+    pub fn months_since(self, rhs: NaiveDate) -> MonthsDelta {
```
djc, line 1090: So what was the use case for adding this?

esheppa: This is mainly for API consistency, we have `Duration` and `TimeDelta`, `Days` and `DateDelta` and here `Months` and `MonthsDelta`. 

Calculating the calendar months between two dates is often useful in financial / accounting style calculations, but it is admittedly a narrow use case, and could be better served in an external library

djc: Unless you have a concrete use case for this in the near future, I would propose that we leave this out for now. We already have a lot to work through in this crate, so better not to add more code unless we have a somewhat strong proof that we need/want it?

esheppa: Sounds good to me

PR #860, opened 2022-11-02
Use std duration v0.4.x

esheppa:


esheppa:
still need to patch 74b087f5d7225bf4d005386151b9ce23f0437540 onto this before its ready for any review

esheppa:
Removed `TimeDelta` as per discussion in https://github.com/chronotope/chrono/pull/858#issuecomment-1307129021

djc:
Huh, not sure how I did that but it wasn't on purpose.

esheppa:
Would this be better as three or four seperate PR's? One each for Naive Date, Time, DateTime and potentially DateTime<Tz>? (I think it may be able to be separated out now that we aren't removing the old code

djc:
Probably easier to digest that way!

esheppa:
Closing in favour of https://github.com/chronotope/chrono/pull/895 (and future PRs for NaiveDate updates and (Naive)DateTime updates)

PR #864, opened 2022-11-06
Add method for computing ratio of two TimeDeltas

rwpenney:
This pull-request adds a method for dividing two TimeDeltas, returning the result as a floating point number. This can be useful for quantizing a DateTime instance relative to a given epoch, for example finding the nearest 45-second interval from a start time of 12:30. The method attempts to minimize round-off error by converting the TimeDelta into either seconds or nanoseconds using two new helper methods `as_seconds()` and `as_nanoseconds()` which also return `f64` types. Some basic unit-tests are also included.

### DIFF
--- a/src/time_delta.rs
+++ b/src/time_delta.rs
@@ -706,7 +706,18 @@ mod tests {
         assert_eq!(TimeDelta::hours(5) / TimeDelta::minutes(15), 20.0);
         assert_eq!(TimeDelta::microseconds(20) / TimeDelta::microseconds(4), 5.0);
         assert_eq!(TimeDelta::hours(2) / TimeDelta::nanoseconds(3), 2.4e12);
-        assert_eq!(TimeDelta::nanoseconds(1) / TimeDelta::nanoseconds(1 << 30), 1.0 / (1024.0 * 1024.0 * 1024.0));
+        assert_eq!(
+            TimeDelta::days(2000) / TimeDelta::nanoseconds(3),
+            (2000.0 * 28800.0 * 1e9)
+        );
+        assert_eq!(
+            TimeDelta::weeks(-4000) / TimeDelta::nanoseconds(2),
+            (-4000.0 * 7.0 * 43200.0 * 1e9)
+        );
+        assert_eq!(
+            TimeDelta::nanoseconds(1) / TimeDelta::nanoseconds(1 << 30),
+            1.0 / (1024.0 * 1024.0 * 1024.0)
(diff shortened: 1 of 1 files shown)


djc:
The `TimeDelta` type as currently defined on main will be going away, potentially in favor of an enum:

```rust
enum TimeDelta {
     Forwards(std::time::Duration),
     Backwards(std::time::Duration),
}
```

This type is intended to be output only, so I don't think we'll want to add operations on it as in this PR.

esheppa:
@rwpenney - potentially [`DurationRound`](https://docs.rs/chrono/latest/chrono/trait.DurationRound.html) or [`SubsecRound`](https://docs.rs/chrono/latest/chrono/trait.SubsecRound.html) may be suitable for your use case? If not can you give a bit more detail as there may be a clean solution in the currently released library

rwpenney:
Thanks @djc & @esheppa for your clarifications - apologies if there are many useful areas of the `chrono` library that I haven't yet explored.

The use-case I'm interested in is something like wanting to find the nearest "clock tick" nearest to a given time. So, suppose that the clock is defined by a start-time t0 (`DateTime<Utc>`) and a time-interval dt (`Duration`), I'd like to be able to take an arbitrary time t1 and compute how many steps of size dt separate it from t0. Algebraically, I'd like a syntax for computing (t1 - t0) / dt as a floating-point number. Obviously, (t1 - t0) and dt can be each be manually converted to a numerical quantity with `.num_nanoseconds()`, but makes this simple operation rather more verbose than it should be.
Any suggestions of a better approach would be very welcome.

djc:
@rwpenney this is arguably a pretty niche use case. I'm comfortable with having to go through ns for this use case.

esheppa:
@rwpenney - from my understanding, the following may be acceptable
1. use one of [`.timestamp()`](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.timestamp) or `timestamp_millis()` / `timestamp_micros()` / etc, on both datetimes
2. calculate the difference
3. divide by the integer number of micros/millis/seconds of your desired tick size

Assuming a tick size of say, 5 seconds:

```rust

let ticks_since = time_a.timestamp().sub(time_b.timestamp()).abs().checked_div(5)?;

```

rwpenney:
Thanks, both, for your helpful suggestions here.
I'd still find it handy to have a built-in method for computing the ratio of two `Duration`s, but am happy to close this P/R.

Issue #875, opened 2022-11-14
Operations with Days over non-Utc DateTime objects cause wrong time shifts


popravich:
Hi all, thanks for great crate!

However I've just encounter an issue using `Days` — when subtracting days from DateTime with non-UTC timezone result's time is decreased by timezone offset.

```rust
use chrono::{Utc, Days, FixedOffset};

let now = Utc::now();
let kyiv_time = now.with_timezone(&FixedOffset::east(7200));
println!("Utc {}; Local: {}", now, kyiv_time);
let d = Days::new(1);
println!("Sub 1: {}; {}", now - d, kyiv_time - d);
println!("Sub 2: {}; {}", (now - d) - d, (kyiv_time - d) - d);
```

esheppa:
Thanks for reporting this @popravich - I should have added tests for this in https://github.com/chronotope/chrono/pull/784/files. I'll put up a PR shortly with a fix

esheppa:
@popravich the fix for this has now been merged - let us know if this has fixed the issue?

popravich:
Thanks! Will check it in a few days.

pitdicker:
Fixed in https://github.com/chronotope/chrono/pull/878.

PR #878, opened 2022-11-15
Fix bug in Add/Sub Days, add tests with DST timezone

esheppa:
This should fix https://github.com/chronotope/chrono/issues/875

The `add/sub_months` imlpls correctly use `.naive_local()` but the `add/sub_days` impls incorrectly used the `datetime` field directly


djc (approved):
Nice, thanks for all the tests!

esheppa:
will look into using the `DstTester` elsewhere in subsequent PRs

djc:
Maybe just clean up the commit message a bit? Looks like it has stale squashing artefacts. :)

esheppa:
I've fixed the commit message. As it was a force push you may want to re-review prior to merge

djc (approved):

PR #879, opened 2022-11-16
Clarify out of range value for from_timestamp_opt

xmo-odoo:
The limits on the secs count has to be hunted down to the bottom of the readme / root document, which is a bit annoying when e.g. migrating from the deprecated `from_timestamp` and wondering why `from_timestamp_opt` is fallible.

The nsecs limits can be inferred from the rest of the text but explicit is better than implicit.


djc (approved):

djc:
Thanks!

PR #882, opened 2022-11-19
Changes to naive::internals to enable more const functions

esheppa:
Not ready for review yet. I've just pushed the code to enable some discussions on https://github.com/chronotope/chrono/issues/815


esheppa:
Using

```rust
struct DateImpl {
    year: i16,
    month: Month,
    day: NonZeroU8,
}
```

might be another option here, especially as requesting the day and month are probably more common than the ordinal.

Either option also resolves https://github.com/chronotope/chrono/pull/811 for `NaiveDate`

djc:
@esheppa you alternative `DateImpl` formulation might be attractive in some ways, just want to point out that it would substantially tighten the available range, since the year range goes from 19 bits to 16. We might decide that's a decent trade-off, just pointing it out. (And probably also means this shouldn't go in 0.4.)

esheppa:
Yes I definitely thing this should be a `0.5`+ thing, it would be very painful to implement at MSRV < 1.48. Also we wouldn't get the benefit of having less fallible APIs unless we can change the year parameters to `i16`. My main intention of this is to start a conversation around whether it is worth narrowing the date range. There are also other tradeoffs here, for example using calculations instead of lookup tables, which worsen performance, but allow validation at compile time when using const generic arguments.

Interestingly the impl with `month` and `day` fields has even worse performance, so I need to do a bit more investigation of why that is.

pitdicker:
@esheppa You did a lot of work here. In https://github.com/chronotope/chrono/pull/1043#issuecomment-1529024460 you thought that there are parts here that can be useful since the alternative approach in https://github.com/chronotope/chrono/pull/1043 was merged?

pitdicker:
Closing for now. Any follow up work is probably going to be a new PR anyway.

PR #883, opened 2022-11-19
add Day<Tz>

esheppa:


src/day.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    ///
+    pub fn succ(&self) -> Option<Day<Tz>> {
+        Some(Day { date: self.date.succ_opt()?, tz: self.tz })
+    }
+
+    ///
+    pub fn pred(&self) -> Option<Day<Tz>> {
+        Some(Day { date: self.date.pred_opt()?, tz: self.tz })
+    }
```
djc, line 106: As mentioned elsewhere, I kinda want to move away from providing these...

src/day.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    ///
+    pub fn start(&self) -> DateTime<Tz> {
```
djc, line 109: Could use some documentation of *why* we have to use a loop here. What problem is this trying to solve?

src/day.rs
```diff
@@ -0,0 +1,231 @@
...
+    Tz: TimeZone + Copy + Display,
+{
+    fn eq(&self, other: &Self) -> bool {
+        self.date.eq(&other.date)
+    }
+}
+
+impl<Tz> Eq for Day<Tz> where Tz: TimeZone + Copy + Display {}
+
+impl<Tz> Day<Tz>
```
djc, line 79: Style nit: would prefer to have the inherent impl before trait impls.

Also I would write `impl<Tz: TimeZone + Copy + Display>` instead of having the `where` clause`, since it's a bit more concise.

src/day.rs
```diff
@@ -0,0 +1,231 @@
...
+            return Some(self);
+        }
+
+        i64::try_from(days.0).ok().and_then(|d| self.diff_days(-d))
+    }
+
+    fn diff_days(self, days: i64) -> Option<Self> {
+        let date = self.date.checked_add_signed(Duration::days(days))?;
+        Some(Day { date, ..self })
+    }
```
djc, line 156: I feel like in general we should start with a minimal API here, we can grow it over time as people actually ask for it.

djc (commented):
Some early feedback.

src/day.rs
esheppa: I've removed, however these are potentially useful when dealing with non-instant style types, for example, `my_day.succ().start()` is an exclusive end time for `my_day` - but this could be implemented directly

src/day.rs
esheppa: I've added some more documentation here including around panics and incorrect results

src/day.rs
esheppa: Agreed, however I added these specifically to use here https://github.com/esheppa/db-dump/blob/7350bfabb9ff48b999ba4d6b0e827e8e2eea275c/examples/industry-coefficient.rs#L37

src/day.rs
Ekleog: One way to have this not require heuristics would be to make the loop slightly differently:
- first, try with NaiveTime::MIN as a fast path
- then, for each 15min segment,
  - if MIN + segment_end is not ok, go to the next segment
  - else, if MIN + segment_end - 1ns is ok too, do a dichotomy between MIN + segment_begin and MIN + segment_end

That said, I personally don't think that doing this would be necessary. Summer time changes across midnight are unlikely enough already, having them not be a multiple-of-15-min offset would be actively harmful to basically any software.

Another idea would be, to delegate the implementation of this to a new trait method on TimeZone, that would have a default implementation of adding NaiveTime::MIN; and on the day chrono_tz gets a summer time transition across midnight the specialization can be added there.

I'm personally partial towards the second idea, as it gets us all of the simplest implementation right now, the most efficient one, and the most future-proof one. Does that make sense?

Ekleog:
(FWIW, I just linked a fuzzer crash from https://github.com/chronotope/chrono/issues/948 ; which needs implementing basically this to fix it)

pitdicker:
As a summary:
`Time<Tz>` was removed in chrono 0.2.0, and `Date<Tz>` was deprecated in 0.4.23.
Operations on these didn't make all that much sense, except as a building block to create a `DateTime<Tz>`.

The type `Day<Tz>` in this PR is to fill a small gap `Date<Tz>` leaves in response to https://github.com/chronotope/chrono/issues/820#issuecomment-1312651118:
`Day<Tz>` can describe a day in some timezone `Tz`, with a well-defined start and end. The length of the day may be less or more than 24 hours. Essentially `Day<Tz>` is a special case of a `Period` https://github.com/chronotope/chrono/issues/380.

In my opinion this type is not generally useful enough to be added. `NaiveDate` and `TimeZone::from_local_datetime` can be used to write code with the same functionality. And a type with these semantics can easily be implemented it in user code.

Also we have not had others except for the one use case in that comment ask for this functionality.

pitdicker:
Closing this PR. If someone wants to push for this feature, please open an issue.

Issue #884, opened 2022-11-19
Incorrect conversion of `TimeZone::from_utc_datetime` 


jedel1043:
Found an interesting bug on the `TimeZone::from_utc_datetime` function. The following code:

```Rust
use chrono::{NaiveDateTime, Utc, TimeZone};

fn main() {
    let date_normal = Utc.with_ymd_and_hms(1899, 1, 1, 0, 0, 0).unwrap();
    let date_cast = Utc.from_utc_datetime(&NaiveDateTime::from_timestamp_millis(-2240524800000).unwrap());
    dbg!(date_normal.timestamp_millis(), date_cast.timestamp_millis());
    dbg!(date_cast, date_normal);
}
```

Prints:

```bash
[src/main.rs:7] date_normal.timestamp_millis() = -2240524800000
[src/main.rs:7] date_cast.timestamp_millis() = -2240524800000
[src/main.rs:8] date_cast = 1898-12-31T23:59:60Z
[src/main.rs:8] date_normal = 1899-01-01T00:00:00Z
```

Both datetimes have the same timestamp and the same timezone but they somehow have different dates.

It can also be reproduced using the `Local` type:

```Rust
use chrono::{NaiveDateTime, TimeZone, Local};

fn main() {
    let date_normal = Local.with_ymd_and_hms(1899, 1, 1, 0, 0, 0).unwrap();
    let date_cast = Local.from_utc_datetime(&NaiveDateTime::from_timestamp_millis(-2240501004000).unwrap());
    dbg!(date_normal.timestamp_millis(), date_cast.timestamp_millis());
    dbg!(date_cast, date_normal);
}
```

Prints:

```
[src/main.rs:6] date_normal.timestamp_millis() = -2240501004000
[src/main.rs:6] date_cast.timestamp_millis() = -2240501004000
[src/main.rs:7] date_cast = 1898-12-31T23:59:60-06:36:36
[src/main.rs:7] date_normal = 1899-01-01T00:00:00-06:36:36
```

Which seems to imply that it's a `TimeZone` problem.

esheppa:
Thanks @jedel1043 for the report, I've added a PR https://github.com/chronotope/chrono/pull/885 to fix this

pitdicker:
Fixed by @esheppa in https://github.com/chronotope/chrono/pull/885

PR #885, opened 2022-11-20
Fix `from_timestamp_millis` implementation and add more tests

esheppa:
Bug reported here: https://github.com/chronotope/chrono/issues/884

### DIFF
--- a/src/naive/datetime/tests.rs
+++ b/src/naive/datetime/tests.rs
@@ -5,6 +5,33 @@ use crate::{Datelike, FixedOffset, Utc};
 use std::i64;
 
 #[test]
+fn test_datetime_from_timestamp_millis() {
+    let valid_map = [
+        (1662921288000, "2022-09-11 18:34:48.000000000"),
+        (1662921288123, "2022-09-11 18:34:48.123000000"),
+        (1662921287890, "2022-09-11 18:34:47.890000000"),
+        (-2208936075000, "1900-01-01 14:38:45.000000000"),
+        (0, "1970-01-01 00:00:00.000000000"),
+        (119731017000, "1973-10-17 18:36:57.000000000"),
+        (1234567890000, "2009-02-13 23:31:30.000000000"),
+        (2034061609000, "2034-06-16 09:06:49.000000000"),
+    ];
+
+    for (timestamp_millis, formatted) in valid_map.iter().cloned() {
+        let naive_datetime = NaiveDateTime::from_timestamp_millis(timestamp_millis);
+        assert_eq!(timestamp_millis, naive_datetime.unwrap().timestamp_millis());
+        assert_eq!(naive_datetime.unwrap().format("%F %T%.9f").to_string(), formatted);
+    }
+
+    let invalid = [i64::MAX, i64::MIN];
+
+    for timestamp_millis in invalid.iter().cloned() {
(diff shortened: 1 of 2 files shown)


src/naive/datetime/tests.rs
```diff
@@ -1,9 +1,37 @@
 use super::NaiveDateTime;
 use crate::oldtime::Duration;
 use crate::NaiveDate;
-use crate::{Datelike, FixedOffset, Utc};
+use crate::{Datelike, FixedOffset, Timelike, Utc};
 use std::i64;
 
+#[test]
+fn test_datetime_from_timestamp_millis() {
+    let timestamp_millis: i64 = 1662921288000; //Sunday, September 11, 2022 6:34:48 PM
```
djc, line 9: So neither of the test cases you added have non-0 subsecond millis. Maybe add some cases for that so we have actual coverage?

src/naive/datetime/tests.rs
```diff
@@ -1,9 +1,37 @@
...
+    let naive_datetime = NaiveDateTime::from_timestamp_millis(timestamp_millis);
+    assert!(naive_datetime.is_some());
+    assert_eq!(timestamp_millis, naive_datetime.unwrap().timestamp_millis());
+    assert_eq!(naive_datetime.unwrap().year(), 2022);
+    assert_eq!(naive_datetime.unwrap().month(), 9);
+    assert_eq!(naive_datetime.unwrap().day(), 11);
+    assert_eq!(naive_datetime.unwrap().hour(), 18);
+    assert_eq!(naive_datetime.unwrap().minute(), 34);
+    assert_eq!(naive_datetime.unwrap().second(), 48);
+    assert_eq!(naive_datetime.unwrap().timestamp_subsec_nanos(), 0);
```
djc, line 19: Maybe cover most of the fields with a single `assert_eq!()` against a string-encoded timestamp? Would be a little more compact and readable.

djc (commented):
Thanks for fixing this!

src/naive/datetime/tests.rs
esheppa: good call, I've added in 0 as well as MIN, MAX and some other interesting cases (these do a good test of `from_timestamp_opt` as well

src/naive/datetime/tests.rs
esheppa: good call - I've switched to comparing against a formatted string

src/naive/datetime/tests.rs
```diff
@@ -4,6 +4,91 @@ use crate::NaiveDate;
 use crate::{Datelike, FixedOffset, Utc};
 use std::i64;
 
+#[test]
+fn test_datetime_from_timestamp_millis() {
+    let timestamp_millis: i64 = 1662921288000; //Sunday, September 11, 2022 6:34:48 PM
```
djc, line 9: Okay now there's kind of a lot of duplication, though. Can we extract the test data out into a slice and loop over that?

Also doesn't seem useful to assert `is_some()` if the next line is going to `unwrap()` anyway?

esheppa: Good point - used a map with a loop, and removed the assert some

djc (approved):

esheppa:
Just fixed the issue on MSRV, you may want to re-review

djc (approved):

PR #889, opened 2022-11-23
Delete github.sh

esheppa:
This file is no longer needed but hasn't yet been deleted


djc:
Should it be removed from 0.4.x, too?

djc (approved):

esheppa:
>  Should it be removed from 0.4.x, too? 
It is removed from there AFAIK

PR #892, opened 2022-11-23
Fix inconsistent documentation for chrono::naive::Days::new

### LINKED ISSUE #891: Confusing documentation surrounding chrono::naive::Days
The documentation for [chrono::naive::Days::new](https://github.com/chronotope/chrono/blob/2aa222112999df57d063b60510f02cc0e7ee4e7f/src/naive/date.rs#L128) says it constructs a new Days from a number of months. Not only is this confusing, it is also incorrect, as a quick test shows that using it with `checked_add_days` only adds one day.

Reference: https://github.com/chronotope/chrono/blob/2aa222112999df57d063b60510f02cc0e7ee4e7f/src/naive/date.rs#L127

Areskiko:
A very simple change to the docs of [chrono::naive::Days::new](https://github.com/chronotope/chrono/blob/2aa222112999df57d063b60510f02cc0e7ee4e7f/src/naive/date.rs#L128) to avoid confusion

While this is a semver compatible change, there seems to be a regression in the docs, as the 4.x branch correctly states that the number refers to days, therefore this targets the main branch.

Closes #891

### DIFF
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -124,7 +124,7 @@ impl NaiveWeek {
 pub struct Days(pub(crate) u64);
 
 impl Days {
-    /// Construct a new `Days` from a number of months
+    /// Construct a new `Days` from a number of days
     pub fn new(num: u64) -> Self {
         Self(num)
     }


djc:
(And merged to main to fix it there.)

PR #893, opened 2022-11-23
convenience functions to create NaiveDates

esheppa:
These functions are to reduce the number of unwraps needed when creating NaiveDates

### DIFF
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -25,7 +25,7 @@ use crate::format::{Item, Numeric, Pad};
 use crate::month::Months;
 use crate::naive::{IsoWeek, NaiveDateTime, NaiveTime};
 use crate::oldtime::Duration as OldDuration;
-use crate::{Datelike, Duration, Weekday};
+use crate::{Datelike, Duration, Month, Weekday};
 
 use super::internals::{self, DateImpl, Mdf, Of, YearFlags};
 use super::isoweek;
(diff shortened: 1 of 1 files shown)


jtmoon79 (changes requested):
These look convenient! 

Please add `#[must_use]` to these functions.

pitdicker:
I am not super convinced of the usefulness of all these methods. My reason is that there are all kinds of operations someone may want to do on dates. Where do you draw the line?

`end_month` and `end_year` seem useful because it they are non-trivial. You should get the current year/month, add 1 (this can cause a panic on overflow), and get the preceding day.

But then you would also want `start_month` and `start_year` for symmetry, which don't feel valuable enough for me.

Would similar methods on `NaiveDateTime` or `DateTime` make sense? Would the end of a month be at the last second? Or the last nanoseond? Or maybe even leap-second? Strictly speaking, the end is the same as the start of the next day, at 0:00:00. But than the date would be different from the one returned by `NaiveDate::end_month`.

src/naive/date.rs
```diff
@@ -236,6 +236,49 @@ fn test_date_bounds() {
...
+    /// Create a `NaiveDate` of the first day of the given year and month
+    ///
+    /// This takes an `i16` rather than `i32` to ensure it is not
+    /// out of the allowed range
+    pub fn start_month(year: i16, month: Month) -> NaiveDate {
+        // ideally this unwrap can be later removed
+        Self::from_ymd_opt(year.into(), month.number_from_month(), 1).unwrap()
+    }
+
+    /// Create a `NaiveDate` of the ;ast day of the given year and month
```
pitdicker, line 266:
```suggestion
    /// Create a `NaiveDate` of the last day of the given year and month
```

src/naive/date.rs
```diff
@@ -236,6 +236,49 @@ fn test_date_bounds() {
 }
 
 impl NaiveDate {
+    /// Create a `NaiveDate` of the first day of the given year
+    ///
+    /// This takes an `i16` rather than `i32` to ensure it is not
+    /// out of the allowed range
+    pub fn start_year(year: i16) -> NaiveDate {
```
pitdicker, line 243: Using an `i16` ties into your work in https://github.com/chronotope/chrono/pull/882. But maybe it is best to accept an `i32` until the decision is made to restrict the range of allowed years?

src/naive/date.rs
```diff
@@ -236,6 +236,49 @@ fn test_date_bounds() {
...
+        Self::from_ymd_opt(year.into(), 1, 1).unwrap()
+    }
+
+    /// Create a `NaiveDate` of the last day of the given year
+    ///
+    /// This takes an `i16` rather than `i32` to ensure it is not
+    /// out of the allowed range
+    pub fn end_year(year: i16) -> NaiveDate {
+        // ideally this unwrap can be later removed
+        Self::from_ymd_opt(i32::from(year) + 1, 1, 1).unwrap().pred_opt().unwrap()
```
pitdicker, line 254: This might fail if `year == MAX_YEAR`.

src/naive/date.rs
```diff
@@ -236,6 +236,49 @@ fn test_date_bounds() {
...
+
+    /// Create a `NaiveDate` of the ;ast day of the given year and month
+    ///
+    /// This takes an `i16` rather than `i32` to ensure it is not
+    /// out of the allowed range
+    pub fn end_month(year: i16, month: Month) -> NaiveDate {
+        // ideally this unwrap can be later removed
+        let start = Self::start_month(year, month);
+
+        (start + Months::new(1)).pred_opt().unwrap()
```
pitdicker, line 274: This might fail if `year == MAX_YEAR && month == 12`.

pitdicker:
Closing. I would be interested to have a method to easily create a `NaiveDate` at the end of a month.

PR #897, opened 2022-11-25
Fix: Typ in docstrings

dandxy89:


djc (approved):
Thanks!

PR #900, opened 2022-11-26
Years type implementation

Rafael-Conde:
I started a simple implementation that I intend to add proper add checks once the problem is actually solved.

I implemented simples tests too, but there is still a lot to do. I'm currently opening the PR so discussions about the implementation can occur here and you guys can follow the progress. I believe this is semver compatible since I'm only adding a new type so I targeted the 0.4.x branch, hopefully, this type will make it into that version. 

I already applied the recommendations @esheppa made in [this comment](https://github.com/chronotope/chrono/issues/416#issuecomment-1328025896) on issue #416

### DIFF
--- a/src/date.rs
+++ b/src/date.rs
@@ -23,6 +23,8 @@ use crate::oldtime::Duration as OldDuration;
 use crate::DateTime;
 use crate::{Datelike, Weekday};
 
+use crate::Years;
+
 /// ISO 8601 calendar date with time zone.
 ///
 /// You almost certainly want to be using a [`NaiveDate`] instead of this type.
@@ -367,6 +369,16 @@ where
         locale: Locale,
     ) -> DelayedFormat<StrftimeItems<'a>> {
         self.format_localized_with_items(StrftimeItems::new_with_locale(fmt, locale), locale)
+    }
+
+    pub fn checked_add_years(mut self, num_years: Years) -> Option<Date<Tz>> {
+        self.date = self.date.checked_add_years(num_years)?;
+        Some(self)
+    }
+
+    pub fn checked_sub_years(mut self, num_years: Years) -> Option<Date<Tz>> {
+        self.date = self.date.checked_sub_years(num_years)?;
+        Some(self)
     }
 }
 
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -27,6 +27,8 @@ use crate::naive::{IsoWeek, NaiveDateTime, NaiveTime};
 use crate::oldtime::Duration as OldDuration;
 use crate::{Datelike, Duration, Weekday};
 
+use crate::Years;
+
 use super::internals::{self, DateImpl, Mdf, Of, YearFlags};
 use super::isoweek;
 
@@ -1248,6 +1250,23 @@ impl NaiveDate {
         NaiveWeek { date: *self, start }
     }
 
+    // I think self here is a copy, but I still have to make sure it's
+    pub fn checked_add_years(self, years: Years) -> Option<NaiveDate> {
+        if years.0 == 0 {
+            return Some(self);
+        }
+        let num_years = i32::try_from(years.0).ok()?;
+        self.with_year(self.year() + num_years)
+    }
+
+    pub fn checked_sub_years(self, years: Years) -> Option<NaiveDate> {
+        if years.0 == 0 {
+            return Some(self);
+        }
+        let num_years = i32::try_from(years.0).ok()?;
+        self.with_year(self.year() - num_years)
+    }
+
     /// The minimum possible `NaiveDate` (January 1, 262145 BCE).
     pub const MIN: NaiveDate = NaiveDate { ymdf: (MIN_YEAR << 13) | (1 << 4) | 0o07 /*FE*/ };
     /// The maximum possible `NaiveDate` (December 31, 262143 CE).
@@ -1684,6 +1703,22 @@ impl Sub<Months> for NaiveDate {
     /// ```
     fn sub(self, months: Months) -> Self::Output {
         self.checked_sub_months(months).unwrap()
+    }
+}
+
+impl Add<Years> for NaiveDate {
+    type Output = NaiveDate;
+
+    fn add(self, years: Years) -> Self::Output {
+        self.checked_add_years(years).unwrap()
+    }
+}
+
+impl Sub<Years> for NaiveDate {
+    type Output = NaiveDate;
+
+    fn sub(self, years: Years) -> Self::Output {
+        self.checked_sub_years(years).unwrap()
     }
 }
 
--- a/src/naive/datetime/mod.rs
+++ b/src/naive/datetime/mod.rs
@@ -24,6 +24,8 @@ use crate::oldtime::Duration as OldDuration;
 use crate::{DateTime, Datelike, LocalResult, Months, TimeZone, Timelike, Weekday};
 use core::cmp::Ordering;
 
+use crate::Years;
+
 #[cfg(feature = "rustc-serialize")]
 pub(super) mod rustc_serialize;
 
@@ -926,6 +928,16 @@ impl NaiveDateTime {
                 NaiveDateTime::from_timestamp_opt(secs, nsecs)
             }
         }
+    }
+
+    pub fn checked_add_years(mut self, num_years: Years) -> Option<NaiveDateTime> {
+        self.date = self.date.checked_add_years(num_years)?;
+        Some(self)
+    }
+
+    pub fn checked_sub_years(mut self, num_years: Years) -> Option<NaiveDateTime> {
+        self.date = self.date.checked_sub_years(num_years)?;
+        Some(self)
     }
 }
 
--- a/src/years.rs
+++ b/src/years.rs
@@ -8,108 +8,51 @@
         clippy::unwrap_used,
         clippy::expect_used)]
 
-use num_integer::Integer;
-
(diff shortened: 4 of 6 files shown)


Rafael-Conde:
I believe I already made the mistake to "commit to main". I branched from main and that's the problem. I'll change that later because I don't have more time now. I decided to open the PR since I said I would. I'll fix that later, wether opening a new PR with the proper branch or another fix for this one.

One thing too, is there a deadline for the 0.4.x version to be out?

djc:
There's no deadline! Thanks for working on this, I'll take a more detailed look after you rebase on 0.4.

Rafael-Conde:
> There's no deadline!

Nice! I'm likely to work on it on the weekends, 

> Thanks for working on this

Not a problem! I appreciate the opportunity to contribute!

> I'll take a more detailed look after you rebase on 0.4.

Ok! I'm still in the middle of the implementation(I couldn't finish it on my first session), I wrote a few lines of code but basically it does nothing so far. But help is always welcome! you guys are very polite when helping out! Thank you for that!

Also I got a bit messed up while rebasing but I think everything should be in order right now. 

If you guys want, I can leave a comment here once I got any of the implementations actually working. To review and improve the code.

Also, during development I plan to leave comments on the code I'm working on to try and complement the code with some thoughts. Other than that I could mark the lines of code here and leave the comment here, where I would otherwise leave in the code.

djc:
I've converted this to a draft PR, please push the "Ready for review" button once you think it's ready!

Rafael-Conde:
Ok! Thank you for that! I believe I should've done it from the beginning, I'm still learning haha

djc:
Yes, CI should pass. It's possible that some failures are unrelated to your changes, but should be unlikely.

src/years.rs
```diff
@@ -0,0 +1,115 @@
...
+            // if divisible by 100 but also divisible by 400 than is leap year
+            (year % 400) == 0
+        }
+    } else {
+        false
+    }
+}
+
+impl NaiveDate {
+    pub fn checked_add_years(self, years: Years) -> Option<NaiveDate> {
```
esheppa, line 66: this impl should be in naive/date.rs - you can import Years over there

src/years.rs
```diff
@@ -0,0 +1,115 @@
...
+            //use this to transform a years addition into the proper addition of days
+            // not sure this is the best implementation
+            // but seems robust since it would rely on the top level implementation
+        }
+    }
+
+    pub fn checked_sub_years(self, num_years: Years) -> Option<NaiveDate> {}
+}
+
+// impl Add<Years> for NaiveDate {}
```
esheppa, line 85: these impls should be in the module where the target type is defined

src/years.rs
```diff
@@ -0,0 +1,115 @@
...
+        year += 1;
+    }
+    while !is_leap_year(year) {
+        year += 1;
+    }
+    year
+}
+
+// almost surely there is a more precise way to implement this
+fn is_leap_year(year: i32) -> bool {
```
esheppa, line 52: you can use ordinal flags and then use the ndays method to figure out whether the target year is a leap year or not. You can construct the ordinal flags directly from the target year itself

esheppa (changes requested):
just leaving some initial comments, mostly on code location

src/years.rs
Rafael-Conde: Thank you for the pointers! I'll surely make the requested changes! This one in special will be of great help! really appreciated!

I'll move the other impl blocks to the proper locations too!

Rafael-Conde:
I finished the implementation, it ended up a lot simpler than I first imagined but I currently don't see why not do that haha

I still need to do the documentation, but the code is compiling and the simple tests I wrote are working. (I made a simple TODO docs just to run but removed from the commit so I don't forget to do it once I'm back at it).

> you can use ordinal flags and then use the ndays method to figure out whether the target year is a leap year or not. You can construct the ordinal flags directly from the target year itself

this helped a lot and ended up pointing me to the solution haha Thank you!!

I'm planning on implementing the addition of Years and I'm still thinking about how to get the difference, in complete years, between 2 dates. 

That specific difference between dates helps me a lot.

src/date.rs
```diff
@@ -23,6 +23,8 @@ use crate::oldtime::Duration as OldDuration;
 use crate::DateTime;
 use crate::{Datelike, Weekday};
 
+use crate::Years;
+
```
pitdicker, line 27: The `Date` type is deprecated, better to not add new methods to it.

src/naive/date.rs
```diff
@@ -1248,6 +1250,23 @@ impl NaiveDate {
         NaiveWeek { date: *self, start }
     }
 
+    // I think self here is a copy, but I still have to make sure it's
```
pitdicker, line 1253: It is a copy, you only want to take `self` by reference for `DateTime<Tz>`.

src/naive/date.rs
```diff
@@ -1248,6 +1250,23 @@ impl NaiveDate {
...
+    pub fn checked_add_years(self, years: Years) -> Option<NaiveDate> {
+        if years.0 == 0 {
+            return Some(self);
+        }
+        let num_years = i32::try_from(years.0).ok()?;
+        self.with_year(self.year() + num_years)
```
pitdicker, line 1259: This will fail if you add 1 year to February 29th. This is different from `checked_add_months`, which will fall back to the last day of the month.

src/naive/datetime/mod.rs
```diff
@@ -927,6 +929,16 @@ impl NaiveDateTime {
...
+    pub fn checked_add_years(mut self, num_years: Years) -> Option<NaiveDateTime> {
+        self.date = self.date.checked_add_years(num_years)?;
+        Some(self)
+    }
+
+    pub fn checked_sub_years(mut self, num_years: Years) -> Option<NaiveDateTime> {
+        self.date = self.date.checked_sub_years(num_years)?;
+        Some(self)
+    }
 }
```
pitdicker, line 942: `NaiveDateTime` could also use `Add` and `Sub` implementations.

src/years.rs
```diff
@@ -0,0 +1,58 @@
+// I usually develop with this clippy flags
+// but I can remove once the this module is done
```
pitdicker, line 2: 👍 when done.

src/years.rs
```diff
@@ -0,0 +1,58 @@
...
+#![warn(clippy::all,
+/*#![warn(*/clippy::pedantic,
+		clippy::perf,
+        clippy::nursery,
+        // clippy::cargo,
+        clippy::unwrap_used,
+        clippy::expect_used)]
+
+/// TODO
```
pitdicker, line 11: Needs documentation.

src/years.rs
```diff
@@ -0,0 +1,58 @@
...
+#[derive(Debug)]
```
pitdicker, line 12: Can you make this match the derives and attributes of `Month`?

src/naive/date.rs
pitdicker: How do you handle integer overflow in for example `self.year() + `i32::MAX`?

src/naive/date.rs
```diff
@@ -1248,6 +1250,23 @@ impl NaiveDate {
...
+    }
+
+    pub fn checked_sub_years(self, years: Years) -> Option<NaiveDate> {
+        if years.0 == 0 {
+            return Some(self);
+        }
+        let num_years = i32::try_from(years.0).ok()?;
+        self.with_year(self.year() - num_years)
```
pitdicker, line 1267: As above.

pitdicker (changes requested):
My personal opinion is that I don't care much for the `Day` and `Month` types. I prefer writing `date.checked_add_years(5)` over `date.checked_add_years(Years::new(5))` or `date + Years::new(5)`. It seems like over-using the type system to me. But others may like it 🤷.

The `DateTime<Tz>` does not have the extra methods yet.

djc:
Yeah, I've been on the fence about the proliferation of types here -- not sure it's the best idea, even though I may have started it. (I think the original idea was that it allowed for using the operators, which does seem like nice API.)

pitdicker:
I was thinking we could add `impl From<u32> for Month` and make `checked_add_month` take `Into<Month>` to make the API a little nicer to use. And the same for `Day`.

We can stuff the types inside a module mostly out of sight :smile:.

pitdicker:
Closing due to inactivity, and because we may want to step away from `Day` and `Month` types, and `Year` as in this PR, in favour of a more feature-complete `CalendarDuration` type.

@Rafael-Conde Thank you for working on this!

Rafael-Conde:
Thank you for notifying my @pitdicker!! 

I recently came back to this PR and saw how much work you did! 

I was really impressed, a lot of time and thought!!

One last recommendation as I leave this, I saw a(apparently old) discussion about what to do with invalid dates(e.g. adding a month to January 31st, should it be February 28 or March 02?)

For this matter I would recommend a cpp conference video of the creator of Chrono describing how he handled that providing invalid dates and methods for resolving it to the preferred way(it was possible to yield both results from the invalid date I mentioned in the example). I'll search for the link and I'll post here or maybe in the new PR, if you folks find it useful. Keep up the good work!!!

pitdicker:
Thank you, and that would be interesting!

I suppose we have the capability to represent an invalid date with the internal `Mdf` type (and wrap into something nicer and public). But then we also have two other places where the result could become invalid: DST transitions and leap seconds. Something to think about :smile:.

PR #902, opened 2022-12-09
Make StrftimeItems::new return Result

### LINKED ISSUE #623: [Bug] Parsing an string with %Z causes the library to panic
I have the following Date string: 
```
"2020-04-12 22:10 ACST"
```

And I do the following: 
```rust
let datetime = chrono::DateTime::parse_from_str(date_str, "%Y-%m-%d %H:%M %Z").unwrap();
```

And I get a panic: 

```
   Compiling playground v0.0.1 (/playground)
    Finished dev [unoptimized + debuginfo] target(s) in 1.67s
     Running `target/debug/playground`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ParseError(NotEnough)', src/main.rs:5:80
stack backtrace:
   0: rust_begin_unwind
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/panicking.rs:517:5
   1: core::panicking::panic_fmt
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/panicking.rs:101:14
   2: core::result::unwrap_failed
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/result.rs:1617:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/result.rs:1299:23
   4: playground::main
             at ./src/main.rs:5:16
   5: core::ops::function::FnOnce::call_once
             at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace
```

This doesn't happen in C with `strftime()`

Playground link: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=bc851bfde4ae1f6e76de8d52e3934bdc

jaggededgedjustice:
This makes the `date_time.format()` function return a `Result` instead of panicking when passed an invalid format string. If any part of the format string is invalid an error is returned.
This is a breaking change for the api.

Fixes #623 
closes #614

### DIFF
--- a/src/format/strftime.rs
+++ b/src/format/strftime.rs
@@ -180,6 +180,8 @@ Notes:
    China Daylight Time.
 */
 
+use crate::format::{ParseError, BAD_FORMAT};
+
 #[cfg(feature = "unstable-locales")]
 extern crate alloc;
 
@@ -233,21 +235,34 @@ pub struct StrftimeItems<'a> {
 
 impl<'a> StrftimeItems<'a> {
     /// Creates a new parsing iterator from the `strftime`-like format string.
-    #[must_use]
-    pub fn new(s: &'a str) -> StrftimeItems<'a> {
-        Self::with_remainer(s)
+    pub fn new(s: &'a str) -> Result<StrftimeItems<'a>, ParseError> {
+        let strftime = Self::with_remainer(s);
+        if strftime.clone().any(|i| i == Item::Error) {
+            return Err(BAD_FORMAT);
+        }
+        Ok(strftime)
     }
 
     /// Creates a new parsing iterator from the `strftime`-like format string.
     #[cfg(feature = "unstable-locales")]
     #[cfg_attr(docsrs, doc(cfg(feature = "unstable-locales")))]
     #[must_use]
-    pub fn new_with_locale(s: &'a str, locale: Locale) -> StrftimeItems<'a> {
-        let d_fmt = StrftimeItems::new(locales::d_fmt(locale)).collect();
-        let d_t_fmt = StrftimeItems::new(locales::d_t_fmt(locale)).collect();
-        let t_fmt = StrftimeItems::new(locales::t_fmt(locale)).collect();
+    pub fn new_with_locale(s: &'a str, locale: Locale) -> Result<StrftimeItems<'a>, ParseError> {
+        let d_fmt = StrftimeItems::new(locales::d_fmt(locale))
+            .expect("ERROR: No date format defined for given format")
+            .collect();
+        let d_t_fmt = StrftimeItems::new(locales::d_t_fmt(locale))
+            .expect("ERROR: No datetime format for the given locale")
+            .collect();
+        let t_fmt = StrftimeItems::new(locales::t_fmt(locale))
+            .expect("ERROR: No time format gor the given locale")
+            .collect();
 
-        StrftimeItems { remainder: s, recons: Vec::new(), d_fmt, d_t_fmt, t_fmt }
+        let strftime = StrftimeItems { remainder: s, recons: Vec::new(), d_fmt, d_t_fmt, t_fmt };
+        if strftime.clone().any(|i| i == Item::Error) {
+            return Err(BAD_FORMAT);
+        }
+        return Ok(strftime);
     }
 
     #[cfg(not(feature = "unstable-locales"))]
@@ -514,9 +529,10 @@ fn test_strftime_items() {
     fn parse_and_collect(s: &str) -> Vec<Item<'_>> {
         // map any error into `[Item::Error]`. useful for easy testing.
         eprintln!("test_strftime_items: parse_and_collect({:?})", s);
-        let items = StrftimeItems::new(s);
-        let items = items.map(|spec| if spec == Item::Error { None } else { Some(spec) });
-        items.collect::<Option<Vec<_>>>().unwrap_or_else(|| vec![Item::Error])
+        match StrftimeItems::new(s) {
+            Err(_) => vec![Item::Error],
+            Ok(items) => items.into_iter().collect(),
+        }
     }
 
     assert_eq!(parse_and_collect(""), []);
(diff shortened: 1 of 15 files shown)


src/format/strftime.rs
```diff
@@ -209,7 +209,7 @@ static D_T_FMT: &[Item<'static>] = &[
 static T_FMT: &[Item<'static>] = &[num0!(Hour), lit!(":"), num0!(Minute), lit!(":"), num0!(Second)];
 
 /// Parsing iterator for `strftime`-like format strings.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Copy)]
```
djc, line 212: Hmm, why did you decide to make this `Copy`? I guess it's technically qualifies, but it seems to be on the larger side to make copying implicit.

djc (approved):
This is great, thanks for fixing this up!

src/format/strftime.rs
jaggededgedjustice: This was done because the check I add in `src/datetime/mod.rs` will take ownership of the `StrftimeItems` preventing it being passed into `format_with_items` and this was the simplest way to solve the problem.
If you think this isn't appropriate I can take a different approach.

djc: Hm, yeah. I think the right way to get here is to remove the `Item::Error` variant and then fix it up so that everything that creates an `Item` becomes fallible instead.

djc: Wait, this code is weirder than I remembered. So it feels to me like `StrftimeItems` should probably contain a single `&[Item<'_>]`, with `new()` eagerly parsing it's `&str` argument. But I don't know how big of a change that is...

jaggededgedjustice:
Ok, I have things mostly working except for some of the configurations. I'm not sure how best to handle the failures though.

src/format/strftime.rs
```diff
@@ -184,11 +186,6 @@ use alloc::vec::Vec;
 use super::{locales, Locale};
 use super::{Fixed, InternalFixed, InternalInternal, Item, Numeric, Pad};
 
-#[cfg(feature = "unstable-locales")]
```
djc, line 187: This is the cause of the test failures. Note that, without `unstable-locales`, this is a slice, not a `Vec`. chrono without default features (and in WASM scenarios) has included some formatting functionality in that case, but `Vec` is not available there. So we have to find a way to use a `&[Item<'a>]` in `StrftimeItems`.

djc (commented):
Thanks, this is looking great!

src/format/strftime.rs
```diff
@@ -211,306 +208,263 @@ static T_FMT: &[Item<'static>] = &[num0!(Hour), lit!(":"), num0!(Minute), lit!("
...
-    }
-
-    /// Creates a new parsing iterator from the `strftime`-like format string.
-    #[cfg(feature = "unstable-locales")]
-    #[cfg_attr(docsrs, doc(cfg(feature = "unstable-locales")))]
-    pub fn new_with_locale(s: &'a str, locale: Locale) -> StrftimeItems<'a> {
-        let d_fmt = StrftimeItems::new(locales::d_fmt(locale)).collect();
-        let d_t_fmt = StrftimeItems::new(locales::d_t_fmt(locale)).collect();
-        let t_fmt = StrftimeItems::new(locales::t_fmt(locale)).collect();
+    pub fn new(s: &'a str) -> Result<StrftimeItems<'a>, ParseError> {
```
esheppa, line 217: Perhaps we should have a new function `parse` or similar for this, and keep the existing function as a lazy iterator? My thinking for this is that if we just want to validate strftime strings, the iterator is more efficient as we can bail out earlier if we hit an error. Alternatively, we may want to collect all the errors rather than the first one

djc: What's the use case for just validating a strftime string? I also think it's pretty unusual for things to collect all errors instead of bailing on the first one (for example, the `FromIterator` impl for `Result`).

src/format/strftime.rs
jaggededgedjustice: This could be a problem. I'm assuming that other collection types like `LinkedList` are also unavailable, and I don't see a good way to do this without a resizeable collection.

We could go back to the system where the iterator generates elements as we go and just run through the string once at creation time to check it's all valid. That feels wasteful though.

jaggededgedjustice: Having spent some time thinking about it, I think we do want to have `StrftimeItems::new` return a `Result` but I can't see any way to use an array as we can't give a size at compile time.
So my current plan is to go back to the system of storing the format string as a `str` and processing it as the iterator is called but iterate through the format string once at creation time to check for problems. Do you have any objections or better ideas?

djc: I'll try to look at it again tomorrow.

djc: No better ideas, so let's do it that way for now.

jaggededgedjustice:
can someone explain the failures? The error message seems to be pointing to a line that doesn't exist in the source.

SomewhereOutInSpace:
If anybody needs a workaround until this is merged, here's what I cooked up:
```rs
// Workaround to avoid panicking when the user-provided format string is invalid.
// Will be obsolete once https://github.com/chronotope/chrono/pull/902 is merged.
std::panic::set_hook(Box::new(|_| ()));
let formatted = std::panic::catch_unwind(|| {
    datetime.format(&description).to_string()
})?;
let _ = std::panic::take_hook();
```
A dummy panic hook is necessary to suppress the panic message. I don't think this actually incurs an allocation (`|_: &PanicInfo| ()` is a ZST) so it should be fine even in a hot loop.

esheppa:
@SomewhereOutInSpace - you can also use a workaround as mentioned here https://github.com/chronotope/chrono/pull/614#issuecomment-1173070176. The main points are using [`StrftimeItems::new()`](https://docs.rs/chrono/latest/chrono/format/strftime/struct.StrftimeItems.html#method.new) and the [`write!`](https://doc.rust-lang.org/std/fmt/fn.write.html) macro depending on the task, both cases exposing the error as a regular `Result` rather than a panic.

jaggededgedjustice:
@esheppa the bit I'm struggling with is finding the line that's throwing the error. Consider, https://github.com/chronotope/chrono/actions/runs/3980518900/jobs/6823655339#logs which says
```
error[E0308]: mismatched types
    --> src/naive/date.rs:1177:42
     |
1177 |         self.format_localized_with_items(StrftimeItems::new_with_locale(fmt, locale), locale)
     |              --------------------------- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `StrftimeItems`, found enum `Result`
     |              |
     |              arguments to this function are incorrect
     |
     = note: expected struct `StrftimeItems<'_>`
                  found enum `Result<StrftimeItems<'_>, ParseError>`
```
The error makes sense for the line shown, but I can't find where that line is. The error message points to `src/naive/date.rs` line `1177` except that line is a comment: https://github.com/jaggededgedjustice/chrono/blob/fac117e8bc4ff092300abd6dc5336156eeb3f137/src/naive/date.rs#L1177

and a search shows no call to `new_with_locale` that doesn't handle the new Return value:
```
grep -r new_with_locale
chrono-handle-invalid-format-strings/src/format/strftime.rs:    pub fn new_with_locale(s: &'a str, locale: Locale) -> Result<StrftimeItems<'a>, ParseError> {
chrono-handle-invalid-format-strings/src/format/mod.rs:    pub fn new_with_locale(
chrono-handle-invalid-format-strings/src/datetime/mod.rs:        Ok(self.format_localized_with_items(StrftimeItems::new_with_locale(fmt, locale)?, locale))
chrono-handle-invalid-format-strings/src/date.rs:        Ok(self.format_localized_with_items(StrftimeItems::new_with_locale(fmt, locale)?, locale))
```

So I have no idea what's going on, unless maybe the worker that ran the last build had an old copy of some code.

jtmoon79 (changes requested):
This needs additional test cases for the path where a `ParseError` is returned.

jaggededgedjustice:
> This needs additional test cases for the path where a `ParseError` is returned.

Do you want a test for every function that can now return a `ParseError`? There is `test_strftime_items()` in `src/format/strftime.rs` which checks that the constructor for `StrftimeItems` will return an error type when given invalid format strings.

jtmoon79:
> > This needs additional test cases for the path where a `ParseError` is returned.
> 
> Do you want a test for every function that can now return a `ParseError`? There is `test_strftime_items()` in `src/format/strftime.rs` which checks that the constructor for `StrftimeItems` will return an error type when given invalid format strings.

If you would please, I'd really appreciate it.  While `chrono` does have decent tests for the public functions, it has few to no unittests on the underlying private component functions. It would be a good project "quality confidence" improvement (Test Engineer lingo) to require tests for private functions.

Hopefully @djc and @esheppa agree. Final decisions should defer to them.

esheppa:
@jtmoon79 that does seem reasonable overall, but as far as I can see there aren't any new (non testing) private functions in this PR? However I think it would be nice to have some basic tests on the `StrftimeItems::new*` functions now that they are fallible

jtmoon79:
> @jtmoon79 that does seem reasonable overall, but as far as I can see there aren't any new (non testing) private functions in this PR? However I think it would be nice to have some basic tests on the `StrftimeItems::new*` functions now that they are fallible

Good point. I had too many PR reviews in my head when I wrote that 🥴

jaggededgedjustice:
Are you happy with the current tests that rely on `is_err()`, or should they check the specific error that gets returned?

src/naive/date.rs
```diff
@@ -2984,4 +3019,12 @@ mod tests {
             }
         }
     }
+
+    #[test]
+    fn test_format() {
+        assert!(
```
jtmoon79, line 3025: Please rename to `fn test_date_format_err()` and add a few more tests within it.

src/naive/datetime/mod.rs
```diff
@@ -1921,3 +1921,16 @@ where
     assert!(from_str(r#"{"date":{"ymdf":20},"time":{"secs":0,"frac":0}}"#).is_err());
     assert!(from_str(r#"null"#).is_err());
 }
+
+#[test]
+fn test_format() {
```
jtmoon79, line 1926: Please rename to `fn test_datetime_format_err()` and add a few more tests within it.

Suggestions:

- `""` (not sure if this should error, you'll have to try it)
- `"%%%"`
- `"%Y%"`
- `"  %Y%   "`
- `"%ぁ"`

jtmoon79 (changes requested):

src/naive/date.rs
jtmoon79: Suggestions:

- `""` (not sure if this should error, you'll have to try it)
- `"%X"`
- `"%%%"`
- `"%Y%"`
- `"  %Y%   "`
- `"%ぁ"`

jaggededgedjustice: How many test cases do think would be required? Personally I'd leave it with just the 1 case as the error is just being passed through from `StrftimeItems::new()` which has `test_strftime_items` to check a wide range of inputs so I don't see the benefit of duplicating the check here.

jtmoon79: Yeah, that's reasonable.

src/naive/time/mod.rs
```diff
@@ -1381,3 +1381,11 @@ where
     assert!(from_str(r#"{"secs":0,"frac":0}"#).is_err());
     assert!(from_str(r#"null"#).is_err());
 }
+
+#[test]
+fn test_format() {
```
jtmoon79, line 1386: Please rename this to `test_time_format_err`.

jtmoon79 (changes requested):

jtmoon79 (approved):

pitdicker:
I am fully behind the goal of this PR.

May I suggest an alternative that works on 0.4.x, and doesn't regress the ergonomics by requiring extra error handling for library users?

- Add new, internal methods to `DelayedFormat` that take `Result<I, ParseError>` instead of just `I`.
- Change the `items` field of `Delayedformat` to store `Result<I, ParseError>`.
- Return the error as part of `fmt::Display`.

Not regressing egnomics when we don't absolutely have to is my primary concern, because formatting is a common operation.

pitdicker:
While I agree this issue is worth fixing, I am not sure the solution here is the best.

My concern are the ergonomics for library users. Also this solution requires parsing the format string twice, which is unfortunate.

---

If I understand it right, the only combination a panic is designed to happen with the current formatting and parsing code is in this snippet:
```rust
let date = NaiveDate::from_ymd_opt(2023, 5, 31).unwrap();
let date_str = date.format("%Y-%M-%D %").to_string();
```

The actual panic doesn't happen in any code that is part of chrono. It happens in `std::string::ToString::to_string`.
When `std::fmt::Display::fmt` returns an error, `to_string` can't pass on the error but panics.

If you use `Display::fmt` with a method that can return an error, it all works as designed:
```rust
let date = NaiveDate::from_ymd_opt(2023, 5, 31).unwrap();
let date_str = format!("{}", date.format("%Y-%M-%D %"))?;
/// or:
println!("{}", date.format("%Y-%M-%D %"))?;
```

This PR requires extra error handling for common cases such as `println` and parsing that were fine without it.

As an alternative (that also works on 0.4.x): what if we had convenience methods `format_to_string` that directly return `fmt::Result<String>`? It would give users directly what they want, a string, and skips the delayed formatting stuff which is useful in this case. And maybe we can add `try_to_string` to `DelayedFormat`?

---

Also there are two cases that are currently not covered by the solution in this PR:

- At the time a format string is created, by `StrftimeItems::new`, it doesn't yet know with which type it will be used for formatting. It may reasonably be used with multiple types. Formatting will return an error when the type does not contain the fields needed for the format string, such as time fields on `NaiveDate`. This would still panic in `to_string`. But that is fixable in this PR by adding more validation in the `format` methods.

- (Contrived) A problem in the format string may not be the only reason formatting can fail, and panic in `to_string`. For example the `RFC2822` formatting item returns an error if the year is negative or more than 4 digits, because that is not supported by that standard. But admittedly this is the only case I found.

pitdicker:
Found some interesting history.

In 2017 `ToString::to_string` was changed to panic on errors ([issue](https://github.com/rust-lang/rust/issues/40103), [PR](https://github.com/rust-lang/rust/pull/40117)).

The documentation for `Display` says, slightly hidden in the [`fmt` module](https://doc.rust-lang.org/nightly/std/fmt/index.html#formatting-traits):

> Additionally, the return value of this function is `fmt::Result` which is a type alias of `Result<(), std::fmt::Error>`. Formatting implementations should ensure that they propagate errors from the Formatter (e.g., when calling `write!`). However, they should never return errors spuriously. **That is, a formatting implementation must and may only return an error if the passed-in `Formatter` returns an error.** This is because, contrary to what the function signature might suggest, string formatting is an infallible operation. This function only returns a result because writing to the underlying stream might fail and it must provide a way to propagate the fact that an error has occurred back up the stack.

So that the `Display` implementation of `DelayedFormat` can return an error is indeed wrong according to this.

pitdicker:
My new thoughts on this PR (I hope I now finally have all the context...)

- Validation should happen in the various `format` methods, as they know the available fields of the type.
- `StrftimeItems::new` is not really the right place for the validation. Doing it here would require running the iterator (and parsing the formatting string) three times. I would keep it as it currently is.
- Validation should check against:
    - `Item::Error`
    - `Internal(InternalFixed { val: InternalInternal::TimezoneOffsetPermissive })`, which currently panics when formatting.
    - Any fields that are not part of the supplied type.
- The `RFC2822` item is not available as a formatting specifier. Maybe we should remove it, and let users call the already existing explicit function to format to this. It is the only formatting specifier that can return an error.

djc:
I feel like this ends up being a lot of pain while not getting us to the API we really want. Please have a look at #1127 and give feedback on the proposed design.

pitdicker:
> I feel like this ends up being a lot of pain while not getting us to the API we really want. Please have a look at #1127 and give feedback on the proposed design.

While only a first step towards a proper solution, we did get `StrftimeItems::parse` which can return an error if the format string is invalid in #1184.

Closing this PR. @jaggededgedjustice thanks for working on this though!

Issue #903, opened 2022-12-13
`NaiveDateTime::from_timestamp_millis(-1000)` does not compare equal to `NaiveDateTime::from_timestamp_opt(-1, 0)`


umanwizard:
Title says it all. This seems to be because the former gets interpreted as -2 seconds and 1 billion nanoseconds, whereas the latter gets interpreted as -1 seconds and 0 nanoseconds.

These feel to me like they should be equal, but I might be missing some subtlety.

esheppa:
This has been fixed by: https://github.com/chronotope/chrono/pull/885 but is yet to be released. If possible do you mind testing against the HEAD of the `0.4.x` branch?

PR #905, opened 2022-12-13
Add test proving that #903 is fixed in 0.4.x head

umanwizard:
Fixes https://github.com/chronotope/chrono/issues/903


src/naive/datetime/tests.rs
```diff
@@ -33,7 +33,7 @@ fn test_datetime_from_timestamp_millis() {
     // Test that the result of `from_timestamp_millis` compares equal to
     // that of `from_timestamp_opt`.
     let secs_test = [0, 1, 2, 1000, 1234, 12345678, -1, -2, -1000, -12345678];
-    for secs in secs_test {
+    for secs in secs_test.iter().cloned() {
```
djc, line 36: Nit: in this case, `copied()` might be nicer.

djc (approved):

PR #901, opened 2022-11-29
Enable clippy lints up to Rust 1.48

LingMan:
Enables clippy to show lints applicable for Rust versions 1.39..=1.48 and fixes all resulting warnings.

PR #907, opened 2022-12-14
Adding automatic dependency management

DarkWanderer:
There is an open security advisory for `time` library which `chrono` is dependent upon: https://github.com/DarkWanderer/DynamicOperations/security/dependabot/4
Having `dependabot` integration will help to react to such issues in timely manner

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: cargo
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,8 +3,12 @@ updates:
   - package-ecosystem: cargo
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
+  - package-ecosystem: cargo
+    directory: "/fuzz/"
+    schedule:
+      interval: "weekly"
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"


djc (approved):
Thanks!

Note that we will not be able to upgrade to time 0.1 in the chrono 0.4.x series due to compatibility constraints but we've already removed it from the main branch.

esheppa:
From what I've seen elsewhere dependabot can often provide a lot of PRs for semver patch updates. I wonder whether we would be better off adding `cargo audit` and similar to our CI runs instead?

esheppa:
Looks like we already have dependabot security updates enabled for this repo, and from what I can see over at [SO](https://stackoverflow.com/questions/64047526/how-to-get-dependabot-to-trigger-for-security-updates-only) this is the only way to get dependabot to do security updates only.

djc:
I usually enable a Dependabot config like this in most of the projects I maintain. It will send PRs for semver-incompatible releases in our dependencies, which I think is useful, but it also has pretty good support for ignoring new releases (like time 0.3.x). It will not send PRs for semver-compatible releases, so I think this would be good addition.

I also usually add cargo-deny to my CI configurations, which will help specifically with the security aspects. I'll submit a PR for that.

LingMan:
From what I've seen Dependabot only sends PRs for semver patch updates if `Cargo.lock` is checked in. Even then it's pretty easy to disable them [like this](https://github.com/rust-lang/cargo-bisect-rustc/commit/72b3b583d86ff6f9de8ac9ba085d4f219aff9509).

I'd only change the interval to `weekly`, since checking for updates every single day is way overkill.

LingMan:
Oh, and it would be good to add an entry covering `fuzz/Cargo.toml` as well.

LingMan:
Don't worry about the failed lint check. That's a new clippy warning that shows up since Rust 1.66 was released a few hours ago. A fix is included in this [PR](https://github.com/chronotope/chrono/pull/901).

LingMan (approved):

.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
+version: 2
+updates:
+  - package-ecosystem: cargo
```
LingMan, line 3: Forgot the quotes around `cargo`. Same in the second entry below.
```suggestion
  - package-ecosystem: "cargo"
```

LingMan (changes requested):

.github/dependabot.yml
DarkWanderer: No, I didn't forget anything, it works this way. YML spec allows for unquoted string values. 

The fact that I listened to suggestions above does not mean that I will spend my time on nitpicks however

LingMan: I only noticed since you did use quotes for `github-actions`. My apologies if I somehow offended you. Personally I'm always happy if someone points out such inconsistencies in my code.

esheppa (approved):
Looks good, thanks @DarkWanderer for the PR and @djc for the clarification

PR #908, opened 2022-12-15
update dependencies

spencerbart:


esheppa:
These look to be mostly patch versions, aside from `arbitrary`. Unless there is a security reason to update, we normally wouldn't increase the patch version unless we needed the specific functionality from that patch, and keeping the patch versions constant gives the best chance of that dependency being reused in the widest number of projects. I'd be happy to merge this with just the `arbitrary` and `libfuzzer-sys` changes however.

There is further discussion at https://github.com/chronotope/chrono/pull/907

djc:
Even the `arbitrary` version is semver-compatible, so I don't think we should update it unless we can show that our usage of `Arbitrary` is incompatible with the explicit version. Would be happy to take the `libfuzzer-sys` bump though.

@spencerbart reminder that in `Cargo.toml` the default is the `^` operator, so chrono's dependencies are already compatible with all of these releases -- that's why changes like these are unnecessary (and can sometimes be counterproductive -- I've seen it happen that someone bumped their dependency requirement to a recent patch version only for that version to be yanked).

PR #920, opened 2022-12-20
When compared to "date" program, the week numbers are off

raphaelroosz:
Hello,

I have been learning rust recently and started a project which would help me practice several features. For doing so I have been writing an equivalent of the "find" command. Goal is to have same API and same output. I have been using "chrono" for the date handling.
When testing it extensively with random dates, I found out discrepancies, which I expected, but by looking deeper, the discrepancy are in the %U and %W format in chrono crate.

When checking the code in findutils-4.9.0/gl/lib/nstrftime.c, one can find the following code for %W:
`(tp->tm_yday        - (tp->tm_wday - 1 + 7) % 7                 + 7) / 7`
The equivalent in chrono is chrono-0.4.23/src/format/mod.rs, one can find the following code for %W:
`(d.ordinal() as i32 - d.weekday().num_days_from_monday() as i32 + 7) / 7`
with num_days_from_monday() being between 0 and 6 so it does match the modulo 7.

Nothing puzzling, math formula seems right except there are dates where it does not match.
Here is why:
the C version is 0 based, the rust version is 1 based, but have same formula !

Pull Request contains:
- as many formats as possible that could be compared and tested extensively against the "date" command,
- the 1 is substracted from calculation (and thus we end up with +6 instead of +7)

### DIFF
--- a/src/format/mod.rs
+++ b/src/format/mod.rs
@@ -521,10 +521,10 @@ fn format_inner<'a>(
             use self::Numeric::*;
 
             let week_from_sun = |d: &NaiveDate| {
-                (d.ordinal() as i32 - d.weekday().num_days_from_sunday() as i32 + 7) / 7
+                (d.ordinal() as i32 - d.weekday().num_days_from_sunday() as i32 + 6) / 7
             };
             let week_from_mon = |d: &NaiveDate| {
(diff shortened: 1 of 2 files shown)


djc:
Thanks for your testing! Could you please retarget this PR at the 0.4.x branch (or open a new PR targeting the 0.4.x branch)?

PR #921, opened 2022-12-20
When compared to "date" program, the week numbers are off

raphaelroosz:
This time PR against the 0.4.x branch as asked.

Hello,

I have been learning rust recently and started a project which would help me practice several features. For doing so I have been writing an equivalent of the "find" command. Goal is to have same API and same output. I have been using "chrono" for the date handling.
When testing it extensively with random dates, I found out discrepancies, which I expected, but by looking deeper, the discrepancy are in the %U and %W format in chrono crate.

When checking the code in findutils-4.9.0/gl/lib/nstrftime.c, one can find the following code for %W:
`(tp->tm_yday        - (tp->tm_wday - 1 + 7) % 7                 + 7) / 7`
The equivalent in chrono is chrono-0.4.23/src/format/mod.rs, one can find the following code for %W:
`(d.ordinal() as i32 - d.weekday().num_days_from_monday() as i32 + 7) / 7`
with num_days_from_monday() being between 0 and 6 so it does match the modulo 7.

Nothing puzzling, math formula seems right except there are dates where it does not match.
Here is why:
the C version is 0 based, the rust version is 1 based, but have same formula !

Pull Request contains:
- as many formats as possible that could be compared and tested extensively against the "date" command,
- the 1 is substracted from calculation (and thus we end up with +6 instead of +7)

### DIFF
--- a/tests/dateutils.rs
+++ b/tests/dateutils.rs
@@ -117,8 +117,8 @@ fn try_verify_against_date_command_format() {
         return;
     }
     let mut date = NaiveDate::from_ymd_opt(1970, 1, 1).unwrap().and_hms_opt(12, 11, 13).unwrap();
-    while date.year() < 2078 {
+    while date.year() < 2008 {
         verify_against_date_command_format_local(date_path, date);
-        date += chrono::Duration::days(1);
+        date += chrono::Duration::days(55);
     }
 }
--- a/tests/dateutils.rs
+++ b/tests/dateutils.rs
@@ -75,7 +75,7 @@ fn try_verify_against_date_command() {
     }
 }
 
-#[cfg(unix)]
+#[cfg(target_os = "linux")]
 fn verify_against_date_command_format_local(path: &'static str, dt: NaiveDateTime) {
     let required_format = "d%d D%D F%F H%H I%I j%j k%k l%l m%m M%M S%S T%T u%u U%U w%w W%W X%X y%y Y%Y z%:z";
     // a%a - depends from localization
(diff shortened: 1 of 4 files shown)


tests/dateutils.rs
```diff
@@ -74,3 +74,51 @@ fn try_verify_against_date_command() {
...
+fn try_verify_against_date_command_format() {
+    let date_path = "/usr/bin/date";
+
+    if !path::Path::new(date_path).exists() {
+        // date command not found, skipping
+        // avoid running this on macOS, which has path /bin/date
+        // as the required CLI arguments are not present in the
+        // macOS build.
+        return;
+    }
```
djc, line 118: If we don't like the `date` command on macOS, how about just guarding the test with `cfg(target_os = "linux")` instead of relying on path differences? The most important thing here is to have this in CI anyway.

tests/dateutils.rs
```diff
@@ -74,3 +74,51 @@ fn try_verify_against_date_command() {
...
+    let mut date = NaiveDate::from_ymd_opt(1970, 1, 1).unwrap().and_hms_opt(12, 11, 13).unwrap();
+    while date.year() < 2078 {
```
djc, line 120: Hmm, how long does this test run? This seems a little intense to run on every CI run.

djc (commented):
Thanks, this is great work!

tests/dateutils.rs
esheppa: This should be ok as it is normally avoided by running `cargo test --lib` and `cargo test --doc`. It will only be run in the `timezones_linux` section of the CI.

tests/dateutils.rs
djc: @raphaelroosz would you mind addressing this so we can merge this? Would be good to get your fix out!

raphaelroosz: Updated on the test I added.

djc (approved):
LGTM modulo the formatting fixes, thanks!

esheppa (approved):
Thanks @raphaelroosz, this looks good! If possible would you mind fixing the formatting changes so that CI passes, then this is good to be merged

esheppa:
While investigating this I think I've discovered an issue with the "%V" specifier as well, but I'll open a separate issue and PR to get that fixed, then add it to the test against the date commands

pitdicker:
These changes where merged in https://github.com/chronotope/chrono/pull/966.

PR #925, opened 2022-12-26
Implement generic assign operations for various types

yottalogical:
Includes:
* `oldtime::Duration`
* `datetime::DateTime`
* `naive::date::NaiveDate`
* `naive::datetime::NaiveDateTime`
* `naive::time::NaiveTime`

Did not implement it for `date::Date`. This is because `Add` expects to be passed the owned object, but `AddAssign` only has access to a mutable reference. Since `Date` doesn't implement `Copy`, that makes it impossible to generically and safely implement `AddAssign` using `Add` without cloning. It might be possible with unsafe code, but that's probably over the top. `date::Date` is deprecated anyways.

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -912,16 +912,6 @@ impl<Tz: TimeZone> Add<OldDuration> for DateTime<Tz> {
     }
 }
 
-impl<Tz: TimeZone> AddAssign<OldDuration> for DateTime<Tz> {
-    #[inline]
-    fn add_assign(&mut self, rhs: OldDuration) {
-        let datetime =
-            self.datetime.checked_add_signed(rhs).expect("`DateTime + Duration` overflowed");
-        let tz = self.timezone();
-        *self = tz.from_utc_datetime(&datetime);
-    }
-}
-
 impl<Tz: TimeZone> Add<Months> for DateTime<Tz> {
     type Output = DateTime<Tz>;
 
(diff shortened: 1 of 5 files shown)


src/datetime/mod.rs
```diff
@@ -982,6 +962,30 @@ impl<Tz: TimeZone> Sub<Days> for DateTime<Tz> {
     }
 }
 
+impl<T, Tz: TimeZone> AddAssign<T> for DateTime<Tz>
+where
+    NaiveDateTime: Add<T, Output = NaiveDateTime>,
```
pitdicker, line 967: This is not correct. The `Add` implementation of `Days` and `Months` with `DateTime` is timezone-aware. I think the correct way to do this is to just add the individual implementations.

src/datetime/mod.rs
```diff
@@ -982,6 +962,30 @@ impl<Tz: TimeZone> Sub<Days> for DateTime<Tz> {
...
+    fn add_assign(&mut self, rhs: T) {
+        let datetime = self.datetime + rhs;
+        let tz = self.timezone();
+        *self = tz.from_utc_datetime(&datetime);
+    }
+}
+
+impl<T, Tz: TimeZone> SubAssign<T> for DateTime<Tz>
+where
+    NaiveDateTime: Sub<T, Output = NaiveDateTime>,
```
pitdicker, line 979: Same for `Sub`

pitdicker (changes requested):
Sorry to add some late comments to this PR.

These implementations make chrono a little nicer to use. Methods such as `check_add_*` have their use, but these `AddAssign` implementations are just convenient. So :+1: from me.

----------

But: a generic implementation will not only affect chrono, but also library users.

If a user has `impl Add<MyType> for NaiveDate` and `impl AddAssign<MyType> for NaiveDate`, that would no longer compile because of the new generic implementation of  `AddAssign` here, right?

This would be a (minor) breaking change for 0.4.x.

I don't think a blanked implementation is a good thing to have in general. Or said differently: if it is just a way to save some code in chrono, it is better to write the individual implementations.

pitdicker:
@yottalogical would you be willing to add the individual assign operations? And @djc would you agree they can be useful?

pitdicker:
Closing due to inactivity. The related issue  to track is https://github.com/chronotope/chrono/issues/70.

PR #926, opened 2022-12-30
Add support for parsing JSON timestamps from strings.

rayhaanj:
This change adds support for parsing string as (numeric) timestamps.

I had a look at the tests but was not sure where I should put the test for this feature, as the current logic to run tests is calling `datetime::test_decodable_json` which does not seem to be exercising the visitor path? Let me know where I should put them and I'll add the tests.

### DIFF
--- a/src/datetime/serde.rs
+++ b/src/datetime/serde.rs
@@ -221,13 +221,16 @@ pub mod ts_nanoseconds {
     ///
     /// let my_s: S = serde_json::from_str(r#"{ "time": -1 }"#)?;
     /// assert_eq!(my_s, S { time: Utc.timestamp_opt(-1, 999_999_999).unwrap() });
+    ///
+    /// let my_s: S = serde_json::from_str(r#"{ "time": "1526522699918355733" }"#)?;
+    /// assert_eq!(my_s, S { time: Utc.timestamp_opt(1526522699, 918355733).unwrap() });
     /// # Ok::<(), serde_json::Error>(())
     /// ```
     pub fn deserialize<'de, D>(d: D) -> Result<DateTime<Utc>, D::Error>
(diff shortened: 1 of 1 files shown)


djc (approved):
Makes sense, thanks!

esheppa:
I think it's probably worthwhile having a few tests to verify this works. They could just be placed at the bottom of the `serde.rs` file along with the other tests there.

jtmoon79 (changes requested):
Needs test cases.

@rayhaanj is it possible to add test cases for "before"?  That is, test cases that exercise some serde code before your changes. Then another commit with your changes + new test cases (or test case changes).

pitdicker:
I'll try to help get this PR ready.
Added some very basic doctests. Good this was requested in previous reviews, as it didn't work yet :smile:.
We now need to use `deserialize_any` instead of `deserialize_i64`.

pitdicker:
Hmm, [`desertialize_any`](https://docs.rs/serde/latest/serde/de/trait.Deserializer.html#tymethod.deserialize_any) may not be the best thing to reach for:
> Require the `Deserializer` to figure out how to drive the visitor based on what data type is in the input.
>
> When implementing `Deserialize`, you should avoid relying on `Deserializer::deserialize_any` unless you need to be told by the `Deserializer` what type is in the input. Know that relying on `Deserializer::deserialize_any` means your data type will be able to deserialize from self-describing formats only, ruling out Postcard and many others.

djc:
Yeah, I don't think we should restrict to self-describing formats only.

pitdicker:
I've opened a question in the serde repo just in case. I think this is not going to work out, but no need to hurry closing this PR yet :smile:.

PR #927, opened 2022-12-30
Allow guaranteed-no-panic conversion from NaiveDate to DateTime

Ekleog:
Hello people,

First, thank you for making chrono! It's very helpful to write code that (hopefully) will work on any timezone.

However, this "make it work on any timezone" makes me scared about doing some stuff that still sounds really reasonable.

So I'm opening this PR to kill two birds with one stone: asking my question ("how can I make a DateTime at the first second of a day"), and providing an easy way for others to know and use the answer.

Basically, I'm wondering whether the implementation of this PR is correct. Is it possible for eg. a summer time change change to happen across midnight?

If the implementation of this PR is always correct, then I think it should land, because otherwise writing the code as an end-user of the library does not make it obvious that it is actually panic-free. And this leads to convoluted ways to work around it. For instance my current solution is I already have a DateTime, and add 86400-its number of seconds since the beginning of the day; and that gives me a very-approximated-during-summer-time-changes but guaranteed no-panic solution.

If the implementation of this PR is incorrect, how would you go about solving the problem at hand, and do you think it would be possible to integrate it into chrono itself?

Either way, have a good end-of-year holiday!

### DIFF
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -25,7 +25,7 @@ use crate::format::{Item, Numeric, Pad};
 use crate::month::Months;
 use crate::naive::{IsoWeek, NaiveDateTime, NaiveTime};
 use crate::oldtime::Duration as OldDuration;
-use crate::{Datelike, Duration, Weekday};
+use crate::{DateTime, Datelike, Duration, TimeZone, Weekday};
 
 use super::internals::{self, DateImpl, Mdf, Of, YearFlags};
 use super::isoweek;
@@ -873,6 +873,12 @@ impl NaiveDate {
         nano: u32,
     ) -> Option<NaiveDateTime> {
         NaiveTime::from_hms_nano_opt(hour, min, sec, nano).map(|time| self.and_time(time))
+    }
+
+    /// Makes a DateTime for the first second of the day in the provided timezone.
+    #[inline]
+    pub fn and_first_second<Tz: TimeZone>(&self, tz: Tz) -> DateTime<Tz> {
+        self.and_hms_opt(0, 0, 0).unwrap().and_local_timezone(tz).unwrap()
     }
 
     /// Returns the packed month-day-flags.


esheppa:
Hi @Ekleog - thanks for the PR.

> Basically, I'm wondering whether the implementation of this PR is correct. Is it possible for eg. a summer time change change to happen across midnight?

Regarding this, I'm not sure whether this does occur in the existing `tzdb` however I don't think there is any restriction against it occurring, and so we have to assume essentially any possible transition could occur (potentially aside from cases like an entire day being skipped).

> For instance my current solution is I already have a DateTime, and add 86400-its number of seconds since the beginning of the day; and that gives me a very-approximated-during-summer-time-changes but guaranteed no-panic solution.

If you already have a valid `DateTime<Tz>` at the start of your desired day, you can do:
(Please note, there is a bug with this which will be fixed in `0.4.24`, in the meantime please use the `0.4.x` branch if you are using this feature)

```rust
// if there is an intervening offset transition, 
// this will add the required number of hours (less or more than 24) 
// to ensure that the output has the same local timestamp as the LHS input
let next_start = my_date_time + Days::new(1);
```

> If the implementation of this PR is incorrect, how would you go about solving the problem at hand, and do you think it would be possible to integrate it into chrono itself?

The PR looks like a good start, and will be correct in some cases, but unfortunately it may be incorrect in timezones which have offset transitions.

Firstly getting the start of day as a NaiveTime should be possible infallibly using the below as of https://github.com/chronotope/chrono/commit/e7d4402853c34bf1fbb7f59dfe257dae517dc7e2.

```rust
let my_naive_date_time = my_naive_date.and_time(NaiveTime::MIN);
```

However adding the timezone is problematic for some timezones as that specific local time may not exist in that timezone. Essentially this works for any `Utc` or `FixedOffset` but not timezones that have offset transitions.

What you can do is the following for UTC (this works because UTC has the same local and UTC time)

```rust
let my_utc_time = Utc.from_utc_datetime(my_naive_date_time); // variable from above

```

And the following for `FixedOffset`

```rust
let my_fixed_offset = FixedOffset::east_opt(7 * 60 * 60); 
let my_fixed_offset_time = my_fixed_offset.and_utc_datetime(my_naive_date_time - my_fixed_offset); // Assuming the naivedatetime is local, translate it to equivalent UTC time
```

Perhaps it would be useful to add equivalent [`MIN_UTC`](https://docs.rs/chrono/latest/chrono/struct.Date.html#associatedconstant.MIN_UTC) and [`MAX_UTC`](https://docs.rs/chrono/latest/chrono/struct.Date.html#associatedconstant.MAX_UTC) to `NaiveDate` given that `Date<Tz>` is now deprecated. Would you be interested in doing that? 

Separately it would be interesting to investigate having a `min_tz` function which finds the earliest valid time on that date in that timezone. We have a basic implementation [here](https://github.com/chronotope/chrono/pull/883/files#diff-219d6c6d839d0a706bca2e506cd44cde210cf3191774f4af85927a1663c3c9c8R55) but as you can see it is not ideal as it has to loop, potentially with some heuristics.

Ekleog:
Awesome, thank you for your answer! I hadn't known of your #883 PR. With it I think my problem would be solved!

I think this PR can thus reasonably be closed, as I don't see anything else actionable here.

Issue #928, opened 2023-01-01
Resetting time to 00:00:00.0


rimutaka:
It takes too many lines to do a simple and commonly used (at least in my case) operation:
```
let x = dt.with_hour(0).unwrap().with_minute(0).unwrap().with_second(0).unwrap();
```
Is there a simpler way of achieving the same?

if not, would you consider a PR to do a reset like that in a single step? E.g.
```
let x = dt.reset_time();
```

esheppa:
Thanks for the question @rimutaka - one possible option is to do:

```rust
let x = NaiveTime::MIN;
```

This works as of https://github.com/chronotope/chrono/commit/e7d4402853c34bf1fbb7f59dfe257dae517dc7e2 and will be released in `0.4.23` but can currently be accessed on branch `0.4.x`.

rimutaka:
@esheppa , Eric, sorry, I probably miss something important here.

**Is there a way to replace the time only portion in `DateTime<Tz>` with `NaiveTime::MIN` as you suggested?**

I have an immutable var with `DateTime<Tz>`, say `let x = Utc::now();`. I need to reset the time portion of that to zeros, e.g. `let y = x.reset_time();` or `x.with_time(NaiveTime::MIN);`.

* x: 2023-01-03T23:19:08.281426984+00:00
* y: 2023-01-03T00:00:00.0+00:00

And I need to do it often enough in my code to find the current multi-step approach of resetting hr, min, sec, ns too verbose.

rimutaka:
@esheppa , Eric, I submitted a draft PR for this feature under #930 . Can you or any other maintainers take a look before I spend any more time polishing it?

**Unrelated**: the pipeline has a step with `cargo clippy --color=always -- -D warnings -A clippy::manual-non-exhaustive`, but there are no warnings related to `clippy::manual-non-exhaustive`. Should that option be removed?

rimutaka:
Looks like the simplest way of resetting the time for `DateTime<Tz>` is this:
```
let x = dt_w_tz.date_naive()
        .and_time(NaiveTime::MIN)
        .and_local_timezone(x.timezone())
        .unwrap();
```
 The maintainers looked into my PR #930 to simplify this, but the gains were marginal and we would still need to use `unwrap()` to avoid a potential panic.

PR #930, opened 2023-01-05
Added reset_time() to Timelike traits

rimutaka:
Comparing or sorting the date only part of timestamps is slow and difficult if the time part is present. It is often faster to reset the time part to `00:00:00.0` before adding to a list or a database.

This PR is an implementation of a shortcut function to reset time on DateTime and NaiveDateTime structures in a single step.

The added function seems to be infallible and does not need to be wrapped into _Option_ or _Result_.

Issue: #928

## Changes

* added the trait itself
* implemented the trait for DateTime, NaiveDateTime and NaiveTime
* implemented tests for all trait implementations

## ToDo

* [ ] Get someone to look at the draft
* [x] Update the docs to include an example

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -914,6 +914,26 @@ impl<Tz: TimeZone> Timelike for DateTime<Tz> {
     fn with_nanosecond(&self, nano: u32) -> Option<DateTime<Tz>> {
         map_local(self, |datetime| datetime.with_nanosecond(nano))
     }
+
+    /// Returns an instance of `DateTime<Tz>` with the same _date_ and _offset_ parts.
+    /// The time part gets reset to `00:00:00.0`.
+    ///
+    /// # Example
+    ///
+    /// * before reset: `2014-11-28T21:45:59.324310806+00:00`
+    /// * after reset:  `2014-11-28T00:00:00.0+00:00`
+    #[inline]
+    fn reset_time(&self) -> DateTime<Tz> {
+        self.clone()
+            .with_hour(0)
+            .unwrap()
+            .with_minute(0)
+            .unwrap()
+            .with_second(0)
+            .unwrap()
+            .with_nanosecond(0)
+            .unwrap()
+    }
 }
 
 // we need them as automatic impls cannot handle associated types
--- a/src/datetime/tests.rs
+++ b/src/datetime/tests.rs
@@ -6,6 +6,7 @@ use crate::offset::Local;
 use crate::offset::{FixedOffset, TimeZone, Utc};
 #[cfg(feature = "clock")]
 use crate::Datelike;
+use crate::Timelike;
 use crate::{
     naive::{NaiveDate, NaiveTime},
     TimeDelta,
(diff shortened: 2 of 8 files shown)


rimutaka:
@esheppa , are you saying there is need for this function because the existing methods are good enough?

rimutaka:
@esheppa , Eric, sorry to bug you with this stuff. Are you saying the existing methods are good enough to achieve the same result?

I decided to submit a PR after updating several projects to 0.4.23. All of them made use of `dt_with_tz.date().and_hms(0, 0, 0)`.

You deprecated `Date` and accessing the same functionality became harder. I can see your reasons for dropping Date, so I thought adding a shortcut for a commonly used set of HMS values that is "safe" and requires no _unwrap()_ would be a good idea. I don't have it in front of me, but resetting the time was mentioned in at least one other recent issue.

Here is an example from one of the projects I had to update ...

Old line:
```
report.date_init = dt_w_tz.date().and_hms(0, 0, 0).to_rfc3339();
```
new line (the simplest way I could find)
```
report.date_init = dt_w_tz.date_naive()
        .and_time(NaiveTime::MIN)
        .and_local_timezone(x.timezone())
        .unwrap()
        .to_rfc3339();
```
It's doable, but is not very ergonomic and having an unhandled _unwrap()_ is a red flag.

Would adding something like `DateTime<Tz>.with_hms_opt(...)` be a better solution because it is more generic?

If someone wants to reset time to zeros or set some values that are guaranteed to resolve to a valid DateTime<Tz> they can use `.unwrap()` without checking. It would be more ergonomic than `.with_hour()....with_minute()...unwrap()...`.

djc:
(This targets the main branch but if you'd like to get this into a 0.4.x release it should probably target the 0.4.x branch.)

rimutaka:
@djc , I see this same pattern in several projects I am involved in. It wasn't me who came up with this. I inherited it from completely unrelated devs.

The common theme there is to know what happened on Day X. E.g. group items by day or select only Day X.
Discarding the time portion allows to compare dates as `==`, not a range. That's a lot faster and easier weather the date is a string or a proper datetime datatype.

On the other hand, using RFC3339 makes it more interoperable - easier to parse and manipulate the dates. It is also unambiguous, while `Date` often depends on the locale, e.g. _DD-MM-YYYY_ vs _MM-DD-YYYY_ vs completely unpredictable `YY-MM-DD`.

In one project both full and zero-time timestamps are passed to the front-end. The script picks one item from the list, resets the time and runs a search for all the items with the matching zero-time timestamp in the list.

I agree that resetting time instead of using `Date` is a hacky approach, but I'm sure it's dictated by the constraints of those other systems. I am just trying to make my life easier and not do multiple unchecked unwraps for a single repetitive op.
The first question in the code review after upgrading to 0.4.23 was _"Can this unwrap() panic?"_. :)

**Would adding `.with_hms_opt(...)` to `DateTime<Tz>` be a better way?**

esheppa:
Thanks for the extra comments, there are a couple of things I'd like to note.

* Using `.and_time(NaiveTime::MIN).and_local_timezone(x.timezone())` will always be fallible as the midnight time may actually not exist on that day in that timezone. 
* Sending the datetimes with a zero timestamp to the frontend makes sense to me because it may be the case that there is no date datatype and only a datetime datatype there - but would naive datetimes be sufficient in this case?
* If you just want the earliest representable time on the day then potentially a `Day<Tz>` as per https://github.com/chronotope/chrono/pull/883 would be useful for you

> Would adding .with_hms_opt(...) to DateTime<Tz> be a better way?

This is possible, and similarly `and_local_time(...)` would also be possible but these would be fallible not just for out of range hours/minutes/seconds but also due to the local time potentially not existing.

Here are some of my thoughts for the issues you described above (all examples assume you start with a `Vec<DateTime<Tz>>`):

> The common theme there is to know what happened on Day X. E.g. group items by day or select only Day X.

For grouping by Day X, I'd probably use a map with a vector, so for example:

```rust
let my_groups = my_datetimetzs.iter().fold(
        BTreeMap::<NaiveDate, Vec<DateTime<Utc>>>::new(),
        |mut acc, x| {
            acc.entry(x.date_naive())
                .and_modify(|ds| ds.push(*x))
                .or_insert(Vec::from([*x]));
            acc
        },
    );
```

For filtering by Day X, I'd use:

```rust 
    let only_on_date = my_datetimetzs
        .iter()
        .filter(|d| d.date_naive() == NaiveDate::from_ymd_opt(2023, 1, 7).unwrap());
```

For creating (potentially locally invalid) NaiveDateTime with zero timestamp I'd use (where `.date_naive().and_time(NaiveTime::MIN)` does the equivalent to `.reset_time()`):

```rust
let my_rfc3339s = my_datetimetzs
        .iter()
        .map(|d| d.date_naive().and_time(NaiveTime::MIN).format("%FT%T").to_string())
        .collect::<Vec<_>>();
```


From your earlier comment:

> It's doable, but is not very ergonomic and having an unhandled unwrap() is a red flag.

This unwrap is due to the assertion that that local time is valid for that timezone, and so there was a potential panic before that was not handled, and now it is just being made explicit. In the case that you want string representations of the date that look similar to rfc3339 but you don't want to have to have a `DateTime<Tz>` first, then you can just make a strftime format string similar to my above code, eg `%FT%T`.

rimutaka:
@djc , @esheppa , thank you for looking into this.

Issue #932, opened 2023-01-05
Weird min/max constants


petar-dambovaliev:
Hey there, i needed min and max constants for UTC timezone and i was a little confused by the API.

```
error[E0283]: type annotations needed
   --> workspace/backend/src/lib.rs:397:27
    |
397 |             let mut min = DateTime::MAX_UTC;
    |                           ^^^^^^^^^^^^^^^^^ cannot infer type of the type parameter `Tz` declared on the struct `DateTime`
    |
    = note: cannot satisfy `_: chrono::TimeZone`
note: required by a bound in `chrono::DateTime::<Tz>::MAX_UTC`
```

Why do i need to write
```
let mut min = DateTime::<Utc>::MAX_UTC;
```
When you read this code, it suggests that there might be a non UTC constant `MAX_UTC`.

It looks like these constants should not be associated or they should be specialized with the same name.

```
DateTime::<Utc>::MAX
```

djc:
The consts are independent of the timezones you pass. Unfortunately I don't think it's possible to elide the type argument when referencing the associated const, even though the const value does not depend on it. Given those constraints, I think this is the best we can do?

petar-dambovaliev:
> If you think there is a better solution, feel free to submit a PR.

Okay, cool. I will.

pitdicker:
Fixed on main with https://github.com/chronotope/chrono/pull/1303.

PR #935, opened 2023-01-09
move MIN_UTC, MAX_UTC to a specialized impl

petar-dambovaliev:
For more info visit the [issue](https://github.com/chronotope/chrono/issues/932)
This is a breaking change.

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -406,7 +406,9 @@ impl<Tz: TimeZone> DateTime<Tz> {
             false => None,
         }
     }
+}
 
+impl DateTime<Utc> {
     /// The minimum possible `DateTime<Utc>`.
     pub const MIN_UTC: DateTime<Utc> = DateTime { datetime: NaiveDateTime::MIN, offset: Utc };
     /// The maximum possible `DateTime<Utc>`.


djc:
So how does this change how they can be referenced? Did you test that it changes? It's not at all obvious to me.

petar-dambovaliev:
> So how does this change how they can be referenced? Did you test that it changes? It's not at all obvious to me.

I tested it out.
Previously, you had to write
```
let mut min: DateTime<Utc> = DateTime::MAX_UTC;
```
Now, it works as
```
let mut min = DateTime::MAX_UTC;
```
The generic argument has been inferred.

djc:
Yes, adding some tests would be useful.

petar-dambovaliev:
> Yes, adding some tests would be useful. As an alternative option, if we name these `MAX` and `MIN` (losing the `_UTC`) suffix this would be a compatible change which can go into 0.4.x.

Yes, removing the postfix makes sense.
Do you mean leaving the original constants where they were and adding these 2 new ones?

djc:
> > Yes, adding some tests would be useful. As an alternative option, if we name these `MAX` and `MIN` (losing the `_UTC`) suffix this would be a compatible change which can go into 0.4.x.
> 
> Yes, removing the postfix makes sense. Do you mean leaving the original constants where they were and adding these 2 new ones?

Yes. You'll likely also want to rebase your change on the 0.4.x branch (and change the target branch for the PR).

petar-dambovaliev:
> > > Yes, adding some tests would be useful. As an alternative option, if we name these `MAX` and `MIN` (losing the `_UTC`) suffix this would be a compatible change which can go into 0.4.x.
> > 
> > 
> > Yes, removing the postfix makes sense. Do you mean leaving the original constants where they were and adding these 2 new ones?
> 
> Yes. You'll likely also want to rebase your change on the 0.4.x branch (and change the target branch for the PR).

Okay, cool.

jtmoon79 (changes requested):
Needs test cases that enforce compile-time expectations.

For example,
```lang-rust
fn test_min_utc() {
     let m: DateTime<Utc> = DateTime::MIN_UTC;
}
```
(IIUC the code)

Ideally, this would be two commits:
1. a commit that adds test cases
2. a commit that adds code changes + (if necessary) test case adjustments

pitdicker:
This is in theory a semver-breaking change, but only for some *very* strange code:
```rust
let dt: DateTime<Utc> = DateTime::<FixedOffset>::MAX_UTC;
```
I.e. you currently have to supply some generic parameter, but then it would just be ignored and the return type is `DateTime<Utc>`.

But a GitHub search shows people are actually writing something like this. For example [1](https://github.com/gunosy/redash-searcher/blob/14074c4ebf850189b9803447c5248c06d2c9c927/redash-searcher-sync/src/app.rs#L215), [2](https://github.com/stagrim/Asta/blob/9302700fd1cdf62def760b8fabd142901f79a338/sasta/src/store/schedule.rs#L170), [3](https://github.com/princeofnubia/dozer/blob/cf49da5510e7d944f9c5f0d8dbc222aea37c1d8e/dozer-sql/src/pipeline/aggregation/min.rs#L135).

What it does show is that this issue is worth fixing.

@djc shall I rebase this on main?

pitdicker:
Merged via https://github.com/chronotope/chrono/pull/1303. @petar-dambovaliev Thanks for working on this!

PR #938, opened 2023-01-18
derive `Hash` for most pub types that also derive `PartialEq`

bruceg:
We needed to store `Item` alongside other data in a hash set, which requires all of the contained data be hashable. While I was adding the derive for that type, I looked for other pub types that could use it.

This is a backport of #852 to the `0.4.x` branch.


djc (approved):

djc:
Thanks!

esheppa (approved):
Looks good!

Issue #931, opened 2023-01-05
Deprecated `from_timestamp` method used in example for `DateTime`


greg-el:
The example given in https://docs.rs/chrono/latest/chrono/struct.DateTime.html#example appears to use the deprecated `from_timestamp`:

```
warning: use of deprecated associated function `chrono::NaiveDateTime::from_timestamp`: use `from_timestamp_opt()` instead
  --> src/main.rs:60:60
   |
60 |             date: DateTime::<Utc>::from_utc(NaiveDateTime::from_timestamp(61, 0), Utc),
   |                                                            ^^^^^^^^^^^^^^
   |
   = note: `#[warn(deprecated)]` on by default
```

I figure this should be as simple as replacing `NaiveDateTime::from_timestamp(61, 0)` with `NaiveDateTime::from_timestamp_opt(61, 0).unwrap()`, happy to open a PR to do so!

djc:
Would be great to have a PR for this! Be sure to target the 0.4.x branch, thanks!

pitdicker:
This issue is fixed by https://github.com/chronotope/chrono/pull/939.

PR #939, opened 2023-01-18
Update deprecated methods in `from_utc` example

greg-el:
`from_timestamp` and `Utc.timestamp` are deprecated, and have been replaced with `from_timestamp_opt` and `Utc.timestamp_opt` respectively, with unwrapping.

Related to https://github.com/chronotope/chrono/issues/931


djc (approved):
Thanks! Not sure why I missed these before.

PR #941, opened 2023-01-22
Fix panic in DateTime::checked_add_days

Ekleog:
Incidentally, add TimeDelta::try_* builders so that it's possible to build them without risking a panic.

This issue was found thanks to cargo-bolero-fuzzing my web server, without actually trying to fuzz chrono itself.

### DIFF
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -644,6 +644,10 @@ impl NaiveDate {
     ///     NaiveDate::from_ymd_opt(2022, 7, 31).unwrap().checked_add_days(Days::new(2)),
     ///     Some(NaiveDate::from_ymd_opt(2022, 8, 2).unwrap())
     /// );
+    /// assert_eq!(
+    ///     NaiveDate::from_ymd_opt(2022, 7, 31).unwrap().checked_add_days(Days::new(1000000000000)),
+    ///     None
+    /// );
     /// ```
     pub fn checked_add_days(self, days: Days) -> Option<Self> {
         if days.0 == 0 {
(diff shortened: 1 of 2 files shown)


djc:
Going to close this. We'll review (and probably merge) #942 (the 0.4.x version of this change) and then eventually that change will make it to the main branch as part of a regular merge of 0.4.x work into main.

djc:
Ahh, sorry, had missed that context. Will open this back up, then we can review it after the other change has made its way into main.

Ekleog:
Awesome thanks! :)

djc:
(Merge is pending in #986.)

djc:
Can you rebase this onto main? Should hopefully not be too painful.

esheppa (commented):
Looks good to me, pending rebase

jtmoon79 (changes requested):
Can you add test cases for the before and after for all changed functions?
e.g. `test_diff_days`, `test_weeks`, `test_try_weeks`, etc.

Specifically, two commits
1. a commit with new test cases
2. a commit of these changes + adjustments to those test cases (if needed)

Ekleog:
Sorry for the delay. I have basically not done any non-work software development for a few months, and have a few personal projects that will be higher priority for me than this PR, so I don't think I'll actually come back to this anytime soon.

Do you prefer if I leave the PR open so that someone else can easily notice it and take it over (and so that it tracks the fact that the panic is currently reachable on `main`), or should I just close it?

pitdicker:
@Ekleog That is very reasonable. If it is ok with you I can take over in a couple of days.

pitdicker:
The new methods are included in https://github.com/chronotope/chrono/pull/1137 together with other features for `Duration`. They are not yet merged however.

PR #942, opened 2023-01-22
Fix panic in DateTime::checked_add_days

Ekleog:
This is a backport of #941, except it needs to work around the fact that we can't modify the `time` crate.


djc (approved):
Thanks, great find!

Ekleog:
Awesome, thank you for the quick review!

esheppa (approved):
Looks good, thanks for the extra docs as well

PR #951, opened 2023-01-30
Apply clippy suggestions for Rust 1.67

djc:


djc:
@esheppa since this should be pretty uncontroversial, going to go ahead and merge this one.

Issue #898, opened 2022-11-26
Documentation of commit style in PRs


esheppa:
As we are using the "Rebase and Merge" strategy rather than the "Squash and Merge" strategy to merge PRs, we require fixes to be done in the underlying commits rather than in new commits during PR review. We should document this and have some recommendations of how to achieve this, something along the lines of usage of:

Redoing commits
* `git rebase -i HEAD~n` and using `edit`, `squash` and `exec`
Splitting commits
* `git reset` to set up the state for subsequently using `git add -i`
* `git add -i` and using the patch mode to stage some but not all of a given diff for a file

With the understanding that there are many ways to achieve the same result and this is just one of the possible ways

Issue #870, opened 2022-11-13
and_local_timezone naming


Ten0:
Hello,

I have a question about the naming of this function.

I don't understand what information the `local` is supposed to bear. What is a "local timezone" compared to a "timezone"?

I'm not particularly familiar with all the timezone-related science, but I think I would have named it simply `NaiveDateTime::in_timezone -> DateTime<Tz>`.

How is the current namig better?

Thanks,

djc:
I think the issue here is that there are two different meanings for combining a `NaiveDateTime` and a `TimeZone`:

* `NaiveDateTime` is already *in* `TimeZone`, clarify this in the type system
* `NaiveDateTime` is actually in `Utc`, need to add/subtract an offset to bring it into `TimeZone`

This is the reason for the `_local` bit. I think the `and_` prefix vs `with_` is simply because this is the convention in chrono, which possibly predates widespread usage of `with_` in the rest of the ecosystem.

Ten0:
Thank you! 

IMO:
- `naive_datetime.and_timezone(tz)` could hardly be misinterpreted as "Assume this is Utc, then convert it to provided timezone". 
- It would seem best (not sure if that's already the case) if the only way to bring a naive datetime that "is actually Utc" in an arbitrary timezone was to first assert it's in Utc by turning it into a timezone aware datetime, and then use the usual timezone conversion features. I don't think that is ever significantly more boilerplatish than having a function that does both, and seems always easier to read.

djc:
For context, see #711. @esheppa any opinions here?

esheppa:
> and we concluded that the clearest naming would be `naive_date.assume_timezone(tz)`

This sounds good to me! One option we had thought about originally was `assert_timezone` but I much prefer `assume_timezone`.

> Internally, we now use naive_date.assume_timezone_fallback_earlier(tz) and naive_date.assume_timezone_fallback_later(tz)

The plan is to eventually have some methods for this within `chrono`, you can see discussions here https://github.com/chronotope/chrono/issues/716 and here https://github.com/chronotope/chrono/pull/830. Would appreciate feedback or alternative thoughts on the ideas there

PR #952, opened 2023-01-31
Naive utc

klnusbaum:
It seems common to use `DateTime<FixedOffset>` as a kind of canonical representation of any given instant. With that said, the following code almost works:

```Rust
fn parse_date(date: &str) -> Result<DateTime<FixedOffset>> {
    let parsed = date
        .parse::<NaiveDate>()?
        .and_time(NaiveTime::default())
        .and_local_timezone(Utc)
        .into()
    Ok(parsed)
}
```

However, `and_local_timezone` returns a `LocalResult` which is kind of irksome in this case. I know the converting to `Utc` won't produce multiple results, but I still either have to call `unwrap` or do something more complicated with the `LocalResult`.

For the case where I know I'm going to Utc, could we just have a special function specifically for Utc? Would folks be opposed to adding the convenience?

### DIFF
--- a/src/naive/datetime/mod.rs
+++ b/src/naive/datetime/mod.rs
@@ -18,6 +18,7 @@ use crate::format::DelayedFormat;
 use crate::format::{parse, ParseError, ParseResult, Parsed, StrftimeItems};
 use crate::format::{Fixed, Item, Numeric, Pad};
 use crate::naive::{Days, IsoWeek, NaiveDate, NaiveTime};
+use crate::offset::Utc;
 use crate::oldtime::Duration as OldDuration;
 use crate::{DateTime, Datelike, LocalResult, Months, TimeZone, Timelike, Weekday};
 
@@ -887,12 +888,29 @@ impl NaiveDateTime {
     /// # Example
     ///
     /// ```
-    /// use chrono::{NaiveDate, Utc};
-    /// let dt = NaiveDate::from_ymd_opt(2015, 9, 5).unwrap().and_hms_opt(23, 56, 4).unwrap().and_local_timezone(Utc).unwrap();
-    /// assert_eq!(dt.timezone(), Utc);
+    /// use chrono::{NaiveDate, FixedOffset};
+    /// let hour = 3600;
+    /// let tz = FixedOffset::east_opt(5 * hour).unwrap();
+    /// let dt = NaiveDate::from_ymd_opt(2015, 9, 5).unwrap().and_hms_opt(23, 56, 4).unwrap().and_local_timezone(tz).unwrap();
+    /// assert_eq!(dt.timezone(), tz);
+    /// ```
     #[must_use]
     pub fn and_local_timezone<Tz: TimeZone>(&self, tz: Tz) -> LocalResult<DateTime<Tz>> {
         tz.from_local_datetime(self)
+    }
+
+    /// Converts the `NaiveDateTime` into the timezone-aware `DateTime<Utc>`.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use chrono::{NaiveDate, NaiveTime, Utc};
+    /// let dt = NaiveDate::from_ymd_opt(2023, 1, 30).unwrap().and_hms_opt(19, 32, 33).unwrap().and_utc();
+    /// assert_eq!(dt.timezone(), Utc);
+    /// ```
+    #[must_use]
+    pub fn and_utc(&self) -> DateTime<Utc> {
+        Utc.from_utc_datetime(self)
     }
 
     /// The minimum possible `NaiveDateTime`.
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -31,6 +31,7 @@ John Nagle <nagle@sitetruth.com>
 Jonas mg <jonasmg@yepmail.net>
 János Illés <ijanos@gmail.com>
 Ken Tossell <ken@tossell.net>
+Kurtis Nusbaum <kcommiter@gmail.com>
 Martin Risell Lilja <martin.risell.lilja@gmail.com>
 Richard Petrie <rap1011@ksu.edu>
 Ryan Lewis <ryansname@gmail.com>
--- a/src/naive/datetime/tests.rs
+++ b/src/naive/datetime/tests.rs
@@ -370,7 +370,7 @@ fn test_nanosecond_range() {
 }
 
 #[test]
-fn test_and_timezone() {
+fn test_and_local_timezone() {
     let ndt = NaiveDate::from_ymd_opt(2022, 6, 15).unwrap().and_hms_opt(18, 59, 36).unwrap();
     let dt_utc = ndt.and_local_timezone(Utc).unwrap();
     assert_eq!(dt_utc.naive_local(), ndt);
@@ -380,4 +380,12 @@ fn test_and_timezone() {
     let dt_offset = ndt.and_local_timezone(offset_tz).unwrap();
     assert_eq!(dt_offset.naive_local(), ndt);
     assert_eq!(dt_offset.timezone(), offset_tz);
+}
+
+#[test]
+fn test_and_utc() {
+    let ndt = NaiveDate::from_ymd_opt(2023, 1, 30).unwrap().and_hms_opt(19, 32, 33).unwrap();
+    let dt_utc = ndt.and_utc();
+    assert_eq!(dt_utc.naive_local(), ndt);
+    assert_eq!(dt_utc.timezone(), Utc);
 }


klnusbaum:
I think technically this is semver compatible with 0.4.x but happy to put it on the 0.5 branch if need be.

src/offset/local/mod.rs
```diff
@@ -51,7 +51,7 @@ mod tz_info;
 /// let dt: DateTime<Local> = Local::now();
 /// let dt: DateTime<Local> = Local.timestamp(0, 0);
 /// ```
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
```
klnusbaum, line 54: I added this so that we could have, IMHO, a better example for the `and_local_timezone` function. However, I can understand if this might be controversial. If we don't want to do this, I can change up the example in `and_local_timezone` to be something different.

klnusbaum:
waiting on #951 to get merged in order for lints to pass.

src/offset/local/mod.rs
esheppa: We could also use 

```rust
assert!(TypeId::of::<DateTime<Local>>() == dt.type_id());
```

Which I think would be preferable here as it is hard to interpret `PartialEq` and `Eq` on `Local`.

esheppa:
Thanks for this @klnusbaum - this is a good idea, and similar approaches have been discussed elsewhere. Fundamentally, it is painful to have to deal with `LocalResult` when using `FixedOffset`, `Utc` or any other implementer of `TimeZone` that doesn't ever change its internal offset/ Due to this, there could also be value in somewhat separating the timezone trait such that there can be a generic `and_fixed_timezone` or similar which works for all timezones which don't change their offset. Let me know your thoughts on this? It does also seem quite reasonable that we could special case `Utc` here and add this function

klnusbaum:
>  there could also be value in somewhat separating the timezone trait such that there can be a generic and_fixed_timezone or similar which works for all timezones which don't change their offset

Ok, I think I'm starting to understand things a little more. So the `Offset` for a time zone can change (like if there's Daylights Savings time)? Are you saying we could make a new, stricter Timezone trait, maybe call it "FixedTimeZone", that, while implementing everything `TimeZone` does, also makes stronger guarantees, namely that it's offset never changes (presumeably `FixedOffset` and `Utc` would implement `FixedTimeZone`)? I guess that could work. Though I'd want to think a lot about it. The whole `Offset` and `TimeZone` thing is already somewhat difficult to grok IMHO and this could add further confusion. That would also be a much bigger change.

Another option would be adding a function called `and_offset` which just creates a `DateTime` from a `NaiveDateTime` with the desired `Offset` (which means it would accept a `FixedOffset` or `Utc`). Though I'm not sure how much sense that makes (i.e. it would technically work, but does it make sense from a human standpoint).

esheppa:
Something else to note here is that `and_local_timezone` was originally added to ease API discovery. The logic being that many of these operations can already be done with methods on the `TimeZone` trait, but an IDE won't necessarily display these unless the trait has been imported (although these days I think rust-analyzer would). This is the same case for UTC where it is not easily discoverable. I think the best option here is that we just add this `and_utc_timezone` function as-is, and continue the discussion on the timezone trait over here: https://github.com/chronotope/chrono/pull/830

src/offset/local/mod.rs
klnusbaum: I'm not sure that would be good given that the purpose of the example (as far as I can tell) is to demonstrate that the created `DateTime` does indeed have the expected timezone.

If we don't want to add `Eq` and `PartialEq` to `Local`, then I think the next best thing is to create our own timezone using `FixedOffset` and use that. I'll put up a change trying that so we can see what it looks like.

src/naive/datetime/mod.rs
```diff
@@ -896,13 +897,27 @@ impl NaiveDateTime {
...
     }
 
+    /// Converts the `NaiveDateTime` into the timezone-aware `DateTime<Utc>`.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use chrono::{NaiveDate, NaiveTime, Utc};
+    /// let dt = NaiveDate::from_ymd_opt(2023, 1, 30).unwrap().and_hms_opt(19, 32, 33).unwrap().and_utc_timezone();
+    /// assert_eq!(dt.timezone(), Utc);
```
jtmoon79, line 916: This docstring code block is not closed, e.g. missing \`\`\`

src/naive/datetime/mod.rs
```diff
@@ -896,13 +897,27 @@ impl NaiveDateTime {
...
     ///
     /// ```
-    /// use chrono::{NaiveDate, Utc};
-    /// let dt = NaiveDate::from_ymd_opt(2015, 9, 5).unwrap().and_hms_opt(23, 56, 4).unwrap().and_local_timezone(Utc).unwrap();
-    /// assert_eq!(dt.timezone(), Utc);
+    /// use chrono::{NaiveDate, FixedOffset};
+    /// let hour = 3600;
+    /// let tz = FixedOffset::east_opt(5 * hour).unwrap();
+    /// let dt = NaiveDate::from_ymd_opt(2015, 9, 5).unwrap().and_hms_opt(23, 56, 4).unwrap().and_local_timezone(tz).unwrap();
+    /// assert_eq!(dt.timezone(), tz);
```
jtmoon79, line 904: This docstring code block is not closed, e.g. missing \`\`\`

src/naive/datetime/mod.rs
```diff
@@ -896,13 +897,27 @@ impl NaiveDateTime {
...
 
+    /// Converts the `NaiveDateTime` into the timezone-aware `DateTime<Utc>`.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use chrono::{NaiveDate, NaiveTime, Utc};
+    /// let dt = NaiveDate::from_ymd_opt(2023, 1, 30).unwrap().and_hms_opt(19, 32, 33).unwrap().and_utc_timezone();
+    /// assert_eq!(dt.timezone(), Utc);
+    pub fn and_utc_timezone(&self) -> DateTime<Utc> {
```
jtmoon79, line 917: Function needs attribute `#[must_use]`

src/naive/datetime/mod.rs
```diff
@@ -896,13 +897,27 @@ impl NaiveDateTime {
...
     /// ```
-    /// use chrono::{NaiveDate, Utc};
-    /// let dt = NaiveDate::from_ymd_opt(2015, 9, 5).unwrap().and_hms_opt(23, 56, 4).unwrap().and_local_timezone(Utc).unwrap();
-    /// assert_eq!(dt.timezone(), Utc);
+    /// use chrono::{NaiveDate, FixedOffset};
+    /// let hour = 3600;
+    /// let tz = FixedOffset::east_opt(5 * hour).unwrap();
+    /// let dt = NaiveDate::from_ymd_opt(2015, 9, 5).unwrap().and_hms_opt(23, 56, 4).unwrap().and_local_timezone(tz).unwrap();
+    /// assert_eq!(dt.timezone(), tz);
     pub fn and_local_timezone<Tz: TimeZone>(&self, tz: Tz) -> LocalResult<DateTime<Tz>> {
```
jtmoon79, line 905: Function needs attribute `#[must_use]`

jtmoon79 (changes requested):

jtmoon79 (approved):

pitdicker:
As a data point, this would make the set of functions to create a `DateTime` from a `NaiveDateTime` and `TimeZone` complete:

- Combine a `TimeZone` and a _local_ `NaiveDateTime`:
  - [`TimeZone::from_local_datetime`](https://docs.rs/chrono/latest/chrono/offset/trait.TimeZone.html#method.from_local_datetime)
  - [`DateTime<Tz: TimeZone>::from_local`](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.from_local)
  - [`NaiveDateTime::and_local_timezone`](https://docs.rs/chrono/latest/chrono/naive/struct.NaiveDateTime.html#method.and_local_timezone)
- Combine a `TimeZone` and a _UTC_ `NaiveDateTime`:
  - [`TimeZone::from_utc_datetime`](https://docs.rs/chrono/latest/chrono/offset/trait.TimeZone.html#method.from_utc_datetime)
  - [`DateTime<Tz: TimeZone>::from_utc`](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.from_utc)
  - `NaiveDateTime::and_utc_timezone` (missing)

pitdicker:
@klnusbaum :+1:  I am not really a reviewer yet, but it would be nice if the commits can be squashed, see https://github.com/chronotope/chrono/issues/898.

https://github.com/chronotope/chrono/issues/870 had the great suggestion that `assume_timezone` would be a better name than `and_local_timezone`. But I suppose that ship is sailed, and `and_utc_timezone` is the best name for the method in this PR.

klnusbaum:
squashed commits as suggested

pitdicker (approved):

klnusbaum:
@esheppa does this look good to you?

src/naive/datetime/mod.rs
```diff
@@ -938,14 +939,31 @@ impl NaiveDateTime {
...
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use chrono::{NaiveDate, NaiveTime, Utc};
+    /// let dt = NaiveDate::from_ymd_opt(2023, 1, 30).unwrap().and_hms_opt(19, 32, 33).unwrap().and_utc_timezone();
+    /// assert_eq!(dt.timezone(), Utc);
+    /// ```
+    #[must_use]
+    pub fn and_utc_timezone(&self) -> DateTime<Utc> {
```
djc, line 963: Naming: should this simply be called `and_utc()`? The `_timezone()` suffix seems superfluous to me. @pitdicker @esheppa @jtmoon79 thoughts?

klnusbaum: sure. seems fine.

djc (approved):

PR #953, opened 2023-01-31
Improve FixedOffset docs

klnusbaum:
The docs for FixedOffset are pointing to deprecated functions. Seems like we should update them to point to the non-deprecated variants.


djc (approved):

djc:
Thanks!

PR #959, opened 2023-02-10
archive derive of PartialEq for rkyv - 0.4.x

mkatychev:
This PR resolves `PartialEq` compilation errors for types that derive `rkyv::Archive`:

```
error[E0277]: can't compare `chrono::datetime::ArchivedDateTime<Utc>` with `DateTime<Utc>`
   --> common/src/transaction.rs:142:5
    |
142 |     rkyv::Archive,
    |     ^^^^^^^^^^^^^ no implementation for `chrono::datetime::ArchivedDateTime<Utc> == DateTime<Utc>`
    |
    = help: the trait `PartialEq<DateTime<Utc>>` is not implemented for `chrono::datetime::ArchivedDateTime<Utc>`
    = help: see issue #48214
    = note: this error originates in the derive macro `rkyv::Archive` (in Nightly builds, run with -Z macro-backtrace for more info)
```

This PR does not address https://github.com/chronotope/chrono/pull/762 since a new archived object is created, for example `ArchivedDateTime<Utc>` is derived from `DateTime<Utc>`, but does preserve feature inclusivity.

I don't think this change merits tests as the object created is not directly accessible for comparison as shown in the rkyv [basic example](https://github.com/rkyv/rkyv#example) but happy to discuss.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -75,7 +75,7 @@ jobs:
       - uses: taiki-e/install-action@cargo-hack
       - uses: Swatinem/rust-cache@v2
       - run: |
-          cargo hack check --feature-powerset --optional-deps serde,rkyv \
+          cargo hack check --feature-powerset --optional-deps serde \
             --skip __internal_bench,iana-time-zone,pure-rust-locales,libc,winapi \
             --all-targets
         # run using `bash` on all platforms for consistent
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ clock = ["std", "winapi", "iana-time-zone", "android-tzdata"]
 oldtime = []
 wasmbind = ["wasm-bindgen", "js-sys"]
 unstable-locales = ["pure-rust-locales"]
+rkyv-validation = ["rkyv/validation"]
 __internal_bench = []
 
 [dependencies]
@@ -33,7 +34,7 @@ num-traits = { version = "0.2", default-features = false }
 rustc-serialize = { version = "0.3.20", optional = true }
 serde = { version = "1.0.99", default-features = false, optional = true }
 pure-rust-locales = { version = "0.7", optional = true }
-rkyv = { version = "0.7", optional = true }
+rkyv = { version = "0.7.41", optional = true }
 arbitrary = { version = "1.0.0", features = ["derive"], optional = true }
 
 [target.'cfg(all(target_arch = "wasm32", not(any(target_os = "emscripten", target_os = "wasi"))))'.dependencies]
--- a/src/date.rs
+++ b/src/date.rs
@@ -479,7 +479,7 @@ impl<Tz: TimeZone> Eq for Date<Tz> {}
 
 impl<Tz: TimeZone> PartialOrd for Date<Tz> {
     fn partial_cmp(&self, other: &Date<Tz>) -> Option<Ordering> {
-        self.date.partial_cmp(&other.date)
+        Some(self.cmp(other))
     }
 }
 
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -51,6 +51,8 @@ mod tests;
 /// [`TimeZone`](./offset/trait.TimeZone.html) implementations.
 #[derive(Clone)]
 #[cfg_attr(feature = "rkyv", derive(Archive, Deserialize, Serialize))]
+#[cfg_attr(feature = "rkyv", archive(compare(PartialEq, PartialOrd)))]
+#[cfg_attr(feature = "rkyv-validation", archive(check_bytes))]
 pub struct DateTime<Tz: TimeZone> {
     datetime: NaiveDateTime,
     offset: Tz::Offset,
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -1506,6 +1506,31 @@ impl<Tz: TimeZone> fmt::Debug for DateTime<Tz> {
     }
 }
 
+// `fmt::Debug` is hand implemented for the `rkyv::Archive` variant of `DateTime` because
+// deriving a trait recursively does not propagate trait defined associated types with their own
+// constraints:
+// In our case `<<Tz as offset::TimeZone>::Offset as Archive>::Archived`
+// cannot be formatted using `{:?}` because it doesn't implement `Debug`.
+// See below for further discussion:
+// * https://github.com/rust-lang/rust/issues/26925
+// * https://github.com/rkyv/rkyv/issues/333
+// * https://github.com/dtolnay/syn/issues/370
+#[cfg(feature = "rkyv-validation")]
+impl<Tz: TimeZone> fmt::Debug for ArchivedDateTime<Tz>
+where
+    Tz: Archive,
+    <Tz as Archive>::Archived: fmt::Debug,
+    <<Tz as TimeZone>::Offset as Archive>::Archived: fmt::Debug,
+    <Tz as TimeZone>::Offset: fmt::Debug + Archive,
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.debug_struct("ArchivedDateTime")
+            .field("datetime", &self.datetime)
+            .field("offset", &self.offset)
+            .finish()
+    }
+}
+
 impl<Tz: TimeZone> fmt::Display for DateTime<Tz>
 where
     Tz::Offset: fmt::Display,
--- a/src/duration.rs
+++ b/src/duration.rs
@@ -54,8 +54,10 @@ macro_rules! try_opt {
 #[cfg_attr(feature = "rkyv", derive(Archive, Deserialize, Serialize))]
 #[cfg_attr(
     feature = "rkyv",
+    archive(compare(PartialEq, PartialOrd)),
     archive_attr(derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug, Hash))
 )]
+#[cfg_attr(feature = "rkyv-validation", archive(check_bytes))]
 pub struct Duration {
     secs: i64,
     nanos: i32, // Always 0 <= nanos < NANOS_PER_SEC
--- a/src/duration.rs
+++ b/src/duration.rs
@@ -797,4 +797,12 @@ mod tests {
             Err(OutOfRangeError(()))
         );
     }
+
+    #[test]
+    #[cfg(feature = "rkyv-validation")]
+    fn test_rkyv_validation() {
+        let duration = Duration::seconds(1);
+        let bytes = rkyv::to_bytes::<_, 16>(&duration).unwrap();
+        assert_eq!(rkyv::from_bytes::<Duration>(&bytes).unwrap(), duration);
+    }
 }
--- a/src/month.rs
+++ b/src/month.rs
@@ -33,8 +33,10 @@ use crate::OutOfRange;
 #[cfg_attr(feature = "rkyv", derive(Archive, Deserialize, Serialize))]
 #[cfg_attr(
     feature = "rkyv",
+    archive(compare(PartialEq)),
     archive_attr(derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug, Hash))
 )]
+#[cfg_attr(feature = "rkyv-validation", archive(check_bytes))]
 #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
 pub enum Month {
     /// January
--- a/src/month.rs
+++ b/src/month.rs
@@ -419,4 +419,12 @@ mod tests {
             from_str::<Month>(string).unwrap_err();
         }
     }
+
+    #[test]
+    #[cfg(feature = "rkyv-validation")]
+    fn test_rkyv_validation() {
+        let month = Month::January;
+        let bytes = rkyv::to_bytes::<_, 1>(&month).unwrap();
+        assert_eq!(rkyv::from_bytes::<Month>(&bytes).unwrap(), month);
+    }
 }
--- a/src/naive/isoweek.rs
+++ b/src/naive/isoweek.rs
@@ -153,6 +153,8 @@ impl fmt::Debug for IsoWeek {
 
 #[cfg(test)]
 mod tests {
+    #[cfg(feature = "rkyv-validation")]
+    use super::IsoWeek;
     use crate::naive::{internals, NaiveDate};
     use crate::Datelike;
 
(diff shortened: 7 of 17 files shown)


esheppa:
Thanks for this @mkatychev - sorry to be a pain but do you mind doing a `git rebase -i HEAD~1` or similar so that this gets a new commit hash and the CI runs (I think the change from `main` to `0.4.x` is currently preventing it running)

mkatychev:
@esheppa  After testing, I had to introduce a couple of changes to enable safe deserlialization. 
Using the [README example](https://github.com/rkyv/rkyv#example) from the rkyv repository as reference, it seems `bytecheck::CheckBytes` has to be derived as well if one wants to avoid using `unsafe`:
```rust
// You can use the safe API for fast zero-copy deserialization
let archived = rkyv::check_archived_root::<Test>(&bytes[..]).unwrap();
assert_eq!(archived, &value);

// Or you can use the unsafe API for maximum performance
let archived = unsafe { rkyv::archived_root::<Test>(&bytes[..]) };
assert_eq!(archived, &value);
```

This seems like a core part of the way rkyv should work so I added a PR upstream to re-export `bytecheck`: https://github.com/rkyv/rkyv/pull/340.

In the meantime however, it made the most sense to add `bytecheck` as an optional dependency to chrono while the PR above is still under review.

mkatychev:
https://github.com/chronotope/chrono/blob/d37b0be87eaf009ff582c69ab35533015572da88/.github/workflows/test.yml#L52-L56

The `dep:feature` is supported in rust 1.60 and upwards. Is there any guidelines for how far behind the msrv should be?

djc:
For the 0.4.x branch, we're sticking with 1.38 for now, so the two optional deps would be the way forward for now. Thanks!

Cargo.toml
```diff
@@ -36,6 +37,7 @@ serde = { version = "1.0.99", default-features = false, optional = true }
 pure-rust-locales = { version = "0.5.2", optional = true }
 criterion = { version = "0.4.0", optional = true }
 rkyv = {version = "0.7", optional = true}
+bytecheck = { version = "~0.6.8", optional = true, default-features = false }
```
djc, line 40: If this is intended to only allow 0.6.x versions, we can leave off the `~` as the `^` operator can be left implicit in Cargo manifests, and it appears that `~` is equivalent to `^` for the major-0 versions.

jtmoon79 (commented):
I visually reviewed. LGTM. However, I know very little about serde internals so I won't be able to spot serde-specific bugs.

Does this change need new test cases?
For example, I see the `rkyv` project implements a some test macros like `pub fn test_archive<T>(value: &T)` https://github.com/rkyv/rkyv/blob/master/rkyv_test/src/util.rs#L4
and test cases using those macros like `fn archive_containers()` https://github.com/rkyv/rkyv/blob/master/rkyv_test/src/test_alloc.rs#L36

Should the definitions here have test cases that exercise the `rkyv` features?

mkatychev:
> I visually reviewed. LGTM. However, I know very little about serde internals so I won't be able to spot serde-specific bugs.
> 
> Does this change need new test cases? For example, I see the `rkyv` project implements a some test macros like `pub fn test_archive<T>(value: &T)` https://github.com/rkyv/rkyv/blob/master/rkyv_test/src/util.rs#L4 and test cases using those macros like `fn archive_containers()` https://github.com/rkyv/rkyv/blob/master/rkyv_test/src/test_alloc.rs#L36
> 
> Should the definitions here have test cases that exercise the `rkyv` features?

I did a bit of upstream work in `bytecheck` and `rkyv` that I have to reflect in this PR. As such I've moved this to draft for when I get around to backporting the changes.

mkatychev:
@jtmoon79  I'm going to wait for `rkyv` to release `0.8` as that allows bytecheck errors to implement `Send + Sync` before marking this ready for review.

Roms1383:
would you mind adding `PartialOrd` too ?
inner types support it (tried locally then I realized a PR was already opened).

mkatychev:
@jtmoon79 [`cargo hack check --feature-powerset`](https://github.com/chronotope/chrono/actions/runs/5424918127/jobs/9864967866?pr=959#step:6:411) should already be covering `rkyv-validation` feature flag

pitdicker:
@mkatychev I have yet to go and read up about rkyv. But can I ask for your opinion/review of https://github.com/chronotope/chrono/pull/762?

mkatychev:
@pitdicker  I'll post my thoughts in this PR, The current approach :
A lot of the primitives 

[two step serialization](https://rkyv.org/architecture/archive.html)

pitdicker:
Thank you for your work!

> @pitdicker what else is needed to move this review forward?

Me needing to get a little experience with rkyv, because I don't understand it yet :smile:. Or a review from @djc.

Cargo.toml
```diff
@@ -26,6 +26,7 @@ clock = ["std", "winapi", "iana-time-zone", "android-tzdata"]
 oldtime = []
 wasmbind = ["wasm-bindgen", "js-sys"]
 unstable-locales = ["pure-rust-locales", "alloc"]
+rkyv-validation = ["rkyv/validation"]
```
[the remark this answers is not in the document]
pitdicker: An extra feature may double our CI times with the `cargo hack check --feature-powerset`, but I can look into that later if necessary.

pitdicker (changes requested):
What I am missing with the `rkyv` feature is that we don't have any tests. That is not a fault of this PR, but can you please add some?

I don't think it needs to be more than a check that all types round-trip with `rkyv::{to_bytes, from_bytes}`.
And maybe include a check that the derived types implement `PartialOrd`, `PartialEq` and `Debug` similar to https://github.com/chronotope/chrono/blob/v0.4.31/src/datetime/tests.rs#L1132.
That should also make coverage happy.

And can you squash the commits?

Cargo.toml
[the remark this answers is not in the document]
mkatychev: Ideally, `cargo hack check` should treat `rkyv/validation` as a superset of `rkyv`. Using `rkyv-validation` en lieu of `rkyv` should increase coverage without adding another row/column to the matrix.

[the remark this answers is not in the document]
pitdicker: That makes me think... Should the `rkyv-validation` feature require the `rkyv` feature?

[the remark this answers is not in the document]
mkatychev: Correct me if I'm wrong but that is already the case: https://doc.rust-lang.org/cargo/reference/features.html#dependency-features

> Features of dependencies can also be enabled in the `[features]` table. `The syntax is "package-name/feature-name"`.

`package-name/feature-name` already implies the requirement of `package-name`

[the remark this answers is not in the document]
pitdicker: You are right :+1:.

Can you add the feature to the list of crate features in `lib.rs` and `README.md`?

Cargo.toml
```diff
@@ -36,6 +37,7 @@ pure-rust-locales = { version = "0.7", optional = true }
 rkyv = { version = "0.7", optional = true }
```
pitdicker, line 37: Should this change the minimum version of rkyv to 0.7.41, the first version with `#[archive(check_bytes)]`?

Cargo.toml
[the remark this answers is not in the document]
mkatychev: This is great feedback. I'll incorporate the comments and add tests later this week.

pitdicker:
> I'll incorporate the comments and add tests later this week.

@mkatychev No hurry, just a reminder.

mkatychev:
@pitdicker apologies for delay, I will start work on this tomorrow.

src/datetime/mod.rs
```diff
@@ -1355,6 +1360,24 @@ impl<Tz: TimeZone> fmt::Debug for DateTime<Tz> {
     }
 }
 
+// This boilerplate is a shortcoming of the `Derive` macro.
```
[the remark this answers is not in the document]
mkatychev: Resolved by https://github.com/chronotope/chrono/pull/959/files#r1328990897.
 `ArchivedDate` should not be handled due to `Date`'s deprecation.

src/offset/utc.rs
```diff
@@ -40,9 +40,14 @@ use crate::{Date, DateTime};
 /// assert_eq!(Utc.timestamp_opt(61, 0).unwrap(), dt);
 /// assert_eq!(Utc.with_ymd_and_hms(1970, 1, 1, 0, 1, 1).unwrap(), dt);
 /// ```
-#[derive(Copy, Clone, PartialEq, Eq, Hash)]
-#[cfg_attr(feature = "rkyv", derive(Archive, Deserialize, Serialize))]
-#[cfg_attr(feature = "rkyv", archive_attr(derive(Clone, Copy, PartialEq, Eq, Debug, Hash)))]
+#[derive(Clone, Copy, Eq, Hash, PartialEq)]
```
mkatychev, line 43: I sorted the derives alphabetically to better compre against the `archive_attr` equivalents.

src/datetime/mod.rs
```diff
@@ -1355,6 +1360,24 @@ impl<Tz: TimeZone> fmt::Debug for DateTime<Tz> {
...
+    Tz: Archive,
+    <Tz as Archive>::Archived: fmt::Debug,
+    <<Tz as TimeZone>::Offset as Archive>::Archived: fmt::Debug,
+    <Tz as TimeZone>::Offset: fmt::Debug + Archive,
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.debug_struct("Date")
+            .field("datetime", &self.datetime)
+            .field("offset", &self.offset)
+            .finish()
```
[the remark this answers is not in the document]
mkatychev: This is meant to simply satisfy a struct that holds `DateTime<Tz>` and itself attempts to derive `Debug` as well as `Archive`, I suppose in that case it is better to create a newtype which is what I have done in the meantime and hand implement the `fmt::Debug` there:
`struct RkyvDateTime(DateTime<Tz>)`

Given the already large surface area of this PR, I think it is better if this manual impl is saved for another time.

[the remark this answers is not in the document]
mkatychev: `ArcivedDateTime` is the struct created by deriving `Archive` on an object: 
https://docs.rs/rkyv/latest/rkyv/trait.Archive.html#associatedtype.Archived
In this case:
```rust
impl<Tz: TimeZone> Archive for DateTime<Tz> {
        type Archived = ArchivedDateTime;
        // ...
}
```

mkatychev:
@pitdicker  this should be ready for review again

mkatychev:
Tagging related issues for `ArchivedDateTime`:
https://github.com/rkyv/rkyv/issues/333
https://github.com/rust-lang/rust/issues/26925

Cargo.toml
```diff
@@ -26,14 +26,15 @@ clock = ["std", "winapi", "iana-time-zone", "android-tzdata"]
...
+rkyv-validation = ["rkyv/validation"]
 __internal_bench = []
 
 [dependencies]
 num-traits = { version = "0.2", default-features = false }
 rustc-serialize = { version = "0.3.20", optional = true }
 serde = { version = "1.0.99", default-features = false, optional = true }
 pure-rust-locales = { version = "0.7", optional = true }
-rkyv = { version = "0.7", optional = true }
+rkyv = { version = "0.7.42", optional = true }
```
pitdicker, line 37:
```suggestion
rkyv = { version = "0.7.41", optional = true }
```

I hope that by setting `0.7.41` as the minimum version and with https://github.com/rkyv/bytecheck/pull/33 that we can get things working with our MSRV.

src/duration.rs
```diff
@@ -50,12 +50,14 @@ macro_rules! try_opt {
...
 ///
 /// This also allows for the negative duration; see individual methods for details.
-#[derive(Clone, Copy, Default, PartialEq, Eq, PartialOrd, Ord, Debug, Hash)]
-#[cfg_attr(feature = "rkyv", derive(Archive, Deserialize, Serialize))]
+#[derive(Clone, Copy, Debug, Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
 #[cfg_attr(
     feature = "rkyv",
-    archive_attr(derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug, Hash))
+    derive(Archive, Deserialize, Serialize),
+    archive(compare(PartialEq)),
```
pitdicker, line 57: Also `PartialOrd`?

src/month.rs
```diff
@@ -221,7 +222,7 @@ impl num_traits::FromPrimitive for Month {
     }
 }
 
-/// A duration in calendar months
+/// A duration calendar months
```
pitdicker, line 225: Please revert this line.

src/naive/isoweek.rs
```diff
@@ -151,6 +153,7 @@ impl fmt::Debug for IsoWeek {
 
 #[cfg(test)]
 mod tests {
+    use super::IsoWeek;
```
pitdicker, line 156:
```suggestion
    #[cfg(feature = "rkyv-validation")]
    use super::IsoWeek;
```

To prevent the unused import warning.

src/date.rs
```diff
@@ -56,7 +56,8 @@ use crate::{Datelike, Weekday};
 ///   even though the raw calculation between `NaiveDate` and `Duration` may not.
 #[deprecated(since = "0.4.23", note = "Use `NaiveDate` or `DateTime<Tz>` instead")]
 #[derive(Clone)]
-#[cfg_attr(feature = "rkyv", derive(Archive, Deserialize, Serialize))]
+#[cfg_attr(feature = "rkyv", derive(Archive, Deserialize, Serialize), archive(compare(PartialEq)))]
```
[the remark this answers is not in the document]
pitdicker: Please revert the changes to this type.

.github/workflows/test.yml
```diff
@@ -75,7 +75,7 @@ jobs:
       - uses: taiki-e/install-action@cargo-hack
       - uses: Swatinem/rust-cache@v2
       - run: |
-          cargo hack check --feature-powerset --optional-deps serde,rkyv \
+          cargo hack check --feature-powerset --optional-deps serde,rkyv-validation \
```
pitdicker, line 78:
```suggestion
          cargo hack check --feature-powerset --optional-deps serde \
```

This fails on CI.

pitdicker (changes requested):
Looking good :smile:. A few comments, and we have trouble with our MSRV.

Can you squash the commits?

Cargo.toml
mkatychev: I'm curious why `0.7.42` was not able to be found 🤔 🤔

.github/workflows/test.yml
mkatychev: I reverted this to `,rykv` instead; seeing if that passes.

pitdicker: As I understand it we only need to put things `--optional-deps` that are not named as features. I tested it locally and `cargo hack` does the combinations we want without `,rkyv`.

pitdicker: And with `,rkyv` it tests double the number of combinations...

pitdicker (approved):

mkatychev:
Thanks for pushing this through @pitdicker !

djc:
This needs to be rebased and split into clean commits so I can review it. Right now there's too much going on when reviewing all the changes at once (notably reordering of derives while also adding/removing derives) to make this efficient to review.

mkatychev:
@djc  I'll rebase the commits later this week, I'll just skip the reordering altogether to make reviewing easier

mkatychev:
@djc  This should be ready for review

djc:
Thanks!

PR #965, opened 2023-02-14
Add `with_ymd` to the `Datelike` trait

### LINKED ISSUE #962: Change any combination of y, m or d at the same time
Imagine I have a `Date` for this date: `2023-2-1` (Y-M-D)
And I want to change the day to 31 and month to January.

AFAIK, I can use the `DateLike` methods, like `with_month` and `with_day`.
And also if I'm not mistaken, there is no `with_md`, `with_ymd` or any other combination.

The problem is that the order of `with`s will be important in this case.

If first I change month, then day, it will be OK.
If first I change day, then month, it will not be OK.

In this example, I know the order, easy. But in most cases those numbers are dynamic.

The same can happen to `TimeLike` methods.

Code-wise, it will be like the code below:

```
use chrono::*;

fn main() {
    let date = NaiveDate::from_ymd_opt(2023, 2, 1).unwrap();
    println!("{}", date.with_month(1).unwrap().with_day(31).unwrap()); // prints `2000-01-31`
    println!("{}", date.with_day(31).expect("wrong day").with_month(1).unwrap()); // panics with `wrong day`
}
```

omid:
fixes #962

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,9 @@ readme: README.md
 
 .PHONY: test
 test:
-	CHANNEL=$(CHANNEL) ./ci/travis.sh
+	@cargo test --all-features --color=always -- --color=always
+	@cargo test --lib --all-features --color=always -- --color=always
+	@cargo test --doc --all-features --color=always -- --color=always
 
 .PHONY: doc
 doc: authors readme
--- a/src/date.rs
+++ b/src/date.rs
@@ -444,7 +444,12 @@ impl<Tz: TimeZone> Datelike for Date<Tz> {
     }
 
     #[inline]
-    fn with_ymd(&self, year: Option<i32>, month: Option<u32>, day: Option<u32>) -> Option<Date<Tz>> {
+    fn with_ymd(
+        &self,
+        year: Option<i32>,
+        month: Option<u32>,
+        day: Option<u32>,
+    ) -> Option<Date<Tz>> {
         map_local(self, |date| date.with_ymd(year, month, day))
     }
 }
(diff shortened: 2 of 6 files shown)


esheppa:
As a thought here, this only saves a few characters compared with just creating a new date and using some `Datelike` methods to replicate the relevant parts of the old date.


```rust
let new_date = date.with_ymd(None, Some(1), Some(5)).unwrap();
```

```rust
let new_date = NaiveDate::from_ymd_opt(date.year(), 1, 5).unwrap();
```

omid:
That's right. So I guess it's just a matter of the API.

It will be easier for types which has time, though, like `NaiveDateTime`.

esheppa:
Good point, in those cases it would be:

```rust
let new_datetime = my_datetime.with_ymd(None, Some(1), Some(5)).unwrap();
```

```rust
let new_datetime = NaiveDate::from_ymd_opt(my_datetime.year(), 1, 5).unwrap().and_time(my_datetime.time());
```

In the case of `DateTime<Tz>` there is a significant saving in verbosity, but there are also issues in that, `with_ymd` may return none even for valid month and day combinations due to the time itself not existing on that day.

omid:
> but there are also issues in that, `with_ymd` may return none even for valid month and day combinations due to the time itself not existing on that day.

Yes. Is this a problem?

And I remembered a case for https://github.com/chronotope/chrono/issues/962#issuecomment-1429390614 when we set clocks forward by one hour in springs.

djc:
With all this, I'm inclined to recommend that you just rely on existing APIs, as @esheppa suggested.

omid:
Yep, I agree. It's complicated.
Another solution can be to remove all `with_*` methods, to avoid confusion and potentially create buggy code.

omid:
Sure, no problem 👍🙏

Issue #961, opened 2023-02-13
%Y-%W-%w parsing for 2019


Bidek56:
`let date = NaiveDate::parse_from_str("2020-01-0", "%Y-%W-%w")?;`
Returns: 2020-01-12
But:
`let date = NaiveDate::parse_from_str("2019-01-0", "%Y-%W-%w")?;`
Returns: Error: ParseError(Impossible)

Any reasons? Thx

djc:
Okay, there seems to be a bug in the logic for [`verify_ordinal`](https://github.com/chronotope/chrono/blob/main/src/format/parsed.rs#L388) that you've hit. My guess is the logic for the `week_from_sun` verification was copied to `week_from_mon`, but the latter version is incorrect in this case (potentially only for the particular case of years starting on a Tuesday?) with an off-by-one error.

I can't quite work out the correct logic right now, please submit a PR if you can (and add the failing case as a test case).

esheppa:
I think the fix is the same as from https://github.com/chronotope/chrono/pull/921/files - I'll push a commit to that PR

FObersteiner:
a related question: why is `%w` needed here? The docs don't say so explicitly and for example `%Y-%m` also works without needing a day, it just defaults to the first day of the month.

I could make a separate issue if this is too much off-topic here ;-)

djc:
%W identifies a week, but not a date, so you can't parse a date out of it.

FObersteiner:
@djc that was my first thought as well. But if you take `%m` for example, similarly  that identifies a month and not a date. Nevertheless, you can parse 2019-01 with `%Y-%m` to a full date. I don't see why this shouldn't work *technically* with `%W`. Just default to the first day of the week. Or, if there is a clear reason why not to do so, be more explicit in the docs.

djc:
I guess if someone wants to contribute code to allow leaving off `%w` I'd be open to merging that.

pitdicker:
Fixed in https://github.com/chronotope/chrono/pull/966.

Allowing default/missing values when parsing is a duplicate of https://github.com/chronotope/chrono/issues/528 and a couple of other issues.

PR #966, opened 2023-02-14
Fixes to parsing and calculation of week numbers

esheppa:
Continuation of https://github.com/chronotope/chrono/pull/921

also fixes a bug reported here: https://github.com/chronotope/chrono/issues/961

### DIFF
--- a/src/format/mod.rs
+++ b/src/format/mod.rs
@@ -534,10 +534,10 @@ fn format_inner<'a>(
             use self::Numeric::*;
 
             let week_from_sun = |d: &NaiveDate| {
-                (d.ordinal() as i32 - d.weekday().num_days_from_sunday() as i32 + 7) / 7
+                (d.ordinal() as i32 - d.weekday().num_days_from_sunday() as i32 + 6) / 7
             };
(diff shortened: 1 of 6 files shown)


djc:
@esheppa thanks for continuing the old PR here! IMO it would be nice to clean up the commit history such that every commit passes tests (squash a bunch of the fixes into the original fixing commit, squash formatting fixes and `cfg` updates into the first commit), and then this should be good to merge.

One question: should we add a `pub(crate)` method to `NaiveDate` for `weeks_from(Weekday)` so that we have one implementation instead of multiple?

esheppa:
> One question: should we add a pub(crate) method to NaiveDate for weeks_from(Weekday) so that we have one implementation instead of multiple? (If so, that could/should probably be a separate commit.)

This is a great idea, I've added in eb75271 but I'd like to add some more tests and squash them into that commit prior to this being merged

djc:
Looks great!

djc:
@esheppa are you able to continue on this or should I take over?

esheppa:
Will get those tests added now

djc (approved):

PR #603, opened 2021-10-14
Fix DateTime::parse_from_str panicking

Vagelis-Prokopiou:
### Thanks for contributing to chrono!

- [ ] Have you added yourself and the change to the [changelog]? (Don't worry
      about adding the PR number)
- [x] If this pull request fixes a bug, does it add a test that verifies that
      we can't reintroduce it?

[changelog]: ../CHANGELOG.md

PR #967, opened 2023-02-14
Fix building chrono-fuzz

LingMan:
- Fixes building chrono-fuzz which was broken in https://github.com/chronotope/chrono/pull/806
- (Sneaks in a minor Dependabot nit that doesn't deserve its own CI run)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,10 +1,10 @@
 version: 2
 updates:
-  - package-ecosystem: cargo
+  - package-ecosystem: "cargo"
     directory: "/"
     schedule:
       interval: "weekly"
-  - package-ecosystem: cargo
+  - package-ecosystem: "cargo"
     directory: "/fuzz/"
     schedule:
       interval: "weekly"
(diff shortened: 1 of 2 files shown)


djc:
I'd like to have most of these changes (except the first one, which I think wouldn't commit) on 0.4.x (from where we'll merge them to main), would you be able to retarget the PR (and rebase the changes)?

LingMan:
Sure, give me a bit to figure it out.

djc:
I recommend `git rebase -i main`, then killing all except the last three non-commented lines.

djc:
Since most development currently happens on 0.4.x, I would still appreciate such a PR for all the applicable changes.

LingMan:
Here you go: https://github.com/chronotope/chrono/pull/968.
I've dropped those commits from this PR so it now mostly fixes the chrono-fuzz build.

jtmoon79 (approved):
LGTM

Also see fuzzing-related Issues in Bug #1010 , PR #603 .

djc:
Thanks!

PR #969, opened 2023-02-15
Remove cargo ignoration `RUSTSEC-2020-0071` on `time`

temeddix:
Now that dependency on `time` is complely removed in version 0.5 with [this commit](https://github.com/chronotope/chrono/commit/4c351b1e720a30a6e32068fe386293b36d0e86fd#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542L21), we can remove this ignoration.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -4,7 +4,6 @@ copyleft = "deny"
 
 [advisories]
 ignore = [
-    "RUSTSEC-2020-0071", # time 0.1, doesn't affect the API we use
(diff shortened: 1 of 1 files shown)


djc:
Thanks! If you'd like, I think we can lose the ignore for rustc_serialize too?

temeddix:
Is that dependency removed as well?

djc:
Yup!

PR #975, opened 2023-02-26
Add parser for tzdata on android

### LINKED ISSUE #922: Local timezones broken on Android
Unfortunately it seems like #756 did not fix timezone determination on Android - UTC is being returned instead.

Part of the problem might be the leading space in the filename:

https://github.com/chronotope/chrono/pull/756/files#diff-568a8cdf0e0b7a4922b13d27771df35ac8e561ac9f25c5229263120edc171486R100

A bigger problem though is that individual files don't appear to be available - `/system/usr/share/zoneinfo` appears to have all data packed into a binary tzdata file, so Chrono will presumably need to be able to parse it. Some Googling reveals the format: https://android.googlesource.com/platform/libcore/+/jb-mr2-release/luni/src/main/java/libcore/util/ZoneInfoDB.java#38, https://go.dev/src/time/zoneinfo_android.go

RumovZ:
Closes #922.

The only way to add a meaningful unit test seems to require a tzdata file, but I don't want to check in a 500 KB binary (67 KB zipped). What do you think?
This commit already had some alpha testing as part of the @ankidroid project.

### DIFF
--- a/src/offset/local/tz_info/mod.rs
+++ b/src/offset/local/tz_info/mod.rs
@@ -24,6 +24,8 @@ pub(crate) enum Error {
     LocalTimeType(&'static str),
     /// Invalid slice for integer conversion
     InvalidSlice(&'static str),
+    /// Invalid tzdata file
+    InvalidTzdataFile(&'static str),
     /// Invalid Tzif file
     InvalidTzFile(&'static str),
     /// Invalid TZ string
@@ -59,6 +61,7 @@ impl fmt::Display for Error {
             LocalTimeType(error) => write!(f, "invalid local time type: {}", error),
             InvalidSlice(error) => error.fmt(f),
             InvalidTzString(error) => write!(f, "invalid TZ string: {}", error),
+            InvalidTzdataFile(error) => error.fmt(f),
             InvalidTzFile(error) => error.fmt(f),
             Io(error) => error.fmt(f),
             OutOfRange(error) => error.fmt(f),
--- a/src/offset/local/tz_info/timezone.rs
+++ b/src/offset/local/tz_info/timezone.rs
@@ -52,6 +52,11 @@ impl TimeZone {
             return Self::from_file(&mut file);
         }
 
+        #[cfg(target_os = "android")]
+        if let Ok(bytes) = super::super::tzdata::find_tz_data(tz_string) {
+            return Self::from_tz_data(&bytes);
+        }
+
         // TZ string extensions are not allowed
         let tz_string = tz_string.trim_matches(|c: char| c.is_ascii_whitespace());
         let rule = TransitionRule::from_tz_string(tz_string.as_bytes(), false)?;
--- a/src/offset/local/tzdata.rs
+++ b/src/offset/local/tzdata.rs
@@ -1,0 +1,131 @@
+//! This module contains functions to locate and parse the tzdata file, which only
+//! exists on Android and is a database of all tz files.
+//!
+//! Logic was mainly ported from https://android.googlesource.com/platform/libcore/+/jb-mr2-release/luni/src/main/java/libcore/util/ZoneInfoDB.java
+
+use core::{cmp::Ordering, convert::TryInto};
+use std::{
+    fs::File,
+    io::{self, ErrorKind, Read, Seek, SeekFrom},
+};
+
+use super::tz_info::Error;
+
+// The database uses 32-bit (4 byte) integers.
+const TZ_INT_SIZE: usize = 4;
+// The first 12 bytes contain a special version string.
+const MAGIC_SIZE: usize = 12;
+const HEADER_SIZE: usize = MAGIC_SIZE + 3 * TZ_INT_SIZE;
+// The database reserves 40 bytes for each id.
+const TZ_NAME_SIZE: usize = 40;
+const INDEX_ENTRY_SIZE: usize = TZ_NAME_SIZE + 3 * TZ_INT_SIZE;
+
+#[derive(Debug, Clone, Copy)]
+struct Header {
+    index_offset: usize,
+    data_offset: usize,
+    _zonetab_offset: usize,
+}
+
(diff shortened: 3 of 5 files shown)


src/offset/local/tz_info/mod.rs
```diff
@@ -24,6 +24,8 @@ pub(crate) enum Error {
     LocalTimeType(&'static str),
     /// Invalid slice for integer conversion
     InvalidSlice(&'static str),
+    /// Invalid tzdata file
+    InvalidTzdataFile(&'static str),
```
djc, line 28: Style nit: I think we'd want to capitalize `Data` here?

src/offset/local/tz_info/timezone.rs
```diff
@@ -52,6 +52,11 @@ impl TimeZone {
             return Self::from_file(&mut file);
         }
 
+        #[cfg(target_os = "android")]
+        if let Ok(bytes) = super::super::tzdata::find_tz_data(tz_string) {
```
djc, line 56: Style: not very fond of `super::super`, maybe use the path from `crate::` instead?

src/offset/local/tzdata.rs
```diff
@@ -0,0 +1,131 @@
...
+#[derive(Debug, Clone, Copy)]
+struct IndexEntry<'a> {
+    _name: &'a [u8],
+    offset: usize,
+    length: usize,
+    _raw_utc_offset: usize,
+}
+
+pub(super) fn find_tz_data(tz_name: &str) -> Result<Vec<u8>, Error> {
+    let mut file = find_file()?;
```
djc, line 42: Since this is the only caller for `find_file()`, I think we could just inline it?

src/offset/local/tzdata.rs
```diff
@@ -0,0 +1,131 @@
...
+    find_tz_data_in_file(&mut file, tz_name)
+}
+
+fn find_file() -> Result<File, Error> {
+    for (env_var, dir) in
+        [("ANDROID_DATA", "/misc/zoneinfo/"), ("ANDROID_ROOT", "/usr/share/zoneinfo/")]
+    {
+        if let Ok(env_value) = std::env::var(env_var) {
+            if let Ok(file) = File::open(env_value + dir + "tzdata") {
```
djc, line 51: Not a great fan of concatenating strings this way, either. Can we just do a `format!()` call instead?

src/offset/local/tzdata.rs
```diff
@@ -0,0 +1,131 @@
...
+                return Ok(file);
+            }
+        }
+    }
+    Err(Error::Io(io::Error::from(io::ErrorKind::NotFound)))
+}
+
+fn find_tz_data_in_file(file: &mut File, tz_name: &str) -> Result<Vec<u8>, Error> {
```
djc, line 59: I would probably also inline this?

src/offset/local/tzdata.rs
```diff
@@ -0,0 +1,131 @@
...
+            _name: bytes[..TZ_NAME_SIZE].splitn(2, |&b| b == 0u8).next().unwrap(),
+            offset: parse_tz_int(bytes, TZ_NAME_SIZE) as usize,
+            length: parse_tz_int(bytes, TZ_NAME_SIZE + TZ_INT_SIZE) as usize,
+            _raw_utc_offset: parse_tz_int(bytes, TZ_NAME_SIZE + 2 * TZ_INT_SIZE) as usize,
+        }
+    }
+}
+
+/// Panics if slice does not contain [TZ_INT_SIZE] bytes beginning at start.
+fn parse_tz_int(slice: &[u8], start: usize) -> u32 {
```
djc, line 129: I'd make this a newtype with a constructor, we can use the type for the `Header` fields.

src/offset/local/tzdata.rs
```diff
@@ -0,0 +1,131 @@
...
+use super::tz_info::Error;
+
+// The database uses 32-bit (4 byte) integers.
+const TZ_INT_SIZE: usize = 4;
+// The first 12 bytes contain a special version string.
+const MAGIC_SIZE: usize = 12;
+const HEADER_SIZE: usize = MAGIC_SIZE + 3 * TZ_INT_SIZE;
+// The database reserves 40 bytes for each id.
+const TZ_NAME_SIZE: usize = 40;
+const INDEX_ENTRY_SIZE: usize = TZ_NAME_SIZE + 3 * TZ_INT_SIZE;
```
djc, line 21: Would prefer to have the constants at the bottom of the module.

djc (commented):
Thanks for working on this!

Here are some suggestions. I wonder if this should maybe remain a separate crate that chrono will depend on only on Android, would that make sense? That would give you more freedom on how to test and avoids the chrono maintainers having to take responsibility for a piece of core infrastructure that is hard to actually exercise in CI (or maybe it isn't?)

Alternatively, for testing we could subset the original file to have like one or two timezones only?

RumovZ:
Thanks for the feedback! I'll come back to it in the upcoming week.
I'm not an Android developer, but will think about the crate idea.

djc:
> To clarify the testing situation, the parsing can be tested easily in CI/on the local platform if a test tzdata file is bundled. Testing on a device/simulator would be a fair bit more complicated - you need to install the Android NDK, cross-compile, then spin up a sim or device to run the code, and it's unlikely to be worth the effort just to confirm the file exists in a standard location.

Sure, the existence of the file and its path is one thing. But since this appears to more or less reverse engineer Android's format, it's not clear to me that we can depend on the stability of the format in the longer term. Unless I'm missing something here?

dae:
There are no guarantees, but as far as I'm aware it's been the same format for a long time. Golang rolled their own version in 2016, and I don't see any changes to the format since then in their tree: https://github.com/golang/go/commits/31881f87873b84709a49ca17195bbe5b3f683acf/src/time/zoneinfo_android.go

PR #978, opened 2023-03-04
Extract timezone info from tzdata file on Android

### LINKED ISSUE #922: Local timezones broken on Android

RumovZ:
Replacement for #975. As per @djc's suggestion, the tzdata parser now resides in its own crate.

Closes #922.

### DIFF
--- a/src/offset/local/tz_info/timezone.rs
+++ b/src/offset/local/tz_info/timezone.rs
@@ -52,6 +52,14 @@ impl TimeZone {
             return Self::from_file(&mut file);
         }
 
+        // attributes are not allowed on if blocks in Rust 1.38
+        #[cfg(target_os = "android")]
+        {
+            if let Ok(bytes) = android_tzdata::find_tz_data(tz_string) {
+                return Self::from_tz_data(&bytes);
+            }
+        }
+
         // TZ string extensions are not allowed
         let tz_string = tz_string.trim_matches(|c: char| c.is_ascii_whitespace());
         let rule = TransitionRule::from_tz_string(tz_string.as_bytes(), false)?;
--- a/src/offset/local/tz_info/timezone.rs
+++ b/src/offset/local/tz_info/timezone.rs
@@ -43,6 +43,14 @@ impl TimeZone {
             return Self::from_tz_data(&fs::read("/etc/localtime")?);
         }
 
+        // attributes are not allowed on if blocks in Rust 1.38
+        #[cfg(target_os = "android")]
+        {
+            if let Ok(bytes) = android_tzdata::find_tz_data(tz_string) {
+                return Self::from_tz_data(&bytes);
+            }
+        }
+
         let mut chars = tz_string.chars();
         if chars.next() == Some(':') {
             return Self::from_file(&mut find_tz_file(chars.as_str())?);
@@ -50,14 +58,6 @@ impl TimeZone {
 
         if let Ok(mut file) = find_tz_file(tz_string) {
             return Self::from_file(&mut file);
-        }
-
-        // attributes are not allowed on if blocks in Rust 1.38
-        #[cfg(target_os = "android")]
-        {
-            if let Ok(bytes) = android_tzdata::find_tz_data(tz_string) {
-                return Self::from_tz_data(&bytes);
-            }
         }
 
         // TZ string extensions are not allowed
--- a/src/offset/local/unix.rs
+++ b/src/offset/local/unix.rs
@@ -68,19 +68,18 @@ struct Cache {
     last_checked: SystemTime,
 }
 
-#[cfg(target_os = "android")]
-const TZDB_LOCATION: &str = " /system/usr/share/zoneinfo";
-
 #[cfg(target_os = "aix")]
 const TZDB_LOCATION: &str = "/usr/share/lib/zoneinfo";
 
-#[allow(dead_code)] // keeps the cfg simpler
 #[cfg(not(any(target_os = "android", target_os = "aix")))]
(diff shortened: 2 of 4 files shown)


dae:
Might be worth closing and then reopening this PR to rerun the tests to see if it was a flake?

dae:
The test passes for me when I run it in [another repo](https://github.com/ankitects/chrono/actions/runs/4357483991/jobs/7616846252). I suspect the rust-cache action is the culprit; try replacing the line in .github/workflows/test.yml with

  with:
    prefix-key: "v1-rust"

RumovZ:
Perfect, thanks for the help. @dae!

src/offset/local/tz_info/timezone.rs
```diff
@@ -8,6 +8,9 @@ use std::{cmp::Ordering, fmt, str};
 use super::rule::{AlternateTime, TransitionRule};
 use super::{parser, Error, DAYS_PER_WEEK, SECONDS_PER_DAY};
 
+#[cfg(target_os = "android")]
```
djc, line 11: I don't think we need this?

djc (approved):

src/offset/local/unix.rs
```diff
@@ -68,19 +68,18 @@ struct Cache {
     last_checked: SystemTime,
 }
 
-#[cfg(target_os = "android")]
```
esheppa, line 71: thanks for cleaning this up

src/offset/local/tz_info/timezone.rs
```diff
@@ -52,6 +52,14 @@ impl TimeZone {
             return Self::from_file(&mut file);
         }
 
+        // attributes are not allowed on if blocks in Rust 1.38
+        #[cfg(target_os = "android")]
```
esheppa, line 56: should this moved up to above line 46? Also what do we do in the case that the `tz_string` is "localtime" on android?

esheppa (commented):
This looks good to me, only thing I'd like to have a further discussion on is the ordering of the options in `from_posix_tz`. Thanks for creating the `android-tzdata` library @RumovZ

src/offset/local/tz_info/timezone.rs
RumovZ: I  have yet to find exhaustive documentation on this, but from what I can tell, `tzdata` doesn't contain a `localtime` entry, and there is `/etc/localtime` like on other Unix systems. Moving the `tzdata` block up sounds good to me.

dae: I've checked an Android device here, and it does not have an /etc/localtime - perhaps they did away with it in later Android releases, or it varies by manufacturer? Android appears to set TZ to the timezone configured in the settings by default, so presumably users would not hit this unless they explicitly overwrote the default value.

src/offset/local/unix.rs
```diff
@@ -68,19 +68,18 @@ struct Cache {
...
-#[allow(dead_code)] // keeps the cfg simpler
 #[cfg(not(any(target_os = "android", target_os = "aix")))]
 const TZDB_LOCATION: &str = "/usr/share/zoneinfo";
 
 fn fallback_timezone() -> Option<TimeZone> {
     let tz_name = iana_time_zone::get_timezone().ok()?;
+    #[cfg(not(target_os = "android"))]
     let bytes = fs::read(format!("{}/{}", TZDB_LOCATION, tz_name)).ok()?;
+    #[cfg(target_os = "android")]
+    let bytes = android_tzdata::find_tz_data(&tz_name).ok()?;
```
jtmoon79, line 82: In PR code for `src/offset/local/tz_info/timezone.rs` (above) the `fn from_posix_tz` code attempts do things the "Android Way" but if that doesn't succeed it falls back to the current implementation. Specifically, if `android_tzdata::find_tz_data(tz_string)` fails it will fall through and use the current general implementation.

Should `fn fallback_timezone` code allow a similar fallback? e.g.

```lang-rust
fn fallback_timezone() -> Option<TimeZone> {
    let tz_name = iana_time_zone::get_timezone().ok()?;
    #[cfg(target_os = "android")]
    if let Ok(bytes) = android_tzdata::find_tz_data(&tz_name).ok() {
        return TimeZone::from_tz_data(&bytes).ok();
    }
    let bytes = fs::read(format!("{}/{}", TZDB_LOCATION, tz_name)).ok()?;
    TimeZone::from_tz_data(&bytes).ok()
}
```

RumovZ: In `from_posix_tz()`, we try to parse the tz string as a transition rule as a fallback, which should or shouldn't work regardless of the OS. On the other hand, trying to find a tz file on Android won't ever work, as far as I know, so I don't think it makes sense as a fallback.

jtmoon79: Okay. 👍

jtmoon79 (approved):

djc:
Thanks!

PR #979, opened 2023-03-06
Fix available range of datetime for `Local.from_utc_datetime` and `Local.from_local_datetime` functions

### LINKED ISSUE #977: Panic in conversion of NaiveDateTime to the local time on Windows
Code like this works fine:
```
let naivedatetime = NaiveDate::from_ymd_opt(1900, 1, 1)
    .unwrap().and_hms_opt(0, 0, 0)
    .unwrap();

let datetime_from_utc = Local.from_utc_datetime(&naivedatetime);
let datetime_from_local = Local.from_local_datetime(&naivedatetime).unwrap();
```
But changing of the first line to 
```
let naivedatetime = NaiveDate::from_ymd_opt(1898, 1, 1)
        .unwrap().and_hms_opt(0, 0, 0)
        .unwrap();
```
causes a panic: `thread 'datetime::tests::test_from_naive_date_time_conversions' panicked at 'attempt to subtract with overflow', src\offset\local\windows.rs:223:18`

I face this problem on Windows only. So what range of datetime should be supported by functions `Local.from_utc_datetime` and `Local.from_local_datetime` on Windows?

vvkostenko:
fixes #977

### DIFF
--- a/src/datetime/tests.rs
+++ b/src/datetime/tests.rs
@@ -954,21 +954,14 @@ fn test_datetime_sub_assign_local() {
 #[test]
 #[cfg(target_os = "windows")]
 fn test_from_naive_date_time_conversions() {
-    let min_year = NaiveDate::from_ymd_opt(1601, 1, 3)
-        .unwrap().and_hms_opt(0, 0, 0)
-        .unwrap();
+    let min_year = NaiveDate::from_ymd_opt(1601, 1, 3).unwrap().and_hms_opt(0, 0, 0).unwrap();
 
-    let max_year = NaiveDate::from_ymd_opt(30827, 12, 29)
-        .unwrap().and_hms_opt(23, 59, 59)
-        .unwrap();
+    let max_year = NaiveDate::from_ymd_opt(30827, 12, 29).unwrap().and_hms_opt(23, 59, 59).unwrap();
 
-    let too_low_year = NaiveDate::from_ymd_opt(1600, 12, 29)
-        .unwrap().and_hms_opt(23, 59, 59)
-        .unwrap();
+    let too_low_year =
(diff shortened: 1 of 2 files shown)


djc:
Hmm, I kind of don't want to take this because the `windows` module code is just so crappy? I think we should almost completely scrap the `windows` module and rebuild it from first principles to implement the `now()` and `naive_to_local()` functions with minimal use of the Windows API (and definitely avoiding all the Unixisms that are there now, like `Timespace`, `Tm` and a lot of strange indirection. Should probably also avoid the `panic!()` in `call()` in favor of yielding `LocalResult::None`.

Would you be interested in working on that? See also #989 which fixes other silliness in `windows`.

nekevss:
I don't know if it's exactly the way to go, but I did get a little OCD over the weekend after removing the panics and tried to clean up the module on this [branch](https://github.com/nekevss/chrono/tree/local-windows-rewrite). Not sure if it's exactly the direction to go as it is mostly just cleaning up the current module and creating wrappers around `windows_sys`.

djc:
A bit of OCD is probably a good thing for this! Your branch seems to still maintain the `Timespec` and `Tm` which I think are complete distractions from the task at hand. Also your branch is targeting the main branch which uses windows-sys, but ideally we'd first land this on 0.4.x which still uses winapi (and then merge a similar change that uses windows-sys to main later). IMO the goal is to make the `windows` module as concise as possible, translating `now()` and `naive_to_local()` as directly as possible to Win32 APIs.

djc:
I think the changes here have been addressed by #992?

jtmoon79:
> I think the changes here have been addressed by https://github.com/chronotope/chrono/pull/992?

Looks like @nekevss covered this in #992 (including the `test_from_naive_date_time_windows`).  Let's defer to that one. This PR should be Closed.

PR #983, opened 2023-03-08
Make eligible functions const.

tormeh:


tormeh:
This pleases my OCD :smile:

djc:
I think we should rebase this on main after we merge in 0.4.x (which now has #984).

djc:
(Merge PR submitted in #986.)

djc:
Would you mind rebasing this on current main? Should hopefully not be too painful.

esheppa (commented):
Looks good once this has been rebased on latest `main`. @djc - in this case as it is passing the tests it should meet MSRV on 1.48 - the only thing it is plausible that there are other missed opportunities to be made const?

tormeh:
Rebased it. I think MSRV is unchanged since last time I looked at this, so it should be good. That was the limiting factor for both PRs.

tormeh:
Seeing this error message on the MSRV build:
```
error: failed to parse manifest at `/home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/syn-2.0.0/Cargo.toml`

Caused by:
  failed to parse the `edition` key

Caused by:
  this version of Cargo is older than the `2021` edition, and only supports `2015` and `2018` editions.
  ```
  
  Sounds like it's not my fault. Not sure what to do about it

tormeh:
Looks like syn is pulled in by serde-derive in dev-dependencies. Either need to up the MSRV or constrain syn version.

tormeh:
> @tormeh - once #994 has been merged that will resolve this issue in the MSRV checking

That PR is for the 0.4 branch, so until it reaches main it won't fix the CI issue for this PR.

tormeh:
@djc If you don't care about the CI I guess you can just merge it now?

jtmoon79 (approved):
I really like this change. Some nice startup time improvements I can easily get from this in [my own project](https://github.com/jtmoon79/super-speedy-syslog-searcher/).

djc:
Thanks!

PR #986, opened 2023-03-09
Merge 0.4

djc:
Note: merge this with a merge commit.


djc:
Going to merge this proactively to avoid it bitrotting again (and unblock other things downstream).

PR #987, opened 2023-03-09
Prepare 0.4.24

djc:
I think it's time!


esheppa (approved):
Looks good to go! There are a bunch of bug fixes here which will be great to have released

PR #988, opened 2023-03-09
Fix locale formatting for `%c` and `%r`

### LINKED ISSUE #985: Invalid datetime format for korean(`ko_KR`)
# Summary

South korea uses [`Year-Month-Day`](https://en.wikipedia.org/wiki/Date_format_by_country)  date format like rest of east asia. however, it's invalidly rendered as `Month-Day-Year` with `%c`.

```
$ date '+%c
2023년 03월 09일 (목) 오전 11시 59분 40초
$ cargo run # chrono +%c
03/09/23 (목) 11:59:36 오전
```

## To reproduce
(tested in `0.4.23`)
```rs
use chrono::{Local, Locale};

fn main() {
    let time = Local::now();
    let display = |fmt: &str| {
        let localized = time.format_localized(fmt, Locale::ko_KR);
        println!("{fmt}: {localized}");
    };
    display("%c");
    display("%r");
}
```
```
%c: 03/09/23 (목) 11:59:36 오전
%r: 11:59:36 오전
 ~/repo   …  locale  date '+%c'
2023년 03월 09일 (목) 오전 11시 59분 40초
 ~/repo   …  locale  date '+%r'
오전 11시 59분 42초
```

scarf005:
# Summary
image

- fixes #985

### `%c` uses locale's date and time (`%x`, `%X`, `%r`)
```rs
assert_eq!(
	dt.format_localized("%c", Locale::ja_JP).to_string(),
	"2001年07月08日 00時34分60秒"
);
assert_eq!(
    dt.format_localized("%c", Locale::ko_KR).to_string(),
    "2001년 07월 08일 (일) 오전 12시 34분 60초"
);
```
###  `%r` uses locale's 12 hour clock time
```rs
assert_eq!(dt.format_localized("%r", Locale::ko_KR).to_string(), "오전 12시 34분 60초");
assert_eq!(dt.format_localized("%r", Locale::ja_JP).to_string(), "午前12時34分60秒");
```

### new macro: `locale_match_ampm!`
uses hardcoded list of locales to find `T_FMT_AMPM` due to https://github.com/cecton/pure-rust-locales/issues/4

### new tests: `test_strftime_localized_korean` and `test_strftime_localized_japanese`
will test some common locale related strftime format specifiers.

### DIFF
--- a/src/format/locales.rs
+++ b/src/format/locales.rs
@@ -31,3 +31,67 @@ pub(crate) const fn d_t_fmt(locale: Locale) -> &'static str {
 pub(crate) const fn t_fmt(locale: Locale) -> &'static str {
     locale_match!(locale => LC_TIME::T_FMT)
 }
+
+macro_rules! expand_ampm {
+    ($item:ident) => {
+        pure_rust_locales::$item::LC_TIME::T_FMT_AMPM
+    };
+}
+
+/// hardcoded `locale_match!` fallback for `T_FMT_AMPM`.
+///
+/// This is because `locale_match!` cannot be used when a locale misses a field.
+/// see [related issue](https://github.com/cecton/pure-rust-locales/issues/4) for details.
+macro_rules! locale_match_ampm {
+    ($locale:expr => ( $($item:ident),* $(,)? )) => {
+        match $locale {
+            $(Locale::$item => {
+                let t_fmt_ampm = expand_ampm!($item);
+
+                if t_fmt_ampm.is_empty() {
+                    expand_ampm!(en_US)
+                } else {
+                    t_fmt_ampm
+                }
+            },)*
+
+            // default fallback is en_US
+            _ => expand_ampm!(en_US),
+        }
+    }
+}
+
+pub(crate) fn t_fmt_ampm(locale: Locale) -> &'static str {
+    #[allow(unused_imports)]
+    use pure_rust_locales::Locale::*;
+
+    locale_match_ampm!(locale => (
+        POSIX,
+        aa_DJ, aa_ER, aa_ER_saaho, aa_ET, af_ZA, agr_PE, ak_GH, am_ET, an_ES, anp_IN, ar_AE, ar_BH, ar_DZ, ar_EG, ar_IN, ar_IQ, ar_JO, ar_KW, ar_LB, ar_LY, ar_MA, ar_OM, ar_QA, ar_SA, ar_SD, ar_SS, ar_SY, ar_TN, ar_YE, as_IN, ast_ES, ayc_PE, az_AZ, az_IR,
+        be_BY, be_BY_latin, bem_ZM, ber_DZ, ber_MA, bg_BG, bhb_IN, bho_IN, bho_NP, bi_VU, bn_BD, bn_IN, bo_CN, bo_IN, br_FR, br_FR_euro, brx_IN, bs_BA, byn_ER,
+        ca_AD, ca_ES, ca_ES_euro, ca_ES_valencia, ca_FR, ca_IT, ce_RU, chr_US, cmn_TW, crh_UA, cs_CZ, csb_PL, cv_RU, cy_GB,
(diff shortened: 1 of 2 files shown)


src/format/locales.rs
```diff
@@ -31,3 +31,67 @@ pub(crate) fn d_t_fmt(locale: Locale) -> &'static str {
...
+                }
+            },)*
+
+            // default fallback is en_US
+            _ => expand_ampm!(en_US),
+        }
+    }
+}
+
+pub(crate) fn t_fmt_ampm(locale: Locale) -> &'static str {
```
djc, line 64: I think this should get a clear comment to explain in high-level terms what it's trying to achieve, as that is pretty unclear to me. I'm not sure I agree that the stacked macro approach is the best option here...

djc (commented):
Thanks for fixing the clippy issue!

src/format/locales.rs
scarf005: alright, i think description for it would be similar to `locale_match_ampm!` macro, but will try splitting it up for high and low level terms.

scarf005: added more docstring at 3f0aafdaac76cdfa3bbbbcf623749b55c4af5f94. could you take a look?

esheppa (approved):
Thanks for this @scarf005

src/format/strftime.rs
```diff
@@ -65,7 +65,7 @@ The following specifiers are available both to formatting and parsing.
 | `%R`  | `00:34`       | Hour-minute format. Same as `%H:%M`.                                  |
 | `%T`  | `00:34:60`    | Hour-minute-second format. Same as `%H:%M:%S`.                        |
 | `%X`  | `00:34:60`    | Locale's time representation (e.g., 23:13:48).                        |
-| `%r`  | `12:34:60 AM` | Hour-minute-second format in 12-hour clocks. Same as `%I:%M:%S %p`.   |
+| `%r`  | `12:34:60 AM` | Locale's 12 hour clock time. (e.g., 11:11:04 PM)                      |
```
esheppa, line 68: Do we ever want to have a possible non-localized `%r` equivalent? Or do we think that it is probably not as useful as the localized version and hence it makes most sense to have the localized version only?

djc:
Reading the [referenced issue](https://github.com/cecton/pure-rust-locales/issues/4), notice the comment in https://github.com/cecton/pure-rust-locales/issues/4#issuecomment-1306556735:

> There is something utterly important that you must know about the current implementation though: I wanted the compiler to optimize away unused code. Every constant or function that is not used in the final binary (the user's binary) should (and must) be stripped by the compiler. This is because we don't want to store all the information for all the languages on every program.

It looks to me like the current setup here ends up referencing a long list of locales in our code, meaning that all of those will essentially be compiled into the chrono library if the localization feature is enabled. So I don't think this is the right approach.

scarf005:
iirc `locale_match_ampm` will expand roughly as same as how `match_locale` macro would have, original locale_match is:
```rs
#[macro_export]
macro_rules! locale_match {
    ($locale:expr => $($item:ident)::+) => {{
        match $locale {
            $crate::Locale::POSIX => $crate::POSIX::$($item)::+,
            $crate::Locale::aa_DJ => $crate::aa_DJ::$($item)::+,
            $crate::Locale::aa_ER => $crate::aa_ER::$($item)::+,
			(...around 290 items more)
```

scarf005:
looks like [rust msrv is too low](https://github.com/chronotope/chrono/actions/runs/4486696197/jobs/7889416160?pr=988) to support 2021, should i raise it?

jtmoon79:
> looks like [rust msrv is too low](https://github.com/chronotope/chrono/actions/runs/4486696197/jobs/7889416160?pr=988) to support 2021, should i raise it?

Relates to #1003 and #995

scarf005:
thanks. i'll rebase the PR once msrv issue is fixed.

pitdicker:
> thanks. i'll rebase the PR once msrv issue is fixed.

The time has come :smile:

djc:
Actually, this PR targets main, which doesn't have the MSRV bump yet.

scarf005:
oh, i see, should've made a PR to `0.4.x`. closing in favor of a new one targeting `0.4.x`.

PR #991, opened 2023-03-15
fix IsoWeek so that its flags are always correct

### LINKED ISSUE #295: IsoWeek comparisons seem off
Intuitively, this test should pass. Is this intentional behavior?

```
#[test]
fn iso_week_lt() {
    use chrono::{Local,TimeZone,Datelike};
    let week1 = Local.ymd(2018,12,31).iso_week();
    let week2 = Local.ymd(2019,01,01).iso_week();
    println!("{:?} < {:?}", week1, week2); // "2019-W01 < 2019-W01"
    assert!(week1 == week2); // fail
    assert!(!(week1 > week2)); // pass
    assert!(!(week1 < week2)); // fail
}
```

moshevds:
This fixes #295. PartialEq and PartialOrd are implemented by directly comparing the internal integer field, which includes the flags value.

### DIFF
--- a/src/naive/isoweek.rs
+++ b/src/naive/isoweek.rs
@@ -46,7 +46,8 @@ pub(super) fn iso_week_from_yof(year: i32, of: Of) -> IsoWeek {
             (year, rawweek)
         }
     };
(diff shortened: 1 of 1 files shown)


djc:
Thanks for fixing this long-standing issue!

PR #992, opened 2023-03-16
Refactor windows module in `Local`

nekevss:
Not sure if this (a new PR) is the best way to go about this, but I reworked the `windows` module as discussed in #979 targeting `0.4.x`. It removes the panics, removes the `Tm` and `Timespec` unix-centric structs to be more windows-focused structs `InnerDuration` and `LocalSysTime`. 

I did adapt the test that @vvkostenko had put together for his PR as a check, and everything does appears to run fine.

Let me know if this is more in line with the updates.


src/offset/local/windows_sys/file_time.rs
```diff
@@ -0,0 +1,50 @@
+use std::{io::Error, result::Result};
```
djc, line 1: Is this stuff that you actually extracted from the windows-sys crate?

I think for now I'd prefer keeping all the code together in the `windows` module, which also simplifies the privacy approaches. It can really be quite minimal.

src/offset/local/windows_sys/system_time.rs
```diff
@@ -0,0 +1,95 @@
...
+            wHour: 0,
+            wMinute: 0,
+            wSecond: 0,
+            wMilliseconds: 0,
+        };
+
+        Self { inner: st }
+    }
+
+    pub(crate) fn from_naive_datetime(dt: &NaiveDateTime) -> Self {
```
djc, line 45: This should yield a `Result` with the `Error` happening if the values are out of range (I'm thinking about the year value in particular).

src/offset/local/windows_sys/file_time.rs
```diff
@@ -0,0 +1,50 @@
...
+use winapi::um::timezoneapi::*;
+
+const HECTONANOSECS_IN_SEC: i64 = 10_000_000;
+const HECTONANOSEC_TO_UNIX_EPOCH: i64 = 11_644_473_600 * HECTONANOSECS_IN_SEC;
+
+pub(crate) struct WinFileTime {
+    inner: FILETIME,
+}
+
+impl WinFileTime {
```
djc, line 16: I don't think we should keep the `WinFileTime` as a separate type internally. As far as I can tell the only use we have for it is to implement `as_unix_seconds()`? So IMO it would be good to implement `as_unix_seconds()` on `WinSystemTime` directly instead (inlining all the necessary operations for creating a `FILETIME` etc).

djc (commented):
This looks like a pretty good direction! I think we should make it much more minimal in order to make it easy to understand/review/verify for correctness, though.

src/offset/local/windows_sys/system_time.rs
nekevss: Do we want to be catching them here or allowing the window's functions to notify us in the `windows_sys_call` macro?

src/offset/local/windows_sys/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+    pub(crate) fn from_unix_seconds(seconds: i64) -> Result<Self, Error> {
+        let t = ((seconds * HECTONANOSECS_IN_SEC) + HECTONANOSEC_TO_UNIX_EPOCH) as u64;
+        let filetime = FILETIME { dwLowDateTime: t as u32, dwHighDateTime: (t >> 32) as u32 };
+        let mut st = WinSystemTime::new();
+        unsafe { windows_sys_call!(FileTimeToSystemTime(&filetime, st.mut_inner()), 0) };
+        Ok(st)
+    }
+
+    /// Return `WinSystemTime`'s inner `SYSTEMTIME` struct.
+    pub(crate) fn inner(&self) -> SYSTEMTIME {
```
djc, line 103: I think we can replace with an `impl From<WinSystemTime> for NaiveDateTime`, which seems like a nicer API?

src/offset/local/windows_sys/file_time.rs
djc: I see you moved all the code from `windows_sys` back into a single module, but I actually meant moving everything into `windows` (just moving the `windows_sys` stuff to the bottom of the module).

src/offset/local/windows_sys/system_time.rs
djc: I'm not completely sure what you mean. I think we should yield `Result` here -- this function is only called by `naive_to_local()` which can return a `LocalResult::None` in case this function fails.

src/offset/local/windows_sys/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+    pub(crate) fn as_time_zone_specific(&self) -> Result<WinSystemTime, Error> {
+        let mut local = WinSystemTime::new();
+        unsafe {
+            windows_sys_call!(
+                SystemTimeToTzSpecificLocalTime(ptr::null(), &self.inner(), local.mut_inner()),
+                0
+            )
+        };
+        Ok(local)
+    }
```
djc, line 134: I think this should also become a free function:

```rust
fn utc_to_local(utc: &WinSystemTime) -> Result<WinSystemTime, Error> {
    ..
}
```

src/offset/local/windows_sys/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+            .field("second", &self.inner.wSecond)
+            .field("ms", &self.inner.wMilliseconds)
+            .finish()
+    }
+}
+
+impl WinSystemTime {
+    /// Creates a new empty `WinSystemTime`.
+    pub(crate) fn new() -> Self {
+        let st = SYSTEMTIME {
```
djc, line 53: I don't like that we're initializing `SYSTEMTIME` with invalid values here. I think it would be maybe more obvious to use `MaybeUninit` in the functions that need it.

src/offset/local/windows_sys/mod.rs
```diff
@@ -0,0 +1,151 @@
+use std::{fmt, io::Error, ptr, result::Result};
+
+use winapi::shared::minwindef::*;
+use winapi::um::minwinbase::SYSTEMTIME;
+use winapi::um::timezoneapi::*;
```
djc, line 5: Would like to get rid of the `*` imports here, can you expand them to what we're actually using?

src/offset/local/windows.rs
```diff
@@ -8,271 +8,105 @@
...
     sec: i64,
     nsec: i32,
 }
 
-impl Timespec {
+impl InnerDuration {
+    /// Constructs a new LocalSysTime representing the current time in UTC
     /// Constructs a timespec representing the current time in UTC.
-    fn now() -> Timespec {
+    fn now_locally() -> LocalSysTime {
```
djc, line 51: It seems like there should be Windows API that would let us get both the current `WinSystemTime` and the current timezone offset in one system call, instead of doing this stuff?

djc (commented):
This is definitely getting better! I think there are still some opportunities to simplify...

djc:
This is great, thanks!

PR #993, opened 2023-03-16
Apply suggestions for clippy 1.68

djc:


djc:
(CI for nightly is broken due to https://github.com/rust-lang/rust/issues/109199.)

PR #997, opened 2023-03-24
Adjusts windows local for millisecond and year

nekevss:
This PR is meant to address a couple fixes/potential gaps in the last request that I found when running the most recent `0.4.x` on `boa` with the test262 Date built-in suite.

Primarily the changes are:
  - Adds support for millisecond, which to be frank was probably a complete oversight on my part.
  - Lowers the potential min year, by checking the year and adding a `shifted` field to `LocalSysTime`.
  - Adjusts the previous test case to account for the above


src/offset/local/windows.rs
```diff
@@ -62,6 +63,7 @@ pub(super) fn naive_to_local(d: &NaiveDateTime, local: bool) -> LocalResult<Date
 struct LocalSysTime {
     inner: SYSTEMTIME,
     offset: i32,
+    shifted: bool,
```
djc, line 66: I think this should get a docstring to explain what it means.

src/offset/local/windows.rs
```diff
@@ -72,41 +74,52 @@ impl LocalSysTime {
...
 
         let offset = FixedOffset::east_opt(self.offset).unwrap();
         DateTime::from_utc(date.and_time(time) - offset, offset)
     }
 }
 
-fn system_time_from_naive_date_time(dt: &NaiveDateTime) -> SYSTEMTIME {
-    SYSTEMTIME {
+fn system_time_from_naive_date_time(dt: &NaiveDateTime) -> (SYSTEMTIME, bool) {
+    let (year, shifted) = if dt.year() < 1601 {
```
djc, line 116: What if `dt.year()` is smaller than zero?

I also think we should shift year by some multiple of 400, to avoid getting into issues with leap year calculations. Ideally we'd cover this behavior with some testing.

src/offset/local/windows.rs
```diff
@@ -72,41 +78,53 @@ impl LocalSysTime {
...
+        Ok(Self { inner: local_time, offset, shifted })
     }
 
     fn datetime(self) -> DateTime<Local> {
         let st = self.inner;
 
-        let date =
-            NaiveDate::from_ymd_opt(st.wYear as i32, st.wMonth as u32, st.wDay as u32).unwrap();
-        let time = NaiveTime::from_hms(st.wHour as u32, st.wMinute as u32, st.wSecond as u32);
+        let year = if self.shifted { st.wYear - 1600 } else { st.wYear };
```
jtmoon79, line 103: Could value `1600` be a file-level `const`? athe docstring explain the significance of year 1600?
Currently, `1600` is an opaque incantation.

src/offset/local/windows.rs
```diff
@@ -72,41 +78,53 @@ impl LocalSysTime {
...
 
         let offset = FixedOffset::east_opt(self.offset).unwrap();
         DateTime::from_utc(date.and_time(time) - offset, offset)
     }
 }
 
-fn system_time_from_naive_date_time(dt: &NaiveDateTime) -> SYSTEMTIME {
-    SYSTEMTIME {
+fn system_time_from_naive_date_time(dt: &NaiveDateTime) -> (SYSTEMTIME, bool) {
+    let (year, shifted) = if dt.year() < 1601 {
```
jtmoon79, line 120: Could value `1601` value be a file-level `const` with a helpful docstring? Currently, value `1601` is an opaque incantation.

src/offset/local/windows.rs
```diff
@@ -72,41 +78,53 @@ impl LocalSysTime {
...
+        .unwrap();
 
         let offset = FixedOffset::east_opt(self.offset).unwrap();
         DateTime::from_utc(date.and_time(time) - offset, offset)
     }
 }
 
-fn system_time_from_naive_date_time(dt: &NaiveDateTime) -> SYSTEMTIME {
-    SYSTEMTIME {
+fn system_time_from_naive_date_time(dt: &NaiveDateTime) -> (SYSTEMTIME, bool) {
```
jtmoon79, line 119: Is it possible to add test cases for only this function `system_time_from_naive_date_time`?
I can see this function is indirectly tested elsewhere but it seems good to me have tests focused solely on this function.

jtmoon79 (changes requested):

src/offset/local/windows.rs
nekevss: There is a docstring down in SYSTEMTIME that designates the valid values for Window's SYSTEMTIME struct (1601-30827). I don't see the exact issue with a `const` though

src/offset/local/windows.rs
```diff
@@ -68,7 +68,12 @@ struct LocalSysTime {
     inner: SYSTEMTIME,
     /// The offset value from UTC
     offset: i32,
-    // Denotes the +/- 400 shifted a year value from invalid to valid
+    // Denotes the multiple of 400 year intervals shifted to create a valid year value. The
```
jtmoon79, line 71: Great comment.  While you're at it, can this be a docstring, e.g. leading `///` ?

nekevss:
Feedback changes made 😄 let me know what you think!

jtmoon79 (approved):

pitdicker:
@nekevss Thank you for working on this! Going to close this PR because https://github.com/chronotope/chrono/pull/1017 is the preferred solution, even though it is not yet merged.

nekevss:
@pitdicker Sounds good to me 😄 Definitely agree that #1017 would be the better approach as well

Issue #995, opened 2023-03-22
The next MSRV bump


djc:
Apart from syn 2: https://github.com/strawlab/iana-time-zone/pull/91 updated its MSRV from 1.38 to 1.48, so unless we pin iana-time-zone we can no longer maintain our 1.38 MSRV for 0.4.x (which is our currently active release branch). The PR also started using windows-sys (which has a 1.48 MSRV) instead of winapi, so it's not trivial to revert to 1.38. @Kijewski and @astraw, would you consider pinging me on future MSRV bumps?

A few days ago, syn 2 was released, with an MSRV of 1.56. Current serde versions depend on syn 2, so we can no longer support serde on our current MSRV. Additionally, iana-time-zone-haiku depends on cxx and cxx-build, cxx depends on cxxbridge-macros, and both cxx-build and cxxbridge-macros depend on syn 2, so iana-time-zone-haiku now also effectively has a 1.56 MSRV. That doesn't necessarily need to affect other platforms, though isolating CI tests from platform-specific dependencies is a bit of a pain.

All things considered, I feel like anything pre-1.56 is pretty much dead in the water at this point, so I'm inclined to bump the MSRV to 1.56 for the next chrono 0.4 release. @esheppa any thoughts? @alex do you have any newer numbers on Rust version adoption in the Python ecosystem/among cryptography users?

alex:
The script I had for summarizing data broke, but at any rate...

Our current MSRV is 1.48; our next release deprecates that, and announces that the following release will have an MSRV of 1.56. So we think going at least that far is fine.

Kijewski:
Sorry, I only looked chrono's  Cargo.toml in the source, not what is actually currently used on crates.io, and I wrongly assumed you are already using windows-sys. :-/

But is iana-time-zone's MSRV bump actually going to be a problem for this project? [#980](https://github.com/chronotope/chrono/pull/980) made it so that iana-time-zone is only used in `cfg(unix)`, and the msrv is only increased for `cfg(windows)`. So I guess unless a windows targetting project is using an old `chrono<=0.4.23` and a new `iana-time-zone==0.1.54`, they should not be affected.

We will give your a heads up in the future if we intend to / need to bump our msrv!

astraw:
>  @Kijewski and @astraw, would you consider pinging me on future MSRV bumps?

Yes, apologies that we did not do this in advance. This was an oversight on my part. I will attempt to ensure we do it differently next time.

djc:
[...]

PR #999, opened 2023-03-24
Month partial ord

### LINKED ISSUE #998: Implement PartialOrd for Month
It would be nice if Month implemented PartialOrd.

For instance, I have a situation where I am aggregating data on a monthly basis, year by year. For that, I have a YearMonth struct like this:

```rust
struct YearMonth {
    year: i32,
    month: Month,
}
```

For the actual calculation, this is then used as key in a `HashMap`, or part of a tuple inside a `Vec`. I then have a situation where I want to find all the `YearMonth` for the last year, which means I have a comparison like `.filter(|year_month| year_month >= current_year_month.month.minus_one_year())` . Since `Month` does not implement `PartialOrd`, I cannot simple derive the implementation for `YearMonth`, but have to implement my own.

Munksgaard:
Also add a few tests to make sure it behaves as expected.

Fixes #998

### DIFF
--- a/src/month.rs
+++ b/src/month.rs
@@ -27,7 +27,7 @@ use rkyv::{Archive, Deserialize, Serialize};
 /// Allows mapping from and to month, from 1-January to 12-December.
 /// Can be Serialized/Deserialized with serde
 // Actual implementation is zero-indexed, API intended as 1-indexed for more intuitive behavior.
-#[derive(PartialEq, Eq, Copy, Clone, Debug, Hash)]
+#[derive(PartialEq, Eq, Copy, Clone, Debug, Hash, PartialOrd)]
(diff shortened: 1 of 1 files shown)


Munksgaard:
I hope I did it correctly by basing it on the 0.4.x branch.

djc (approved):
Yeah, this is great! The MSRV test will fail, but we can ignore that (#995).

totikom:
Is there a reason why `PartialOrd` is derived but `Ord` isn't?

Munksgaard:
> Is there a reason why `PartialOrd` is derived but `Ord` isn't?

Nothing in particular, I think. Shouldn't be hard to add if you need it? Don't know if @djc wants it though.

djc:
Sounds okay to me.

totikom:
Done in https://github.com/chronotope/chrono/pull/1128

Issue #835, opened 2022-10-05
parsing timezone should support Unicode "MINUS SIGN" (U+2212)


jtmoon79:
According to [Wikipedia](https://en.wikipedia.org/w/index.php?title=ISO_8601&oldid=1114309368#Time_offsets_from_UTC)

> To represent a negative offset, ISO 8601 specifies using a [minus sign](https://en.wikipedia.org/wiki/Minus_sign), (`−`). If the interchange character set is limited and does not have a minus sign character, then the [hyphen-minus](https://en.wikipedia.org/wiki/Hyphen-minus) should be used, (`-`). [ASCII](https://en.wikipedia.org/wiki/ASCII) does not have a minus sign, so its hyphen-minus character (code is 45 decimal or 2D hexadecimal) would be used. If the character set has a minus sign, then that character should be used. [Unicode](https://en.wikipedia.org/wiki/Unicode) has a minus sign, and its character code is U+2212 (2212 hexadecimal); the [HTML character entity](https://en.wikipedia.org/wiki/HTML_entity) invocation is `&minus;`.

## Problem

chrono `parse_from_str` fails to parse a Unicode "minus sign" character `−` (`U+2212`).

[rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=650c8bd417e148bbda1699d24105766b)
```lang-rust
extern crate chrono;
use chrono::DateTime;

fn main() {
    let dt_pattern = r"%Y %b %d %H%M%S %:z";

    // use ASCII "hyphen-minus"
    let data = r"2019 Aug 30 125959 -07:00";
    let dt = DateTime::parse_from_str(data, dt_pattern);
    dbg!(&dt);

    // use Unicode "minus" character
    let data = r"2019 Aug 30 125959 −07:00";
    let dt = DateTime::parse_from_str(data, dt_pattern);
    dbg!(&dt);
}
```

results in output
```text
[src/main.rs:10] &dt = Ok(
    2019-08-30T12:59:59-07:00,
)
[src/main.rs:15] &dt = Err(
    ParseError(
        Invalid,
    ),
)
```

(using chrono version *0.4.22*).

In some circumstances, chrono users must manually transform a found "minus sign" character `−` to "hyphen-minus" character `-`.

## Solution

To more fully support the ISO 8601 standard, chrono should parse Unicode "minus sign" character `−` (`U+2212`) in numeric timezone offsets, e.g. `−07:00`.

djc:
[...]

PR #1001, opened 2023-03-28
Issue 835 timezone offset allow MINUS SIGN (U+2212) (branch main)

jtmoon79:
Add parsing for [MINUS SIGN (U+2212)](https://www.compart.com/en/unicode/U+2212) timezone offset signage as specified in ISO 8601 and RFC 3339.

According to [Wikipedia](https://en.wikipedia.org/w/index.php?title=ISO_8601&oldid=1114309368#Time_offsets_from_UTC)

> To represent a negative offset, ISO 8601 specifies using a [minus sign](https://en.wikipedia.org/wiki/Minus_sign), (`−`). If the interchange character set is limited and does not have a minus sign character, then the [hyphen-minus](https://en.wikipedia.org/wiki/Hyphen-minus) should be used, (`-`). [ASCII](https://en.wikipedia.org/wiki/ASCII) does not have a minus sign, so its hyphen-minus character (code is 45 decimal or 2D hexadecimal) would be used. If the character set has a minus sign, then that character should be used. [Unicode](https://en.wikipedia.org/wiki/Unicode) has a minus sign, and its character code is U+2212 (2212 hexadecimal); the [HTML character entity](https://en.wikipedia.org/wiki/HTML_entity) invocation is `&minus;`.

Issue #835

---

Two extra commits I'd like to throw into this PR:

* small extra tests around nanosecond parsing
* consolidate path to `date`.

When running tests, there are two very long running tests (tens of minutes) that require `/usr/bin/date` to run (return early if `/usr/bin/date` is not present).
I often have to manually edit the path to `date` in both places to force the tests to return early (effectively skip).
That is, I manually changed
```lang-rust
   let date_path = "/usr/bin/date";
```
to
```lang-rust
   let date_path = "/usr/bin/date DO NOT RUN THIS TEST!";
```
Consolidating two local `data_path` variables to one file-wide `const PATH_DATE = ...` makes this workaround editing a little easier.

Additionally, consolidating the path follows the DRY principle.

### DIFF
--- a/src/format/parse.rs
+++ b/src/format/parse.rs
@@ -665,7 +665,10 @@ fn test_parse() {
     check!("",    [lit!("a")]; TOO_SHORT);
     check!(" ",   [lit!("a")]; INVALID);
     check!("a",   [lit!("a")]; );
-    // a Literal may contain whitespace and match whitespace, but this should not be done
+    check!("+",   [lit!("+")]; );
+    check!("-",   [lit!("-")]; );
+    check!("−",   [lit!("−")]; ); // MINUS SIGN (U+2212)
+                                  // a Literal may contain whitespace and match whitespace, but this should not be done
     check!(" ",   [lit!(" ")]; );
     check!("aa",  [lit!("a")]; TOO_LONG);
     check!("🤠",  [lit!("a")]; INVALID);
@@ -681,6 +684,8 @@ fn test_parse() {
     check!("1",   [lit!("1")]; );
     check!("1234", [lit!("1234")]; );
     check!("+1234", [lit!("+1234")]; );
+    check!("-1234", [lit!("-1234")]; );
+    check!("−1234", [lit!("−1234")]; ); // MINUS SIGN (U+2212)
     check!("PST", [lit!("PST")]; );
     check!("🤠",  [lit!("🤠")]; );
     check!("🤠a", [lit!("🤠"), lit!("a")]; );
@@ -747,6 +752,7 @@ fn test_parse() {
     check!("-0042",       [num!(Year)]; year: -42);
     check!("+0042",       [num!(Year)]; year: 42);
     check!("-42195",      [num!(Year)]; year: -42195);
+    check!("−42195",      [num!(Year)]; INVALID); // MINUS SIGN (U+2212)
     check!("+42195",      [num!(Year)]; year: 42195);
     check!(" -42195",     [num!(Year)]; INVALID);
     check!(" +42195",     [num!(Year)]; INVALID);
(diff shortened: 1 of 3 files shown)


src/format/parse.rs
```diff
@@ -634,6 +634,9 @@ fn test_parse() {
     check!("",    [lit!("a")]; TOO_SHORT);
     check!(" ",   [lit!("a")]; INVALID);
     check!("a",   [lit!("a")]; );
+    check!("+",   [lit!("+")]; );
+    check!("-",   [lit!("-")]; );
+    check!("−",   [lit!("−")]; );
```
esheppa, line 639: is it worth using a unicode escape here to make it clear that this is the unicode minus sign `"\u{2212}"`? I feel like otherwise this could be subtly deleted later

src/format/scan.rs
```diff
@@ -289,12 +299,33 @@ where
         }
     }
     let negative = match s.as_bytes().first() {
```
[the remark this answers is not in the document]
esheppa: and that would allow us to compress the allow_tz_minus_sign into the match also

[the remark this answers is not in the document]
jtmoon79: `chars()` worked best here.  Good suggestion.

For consistency, I used `len_utf8()` for each arm of this `match` statement. It's a `const` so will precompute the value at compile time (so no runtime cost).

src/format/parse.rs
jtmoon79: I tried using `\u{2212}` but I found code became difficult to grok.
I code commented all `'−'` with some variation of `// MINUS SIGN (U+2212)`.

If you'd strongly prefer `\u{2212}` then I can change it.  But I found the code comment and "native" `−` easier to grok.

src/format/scan.rs
```diff
@@ -388,7 +418,7 @@ pub(super) fn timezone_offset_2822(s: &str) -> ParseResult<(&str, Option<i32>)>
             Ok((s, None)) // recommended by RFC 2822: consume but treat it as -0000
         }
     } else {
-        let (s_, offset) = timezone_offset(s, |s| Ok(s))?;
+        let (s_, offset) = timezone_offset_internal(s, |s| Ok(s), false, false)?;
```
pitdicker, line 421: RFC 2822 doesn't say U+2212 is acceptable. But that is because it specifies an ASCII-only email format.

I think it would be helpful if chrono was a little less strict in interpreting the standard, and can also parse RFC 2822 from Unicode sources instead of only bare email.

I'm not sure when Word and similar software automatically replace `-` with U+2212 when used near/between a numbers. But it is not uncommon for typographical reasons.

pitdicker (approved):
Very nice! Anything that makes date parsing accept more valid date strings is a big :+1: from me.

Needs a rebase though.

src/format/scan.rs
jtmoon79: It's a reasonable point. In this case, I lean toward remaining strict about it.

jtmoon79:
I'm unable to replicate the failed `cargo fmt --check -- --color=always`.  WORKSFORME...

jtmoon79:
Any chance on this PR?

I can submit another targeting branch `0.4.x` when this PR is merged.

djc:
The changes here look okay to me, would you mind cleaning up the commit history? The `timezone_offset_internal()` refactor should happen before the original change to allow U+2212, the formatting fixes should be squashed into the commits that introduced them, and tests using U+2212 should be annotated early in the PR so that any new tests using it can add comment directly when they're added.

Also, please rebase on 0.4.x so that the MSRV CI issue gets solved.

jtmoon79:
PR for branch `0.4.x` #1087

djc:
Given that #1087 will get merged to main, do we still need this PR?

jtmoon79:
> Given that #1087 will get merged to main, do we still need this PR?

No. Closed.

PR #1003, opened 2023-03-30
Merge 0.4.x msrv ci issue

esheppa:
My intention here is to bring across the CI fix, but having looked into more detail at 1f1e2f8ff0e166ffd80ae95218a80b54fe26e003 I think that there may be some possibility to bring some or all of that work into 0.5 which is using `windows-sys` instead, so perhaps its better for this not to be merged for now. Question is, should that work be done in the merge here, or in a seperate PR directly to `main`?


kennykerr:
I don't recommend using `windows-bindgen` directly. I think your best bet is to stick with `windows-sys`. 

https://kennykerr.ca/rust-getting-started/standalone-code-generation.html

pitdicker:
Now that the MSRV is raised to 1.56 in https://github.com/chronotope/chrono/pull/1053 we would no longer need a workaround.

esheppa:
That is a good point @pitdicker - and as the MSRV now exceeds `1.48` everywhere, we can use the same impl for offset/local/windows.rs in both `0.4.x` and `main`.

pitdicker:
I opened https://github.com/chronotope/chrono/issues/1190 for the suggestion to switch to `windows-sys` like on main.

PR #1004, opened 2023-03-30
Inline a handful of functions from num-integer

alex:
The motivation for this is that we use a very small portion of what num-integer provides, but it's a relatively heavy dependency -- it brings in several crates with build scripts. The goal here is to reduce compilation time for projects using chrono.

Let me know if you'd prefer I rebase this onto the 0.4.x branch.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -1,0 +1,89 @@
+// All code cribbed from num-integer. We do this to minimize dependencies and
+// build-time.
+//
+// num-integer is released under the terms of the MIT OR Apache-2.0 licenses,
+// as is this code. These are the same licenses as Chrono itself.
+
+pub(crate) trait Integer: Copy {
+    fn div_rem(self, other: Self) -> (Self, Self);
+    fn div_floor(self, other: Self) -> Self;
+    fn mod_floor(self, other: Self) -> Self;
+}
+
+// Implements the Integer trait for a signed integer type.
+macro_rules! impl_integer_signed {
+    ($t: ty) => {
+        impl Integer for $t {
+            #[inline]
+            fn div_rem(self, other: Self) -> (Self, Self) {
+                (self / other, self % other)
+            }
+
+            #[inline]
+            fn div_floor(self, other: Self) -> Self {
+                // Algorithm from [Daan Leijen. _Division and Modulus for Computer Scientists_,
+                // December 2001](http://research.microsoft.com/pubs/151917/divmodnote-letter.pdf)
+                let (d, r) = self.div_rem(other);
+                if (r > 0 && other < 0) || (r < 0 && other > 0) {
+                    d - 1
+                } else {
+                    d
+                }
+            }
+
+            #[inline]
+            fn mod_floor(self, other: Self) -> Self {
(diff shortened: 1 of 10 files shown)


src/utils.rs
```diff
@@ -0,0 +1,84 @@
+// All code cribbed from num-integer. We do this to minimize dependencies and
```
jtmoon79, line 1: When cribbing other software, something to be cautious around is declaring licenses. On advice of a lawyer, it was adequate to surround code with explicit comments about the source, and the license*.

Here is an example from my own project https://github.com/jtmoon79/super-speedy-syslog-searcher/blob/0.5.58/src/data/utmpx.rs#L134-L186

I recommend doing the same here. The top of the file should include a docstring stating something like

```lang-rust
//! Number utility functions.
//!
//! Code in this file `src/utils.rs` was copied from project [`num-integer`].
//! That project is [licensed under the MIT license].
//!
//! [`num-integer`]: https://github.com/rust-num/num-integer
//! [licensed under the MIT license]: https://github.com/rust-num/num-integer/blob/num-integer-0.1.45/LICENSE-MIT
```

_* not formal legal advice, just a recommendation_

src/utils.rs
```diff
@@ -0,0 +1,84 @@
...
+// build-time.
+
+pub(crate) trait Integer: Copy {
+    fn div_rem(self, other: Self) -> (Self, Self);
+    fn div_floor(self, other: Self) -> Self;
+    fn mod_floor(self, other: Self) -> Self;
+}
+
+macro_rules! impl_integer_signed {
```
jtmoon79, line 10: Could `impl_integer_signed` have a docstring explaining it?

src/utils.rs
```diff
@@ -0,0 +1,84 @@
...
+                    r + other
+                } else {
+                    r
+                }
+            }
+        }
+    };
+}
+
+macro_rules! impl_integer_unsigned {
```
jtmoon79, line 45: Could `impl_integer_unsigned` have a docstring explaining it?

jtmoon79 (changes requested):
Removing dependencies is good.

jtmoon79 (approved):
Thanks for the update. I prefer `///` so those comments also appear in online docs. But code comments are adequate.

alex:
The module isn't public, so I don't think it'd show up in docs either way?

jtmoon79:
> The module isn't public, so I don't think it'd show up in docs either way?

Good point!

djc:
Oh, good idea. It would be nice to have this on 0.4.x as well, would you mind rebasing?

alex:
@djc hmm, So 0.4.x has `impl num_traits::FromPrimitive for Month`, which makes it part of the public API. I'm not sure if we can remove it without a compatibility break there?

alex:
https://github.com/chronotope/chrono/pull/794 is the PR that removed the num-traits usage in the API, FWIW.

djc:
Ah yeah, that would be the reason we haven't done this on 0.4. Okay, let's just do this on 0.5?

alex:
Sounds good. I think this is ready for review then (modulo the CI issues that are also present on main.)

djc (approved):

PR #1007, opened 2023-03-31
Add #[must_use] to some methods

aceArt-GmbH:
## Problem:

```rust
some_date.checked_add_days(Days::new(1)).unwrap();
```

## Expected:

`some_date` to change by 1 day, or a warning to be thrown.
Currently neither happens.

## After:

```
warning: unused return value of `NaiveDate::checked_add_days` that must be used
   --> src/main.rs:10:9
    |
10  |         some_date.checked_add_days(Days::new(1));
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(unused_must_use)]` on by default
```

Adding [`#[must_use]`](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute) indicates that the method does not change the original object and the return value has to be used.

As I am not using chrono too much, I only added must_use to some methods and not all that should have it

### DIFF
--- a/src/date.rs
+++ b/src/date.rs
@@ -77,6 +77,7 @@ impl<Tz: TimeZone> Date<Tz> {
     //
     // note: this constructor is purposely not named to `new` to discourage the direct usage.
     #[inline]
+    #[must_use]
     pub fn from_utc(date: NaiveDate, offset: Tz::Offset) -> Date<Tz> {
         Date { date, offset }
     }
@@ -86,6 +87,7 @@ impl<Tz: TimeZone> Date<Tz> {
     ///
     /// Panics on invalid datetime.
     #[inline]
+    #[must_use]
     pub fn and_time(&self, time: NaiveTime) -> Option<DateTime<Tz>> {
         let localdt = self.naive_local().and_time(time);
         self.timezone().from_local_datetime(&localdt).single()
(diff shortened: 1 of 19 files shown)


jtmoon79:
This is a good idea.
I reviewed other code. It seems adding `#[must_use]` in only one file might be inconsistent. Here is [one of many examples](https://github.com/chronotope/chrono/blob/v0.4.23/src/datetime/mod.rs#L159).

@djc my own recommendation is to allow this PR and then create a follow-up Issue like _add must_use attribute_ that addresses the remaining code base. So this partially-complete work is not forgotten. 

Or maybe @aceArt-GmbH would like to make an attempt to apply `#[must_use]` to the entire code-base for this PR? IMO, `#[must_use]` is only needed for `pub` functions that take `&self` (and of course that 1. return something new 2. don't modify `self`).

aceArt-GmbH:
thanks @jtmoon79 for your feedback.
I added `#[must_use]` to all `pub fn`. Didn't see any that changed `self`

jtmoon79 (approved):
I visually reviewed the changes. It all looks correct to me.

This is a great usability improvement. Thanks so much @aceArt-GmbH !

djc:
Thanks for working on this! Can this be put on the 0.4.x branch? I think that would probably make sense.

aceArt-GmbH:
I switched the branch

djc (approved):

djc:
Thanks!

PR #1011, opened 2023-04-05
Add parse_and_remainder methods

pitdicker:
Functions like `DateTime::parse_from_str` fail with `ParseErrorKind::TooLong` if a string contains content after the date(/time).
It would be very helpful for me if there was a function that would return the parsed value and remaining string slice.

`format::parse::parse_internal` already supports this functionality, so this would merely be exposing it.

The alternative for other libraries would be to re-implement a parser, or to first extract the relevant part of the string with something like a regex.

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -23,7 +23,7 @@ use std::time::{SystemTime, UNIX_EPOCH};
 use crate::format::DelayedFormat;
 #[cfg(feature = "unstable-locales")]
 use crate::format::Locale;
-use crate::format::{parse, ParseError, ParseResult, Parsed, StrftimeItems};
+use crate::format::{parse, parse_and_remainder, ParseError, ParseResult, Parsed, StrftimeItems};
 use crate::format::{Fixed, Item};
 use crate::naive::{Days, IsoWeek, NaiveDate, NaiveDateTime, NaiveTime};
 #[cfg(feature = "clock")]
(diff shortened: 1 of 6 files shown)


src/naive/datetime/mod.rs
```diff
@@ -328,6 +328,16 @@ impl NaiveDateTime {
         parsed.to_naive_datetime_with_offset(0) // no offset adjustment
     }
 
+    /// Parses a string with the specified format string and returns a new `NaiveDateTime`, and a
+    /// slice with the remaining portion of the string.
+    /// See the [`format::strftime` module](../format/strftime/index.html)
+    /// on the supported escape sequences.
```
jtmoon79, line 334: I noticed in some cases the original `fn parse_from_str` has a larger and longer docstring with further details. Could you hint to the user that `fn parse_and_remainder` is very similar to `fn parse_from_str` in the docstring?

Could you add a separate line like:

```lang-rust
///
/// Similar to [`parse_from_str`](parse_from_str)
```

Do this for all the `fn parse_from_str` docstrings.

src/datetime/mod.rs
```diff
@@ -604,6 +604,25 @@ impl DateTime<FixedOffset> {
         parse(&mut parsed, s, StrftimeItems::new(fmt))?;
         parsed.to_datetime()
     }
+
+    /// Parses a string from a user-specified format into a `DateTime<FixedOffset>` value, and a
+    /// slice with the remaining portion of the string.
+    ///
+    /// Note that this method *requires a timezone* in the input string. See
+    /// [`NaiveDateTime::parse_from_str`](./naive/struct.NaiveDateTime.html#method.parse_from_str)
+    /// for a version that does not require a timezone in the to-be-parsed str. The returned
```
jtmoon79, line 613: > `/// ... does not require a timezone in the to-be-parsed str. The returned`

Somewhat awkward phrase.

How about text

```lang-rust
/// ... does not require a timezone in `s`. The returned
```

src/datetime/mod.rs
```diff
@@ -604,6 +604,25 @@ impl DateTime<FixedOffset> {
...
+    /// [`DateTime`] value will have a [`FixedOffset`] reflecting the parsed timezone.
+    ///
+    /// See the [`format::strftime` module](./format/strftime/index.html) for supported format
+    /// sequences.
+    pub fn parse_and_remainder<'a>(
```
jtmoon79, line 618: Please add `#[must_use]` where appropriate.  I think all of these new functions need that.

There is a closed PR on `0.5` branch that adds `#[must_use]` so this gets ahead of some extra work. https://github.com/chronotope/chrono/pull/1007

jtmoon79 (approved):
Overall, why not?  As the Op wrote, the functions are already there.

Some nitpicks in my review.

@djc since this adds new API endpoints, should this target branch `main` (e.g. release 0.5) instead of branch `0.4.x`?

src/datetime/mod.rs
pitdicker: All these methods return a `ParseResult` which already gives these warnings.

pitdicker:
Thank you, updated.

jtmoon79 (approved):

esheppa:
This should be ok on `0.4.x` as it doesn't change any existing API surface and only adds new functions which is fine. If possible @pitdicker it would be great to have a couple of tests that show the new behavior - in this case doc tests are probably fine

pitdicker:
Sorry for the slow reply. I added some doctests.

jtmoon79 (approved):

src/datetime/mod.rs
```diff
@@ -603,6 +603,25 @@ impl DateTime<FixedOffset> {
         parse(&mut parsed, s, StrftimeItems::new(fmt))?;
         parsed.to_datetime()
     }
+
+    /// Parses a string from a user-specified format into a `DateTime<FixedOffset>` value, and a
+    /// slice with the remaining portion of the string.
+    ///
+    /// Note that this method *requires a timezone* in the input string. See
+    /// [`NaiveDateTime::parse_from_str`](./naive/struct.NaiveDateTime.html#method.parse_from_str)
```
djc: I suppose the comment here should refer to `NaiveDate::parse_and_remainder()`?

djc (commented):
Please squash the review feedback into the originating commit(s).

pitdicker:
:+1: Done

djc (approved):

djc:
Just needs a rebase.

pitdicker:
Rebased.

PR #1012, opened 2023-04-06
add impl From<NaiveDateTime> for NaiveDate

### LINKED ISSUE #1006: Why no From<NaiveDateTime> for NaiveDate?
Looks like the canonical way to convert a `NaiveDateTime` into a `NaiveDate` is with [`NaiveDateTime::date`](https://docs.rs/chrono/latest/chrono/naive/struct.NaiveDateTime.html#method.date). But I wonder, why is this conversion only implemented as a method on `NaiveDateTime`, and not as an implementation of `From<NaiveDateTime> for NaiveDate`? It seems to me that in general, the standard place to look for infallible type conversions is [`trait From`](https://doc.rust-lang.org/std/convert/trait.From.html), not in the from-type's regular methods. 

Is there a reason why there is no `impl From<NaiveDateTime> for NaiveDate`?

pezcore:
Add `impl From<NaiveDateTime> for NaiveDate`. Fixes #1006

### DIFF
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -1828,6 +1828,12 @@ impl Sub<NaiveDate> for NaiveDate {
     }
 }
 
+impl From<NaiveDateTime> for NaiveDate {
+    fn from(naive_datetime: NaiveDateTime) -> Self {
+        naive_datetime.date()
+    }
+}
+
 /// Iterator over `NaiveDate` with a step size of one day.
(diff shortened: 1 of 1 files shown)


jtmoon79 (commented):
Can this use `#[must_use]` ?

pitdicker (approved):

djc:
Thanks, sorry for the delay!

PR #1013, opened 2023-04-12
Fix military timezones in RFC 2822 parsing

pitdicker:
Currently RFC 2822 formatted datetimes with a single letter as timezone fail to parse, like `Tue, 20 Jan 2015 17:35:20 Z`.

The existing code had a comment `recommended by RFC 2822: consume but treat it as -0000`.
But then it would do the opposite and return an error.

I changed it to allow all acceptable military timezones (A-I and K-Z), but keep failing on other input.

### DIFF
--- a/src/format/parse.rs
+++ b/src/format/parse.rs
@@ -854,6 +854,20 @@ fn test_rfc2822() {
         ("Tue, 20 Jan 2015 17:35:20 -0890", Err(OUT_OF_RANGE)), // bad offset
         ("6 Jun 1944 04:00:00Z", Err(INVALID)),            // bad offset (zulu not allowed)
         ("Tue, 20 Jan 2015 17:35:20 HAS", Err(NOT_ENOUGH)), // bad named time zone
+        // named timezones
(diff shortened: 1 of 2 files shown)


djc:
Where are these military timezones defined, and what do they mean? I think adding support for `Z` makes sense, but I'm not convinced adding support for other one-letter zones in chrono itself makes sense.

pitdicker:
To be honest, `Z` is the only one I care about.

[RFC 2822](https://www.rfc-editor.org/rfc/rfc2822#page-31) defines them as:

> The 1 character military time zones were defined in a non-standard
way in [RFC822] and are therefore unpredictable in their meaning.
The original definitions of the military zones "A" through "I" are
equivalent to "+0100" through "+0900" respectively; "K", "L", and "M"
are equivalent to  "+1000", "+1100", and "+1200" respectively; "N"
through "Y" are equivalent to "-0100" through "-1200" respectively;
and "Z" is equivalent to "+0000".  However, because of the error in
[RFC822], they SHOULD all be considered equivalent to "-0000" unless
there is out-of-band information confirming their meaning.

I thought it is best to support the complete format defined by RFC 2822 just to be on the safe side. This is also what `time` happens to do: https://github.com/time-rs/time/blob/main/time/src/parsing/parsable.rs#L243:L264

Issue #716, opened 2022-06-21
Consider new helper methods for `LocalResult`


botahamec:
From #687 

> Furthermore, while the LocalResult is of course correct, I think it would be nice to have some kind of "lossy" output. I would assume that most of the time there is only one result (as the DST transitions are mostly in the night anyway, where they have the least impact). Currently the user has to handle this situation every time, while most of the time the conversion will be correct. Anyway, I think it would be good for a time library to have some sensible default how to work with this, as not every user might know what to do in this case.
>
> I checked on the Java time API, and they handle it like this (https://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html#atZone-java.time.ZoneId-):
>
> > In most cases, there is only one valid offset for a local date-time. In the case of an overlap, where clocks are set back, there are two valid offsets. This method uses the earlier offset typically corresponding to "summer".
> >
> > In the case of a gap, where clocks jump forward, there is no valid offset. Instead, the local date-time is adjusted to be later by the length of the gap. For a typical one hour daylight savings change, the local date-time will be moved one hour later into the offset typically corresponding to "summer".
> >
> > To obtain the later offset during an overlap, call [ZonedDateTime.withLaterOffsetAtOverlap()](https://docs.oracle.com/javase/8/docs/api/java/time/ZonedDateTime.html#withLaterOffsetAtOverlap--) on the result of this method. To throw an exception when there is a gap or overlap, use [ZonedDateTime.ofStrict(LocalDateTime, ZoneOffset, ZoneId)](https://docs.oracle.com/javase/8/docs/api/java/time/ZonedDateTime.html#ofStrict-java.time.LocalDateTime-java.time.ZoneOffset-java.time.ZoneId-).
>
> Why not create an additional method on LocalResult which returns something like the Java handling (not quite sure how other time libraries handle this situation)?

I like what we have already (the `earliest` and `latest` methods might just be as far as we want to go), but I wanted to see what people think about this. I don't think there's anything we could do without a breaking change. One idea is to have the implementor of the `TimeZone` interface provide a later or earlier time that would work.

esheppa:
Thanks for adding this @botahamec - and thanks @gerritsangel for the original comments  - I think this is a great idea.

[...]

Issue #651, opened 2022-02-17
SystemTimeToTzSpecificLocalTime failed with: The parameter is incorrect. (os error 87)


andrewbanchich:
We're getting a panic from chrono [here](https://github.com/chronotope/chrono/blob/main/src/sys/windows.rs#L97) when run on a Windows 2012 server:

```
thread '<unnamed>' panicked at 'SystemTimeToTzSpecificLocalTime failed with: The parameter is incorrect. (os error 87)',
/cargo/registry/src/github.com-1ecc6299db9ec823/chrono-0.4.19/src/sys/windows.rs:97:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b\/library\std\src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b\/library\core\src/panicking.rs:107:14
   2: chrono::sys::inner::time_to_local_tm
   3: <chrono::offset::local::Local as chrono::offset::TimeZone>::from_utc_datetime
   4: <chrono::datetime::DateTime<chrono::offset::local::Local> as core::convert::From<std::time::SystemTime>>::from
...
```

We debugged the `SystemTime` we were getting and it was:

```rust
SystemTime {
    intervals: 147221225472,
}
```

A normal `intervals` value (e.g. `132790361186042732`) is much larger.

I'm assuming there's something about this unusually short `SystemTime` that's causing chrono to panic.

andrewbanchich:
This seems to be an issue with something lower level having to do with Windows, so I'm going to close this.

We [opened an issue](https://github.com/microsoft/windows-rs/issues/1549) in the official `windows` crate repo.

andrewbanchich:
Actually, I'm reopening this since based on [the feedback](https://github.com/microsoft/windows-rs/issues/1549#issuecomment-1044861209) from that issue it looks like the problem might be more in chrono's domain.

djc:
I'm unable to test on Windows myself. I'd be happy to review any code changes that are supported with documentation and/or explanation why they improve the current situation, but I probably won't be able to work on this proactively.

Let me know if you need any guidance on how to fix chrono to support this scenario better.

andrewbanchich:
@annmarie-switzer should have the fix for this. Unfortunately, I don't have the code and also am unable to test on Windows.

rgwood:
[...]

PR #1017, opened 2023-04-17
Windows: base implementation on `GetTimeZoneInformationForYear`

pitdicker:
Issues caused by the current use of Windows API's:
- `TzSpecificLocalTimeToSystemTime` doesn't let us choose how to handle ambiguous cases during a DST transition. It always returns the earliest option.
- using `GetLocalTime` with `TzSpecificLocalTimeToSystemTime` can give wrong results during a DST transition  because the issue mentioned above.
- `TzSpecificLocalTimeToSystemTime` always returns a result, also for non-existing datetimes during a DST transition.
- The Windows API restricts years to a range of `1601..30828`. 

The alternative in this PR is to calculate the offset ourselves using the information provided by `GetTimeZoneInformationForYear`.This is similar to what we do on Unix.

The tricky bits to determine the correct offset between DST transitions is copied from the Unix implementation. Initially I wanted to share the code between the two systems, but that didn't make the Unix implementation any clearer.

Part of the old Windows implementation have been moved to a test, so our implementation can be compared to `TzSpecificLocalTimeToSystemTime`. I have set my computer timezone to ca. 20 'interesting' timezones (determined by looking at data of the entries enumerated by [`EnumDynamicTimeZoneInformation`](https://learn.microsoft.com/en-us/windows/win32/api/timezoneapi/nf-timezoneapi-enumdynamictimezoneinformation)) and found no differences.

---

I ended up writing this PR because of a yak shaving. In order to remove some `unwrap`s from my code, I need some progress on https://github.com/chronotope/chrono/issues/716. To show we need a fourth enum variant for `LocalResult`, I started implementing a proof of concept for `DateTime<Local>`. And it turned out the Windows implementation had no place to hook into...

Fixes https://github.com/chronotope/chrono/issues/1150, https://github.com/chronotope/chrono/issues/651.


djc:
Did you see #997 already?

esheppa:
It looks like these two PRs should be somewhat orthogonal, but this is also related to https://github.com/chronotope/chrono/pull/750 which is great as by extending this we should be able to extract the name of the timezone as well

djc:
> The tricky bits to determine the correct offset between DST transitions is copied from the Unix implementation. Initially I wanted to share the code between the two systems, but that didn't make the Unix implementation any clearer.

Can you add that here as a separate commit? It does sound like something we'd want.

pitdicker:
> Did you see #997 already?

Somewhat, I forgot about it. But the PR before by @nekevss was very useful for feeling my way in the Windows implementation.

> Can you add that here as a separate commit? It does sound like something we'd want.

I can dig up the commit, but it will probably be tomorrow because it needs some cleaning up.

esheppa:
We could also review and merge this as is, before enhancing to get the timezone name via another PR if that is more suitable @djc?

src/offset/local/tz_info/timezone.rs
```diff
@@ -572,8 +584,11 @@ impl LocalTimeType {
     }
 
     /// Returns offset from UTC in seconds
-    pub(crate) const fn offset(&self) -> i32 {
-        self.ut_offset
+    pub(crate) const fn offset(&self) -> Result<FixedOffset, Error> {
+        match FixedOffset::east_opt(self.ut_offset) {
```
djc, line 588: This is kind of fake, right? Don't we already know, by here, that `ut_offset` is in bounds?

src/offset/local/mod.rs
```diff
@@ -186,6 +189,102 @@ impl TimeZone for Local {
     }
 }
 
+#[cfg(any(unix, windows))]
```
djc, line 192: I'd like the three Unix-refactoring commits appear before the ones that add new Windows stuff, can you fix that up?

djc (commented):
Thanks!

src/offset/local/tz_info/timezone.rs
pitdicker: If it was my personal code I would just trust the timezone data to be valid an `unwrap` all over the place :smile:. It is mostly there because `unwrap` and `expect` are not available in `const`.

This preserves the check that was in https://github.com/chronotope/chrono/blob/main/src/offset/local/unix.rs#L156

pitdicker: I'll try to find a better way to structure this.

djc: If we execute on the MSRV bump discussed in #995 (which we should be doing anyway -- hint hint), `unwrap()` might become available in `const` contexts. However, the issue with `unwrap()` in `const` is that, while I'm fine with `unwrap()` being called in an actual `const` context (that is, at compile time), `const` functions can actually still also be called at run time, so we still don't want to panic in that case.

pitdicker: I have made `LocalTimeType` store a `FixedOffset`, so validation is now moved to creation time instead of use.

pitdicker:
Just noticed issue https://github.com/chronotope/chrono/issues/651. That issue would be fixed by this PR, added a test.

src/offset/local/windows.rs
```diff
@@ -8,158 +8,222 @@
...
-
-        Self::from_local_time(st).expect("Current local time must exist")
+// The basis for Windows timezone and DST support has been in place since Windows 2000. It does not
+// allow for complex rules like the IANA timezone database:
+// - A timezone has the same base offset the whole year.
+// - There are either zero or two DST transitions.
+// - As of Vista(?) only years from 2004 until a few years into the future are supported.
+// - All other years get the base settings, which seem to be that of the current year.
+impl TzInfo {
+    fn get_current_for_year(year: i32) -> Option<Self> {
```
nekevss, line 52: So this would still cause windows to return `LocalResult::None` in cases outside of the 1601-30827 range, right?

Is it possible to adjust for handling dates outside of that range so that the behavior is stable whether it is used on linux or windows.

pitdicker: That is what the comment below it is about :smile:. The Windows API has those limitations, so we clamp the year for values this far into the past or future. Whatever timezone rules are extrapolated for these years (and Windows extrapolates it really naive...) is what can be extrapolated for years beyond. But the clamped `ref_year` is only used for getting the timezone info, not for calculating the local datetime.

nekevss: Ah, sorry. I really should've double checked `clamp` (100% my bad). 

In this case though, since we can be certain that `ref_year` should always be valid, would it be better to actually panic when `GetTimeZoneInformationForYear` returns 0 instead of returning `None` and hiding the error?

src/datetime/tests.rs
```diff
@@ -857,6 +857,15 @@ fn test_datetime_from_local() {
...
 
+#[test]
+fn test_issue_651() {
+    // dt corresponds to `FILETIME = 147221225472` from the issue.
+    // This used to fail on Windows for timezones with an offset of -5:00 or greater.
+    let dt = NaiveDate::from_ymd_opt(1601, 1, 1).unwrap().and_hms_milli_opt(4, 5, 22, 122).unwrap();
+    let local_dt = Local.from_utc_datetime(&dt);
+    dbg!(local_dt);
+}
+
```
nekevss, line 868: Would it be of any added value to check the leap year information for 1500 and 1600 on windows?

pitdicker: Handling milliseconds and dealing with extreme years where you have to account for the 400-year leap cycle were your efforts in https://github.com/chronotope/chrono/pull/997. I am sorry this PR crosses that and takes a different approach.

We are not doing anything special anymore in that area, it should be as valid as other leap year handling in chrono.

pitdicker:
I noticed `Utc::now` uses the standard library to get the current date, while for `Local::now` we call a Windows API ourselves. Because on both Unix and Windows we get an UTC timestamp from the OS and then convert it to local time, we may just as well always use `Utc::now` inside `Local::now` and be consistent.

On Windows the standard library uses [`QueryPerformanceCounter`](https://learn.microsoft.com/en-us/windows/win32/api/profileapi/nf-profileapi-queryperformancecounter), which is guaranteed to provide sub-microsecond values. What we used for `Local::now` returned a `SYSTEMTIME` with at most millisecond precision.

pitdicker:
The benchmark results have improved a bit.
Before:
```
     Running benches\chrono.rs (target\release\deps\chrono-3f8222dc2c00b8c6.exe)
bench_get_local_time    time:   [1.4577 µs 1.5100 µs 1.5519 µs]
```

After:
```
     Running benches\chrono.rs (target\release\deps\chrono-3f8222dc2c00b8c6.exe)
bench_get_local_time    time:   [1.2091 µs 1.2737 µs 1.3304 µs]
                        change: [-17.784% -13.853% -9.6609%] (p = 0.00 < 0.05)
                        Performance has improved.
```

src/offset/local/windows.rs
pitdicker: I hope that in the future when chrono uses `Result`s instead of `Option`s for fallible operations this can return something like `Err(ChronoError::LocalTzLoadingError)`. Currently `None`, which gets converted to `LocalResult::None`, is the only way to return an error, similar to the Unix implementation.

nekevss: My thought was that you could panic instead of returning `None`, which might be the better option in this scenario, In theory, this should never fail from an incorrect parameter error, so the source of the error in that case may warrant panicking instead of swallowing the error with `None`.

nekevss: Totally agree with that. I was looking for other errors and happened upon someone mentioning that another syscall in `GetTimeZoneInformationForYear` can throw an oom error, but other than that I've found no reference to this call being fallible if the years are valid.

Also maybe this [doc](https://github.com/MicrosoftDocs/sdk-api/blob/docs/sdk-api-src/content/timezoneapi/nf-timezoneapi-gettimezoneinformationforyear.md) might be helpful

src/offset/local/mod.rs
```diff
@@ -186,6 +190,102 @@ impl TimeZone for Local {
...
+pub(crate) struct TzInfo {
+    std_offset: FixedOffset,
+    dst_offset: FixedOffset,
+    std_transition: Option<NaiveDateTime>,
+    dst_transition: Option<NaiveDateTime>,
+}
+
+#[cfg(any(unix, windows))]
+impl TzInfo {
+    fn lookup_with_dst_transitions(&self, dt: &NaiveDateTime) -> LocalResult<FixedOffset> {
```
pitdicker, line 203: It took me hours to convince myself the logic in this function was correct in the Unix version, so I simplified it a bit here.

If the first transition is to DST, the year starts in STD, transitions to DST, and ends with STD.
If the first transition is to STD, the opposite.

For every transition we take in local time the earliest clock value, and latest clock value (`transition_min` and `transition_max`).
- If the transition crates a gap, all times exclusive do not exist `(*_transition_min..*_transition_max)`.
- If the transition crates an overlap, all times inclusive do not exist `[*_transition_min..*_transition_max]`.
- The time in between the transitions should have the remaining comparison operator (i.e. `>` if the transition uses `<=`).

The result to return in `LocalResult::Single` is straightforward, `dst_offset` or `std_offset` as appropriate.

The correct order in `LocalResult::Ambiguous` is the offset right before the transition, then the offset right after.

Hope that helps for the review.

src/offset/local/tz_info/rule.rs
```diff
@@ -228,123 +230,32 @@ impl AlternateTime {
         }
     }
 
-    fn find_local_time_type_from_local(
+    fn find_local_offset_from_local(
```
jtmoon79, line 233: Can this function also have a 1:1 test case? `fn test_find_local_offset_from_local()` with many asserts?
I'd have more confidence if it did.

src/offset/local/windows.rs
```diff
@@ -8,158 +8,209 @@
...
-            SystemTimeToTzSpecificLocalTime(ptr::null(), utc_time, local.as_mut_ptr()),
-            0
-        )
-    };
-    // SAFETY: SystemTimeToTzSpecificLocalTime must have succeeded at this point, so we can
-    // assume the value is initialized.
-    Ok(unsafe { local.assume_init() })
-}
+#[test]
+fn verify_against_tz_specific_local_time_to_system_time() {
```
jtmoon79, line 145: Leave a code comment describing what this somewhat involved test case does.

jtmoon79 (changes requested):

pitdicker:
@jtmoon79 Thank you for the review again!

I am a slow learning when it comes to adding tests :rofl:. This will take some work.

src/offset/local/mod.rs
pitdicker: Added a test :+1:

src/offset/local/tz_info/rule.rs
pitdicker: This function is covered by the same test as `TzInfo::lookup_with_dst_transitions`

pitdicker:
Now that Windows supports for greater year ranges, it might overflow when near the limits of `NaiveDateTime`.
When adding a test for creating a `DateTime<Local>` with `NaiveDateTime::{MIN, MAX}` I hit multiple fun issues.
- It panicked on overflow instead of returning `LocalResult::None`.
- The `Debug` implementation of `DateTime` can panic (not something I want to fix in this PR).
- `FixedOffset` has the same issue.

This can be fixed inside the provided implementation of `TimeZone::from_local_datetime`, which I did.

But `Local` doesn't use that, the code is duplicated.
`Local` also does other things in a roundabout way, creating a `DateTime` just to extract the offset in `offset_from_local_datetime` for example. I would like to clean this up, but that would complicate this PR even more. I'll wait until after this is merged.

pitdicker:
> `Local` also does other things in a roundabout way, creating a `DateTime` just to extract the offset in `offset_from_local_datetime` for example. I would like to clean this up, but that would complicate this PR even more. I'll wait until after this is merged.

Thought better of this. I have removed the commit this fixes handling overflow when near the limits of `NaiveDateTime`.

I have added a commit to this PR removing all the roundabout stuff in the `local` module, including in `stub.rs`. In my opinion it is quite a bit more readable.

Now that all inner modules return offsets, any fix for overflow handling at the `TimeZone` trait level will automatically be picked up without merge conflicts.

pitdicker:
Split out the two commits that refactor `Local` to https://github.com/chronotope/chrono/pull/1072.

src/offset/fixed.rs
```diff
@@ -24,7 +24,7 @@ use crate::Timelike;
 #[derive(PartialEq, Eq, Hash, Copy, Clone)]
 #[cfg_attr(feature = "rkyv", derive(Archive, Deserialize, Serialize))]
 pub struct FixedOffset {
-    local_minus_utc: i32,
+    pub(crate) local_minus_utc: i32,
```
djc, line 27: This commit should be split in a number of smaller ones:

* Use the `offset()` method instead of directly accessing the `ut_offset` field
* Change `LocalResult` return values from `LocalTimeType` to offset only

Renaming any identifier should be a commit by itself.

pitdicker:
I have split the changes into a couple more commits, and updated them to fit with the developments on 0.4.x (`MSRV bump, `TryFrom` instead of `ToPrimitive`).

src/offset/local/tz_info/timezone.rs
```diff
@@ -540,39 +552,56 @@ impl fmt::Debug for TimeZoneName {
...
-    pub(super) ut_offset: i32,
+    ut_offset: FixedOffset,
     /// Daylight Saving Time indicator
     is_dst: bool,
     /// Time zone name
     name: Option<TimeZoneName>,
 }
 
+// Remove this hack once the MSRV >= 1.57
+macro_rules! const_unwrap {
```
jtmoon79, line 563: clever macro 🥷

jtmoon79 (approved):
LGTM

pitdicker:
After the refactor in https://github.com/chronotope/chrono/pull/1072 this PR is now 20% smaller. The rebase was pretty difficult though.

src/offset/local/tz_info/timezone.rs
```diff
@@ -547,48 +548,63 @@ impl fmt::Debug for TimeZoneName {
...
+    ($e:expr $(,)*) => {
+        match $e {
+            ::core::option::Option::Some(x) => x,
```
djc, line 574: Can we do without the `::core::option::Option` verbosity? This is just an internal macro so it doesn't seem like we need it.

src/offset/local/mod.rs
djc: Can you lay your explanation here down in a comment?

If this is created to replace existing logic for the Unix time zone logic, I think we should switch the Unix implementation in the same commit so that it's easier for me to compare the new code with the old code.

src/offset/local/mod.rs
```diff
@@ -147,6 +150,102 @@ impl TimeZone for Local {
     }
 }
 
+#[cfg(any(unix, windows))]
+pub(crate) struct TzInfo {
+    std_offset: FixedOffset,
+    dst_offset: FixedOffset,
+    std_transition: Option<NaiveDateTime>,
```
djc, line 157: Is `transition` start or end? Maybe there should be a nested type here, so that we can have `std` and `dst` fields of a type `OffsetChange` (name TBD) with `offset` and `start` fields?

src/datetime/tests.rs
```diff
@@ -873,6 +873,16 @@ fn test_datetime_from_local() {
     assert_eq!(datetime_west, datetime_utc.with_timezone(&timezone_west));
 }
 
+#[test]
+#[cfg(feature = "clock")]
+fn test_issue_651() {
```
djc, line 878: Please give this test a name that describes what it's testing for, and a comment/docstring containing a link to the issue.

src/offset/local/windows.rs
```diff
@@ -119,3 +119,82 @@ fn systemtime_to_naive_dt(st: SYSTEMTIME, year: i32) -> Option<NaiveDateTime> {
     };
     Some(date.and_time(time))
 }
+
+#[test]
```
djc, line 123: I'd like to have this squashed into the previous commit so that I can compare the test and the previous code more easily.

djc (commented):
This is looking good!

pitdicker:
One thing I forgot to check is the benchmark on Linux.
The logic in `lookup_with_dst_transitions` now works with `NaiveDateTime`s instead of simple Unix timestamps. Does that effect performance?

Before:
```
bench_get_local_time    time:   [91.464 ns 91.902 ns 92.373 ns]
Found 11 outliers among 100 measurements (11.00%)
  11 (11.00%) high mild
```

After:
```
bench_get_local_time    time:   [91.522 ns 91.786 ns 92.072 ns]
                        change: [-2.3806% -1.6025% -0.8457%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) high mild
  2 (2.00%) high severe
```

No interesting changes.

pitdicker:
During one of the rebases the test for `lookup_with_dst_transitions` disappeared. I have added it back. Sorry!

pitdicker:
I have found a way to make `lookup_with_dst_transitions` *much* simpler and able to work with an arbitrary number of transitions per year.
This would be needed to fix problem 1 in https://github.com/chronotope/chrono/issues/1153 (for a later PR), and happens to fix problem 3 in https://github.com/chronotope/chrono/issues/1153 by sorting the transition dates in UTC.

@djc The change is attached as a separate commit. Do you want me to squash it into the right commits here, or remove it and open a followup PR?

djc:
> @djc The change is attached as a separate commit. Do you want me to squash it into the right commits here, or remove it and open a followup PR?

How orthogonal is it to the other changes in this PR? Maybe it should go in a separate PR to land before this one?

pitdicker:
If you have not reviewed `lookup_with_dst_transitions` yet, these changes can make it easier. But otherwise I am fine with moving it to a follow-up PR.

src/offset/local/tz_info/timezone.rs
pitdicker: Can be removed now :tada:

pitdicker:
@djc I am not really happy with this PR. The Windows implementation that this started with can be quite simple and be written in 200 lines. And it adds an equal number of lines for tests.

But the Unix `tz_info` module is not all that integrated in chrono. It has its own types and logic that partly duplicate what chrono has. Changing it just enough so that we can have a shared lookup function for Windows and Unix makes this PR more difficult and leaves the Unix module in a state that does not feel better.

If I remove everything except the changes needed for Windows, would that help with review?
And we can refactor `tz_info` later?

src/offset/local/windows.rs
```diff
@@ -8,132 +8,195 @@
...
-    Ok(FixedOffset::east_opt(offset).unwrap())
 }
 
-fn from_local_time(local_time: SYSTEMTIME) -> Result<FixedOffset, Error> {
-    let utc_time = local_to_utc_time(&local_time)?;
-    let utc_secs = system_time_as_unix_seconds(&utc_time)?;
-    let local_secs = system_time_as_unix_seconds(&local_time)?;
-    let offset = (local_secs - utc_secs) as i32;
-    Ok(FixedOffset::east_opt(offset).unwrap())
+// The basis for Windows timezone and DST support has been in place since Windows 2000. It does not
```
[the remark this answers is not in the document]
ChrisDenton: All these APIs use the same underlying data so have the similar limitations. The data is stored in the registry; the structure of which is documented [here](https://learn.microsoft.com/en-us/windows/win32/api/timezoneapi/ns-timezoneapi-time_zone_information#remarks) and [here](https://learn.microsoft.com/en-us/windows/win32/api/timezoneapi/ns-timezoneapi-dynamic_time_zone_information#remarks). I can't find documentation on the 2004 limitation but that appears to be accurate based on my system (though it is possible to manually edit the registry to add more).

Modern versions of Windows do have [icu.dll](https://learn.microsoft.com/en-us/windows/win32/intl/international-components-for-unicode--icu-) which (amongst over things) contains ICU timezone data exposed via C APIs. This might be worth adding to Chrono at some point but even if so then they'll still need to be a fallback so that would be a whole other PR (imho).

ChrisDenton:
Hi @pitdicker, would you mind if I helped to move this PR along? I can throw up some commits addressing @djc's review comments which you can take or leave however you like?

pitdicker:
> Hi @pitdicker, would you mind if I helped to move this PR along? I can throw up some commits addressing @djc's review comments which you can take or leave however you like?

@ChrisDenton Thank you for helping out and pushing this! To be honest I have been completely out of it for almost 4 months, but should be back to working on chrono again. Sorry for not replying yesterday, before you did the work in https://github.com/chronotope/chrono/pull/1391.

What wasn't clear in this PR is that I have a couple of related branches that should make the code simpler and handle more cases, and I wanted to see what was salvageable first. I had to drop the last simplification (https://github.com/chronotope/chrono/pull/1017#issuecomment-1606137552) because of a simple mistake I couldn't find when rebasing for (https://github.com/chronotope/chrono/issues/1364#issuecomment-1828285543). And there was the idea to limit this PR to just the changes to Windows (https://github.com/chronotope/chrono/pull/1017#issuecomment-1675717385), as I have another branch that takes the cleanup on Unix further (https://github.com/pitdicker/chrono/tree/unix_tz_rule).

The latest review comments by @djc point out some of the parts I fixed in those later commits that were missing from this PR. All to say... It takes a little effort for me to get everything sorted again. But I'd like to take some of your commits from https://github.com/chronotope/chrono/pull/1391, thank you!

Just finished getting it to a reasonably clean base, but it needs a few more hours of work.

pitdicker:
@ChrisDenton Can I ask for your review when this is ready again? (out of Draft)

pitdicker (commented):
Added some small comments about what I want to look into.

ChrisDenton:
Thank you so much for working on this. The lasted update does simplify things and I find it much easier to follow.

src/offset/local/mod.rs
```diff
@@ -264,6 +346,119 @@ mod tests {
...
+
+        // offset stays the same
+        let std = FixedOffset::east_opt(3 * 60 * 60).unwrap();
+        let transitions = [
+            Transition::new(ymdhms(2023, 3, 26, 2, 0, 0), std, std),
+            Transition::new(ymdhms(2023, 10, 29, 3, 0, 0), std, std),
+        ];
+        compare_lookup(&transitions, 2023, 3, 26, 2, 0, 0, LocalResult::Single(std));
+        compare_lookup(&transitions, 2023, 10, 29, 3, 0, 0, LocalResult::Single(std));
+    }
```
jtmoon79, line 460: Is it worthwhile adding edge cases to this test function `test_lookup_with_dst_transitions`? i.e. should this also test transitions at or near `DateTime::MIN` and `DateTime::MAX` (or whatever the minimum and maximum should on Windows)?  At first glance, that looks important to me.

src/offset/local/win_bindings.rs
```diff
@@ -1,10 +1,30 @@
...
 
 #![allow(non_snake_case, non_upper_case_globals, non_camel_case_types, dead_code, clippy::all)]
+::windows_targets::link!("kernel32.dll" "system" fn GetTimeZoneInformationForYear(wyear : u16, pdtzi : *const DYNAMIC_TIME_ZONE_INFORMATION, ptzi : *mut TIME_ZONE_INFORMATION) -> BOOL);
 ::windows_targets::link!("kernel32.dll" "system" fn SystemTimeToFileTime(lpsystemtime : *const SYSTEMTIME, lpfiletime : *mut FILETIME) -> BOOL);
 ::windows_targets::link!("kernel32.dll" "system" fn SystemTimeToTzSpecificLocalTime(lptimezoneinformation : *const TIME_ZONE_INFORMATION, lpuniversaltime : *const SYSTEMTIME, lplocaltime : *mut SYSTEMTIME) -> BOOL);
 ::windows_targets::link!("kernel32.dll" "system" fn TzSpecificLocalTimeToSystemTime(lptimezoneinformation : *const TIME_ZONE_INFORMATION, lplocaltime : *const SYSTEMTIME, lpuniversaltime : *mut SYSTEMTIME) -> BOOL);
 pub type BOOL = i32;
+pub type BOOLEAN = u8;
+#[repr(C)]
+pub struct DYNAMIC_TIME_ZONE_INFORMATION {
```
jtmoon79, line 11: This struct seems to have dropped out of the sky.

I prefer to give the reader some help finding the source material for magic values and structures, e.g.

    /// struct from Windows Win32 API
    /// <https://learn.microsoft.com/en-us/windows/win32/api/timezoneapi/ns-timezoneapi-dynamic_time_zone_information>

src/offset/local/windows.rs
```diff
@@ -8,132 +8,238 @@
...
-        }
-    }
-}
-
-const HECTONANOSECS_IN_SEC: i64 = 10_000_000;
-const HECTONANOSEC_TO_UNIX_EPOCH: i64 = 11_644_473_600 * HECTONANOSECS_IN_SEC;
+use crate::offset::local::{lookup_with_dst_transitions, Transition};
+use crate::{Datelike, FixedOffset, LocalResult, NaiveDate, NaiveDateTime, NaiveTime, Weekday};
 
+// We don't use `SystemTimeToTzSpecificLocalTime` because it doesn't support the same range of dates
```
jtmoon79, line 20: This is a great comment.

src/offset/local/windows.rs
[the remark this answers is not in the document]
jtmoon79: Great info @ChrisDenton .  I suggest placing your reply into the code as a code comment including the links.

src/offset/local/win_bindings.rs
pitdicker: > This struct seems to have dropped out of the sky.

Quite close, the entire file gets generated by windows-bindgen.

src/offset/local/windows.rs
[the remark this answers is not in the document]
pitdicker: > Do we have references for this? Does this PR make our timezone support worse than what we had before? Seems like a pretty limited implementation...

It is very limited, yes. But it is what the `Local` time on Windows PCs is. We are not getting worse in any way, just using the same underlying data as @ChrisDenton says.

I did have some more references but have misplaced them...

Note that the details here don't matter all that much. `GetTimeZoneInformationForYear` just provides a couple of offsets and transition dates for a particular year, and we have to look up where a datetime fails between the transitions.

pitdicker:
I think the logic is pretty much what I would want it to be. The logic in `lookup_with_dst_transitions` is reusable on Unix, I have another branch to do so (but want to keep this one focused on just Windows).

If anyone wants to help out with adding some tests I would deeply appreciate it. But otherwise I'll get to it.
I think it would be nice to test `lookup_with_dst_transitions` with just one transition date, and with three.
And to add a pretty artificial test for `lookup_with_dst_transitions` with transitions right around midnight at `NaiveDateTime::MIN|MAX`.

pitdicker:
Added tests for `lookup_with_dst_transitions` with transitions right around midnight at `NaiveDateTime::MIN|MAX`, and with a single transitions. Why did I put this off, only an hour work? Anyway ready for review.

src/offset/local/windows.rs
```diff
@@ -8,132 +8,253 @@
...
-fn from_utc_time(utc_time: SYSTEMTIME) -> Result<FixedOffset, Error> {
-    let local_time = utc_to_local_time(&utc_time)?;
-    let utc_secs = system_time_as_unix_seconds(&utc_time)?;
-    let local_secs = system_time_as_unix_seconds(&local_time)?;
-    let offset = (local_secs - utc_secs) as i32;
-    Ok(FixedOffset::east_opt(offset).unwrap())
+// The basis for Windows timezone and DST support has been in place since Windows 2000. It does not
+// allow for complex rules like the IANA timezone database:
+// - A timezone has the same base offset the whole year.
+// - There some to be either zero or two DST transitions (but we support having just one).
```
djc, line 113: "There some to be"?

src/offset/local/windows.rs
```diff
@@ -8,132 +8,253 @@
...
+    dst_transition: Option<NaiveDateTime>,
 }
 
-fn from_local_time(local_time: SYSTEMTIME) -> Result<FixedOffset, Error> {
-    let utc_time = local_to_utc_time(&local_time)?;
-    let utc_secs = system_time_as_unix_seconds(&utc_time)?;
-    let local_secs = system_time_as_unix_seconds(&local_time)?;
-    let offset = (local_secs - utc_secs) as i32;
-    Ok(FixedOffset::east_opt(offset).unwrap())
+fn tz_info_for_year(year: i32) -> Option<TzInfo> {
```
djc, line 130: Let's make this a method? `TzInfo::for_year()`?

src/offset/local/windows.rs
```diff
@@ -8,132 +8,253 @@
...
-        wDay: dt.day() as u16,
-        // Valid values: 0-23
-        wHour: dt.hour() as u16,
-        // Valid values: 0-59
-        wMinute: dt.minute() as u16,
-        // Valid values: 0-59
-        wSecond: dt.second() as u16,
-        // Valid values: 0-999
-        wMilliseconds: 0,
+fn systemtime_to_naive_dt(st: SYSTEMTIME, year: i32) -> Option<NaiveDateTime> {
```
djc, line 151: I think I preferred the old name? (And don't see a strong reason to change it.)

src/offset/local/windows.rs
```diff
@@ -8,132 +8,253 @@
...
+    let time = NaiveTime::from_hms_milli_opt(
+        st.wHour as u32,
+        st.wMinute as u32,
+        st.wSecond as u32,
+        st.wMilliseconds as u32,
+    )?;
+    // In Chrono's Weekday, Monday is 0 whereas in SYSTEMTIME Monday is 1 and Sunday is 0.
+    // Therefore we move back one day after converting the u16 value to a Weekday.
+    let day_of_week = Weekday::try_from(u8::try_from(st.wDayOfWeek).ok()?).ok()?.pred();
+    let date = if st.wYear == 0 {
```
djc, line 164: Invert this condition so we can return early?

src/offset/local/windows.rs
```diff
@@ -8,132 +8,253 @@
...
-    unsafe { windows_sys_call!(SystemTimeToFileTime(st, init.as_mut_ptr()), 0) }
-    // SystemTimeToFileTime must have succeeded at this point, so we can assum the value is
-    // initalized.
-    let filetime = unsafe { init.assume_init() };
-    let bit_shift = ((filetime.dwHighDateTime as u64) << 32) | (filetime.dwLowDateTime as u64);
-    let unix_secs = (bit_shift as i64 - HECTONANOSEC_TO_UNIX_EPOCH) / HECTONANOSECS_IN_SEC;
-    Ok(unix_secs)
+    #[test]
+    fn verify_against_tz_specific_local_time_to_system_time() {
+        // The implementation in Windows itself is the source of truth on how to work with the OS
```
djc, line 192: Nice!

src/offset/local/mod.rs
```diff
@@ -183,11 +186,95 @@ impl TimeZone for Local {
     }
 }
 
+#[cfg(windows)]
+#[derive(Copy, Clone, Eq, PartialEq)]
+struct Transition {
```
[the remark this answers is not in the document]
djc: What's the rationale for keeping these in `local/mod.rs` when they're Windows-specific?

src/offset/local/mod.rs
```diff
@@ -293,7 +293,7 @@ mod tests {
     #[test]
     fn verify_correct_offsets_distant_past() {
         // let distant_past = Local::now() - Duration::days(365 * 100);
-        let distant_past = Local::now() - Duration::days(250 * 31);
+        let distant_past = Local::now() - Duration::days(365 * 500);
```
djc, line 296: Nit, commit message: "Extent" -> "Extend"

src/datetime/tests.rs
```diff
@@ -1265,6 +1265,17 @@ fn test_datetime_from_timestamp_millis() {
     );
 }
 
+#[test]
+#[cfg(feature = "clock")]
+fn test_datetime_before_windows_api_limits() {
+    // dt corresponds to `FILETIME = 147221225472` from issue 651.
```
djc, line 1271: Can you refer to the issue by making it a link?

src/offset/local/windows.rs
pitdicker: We didn't have anything like this function before. I don't mind changing the name to something else.

src/offset/local/mod.rs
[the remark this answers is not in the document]
pitdicker: After this is merged I would like to open a PR with the refactoring of the Unix side that use these. (May not be right the same week though.)

src/offset/local/windows.rs
pitdicker: Done.

src/offset/local/windows.rs
```diff
@@ -8,132 +8,255 @@
...
-        // Valid values: 0-23
-        wHour: dt.hour() as u16,
-        // Valid values: 0-59
-        wMinute: dt.minute() as u16,
-        // Valid values: 0-59
-        wSecond: dt.second() as u16,
-        // Valid values: 0-999
-        wMilliseconds: 0,
+impl TzInfo {
+    fn tz_info_for_year(year: i32) -> Option<TzInfo> {
```
djc, line 131: As suggested previously, let's call this `for_year()` -- no need to duplicate the type name in the method name.

djc (approved):

src/offset/local/windows.rs
pitdicker: Sorry, only now I get your intention. Yes, that is better.

pitdicker:
# :tada:

Happy to see this completed. Thank you @djc, @ChrisDenton and others!

PR #1019, opened 2023-04-17
Add an example to Weekday

pitdicker:
Using `Weekday` was not convenient at all until I noticed the `FromPrimitive` implementation. Added this and some more examples to the type documentation.


djc (approved):
We should probably add `TryFrom` impls for `Weekday` and other enums like it (I believe we have these on main already), and `#[deprecate]` the `FromPrimitive` impls.

Issue #551, opened 2021-04-07
Inconsistency in `chrono::format::Parsed::to_naive_date` for instances with week numbers


weiznich:
The `chrono::format::Parsed::to_naive_date` method returns depending on the year (e.g. 2002) a `ParseError(Impossible)` even if the fields `year`, `week_from_mon` and `weekday` are correctly populated as described in the [documentation](https://docs.rs/chrono/0.4.19/chrono/format/struct.Parsed.html#method.to_naive_date) of the corresponding method. This does not happen for all years, for example setting the year to 2001 yields a valid date.

See [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e897f109e3810c242a8b2cb93a7dd58b) for a minimal example.

pitdicker:
Fixed by https://github.com/chronotope/chrono/pull/966.

PR #1020, opened 2023-04-17
Add test for issue 551

pitdicker:
Issue https://github.com/chronotope/chrono/issues/551 was fixed by https://github.com/chronotope/chrono/pull/966.


djc:
Thanks!

Issue #388, opened 2020-01-20
Implement `NaiveDateTime::with_timezone`?


dario23:
hi,

i noticed that for `NaiveDateTime`, the `with_timezone` method is missing; where the `Naive` types are where i think i need this functionality most, to add a timezone i know from somewhere else to a timezone-less value i just parsed.

was this forgotten, or is there a reason this method cannot/should not exist? happy to open a PR if that has any chance of getting accepted :-)

hei-pa:
Yes I need something like this too.

I have parsed a String to a `NaiveDateTime`, now my timezone comes from somewhere else and I need to contruct a `DateTime` out of this `NaiveDateTime`. It would be great to have a `with_timezone(x)` function on `NaiveDateTime` with a `DateTime` as result.

pitdicker:
Added `with_timezone` to `NaiveDateTime` in https://github.com/chronotope/chrono/pull/1021.

pitdicker:
https://github.com/chronotope/chrono/pull/1021 was reverted in https://github.com/chronotope/chrono/pull/1029 because we already have this functionality as [`NaiveDateTime::and_local_timezone`](https://docs.rs/chrono/latest/chrono/naive/struct.NaiveDateTime.html#method.and_local_timezone).

PR #1021, opened 2023-04-18
Add with_timezone to NaiveDateTime

pitdicker:
As requested in https://github.com/chronotope/chrono/issues/388.

Combining a `NaiveDateTime` and `TimeZone` is already possible with `TimeZone::from_local_datetime`. But it makes sense to me to add the equivalent method to `NaiveDateTime`.

Note that `DateTime::with_timezone` already exists, this gives `NaiveDateTime` a similar method.

Closes https://github.com/chronotope/chrono/issues/388.

### DIFF
--- a/src/naive/datetime/mod.rs
+++ b/src/naive/datetime/mod.rs
@@ -554,6 +554,28 @@ impl NaiveDateTime {
         self.time.nanosecond()
     }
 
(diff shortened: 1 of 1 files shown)


djc (approved):
Thanks!

Issue #1010, opened 2023-04-05
Some panic cases found by afl.rs


Koral77:
I've using [afl.rs](https://github.com/rust-fuzz/afl.rs) to fuzz this crate. And I've found several APIs may panic due to arithmetic overflow. 

The code to replay these panics are
```
let _local0 = chrono::naive::NaiveDateTime::from_timestamp_opt(-4227854320, 1678774288);
let _local1 = chrono::Duration::microseconds(-7019067213869040);
let _local2_param0_helper1 = _local0.unwrap();
chrono::DurationRound::duration_trunc(_local2_param0_helper1, _local1);
```
```
let _local0 = chrono::naive::NaiveDateTime::from_timestamp_opt(320041586, 1920103021);
let _local1 = chrono::Duration::nanoseconds(-8923838508697114584);
let _local2_param0_helper1 = _local0.unwrap();
chrono::DurationRound::duration_round(_local2_param0_helper1, _local1);
```
```
let _local0 = chrono::naive::NaiveDateTime::from_timestamp_opt(-2621440, 0);
let _local1 = chrono::Duration::nanoseconds(-9223372036854771421);
let _local2_param0_helper1 = _local0.unwrap();
chrono::DurationRound::duration_round(_local2_param0_helper1, _local1);
```
```
let _local0 = chrono::naive::NaiveDateTime::from_timestamp_opt(-502509993984, 64);
let _local1_param0_helper1 = _local0.unwrap();
chrono::Datelike::with_ordinal0(&(_local1_param0_helper1), 4294967295);
```
```
let _local0 = chrono::naive::NaiveDateTime::from_timestamp_opt(-754576364, 336909572);
let _local1_param0_helper1 = _local0.unwrap();
chrono::Datelike::with_day0(&(_local1_param0_helper1), 4294967295);
```
```
let _local0 = chrono::naive::NaiveDateTime::from_timestamp(-8377300, 742391807);
chrono::Datelike::with_month0(&(_local0), 4294967295);
chrono::Datelike::with_month(&(_local0), 4294967295);
```
```
let _local0 = chrono::naive::NaiveDateTime::from_timestamp(-11676614656, 15282199);
chrono::naive::NaiveDateTime::timestamp_nanos(&(&_local0));
```

Besides, I've found an not a char boundary error. The code to replay the panic is
```
[...]

PR #1024, opened 2023-04-18
Prevent string slicing inside char boundaries

pitdicker:
This takes care of the problematic string slicing found in https://github.com/chronotope/chrono/issues/1010 replay 4.

I believe all string slicing on arbitrary boundaries in this module now happen on `u8` slices.

### DIFF
--- a/src/format/parse.rs
+++ b/src/format/parse.rs
@@ -997,3 +997,11 @@ fn test_rfc3339() {
         }
     }
 }
+
+#[cfg(test)]
+#[test]
+fn test_issue_1010() {
+    let dt = crate::NaiveDateTime::parse_from_str("\u{c}SUN\u{e}\u{3000}\0m@J\u{3000}\0\u{3000}\0m\u{c}!\u{c}\u{b}\u{c}\u{c}\u{c}\u{c}%A\u{c}\u{b}\0SU\u{c}\u{c}",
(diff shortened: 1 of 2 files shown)


pitdicker:
The failed CI run is unrelated to this PR.

djc:
Can you just explain why this fixes it? It's not obvious to me why we even get the incorrect boundaries in there.

djc:
Don't we know up front that the "possible match" is up to a valid character boundary?

pitdicker:
Not with the way the code currently works.

For example https://github.com/chronotope/chrono/blob/main/src/format/scan.rs#L167:
```rust
pub(super) fn short_or_long_weekday(s: &str) -> ParseResult<(&str, Weekday)> {
    // lowercased weekday names, minus first three chars
    static LONG_WEEKDAY_SUFFIXES: [&str; 7] =
        ["day", "sday", "nesday", "rsday", "day", "urday", "day"];

    let (mut s, weekday) = short_weekday(s)?;

    // tries to consume the suffix if possible
    let suffix = LONG_WEEKDAY_SUFFIXES[weekday.num_days_from_monday() as usize];
    if s.len() >= suffix.len() && equals(&s[..suffix.len()], suffix) {
        s = &s[suffix.len()..];
    }

    Ok((s, weekday))
}
```

This will take up to 7 times a slice out of `s`, with the lengths in bytes of 3, 4, 6, 5, etc. for the possible matches in `LONG_WEEKDAY_SUFFIXES`. To say it differently, the length of the slice is not determined by the input `s`, but by the length of the possible match.

djc:
Thanks!

PR #1026, opened 2023-04-18
Apply comments from MSRV bump

pitdicker:
Apply three comments that could be done when the MSRV was increased to 1.48.

I am not sure about applying `#[non_exhaustive]` to `ParseErrorKind` instead of a `__Nonexhaustive` variant. Would that be a breaking change?

Or would this be the time to increase the MSRV further?

### DIFF
--- a/src/format/mod.rs
+++ b/src/format/mod.rs
@@ -386,7 +386,7 @@ pub enum ParseErrorKind {
     /// There was an error on the formatting string, or there were non-supported formating items.
     BadFormat,
 
-    // TODO: Change this to `#[non_exhaustive]` (on the enum) when MSRV is increased
+    // TODO: Change this to `#[non_exhaustive]` (on the enum) with the next breaking release.
     #[doc(hidden)]
     __Nonexhaustive,
 }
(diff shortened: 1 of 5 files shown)


src/format/mod.rs
```diff
@@ -357,6 +357,7 @@ impl ParseError {
 
 /// The category of parse error
 #[derive(Debug, Clone, PartialEq, Eq, Copy, Hash)]
+#[non_exhaustive]
```
djc, line 360: Can you have a look at the other uses of `non_exhaustive` on main, and backport them to 0.4.x?

djc (approved):
Good stuff!

src/format/mod.rs
pitdicker: There is only one other case, on `Parsed`. But I have my doubts on that one, because of [this comment](https://github.com/chronotope/chrono/blob/0.4.x/src/format/parsed.rs#L108):
```rust
    /// A dummy field to make this type not fully destructible (required for API stability).
    _dummy: (),
```

pitdicker: Sorry, reading it again the comment is just another way of saying the same thing...

djc: Are you worried that `non_exhaustive` doesn't prevent destruction? I think it does...

pitdicker:
:+1: Updated the comment.

Issue #1022, opened 2023-04-18
How to create local date time with zero hour/minute/second


DoumanAsh:
Due to tons of unecessary deprecations it became impossible to use this library for our use case which is:

1. Get local date ('chrono::Local::today()`)
2. Set hour/minute/second to 0:0:0 (`and_hms(0, 0, 0)`)

There is no need to deprecate useful things unless you have alternative for their usage?

`chrono::Local::today` cannot be replaced with `chrono::Local::now` as it is.
It returns two completely different things

Something simple as creating midnight timestamp should be error-less, yet you force user unwrapping all these infallible operations because you cannot even provide const constructors

pitdicker:
Ai, I can understand your frustration. That was non-trivial.

```rust
use chrono::{Local, TimeZone};

fn main() {
    let today_date_naive = Local::now().date_naive();
    let midnight_naive = today_date_naive.and_hms_opt(0, 0, 0).unwrap();
    let midnight_local = Local.from_local_datetime(&midnight_naive).unwrap();
    println!("{:?}", midnight_local);
}
```

The deprecation of methods that are infallible as long as you pass values that are in range (like `and_hms`) is something that I as a user also don't really like, but that is another issue.

It is maybe interesting to know if you wanted a time 1 minute after midnight today this program could have panicked in some locales, because it is not uncommon to have a DST transition at midnight. Than that time wouldn't exist.

I wonder if a `set_time` method on `DateTime` would be possible. Than maybe something like `Local::now().set_time(time)` could work.

DoumanAsh:
> It is maybe interesting to know that if you wanted a time 1 minute after midnight today this program could have panicked in some locales, because it is not uncommon to have a DST transition at midnight. Than that time wouldn't exist.

I'm not sure how is that possible?
Just because DST exist, doesn't make `00:01` impossible time I think.
Every timezone exists within range of 0 to 24 hours etc

pitdicker:
Just for fun, may I suggest you run the following code on a Unix system with the date and timezone set to 2022-03-25 Damascus, Syria (to pick an example)?
```rust
use chrono::{Local, TimeZone};

fn main() {
    let today_date_naive = Local::now().date_naive();
    let midnight_naive = today_date_naive.and_hms_opt(0, 0, 1).unwrap();
    let midnight_local = Local.from_local_datetime(&midnight_naive).unwrap();
    println!("{:?}", midnight_local);
}
```

That will fail with:
```
[...]

PR #1027, opened 2023-04-18
Add a `set_time` method to `DateTime`

pitdicker:
Example of where this would be useful: https://github.com/chronotope/chrono/issues/1022.

Workarounds such as substracting the current time would not be correct.
You have to extract the current date as `NaiveDate`, add the desired time to get a `NaiveDateTime`, extract the original timezone, and construct a new `DateTime`.

Because it is definitely non-trivial to figure out, I propose to add this method.

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -33,7 +33,7 @@ use crate::oldtime::Duration as OldDuration;
 #[allow(deprecated)]
 use crate::Date;
 use crate::Months;
-use crate::{Datelike, Timelike, Weekday};
+use crate::{Datelike, LocalResult, Timelike, Weekday};
 
 #[cfg(feature = "rkyv")]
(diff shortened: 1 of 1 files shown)


pitdicker:
Thinking about this a day later, it feels a bit like a hack. Or at least, incomplete.
When I think about `DateTime`, I see it as a combination of `NaiveDate`, `NaiveTime`, and something implementing `TimeZone` . It is already possible to swap out the timezone with `with_timezone`. This PR would add the ability to replace the time part. With a `set_date` method the set would be complete.

But I am starting to learn that with chrono you should have the mindset: when you want to do something non-obvious with `DateTime`, just drop down to its components such as `NaiveDate` and `TimeZone`, change what is needed, and recombine them.

Feel free to close this PR if you agree.

djc:
Yeah, I think that makes more sense.

PR #1028, opened 2023-04-19
Fix tests on AIX

ecnelises:
- Add AIX timezone database path to search list.
- Use date program from freeware to do the testing.

Unfortunately AIX has not been supported by CI, but after this fix, all tests pass on AIX.


pitdicker (approved):
LGTM

PR #1033, opened 2023-04-22
add cargo clean

esheppa:


djc:
What are you trying to do here?

esheppa:
Should have put more detailed comments, updating now. Just trying to figure out why the MSRV check is failing in `0.4.x` CI. Should be solved now

esheppa:
It looks like only `serde_derive` needs to be locked at a specific version here, but `serde_json` may move to use that version at some point

LingMan:
Wouldn't it be better to `cargo update -p serde_derive --precise 1.0.156` in the MSRV job? That would fix CI without forcing everyone to use old versions of serde. You could even re-enable the serde feature in the test.

jtmoon79 (approved):
@esheppa are you still working on this PR?  If so, should it be put back to a "Draft"?

Otherwise, LGTM. (Full Disclosure: I didn't look into the versioning in-depth, just taking OPs comments at face value)

esheppa:
@jtmoon79 yes still keen to get this merged if the CI failures are still happening as its a bit awkward to merge with any failed tests.

@LingMan we could also do this, however these are only for dev dependencies so we are not forcing anyone to use old versions unless they are running the tests locally. One risk with the `update` approach is that a test might make use of a feature not available in those versions, while this would immediately be caught by CI, it is less ergonomic. 

That said, re-enabling the serde feature is a good idea. I've added in 07749dc

pitdicker:
Now that the MSRV is raised to 1.56 in https://github.com/chronotope/chrono/pull/1053 we would no longer need a workaround.

esheppa:
Thanks - looks like this is not needed which is ideal!

Issue #544, opened 2021-03-11
chrono assumes TimeZone's offset's debug format is valid RFC3339


nkovacs:
I originally opened [this issue](https://github.com/kennytm/tzfile/issues/2) in tzfile, but I think it might be a chrono issue.

[Chrono assumes that the debug format of a TimeZone's offset will be valid RFC3339](https://docs.rs/chrono/0.4.19/src/chrono/datetime.rs.html#2066), but this isn't the case for tzfile's ArcTz. Would it not make more sense to use to_rfc3339, or to use offset.fix()?

pitdicker:
Fixed in https://github.com/chronotope/chrono/pull/1035.

PR #1035, opened 2023-04-24
Do not use Offset's Debug impl when serializing DateTime

pitdicker:
The `Debug` implementation of `Offset` may not be valid RFC3339. For real timezones it makes sense for `Offset` to the print the name of the timezone, like `tz_file` and `chrono_tz` do.

So the `Serialize` implementation of `DateTime` should not rely on the `Debug` implementation of `Offset`, but convert the `Offset` first to a `FixedOffset`.

Fixes https://github.com/chronotope/chrono/issues/544.

### DIFF
--- a/src/datetime/serde.rs
+++ b/src/datetime/serde.rs
@@ -3,11 +3,12 @@
 use core::fmt;
 use serde::{de, ser};
 
-use super::DateTime;
+use super::{DateTime, SecondsFormat};
+use crate::format::write_rfc3339;
 use crate::naive::datetime::serde::serde_from;
 #[cfg(feature = "clock")]
 use crate::offset::Local;
-use crate::offset::{FixedOffset, TimeZone, Utc};
+use crate::offset::{FixedOffset, Offset, TimeZone, Utc};
 
 #[doc(hidden)]
 #[derive(Debug)]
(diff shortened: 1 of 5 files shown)


src/datetime/serde.rs
```diff
@@ -25,7 +26,9 @@ pub struct MicroSecondsTimestampVisitor;
 #[derive(Debug)]
 pub struct MilliSecondsTimestampVisitor;
 
-/// Serialize into a rfc3339 time string
+/// Serialize into a rfc3339 time string.
+/// In addition to rfc3339 the offset may contain seconds when present, instead of only hours and
+/// minutes.
```
pitdicker, line 31: This documents the current behavior, that I had to re-implement to make a test pass.

But deserialization is unable to read on offset with seconds. What should we do here?
Being able to round-trip is worth something, writing correct rfc3339 time strings is also good.

pitdicker: I opened https://github.com/chronotope/chrono/pull/1036 with a proposal to deny offsets with seconds, which should never exist anyway.

jtmoon79 (approved):
LGTM

pitdicker:
@jtmoon79 Thank you for all the review work!

pitdicker:
This currently fails on Ci, because the logic for serde should be replicated for rustc_serialize.

I don't really want to do that here, but want to extend the formatting capabilities of chrono for offsets a bit. Wanting to format `+00:00` as `Z` seems pretty useful in general. Once that works out I'll continue this PR.

pitdicker:
The formatting changes I would like to use are part of https://github.com/chronotope/chrono/pull/1082.

pitdicker:
With https://github.com/chronotope/chrono/pull/1200 this would *finally* be easy to fix.

djc (approved):

PR #1040, opened 2023-04-27
Documentation for `DateTime` and `TimeZone`

pitdicker:
If the general direction of this documentation seems useful, I can write it for more types.

My idea is to give a tour of the available methods on a type, similar to [`std::Option`](https://doc.rust-lang.org/std/option/index.html).

Useful examples could also be a plus, but this is just a start.


src/datetime/mod.rs
```diff
@@ -79,11 +79,141 @@ pub enum SecondsFormat {
...
+///   - [`TimeZone::with_ymd_and_hms`]
+///
+/// ## From the current time
+/// (Requires the default `clock` feature.)
+/// - [`Local.now()`](offset/struct.Local.html#method.now)
+/// - [`Utc.now()`](offset/struct.Utc.html#method.now)
+///
+/// ## Parsing
+/// - [`parse_from_str`](#method.parse_from_str) for parsing with a format string.
+/// - [`parse_from_rfc2822`](#method.parse_from_rfc2822) for parsing a string such as
```
jtmoon79, line 134: A precise link to the official RFC will lend trustworthiness to these docs.

```rust
/// ... for parsing an [RFC 2822] compliant string such as ...
///
/// [RFC 2822]: https://www.rfc-editor.org/rfc/rfc2822.html#section-3.3
```

src/datetime/mod.rs
```diff
@@ -79,11 +79,141 @@ pub enum SecondsFormat {
...
+///   `Tue, 1 Jul 2003 10:52:37 +0200` to a [`DateTime<FixedOffset>`].
+/// - [`parse_from_rfc3339`](#method.parse_from_rfc3339) for parsing a string such as
```
jtmoon79, line 136: A precise link to the official RFC will lend trustworthiness to these docs.

```rust
/// ... for parsing an [RFC 3339] compliant string such as
///
/// [RFC 3339]: https://www.rfc-editor.org/rfc/rfc3339.html#section-5.6
```

src/datetime/mod.rs
```diff
@@ -79,11 +79,141 @@ pub enum SecondsFormat {
...
+/// - [`TimeZone::datetime_from_str`] parsing from a string without timezone information
+///   (or with an offset that must match that of the timezone).
+///
+/// ## From common external types:
+/// - Combine a [`TimeZone`] with a UTC Unix timestamp:
+///   - [`TimeZone::timestamp_opt`]
+///   - [`TimeZone::timestamp_millis_opt`]
+///   - [`TimeZone::timestamp_nanos`]
+/// - From the standard library [`SystemTime`](
+///   https://doc.rust-lang.org/std/time/struct.SystemTime.html) (requires the default `std`
```
jtmoon79, line 147: I prefer separate links. The raw source docs have cleaner widths and are easier to read/verify.  Also, std library things don't need leading `http://doc....`.

```rust
/// From the standard library [`SystemTime`]:
///
/// [`SystemTime`]: std::time::SystemTime
```

src/datetime/mod.rs
```diff
@@ -79,11 +79,141 @@ pub enum SecondsFormat {
...
+/// - [`Local`] matches the current timezone settings of the operating system. That may however be a
+///   simplified implementation, not describing all complexities like the timezone database does.
+/// - [`FixedOffset`] stores only the offset from UTC of a specific datetime, without knowing the
+///   timezone. This is the preferred type for an end result, a timestamp.
+/// - [`Utc`] is the universal time, carrying no information about a local time at all. It is
+///   continuous, without gaps or ambiguities.
+///
+/// # Creating a `DateTime`
+///
+/// ## Combine from primitive types
```
jtmoon79, line 115: Markdown linters will request empty newline after all headers

src/datetime/mod.rs
```diff
@@ -79,11 +79,141 @@ pub enum SecondsFormat {
...
+/// - If the subsecond part should be included in the timestamp (reducing the range of available
+///   years), consider:
+///   - [`timestamp_millis`](#method.timestamp_millis)
+///   - [`timestamp_micros`](#method.timestamp_micros)
+///   - [`timestamp_nanos`](#method.timestamp_nanos)
+///
+/// ## Into the standard library `SystemTime`
+/// (Requires the default `std` feature.)
+///   - [`SystemTime::from`](#impl-From<DateTime<Tz>>-for-SystemTime) to convert to a UTC
+///     [`SystemTime`](https://doc.rust-lang.org/std/time/struct.SystemTime.html).
```
jtmoon79, line 215: Use separate links, then the link only needs to be decalred once. e.g.

```rust
/// ... blah [`SystemTime`] blah ...
/// ...
/// [`SystemTime`] blah ...
///
/// [`SystemTime`]: std::time::SystemTime
```

jtmoon79 (approved):
This is great!

Feedback is recommended but not required.

src/datetime/mod.rs
pitdicker: Excellent idea.

src/datetime/mod.rs
pitdicker: Those exist? 😆

pitdicker:
Thank you for the good suggestions!

I will add new commits to address review comments, and can squash them before later.

jtmoon79 (approved):

src/datetime/mod.rs
```diff
@@ -79,11 +79,141 @@ pub enum SecondsFormat {
     __NonExhaustive,
 }
 
-/// ISO 8601 combined date and time with time zone.
+/// `DateTime` is the preferred type for working with real-world timestamps. It contains all the
```
djc, line 82: For best presentation in rustdoc, the first line should be a high-level, self-contained description of the type. I usually prefer not to end that line with a period, since it's usually more like a sentence fragment than a full sentence. Add an empty line after that, then start paragraphs of more detailed documentation.

src/datetime/mod.rs
```diff
@@ -79,11 +79,141 @@ pub enum SecondsFormat {
...
+/// information needed to exactly specify a point in time, and to do calculations on it: a date,
```
djc, line 83: I'm not sure it is correct to say that `DateTime` "contains" a "timezone". It contains the offset, but not the timezone itself. Maybe that can be worded differently?

src/offset/mod.rs
```diff
@@ -246,7 +246,8 @@ pub trait Offset: Sized + Clone + fmt::Debug {
 ///
 /// ## Example
 ///
-/// ```
+#[cfg_attr(not(feature = "clock"), doc = "```ignore")]
```
djc, line 249: This seems unrelated to the commit message.

src/datetime/mod.rs
```diff
@@ -174,14 +174,19 @@ pub enum SecondsFormat {
 /// the time.
 ///
 /// Timezone-aware calculations on a [`DateTime`] can be done with:
-/// - [`Sub<DateTime>`] to get a [`Duration`](struct.Duration.html) with the difference between two
-///   datetimes.
-/// - [`Add<Duration>`] to add a [`Duration`](struct.Duration.html) of an absolute number of
+/// - [`signed_duration_since`](#method.signed_duration_since) to get a [`Duration`](
```
djc, line 177: I think we should mention both the operator and the checked methods here.

src/datetime/mod.rs
pitdicker: I am not sure how to word it better. Changed it to "and associated time zone" for now. But the paragraph below explains that the time zone is tracked by the type system.

src/datetime/mod.rs
pitdicker: The line below says "The same functionality is also available with implementations of `Add` and `Sub`."
I initially made those more prominent, but they are not a great idea to use because they can panic in pretty common scenarios.

To make them a bit more visible I added an example of working with `DateTimes`.

src/datetime/mod.rs
djc: As I explained, my beef is with the word "contains". The next paragraph goes into a lot more detail, so I think we should get this first, high-level description right.

pitdicker:
*Finally* updated. Sorry to let this linger for 3+ months.

PR #1041, opened 2023-04-27
Use correct offset in conversion from `Local` to `FixedOffset`

pitdicker:
I can't see any good reason to forget the local offset when converting from `DateTime<Local>` to `DateTime<FixedOffset>`.
Seems to be an oversight in the initial implementation https://github.com/chronotope/chrono/pull/271.


jtmoon79 (approved):
Great catch!

pitdicker:
:+1: Added a test.

PR #1042, opened 2023-04-28
Preserve `-00:00` offset

pitdicker:
RFC 3339 and RFC 2822 make a distinction between an offset of `+00:00` and `-00:00`.

RFC 2822:
> The form "+0000" SHOULD be used to indicate a time zone at Universal Time. Though "-0000" also indicates Universal Time, it is used to indicate that the time was generated on a system that may be in a local time zone other than Universal Time and therefore indicates that the date-time contains no information about the local time zone.

From RFC3339:
> If the time in UTC is known, but the offset to local time is unknown, this can be represented with an offset of "-00:00". This differs semantically from an offset of "Z" or "+00:00", which imply that UTC is the preferred reference point for the specified time.

Currently chrono drops the distinction, but I would like to preserve it.

- I noticed `format/scan.rs` was already written with the functionality of this RFCs in mind, but did not go all the way. It now follows the specs better and can sometimes give better parser errors.
- Modifying `Parsed` to preserve this information is a bit messy as all the methods are public. Adding a public constant `NO_OFFSET_INFO` seemed the cleanest solution. Parsing doesn't support more than two digits for the offset hours, limiting the values that are passed to `Parsed::set_offset` to `99 * 3600 + 59 * 60 = 359940`. The value of `NO_OFFSET_INFO` is way outside that range, so no chance for conflicts.
- `FixedOffset` can be modified in a reasonably clean way to encode `-0`: shift the value 2 bits left and encode this special state as an `OFFSET_UNKNOWN` flag in the rightmost bits. Advantage of this encoding is that it only takes a single shift right to get the offset in seconds, so this should not measurably impact performance.
- The same shifting trick enables using `NonZeroI32`, as long as we set one of the rightmost bits to something non-null with the `OFFSET_NORMAL` flag. This gives `FixedOffset` a niche. `Option<DateTime<FixedOffset>>` is now the same size as `DateTime<FixedOffset>`: 16 bytes instead of 20.
- For comparisons and hashing the flags are ignored. This way `-00:00` is treated as equal to `+00:00`, which is usually what is expected. Also this way we remain backwards compatible.
- `FixedOffset::no_offset_info` returns whether the offset is `-00:00`.
- `FixedOffset::OFFSET_UNKNOWN` can be used to initialize an offset to `-00:00`.

### DIFF
--- a/src/datetime/tests.rs
+++ b/src/datetime/tests.rs
@@ -518,7 +518,7 @@ fn test_datetime_rfc2822() {
     );
     assert_eq!(
         DateTime::parse_from_rfc2822("Wed, 18 Feb 2015 23:16:09 -0000"),
-        Ok(FixedOffset::east_opt(0).unwrap().with_ymd_and_hms(2015, 2, 18, 23, 16, 9).unwrap())
+        Ok(FixedOffset::OFFSET_UNKNOWN.with_ymd_and_hms(2015, 2, 18, 23, 16, 9).unwrap())
     );
     assert_eq!(
         ymdhms_micro(&edt, 2015, 2, 18, 23, 59, 59, 1_234_567).to_rfc2822(),
--- a/src/format/parse.rs
+++ b/src/format/parse.rs
@@ -12,6 +12,7 @@ use super::scan;
 use super::{Fixed, InternalFixed, InternalInternal, Item, Numeric, Pad, Parsed};
 use super::{ParseError, ParseResult};
 use super::{BAD_FORMAT, INVALID, OUT_OF_RANGE, TOO_LONG, TOO_SHORT};
+use crate::format::parsed::NO_OFFSET_INFO;
 use crate::{DateTime, FixedOffset, Weekday};
 
 fn set_weekday_with_num_days_from_sunday(p: &mut Parsed, v: i64) -> ParseResult<()> {
@@ -142,7 +143,8 @@ fn parse_rfc2822<'a>(parsed: &mut Parsed, mut s: &'a str) -> ParseResult<(&'a st
     }
 
     s = scan::space(s)?; // mandatory
-    parsed.set_offset(i64::from(try_consume!(scan::timezone_offset_2822(s))))?;
+    let offset = try_consume!(scan::timezone_offset_2822(s));
+    parsed.set_offset(i64::from(offset.unwrap_or(NO_OFFSET_INFO)))?;
 
     // optional comments
     while let Ok((s_out, ())) = scan::comment_2822(s) {
@@ -213,16 +215,17 @@ pub(crate) fn parse_rfc3339<'a>(parsed: &mut Parsed, mut s: &'a str) -> ParseRes
     }
 
     let offset = try_consume!(scan::timezone_offset(s, |s| scan::char(s, b':'), true, false, true));
-    // This range check is similar to the one in `FixedOffset::east_opt`, so it would be redundant.
-    // But it is possible to read the offset directly from `Parsed`. We want to only successfully
-    // populate `Parsed` if the input is fully valid RFC 3339.
-    // Max for the hours field is `23`, and for the minutes field `59`.
-    const MAX_RFC3339_OFFSET: i32 = (23 * 60 + 59) * 60;
-    if !(-MAX_RFC3339_OFFSET..=MAX_RFC3339_OFFSET).contains(&offset) {
-        return Err(OUT_OF_RANGE);
+    if let Some(offset) = offset {
+        // This range check is similar to the one in `FixedOffset::east_opt`, so it would be
+        // redundant. But it is possible to read the offset directly from `Parsed`. We want to only
+        // successfully populate `Parsed` if the input is fully valid RFC 3339.
+        // Max for the hours field is `23`, and for the minutes field `59`.
+        const MAX_RFC3339_OFFSET: i32 = (23 * 60 + 59) * 60;
+        if !(-MAX_RFC3339_OFFSET..=MAX_RFC3339_OFFSET).contains(&offset) {
+            return Err(OUT_OF_RANGE);
+        }
     }
-    parsed.set_offset(i64::from(offset))?;
-
+    parsed.set_offset(i64::from(offset.unwrap_or(NO_OFFSET_INFO)))?;
     Ok((s, ()))
 }
 
@@ -463,7 +466,7 @@ where
                             false,
                             true,
                         ));
-                        parsed.set_offset(i64::from(offset))?;
+                        parsed.set_offset(i64::from(offset.unwrap_or(NO_OFFSET_INFO)))?;
                     }
 
                     &TimezoneOffsetColonZ | &TimezoneOffsetZ => {
@@ -474,7 +477,7 @@ where
                             false,
                             true,
                         ));
-                        parsed.set_offset(i64::from(offset))?;
+                        parsed.set_offset(i64::from(offset.unwrap_or(NO_OFFSET_INFO)))?;
                     }
                     &Internal(InternalFixed {
                         val: InternalInternal::TimezoneOffsetPermissive,
@@ -486,7 +489,7 @@ where
                             true,
                             true,
                         ));
-                        parsed.set_offset(i64::from(offset))?;
+                        parsed.set_offset(i64::from(offset.unwrap_or(NO_OFFSET_INFO)))?;
                     }
 
                     &RFC2822 => try_consume!(parse_rfc2822(parsed, s)),
@@ -575,11 +578,11 @@ fn parse_rfc3339_relaxed<'a>(parsed: &mut Parsed, mut s: &'a str) -> ParseResult
     s = parse_internal(parsed, s, TIME_ITEMS.iter())?;
     s = s.trim_start();
     let (s, offset) = if s.len() >= 3 && "UTC".as_bytes().eq_ignore_ascii_case(&s.as_bytes()[..3]) {
-        (&s[3..], 0)
+        (&s[3..], Some(0))
     } else {
         scan::timezone_offset(s, scan::colon_or_space, true, false, true)?
     };
-    parsed.set_offset(i64::from(offset))?;
+    parsed.set_offset(i64::from(offset.unwrap_or(NO_OFFSET_INFO)))?;
     Ok((s, ()))
 }
 
@@ -1041,8 +1044,8 @@ mod tests {
         check("+1234:56", &[fixed(TimezoneOffset)], Err(TOO_LONG));
         check("+1234:567", &[fixed(TimezoneOffset)], Err(TOO_LONG));
         check("+00:00", &[fixed(TimezoneOffset)], parsed!(offset: 0));
-        check("-00:00", &[fixed(TimezoneOffset)], parsed!(offset: 0));
-        check("−00:00", &[fixed(TimezoneOffset)], parsed!(offset: 0)); // MINUS SIGN (U+2212)
+        check("-00:00", &[fixed(TimezoneOffset)], parsed!(offset: NO_OFFSET_INFO));
+        check("−00:00", &[fixed(TimezoneOffset)], parsed!(offset: NO_OFFSET_INFO)); // MINUS SIGN (U+2212)
         check("+00:01", &[fixed(TimezoneOffset)], parsed!(offset: 60));
         check("-00:01", &[fixed(TimezoneOffset)], parsed!(offset: -60));
         check("+00:30", &[fixed(TimezoneOffset)], parsed!(offset: 1_800));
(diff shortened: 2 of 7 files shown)


src/format/parse.rs
```diff
@@ -216,10 +215,7 @@ fn parse_rfc3339<'a>(parsed: &mut Parsed, mut s: &'a str) -> ParseResult<(&'a st
     }
 
     let offset = try_consume!(scan::timezone_offset_zulu(s, |s| scan::char(s, b':')));
-    if offset <= -86_400 || offset >= 86_400 {
```
pitdicker, line 219: This check was redundant, because it is also checked in `FixedOffset_east_opt`. So I made it the same as the other places that use `Parsed::set_offset`. They can return the same `OUT_OF_RANGE`, just a little later.

djc:
> Currently chrono drops the distinction, but I would like to preserve it.

Do you have an actual use case for this?

pitdicker:
Ai, why do you have to ask that :rofl: ?
As a hobby I am making my own feed reader, and I would like to know the time an entry was published by the author in his own timezone. If the timestamp indicates it knows the time in UTC but the relation to the local time is unknown, I find that interesting. But currently chrono offers no easy way to know.

pitdicker:
As a better reason: we would follow the RFCs more closely, and be able to round-trip these timestamps.

djc:
In general, I would concur we want to follow the RFC closely. But at the same time, "you find it interesting" to know this doesn't seem like a strong justification for adding ~90 lines of code on net.

pitdicker:
That is fair.

Honestly the changes felt small, so i did some counting:
- 17 lines for testing the size of common types in chrono.
- 30 would also be there if we wanted to use the part of encoding an offset in an `NonZeroI32` to get the niche optimization.
- that leaves ca 45 netto for the feature and documentation

src/format/parse.rs
```diff
@@ -853,7 +855,7 @@ fn test_rfc2822() {
         ("Tue, 20 Jan 2015 17:35:90 -0800", Err(OUT_OF_RANGE)), // bad second
         ("Tue, 20 Jan 2015 17:35:20 -0890", Err(OUT_OF_RANGE)), // bad offset
         ("6 Jun 1944 04:00:00Z", Err(INVALID)),            // bad offset (zulu not allowed)
-        ("Tue, 20 Jan 2015 17:35:20 HAS", Err(NOT_ENOUGH)), // bad named time zone
+        ("Tue, 20 Jan 2015 17:35:20 HAS", Err(INVALID)),   // bad named time zone
```
pitdicker, line 858: We would now report the correct error here.

src/format/parse.rs
```diff
@@ -868,14 +870,15 @@ fn test_rfc2822() {
...
-        ("Tue, 20 Jan 2015 17:35:20 K", Ok("Tue, 20 Jan 2015 17:35:20 +0000")),
-        ("Tue, 20 Jan 2015 17:35:20 k", Ok("Tue, 20 Jan 2015 17:35:20 +0000")),
+        // other named single-letter military timezones must fallback to -0000
+        ("Tue, 20 Jan 2015 17:35:20 A", Ok("Tue, 20 Jan 2015 17:35:20 -0000")),
+        ("Tue, 20 Jan 2015 17:35:20 a", Ok("Tue, 20 Jan 2015 17:35:20 -0000")),
+        ("Tue, 20 Jan 2015 17:35:20 K", Ok("Tue, 20 Jan 2015 17:35:20 -0000")),
+        ("Tue, 20 Jan 2015 17:35:20 k", Ok("Tue, 20 Jan 2015 17:35:20 -0000")),
         // named single-letter timezone "J" is specifically not valid
-        ("Tue, 20 Jan 2015 17:35:20 J", Err(NOT_ENOUGH)),
+        ("Tue, 20 Jan 2015 17:35:20 J", Err(INVALID)),
```
pitdicker, line 881: And here.

src/format/parse.rs
```diff
@@ -737,7 +739,7 @@ fn test_parse() {
 
     // fixed: timezone offsets
     check!("+00:00",    [fix!(TimezoneOffset)]; offset: 0);
-    check!("-00:00",    [fix!(TimezoneOffset)]; offset: 0);
+    check!("-00:00",    [fix!(TimezoneOffset)]; offset: NO_OFFSET_INFO);
```
jtmoon79, line 742: Please also add asserts for variant

```rust
        check!(" -00:00",    [fix!(TimezoneOffset)]; offset: NO_OFFSET_INFO);
        check!("+00:01",    [fix!(TimezoneOffset)]; offset: 1);
        check!("-00:01",    [fix!(TimezoneOffset)]; offset: -1);
```

Somewhat tacked onto this PR. But still good-to-have.

Up to you.

src/format/parse.rs
jtmoon79, line 858: Can you add extra asserts
```rust
("20 Jan 2015 17:35:20 +0000", Ok("Tue, 20 Jan 2015 17:35:20 +0000")),
("20 Jan 2015 17:35:20 -0001", Ok("Tue, 20 Jan 2015 17:35:20 -0001")),
 ("Tue, 20 Jan 2015 17:35:20 -9900", Err(OUT_OF_RANGE)),  // bad offset
```
(adjusted to whatever the correct value is)

src/format/scan.rs
```diff
@@ -325,30 +332,27 @@ pub(super) fn timezone_offset_2822(s: &str) -> ParseResult<(&str, Option<i32>)>
         let name = &s.as_bytes()[..upto];
         let s = &s[upto..];
         let offset_hours = |o| Ok((s, Some(o * 3600)));
-        if equals(name, "gmt") || equals(name, "ut") {
```
jtmoon79, line 328: Can you add a code comment

```rust
// RFC 2822 requires support for some named North America timezones, a small subset of all named timezones
```

Also a tacked-on change.

src/lib.rs
```diff
@@ -525,3 +525,20 @@ macro_rules! matches {
         }
     }
 }
+
+#[test]
+#[allow(deprecated)]
+fn test_type_sizes() {
```
jtmoon79, line 531: I'm curious if this has been tested on an x86 (32 bit) platform?

IME FreeBSD x86 is always little weird about sizings of things (different than typical Linux anyway).

If I get around to it, I'll try to compile this PR on a FreeBSD 13 VM.

src/offset/fixed.rs
```diff
@@ -16,16 +17,50 @@ use crate::oldtime::Duration as OldDuration;
...
-    local_minus_utc: i32,
+    // Encodes an offset in seconds, with the value shifted two bits to the left.
+    // The remaining bits encode flags:
+    // - `OFFSET_NORMAL` sets one flag to a non-zero value, so we can encode it in a `NonZeroI32`
+    //   and get niche optimization.
+    // - `OFFSET_UNKNOWN` to preserve the difference RFC 2822 and RFC 3339 make between an offset
+    //    of +0:00 and -0:00.
+    // Advantage of this encoding is that it only takes a single shift right to get offset in
+    // seconds.
+    local_minus_utc: NonZeroI32,
```
jtmoon79, line 40: Can you make this code comment a docstring?
In the docstring, would you tell the user what helper functions they should call to extract various infos?  Can you provide those helper functions? Even if it's only chrono developers that need to know, it's still worthwhile to write down.
Otherwise I might assume I have to do the bit shifts and masks myself... and those are programming approaches I would like everyone to move on from. 😬

src/offset/fixed.rs
```diff
@@ -138,7 +188,10 @@ impl Offset for FixedOffset {
 
 impl fmt::Debug for FixedOffset {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let offset = self.local_minus_utc;
+        if self.no_offset_info() {
+            return write!(f, "-00:00");
```
jtmoon79, line 192: can you code comment why you're putting `-00:00` here?

jtmoon79 (changes requested):
I'm very compelled by this standards-following PR. I thought something wasn't quite right about how the various RFC and `+00:00` and `-00:00` were being handled but I couldn't put my finger on it (and it seemed like a difficult bramblebush to crawl through).

Thanks much @pitdicker , this is a great change!

jtmoon79:
> RFC 3339 and RFC 2822 make a distinction between an offset of `+00:00` and `-00:00`.
> 
> RFC 2822:
> 
> > The form "+0000" SHOULD be used to indicate a time zone at Universal Time. Though "-0000" also indicates Universal Time, it is used to indicate that the time was generated on a system that may be in a local time zone other than Universal Time and therefore indicates that the date-time contains no information about the local time zone.
> 
> From RFC3339:
> 
> > If the time in UTC is known, but the offset to local time is unknown, this can be represented with an offset of "-00:00". This differs semantically from an offset of "Z" or "+00:00", which imply that UTC is the preferred reference point for the specified time.
> 
> Currently chrono drops the distinction, but I would like to preserve it.
> 
>     * I noticed `format/scan.rs` was already written with the functionality of this RFCs in mind, but did not go all the way. It now follows the specs better and can sometimes give better parser errors.
> 
>     * Modifying `Parsed` to preserve this information is a bit messy as all the methods are public. Adding a public constant `NO_OFFSET_INFO` seemed the cleanest solution. Parsing doesn't support more than two digits for the offset hours, limiting the values that are passed to `Parsed::set_offset` to `99 * 3600 + 59 * 60 = 359940`. The value of `NO_OFFSET_INFO` is way outside that range, so no chance for conflicts.
> 
>     * `FixedOffset` can be modified in a reasonably clean way to encode `-0`: shift the value 2 bits left and encode this special state as an `OFFSET_UNKNOWN` flag in the rightmost bits. Advantage of this encoding is that it only takes a single shift right to get the offset in seconds, so this should not measurably impact performance.
> 
>     * The same shifting trick enables using `NonZeroI32`, as long as we set one of the rightmost bits to something non-null with the `OFFSET_NORMAL` flag. This gives `FixedOffset` a niche. `Option<DateTime<FixedOffset>>` is now the same size as `DateTime<FixedOffset>`: 16 bytes instead of 20.
> 
>     * For comparisons and hashing the flags are ignored. This way `-00:00` is treated as equal to `+00:00`, which is usually what is expected. Also this way we remain backwards compatible.
> 
>     * `FixedOffset::no_offset_info` returns whether the offset is `-00:00`.
> 
>     * `FixedOffset::OFFSET_UNKNOWN` can be used to initialize an offset to `-00:00`.

Can you find an appropriate docstring to place this information (especiially the quoted RFC stuff)?  This is a great explanation of little finicky area that I suspect a few users will run into.

esheppa:
Interestingly, this was a common case that was causing failures in [`tests/dateutils.rs::verify_against_date_command_format_local`](https://github.com/chronotope/chrono/blob/0a16a655416b07e011d9f10de24220103be7044a/tests/dateutils.rs#L79), if this is merged, we may be able to ensure the offsets match as well, rather than just the date/time values.

src/format/parse.rs
pitdicker: These checks are already right above:
```rust
        check!("+00:01",    [fix!(TimezoneOffset)]; offset: 60);
        check!("-00:01",    [fix!(TimezoneOffset)]; offset: -60);
```

But I added `" -00:00"`

src/offset/fixed.rs
pitdicker: > honestly, I'd prefer to have discrete struct members for each piece of information embedded here. I'm not persuaded the runtime efficiency-savings is worth the mental overhead.

This is somewhat comparable to `NaiveDate`, which fits a year, ordinal, leap-year flag and weekday flags in an i32.

I agree that using individual struct members is usually better. But chrono's `DateTime<FixedOffset>` is a pretty fundamental type in the rust ecosystem. It should not increase in size without very good reason.

> Also, I'm not sure, but I'm concerned it might be a problem on less-common Big Endian platforms (I don't recall where and when Rust gives guarantees around bit endianness).

In safe Rust, we do not have to worry about this. As soons as we start transmuting it becomes messy.

pitdicker:
> Can you find an appropriate docstring to place this information (especially the quoted RFC stuff)? This is a great explanation of little finicky area that I suspect quite a few users will run into.
> You might find it appropriate to split up the info and place within various user-facing functions. Whatever you think is good. So long as this subtlety is explained to the user.

Good idea, I have written an introduction for the `FixedOffset` type, and included a small reference in the docstring of the `Parsed::offset` field.

src/datetime/tests.rs
```diff
@@ -471,7 +471,7 @@ fn test_datetime_rfc2822() {
     );
     assert_eq!(
         DateTime::parse_from_rfc2822("Wed, 18 Feb 2015 23:16:09 -0000"),
-        Ok(FixedOffset::east_opt(0).unwrap().with_ymd_and_hms(2015, 2, 18, 23, 16, 9).unwrap())
+        Ok(FixedOffset::OFFSET_UNKNOWN.with_ymd_and_hms(2015, 2, 18, 23, 16, 9).unwrap())
```
djc, line 474: This commit has way too much going on for me to review. Please split it into smaller parts.

pitdicker: Done.

jtmoon79 (approved):
LGTM

src/format/parse.rs
pitdicker: Added a comment instead.

src/offset/fixed.rs
```diff
@@ -15,16 +16,85 @@ use crate::oldtime::Duration as OldDuration;
 use crate::DateTime;
 use crate::Timelike;
 
-/// The time zone with fixed offset, from UTC-23:59:59 to UTC+23:59:59.
+const OFFSET_NORMAL: i32 = 1;
+const OFFSET_UNKNOWN: i32 = 2;
+
+/// A fixed offset which can be used to indicate a time zone.
```
jtmoon79, line 22: Suggested opening sentence.

    /// A fixed offset of clock time from [Coordinated Universal Time] (UTC) stored to seconds granularity.
    /// A `FixedOffset` can be used to indicate a time zone.
    ///
    /// [Coordinated Universal Time]: https://en.wikipedia.org/wiki/Coordinated_Universal_Time

jtmoon79 (approved):
Approve still, with suggestion.

src/offset/fixed.rs
pitdicker: Good suggestion.

PR #1043, opened 2023-04-29
Make functions in internals const

pitdicker:
@esheppa My intention was to understand why you would need to overhaul the internals of chono in https://github.com/chronotope/chrono/pull/882 to make functions const. So I started sprinkling `const` around to see what breaks.

Everything in `internals.rs` can easily be made const, even on the 0.4.x branch. I made this PR because the work was done anyway, no use in letting someone else do it again.

@esheppa Would you be willing to review?
I also did a good number of the functions from `NaiveDate` just to prove for myself there are no major obstacles, but kept them out of this PR to keep it reviewable. There are a lot more choices that can be made there, so they deserve an own PR in the future.

I'll add some comments on the required changes.

### DIFF
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -241,19 +241,35 @@ impl NaiveDate {
     pub(crate) fn weeks_from(&self, day: Weekday) -> i32 {
         (self.ordinal() as i32 - self.weekday().num_days_from(day) as i32 + 6) / 7
     }
-    /// Makes a new `NaiveDate` from year and packed ordinal-flags, with a verification.
-    fn from_of(year: i32, of: Of) -> Option<NaiveDate> {
-        if (MIN_YEAR..=MAX_YEAR).contains(&year) {
-            let of = of.inner();
-            Some(NaiveDate { ymdf: (year << 13) | (of as DateImpl) })
-        } else {
-            None
+
+    /// Makes a new `NaiveDate` from year, ordinal and flags.
+    /// Does not check whether the flags are correct for the provided year.
+    const fn from_ordinal_and_flags(
+        year: i32,
+        ordinal: u32,
+        flags: YearFlags,
+    ) -> Option<NaiveDate> {
+        if year < MIN_YEAR || year > MAX_YEAR {
+            return None; // Out-of-range
+        }
+        // Enable debug check once the MSRV >= 1.57 (panicking in const feature)
+        // debug_assert!(YearFlags::from_year(year).0 == flags.0);
+        match Of::new(ordinal, flags) {
+            Some(of) => Some(NaiveDate { ymdf: (year << 13) | (of.inner() as DateImpl) }),
+            None => None, // Invalid: Ordinal outside of the nr of days in a year with those flags.
         }
     }
 
-    /// Makes a new `NaiveDate` from year and packed month-day-flags, with a verification.
-    fn from_mdf(year: i32, mdf: Mdf) -> Option<NaiveDate> {
-        NaiveDate::from_of(year, mdf.to_of()?)
+    /// Makes a new `NaiveDate` from year and packed month-day-flags.
+    /// Does not check whether the flags are correct for the provided year.
+    const fn from_mdf(year: i32, mdf: Mdf) -> Option<NaiveDate> {
+        if year < MIN_YEAR || year > MAX_YEAR {
+            return None; // Out-of-range
+        }
+        match mdf.to_of() {
+            Some(of) => Some(NaiveDate { ymdf: (year << 13) | (of.inner() as DateImpl) }),
+            None => None, // Non-existing date
+        }
     }
 
     /// Makes a new `NaiveDate` from the [calendar date](#calendar-date)
@@ -324,7 +340,7 @@ impl NaiveDate {
     #[must_use]
     pub fn from_yo_opt(year: i32, ordinal: u32) -> Option<NaiveDate> {
         let flags = YearFlags::from_year(year);
-        NaiveDate::from_of(year, Of::new(ordinal, flags)?)
+        NaiveDate::from_ordinal_and_flags(year, ordinal, flags)
     }
 
     /// Makes a new `NaiveDate` from the [ISO week date](#week-date)
@@ -393,20 +409,21 @@ impl NaiveDate {
             if weekord <= delta {
                 // ordinal < 1, previous year
                 let prevflags = YearFlags::from_year(year - 1);
-                NaiveDate::from_of(
+                NaiveDate::from_ordinal_and_flags(
                     year - 1,
-                    Of::new(weekord + prevflags.ndays() - delta, prevflags)?,
+                    weekord + prevflags.ndays() - delta,
+                    prevflags,
                 )
             } else {
                 let ordinal = weekord - delta;
                 let ndays = flags.ndays();
                 if ordinal <= ndays {
                     // this year
-                    NaiveDate::from_of(year, Of::new(ordinal, flags)?)
+                    NaiveDate::from_ordinal_and_flags(year, ordinal, flags)
                 } else {
                     // ordinal > ndays, next year
                     let nextflags = YearFlags::from_year(year + 1);
-                    NaiveDate::from_of(year + 1, Of::new(ordinal - ndays, nextflags)?)
+                    NaiveDate::from_ordinal_and_flags(year + 1, ordinal - ndays, nextflags)
                 }
             }
         } else {
@@ -451,7 +468,7 @@ impl NaiveDate {
         let (year_div_400, cycle) = div_mod_floor(days, 146_097);
         let (year_mod_400, ordinal) = internals::cycle_to_yo(cycle as u32);
         let flags = YearFlags::from_year_mod_400(year_mod_400 as i32);
-        NaiveDate::from_of(year_div_400 * 400 + year_mod_400 as i32, Of::new(ordinal, flags)?)
+        NaiveDate::from_ordinal_and_flags(year_div_400 * 400 + year_mod_400 as i32, ordinal, flags)
     }
 
     /// Makes a new `NaiveDate` by counting the number of occurrences of a particular day-of-week
@@ -1037,7 +1054,7 @@ impl NaiveDate {
 
         let (year_mod_400, ordinal) = internals::cycle_to_yo(cycle as u32);
         let flags = YearFlags::from_year_mod_400(year_mod_400 as i32);
-        NaiveDate::from_of(year_div_400 * 400 + year_mod_400 as i32, Of::new(ordinal, flags)?)
+        NaiveDate::from_ordinal_and_flags(year_div_400 * 400 + year_mod_400 as i32, ordinal, flags)
     }
 
     /// Subtracts the `days` part of given `Duration` from the current date.
@@ -1069,7 +1086,7 @@ impl NaiveDate {
 
         let (year_mod_400, ordinal) = internals::cycle_to_yo(cycle as u32);
         let flags = YearFlags::from_year_mod_400(year_mod_400 as i32);
-        NaiveDate::from_of(year_div_400 * 400 + year_mod_400 as i32, Of::new(ordinal, flags)?)
+        NaiveDate::from_ordinal_and_flags(year_div_400 * 400 + year_mod_400 as i32, ordinal, flags)
     }
 
     /// Subtracts another `NaiveDate` from the current date.
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -243,8 +243,8 @@ impl NaiveDate {
     }
     /// Makes a new `NaiveDate` from year and packed ordinal-flags, with a verification.
     fn from_of(year: i32, of: Of) -> Option<NaiveDate> {
-        if (MIN_YEAR..=MAX_YEAR).contains(&year) && of.valid() {
-            let Of(of) = of;
+        if (MIN_YEAR..=MAX_YEAR).contains(&year) {
+            let of = of.inner();
             Some(NaiveDate { ymdf: (year << 13) | (of as DateImpl) })
         } else {
             None
@@ -253,7 +253,7 @@ impl NaiveDate {
 
     /// Makes a new `NaiveDate` from year and packed month-day-flags, with a verification.
     fn from_mdf(year: i32, mdf: Mdf) -> Option<NaiveDate> {
-        NaiveDate::from_of(year, mdf.to_of())
+        NaiveDate::from_of(year, mdf.to_of()?)
     }
 
     /// Makes a new `NaiveDate` from the [calendar date](#calendar-date)
@@ -924,7 +924,7 @@ impl NaiveDate {
     /// Returns the packed ordinal-flags.
     #[inline]
     const fn of(&self) -> Of {
-        Of((self.ymdf & 0b1_1111_1111_1111) as u32)
+        Of::from_date_impl(self.ymdf)
     }
 
     /// Makes a new `NaiveDate` with the packed month-day-flags changed.
@@ -932,20 +932,16 @@ impl NaiveDate {
     /// Returns `None` when the resulting `NaiveDate` would be invalid.
     #[inline]
     fn with_mdf(&self, mdf: Mdf) -> Option<NaiveDate> {
-        self.with_of(mdf.to_of())
+        Some(self.with_of(mdf.to_of()?))
     }
 
     /// Makes a new `NaiveDate` with the packed ordinal-flags changed.
     ///
     /// Returns `None` when the resulting `NaiveDate` would be invalid.
+    /// Does not check if the year flags match the year.
     #[inline]
-    fn with_of(&self, of: Of) -> Option<NaiveDate> {
-        if of.valid() {
-            let Of(of) = of;
-            Some(NaiveDate { ymdf: (self.ymdf & !0b1_1111_1111_1111) | of as DateImpl })
-        } else {
-            None
-        }
+    const fn with_of(&self, of: Of) -> NaiveDate {
+        NaiveDate { ymdf: (self.ymdf & !0b1_1111_1111_1111) | of.inner() as DateImpl }
     }
 
     /// Makes a new `NaiveDate` for the next calendar date.
@@ -974,7 +970,10 @@ impl NaiveDate {
     #[inline]
     #[must_use]
     pub fn succ_opt(&self) -> Option<NaiveDate> {
-        self.with_of(self.of().succ()).or_else(|| NaiveDate::from_ymd_opt(self.year() + 1, 1, 1))
+        match self.of().succ() {
+            Some(of) => Some(self.with_of(of)),
+            None => NaiveDate::from_ymd_opt(self.year() + 1, 1, 1),
+        }
     }
 
     /// Makes a new `NaiveDate` for the previous calendar date.
@@ -1003,7 +1002,10 @@ impl NaiveDate {
     #[inline]
     #[must_use]
     pub fn pred_opt(&self) -> Option<NaiveDate> {
-        self.with_of(self.of().pred()).or_else(|| NaiveDate::from_ymd_opt(self.year() - 1, 12, 31))
+        match self.of().pred() {
+            Some(of) => Some(self.with_of(of)),
+            None => NaiveDate::from_ymd_opt(self.year() - 1, 12, 31),
+        }
     }
 
     /// Adds the `days` part of given `Duration` to the current date.
@@ -1622,7 +1624,7 @@ impl Datelike for NaiveDate {
     /// ```
     #[inline]
     fn with_ordinal(&self, ordinal: u32) -> Option<NaiveDate> {
-        self.with_of(self.of().with_ordinal(ordinal)?)
+        self.of().with_ordinal(ordinal).map(|of| self.with_of(of))
     }
 
     /// Makes a new `NaiveDate` with the day of year (starting from 0) changed.
@@ -1647,7 +1649,7 @@ impl Datelike for NaiveDate {
     #[inline]
     fn with_ordinal0(&self, ordinal0: u32) -> Option<NaiveDate> {
         let ordinal = ordinal0.checked_add(1)?;
-        self.with_of(self.of().with_ordinal(ordinal)?)
+        self.with_ordinal(ordinal)
     }
 }
 
--- a/src/naive/internals.rs
+++ b/src/naive/internals.rs
@@ -52,7 +52,7 @@ pub(super) const FE: YearFlags = YearFlags(0o07);
 pub(super) const G: YearFlags = YearFlags(0o16);
 pub(super) const GF: YearFlags = YearFlags(0o06);
 
-static YEAR_TO_FLAGS: [YearFlags; 400] = [
+const YEAR_TO_FLAGS: &[YearFlags; 400] = &[
     BA, G, F, E, DC, B, A, G, FE, D, C, B, AG, F, E, D, CB, A, G, F, ED, C, B, A, GF, E, D, C, BA,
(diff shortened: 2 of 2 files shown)


src/naive/internals.rs
```diff
@@ -53,7 +51,7 @@ pub(super) const FE: YearFlags = YearFlags(0o07);
 pub(super) const G: YearFlags = YearFlags(0o16);
 pub(super) const GF: YearFlags = YearFlags(0o06);
 
-static YEAR_TO_FLAGS: [YearFlags; 400] = [
+const YEAR_TO_FLAGS: &[YearFlags; 400] = &[
```
pitdicker, line 54: A static is only included once in the binary. To prevent duplication of this array on every use as const, It has to be put behind a reference (see https://users.rust-lang.org/t/const-vs-static/52951/12). It can still be duplicated with multiple codegen units, but the statics are not _that_ big that it should be a problem.

src/naive/internals.rs
```diff
@@ -94,47 +92,48 @@ static YEAR_DELTAS: [u8; 401] = [
...
-    let (mut year_mod_400, mut ordinal0) = div_rem(cycle, 365);
-    let delta = u32::from(YEAR_DELTAS[year_mod_400 as usize]);
+pub(super) const fn cycle_to_yo(cycle: u32) -> (u32, u32) {
+    let mut year_mod_400 = cycle / 365;
+    let mut ordinal0 = cycle % 365;
+    let delta = YEAR_DELTAS[year_mod_400 as usize] as u32;
     if ordinal0 < delta {
         year_mod_400 -= 1;
-        ordinal0 += 365 - u32::from(YEAR_DELTAS[year_mod_400 as usize]);
+        ordinal0 += 365 - YEAR_DELTAS[year_mod_400 as usize] as u32;
```
pitdicker, line 101: Everything old is new again... `u32::from` is not available in const, we have to use old casts.

src/naive/internals.rs
```diff
@@ -271,27 +270,30 @@ pub(super) struct Of(pub(crate) u32);
...
         }
     }
 
     #[inline]
-    pub(super) fn from_mdf(Mdf(mdf): Mdf) -> Of {
+    pub(super) const fn from_mdf(Mdf(mdf): Mdf) -> Of {
         let mdl = mdf >> 3;
-        match MDL_TO_OL.get(mdl as usize) {
-            Some(&v) => Of(mdf.wrapping_sub((i32::from(v) as u32 & 0x3ff) << 3)),
-            None => Of(0),
```
pitdicker, line 286: These three lines where using `array.get()` and returning a safe value instead of panicking on out-of-bounds. Yet the index value is created in this module, and always in bounds.

 I suspect the reason here is not just to avoid a potential panic, but to avoid the creation of landings pads if the compiler can't elide the bounds check.

In my opinion it would be fine to simplify this code to direct array indexing, but I kept the functionality. Note that `array.get()` is unavailable in const, so this needs a manual bounds check.

src/naive/internals.rs
```diff
@@ -368,33 +370,37 @@ impl fmt::Debug for Of {
 /// (month, day of month and leap flag),
 /// which is an index to the `MDL_TO_OL` lookup table.
 #[derive(PartialEq, PartialOrd, Copy, Clone)]
-pub(super) struct Mdf(pub(super) u32);
```
pitdicker, line 371: The value didn't need to be public. Making it private makes it clearer the value is always correct, and thus in bounds for array indexing.

src/naive/internals.rs
```diff
@@ -317,22 +319,22 @@ impl Of {
     }
 
     #[inline]
-    pub(super) fn weekday(&self) -> Weekday {
+    pub(super) const fn weekday(&self) -> Weekday {
         let Of(of) = *self;
-        Weekday::from_u32(((of >> 4) + (of & 0b111)) % 7).unwrap()
+        Weekday::from_u32_mod7((of >> 4) + (of & 0b111))
```
pitdicker, line 324: Can't use `Weekday::from_u32` because that is part of the `num_traits::FromPrimitive`, and trait methods can't yet be const.

src/weekday.rs
```diff
@@ -50,14 +50,29 @@ pub enum Weekday {
 }
 
 impl Weekday {
+    /// Create a `Weekday` from an `u32`, with Monday = 0.
+    /// Infallible, takes `n & 7`.
+    #[inline]
+    pub(crate) const fn from_u32_mod7(n: u32) -> Weekday {
```
pitdicker, line 56: For use in this crate it would be nice if there was a const, infallible way to create a `Weekday`.

Because the two methods construction a `Weekday` in `internals.rs` passed `n % 7`, making `from_u32_mod7` seemed like a good way to have an infallible function and avoid a little code duplication.

pitdicker:
@RagibHasin You expressed an interest to help with making the API const. Would you also be willing to review?

And maybe we can work together?

It seems easy to avoid one massive `make everything const`-PR, because a lot depends on each other.`DateTime` depends on `NaiveDateTime`, `NaiveDateTime` depends on `NaiveDate` and `NaiveTime`...

`IsoWeek`, `Month`, `NaiveTime`, `Duration` and `FixedOffset` should be pretty self contained. Do you want to make PRs for any or all of those?

RagibHasin:
Thanks, @pitdicker. I expect to have some free time in around 12 hours. I would try to review it then.

Very excited about this progress.

pitdicker:
https://github.com/chronotope/chrono/pull/638 Already worked on `Duration` about 2 years ago, but hit the problem `Option::unwrap` and `Option::expect` are unavailable in const.

Maybe we can use some ugly macro that imitates them but works in const by doing an implicit panic, like https://users.rust-lang.org/t/compile-time-const-unwrapping/51619/6?

Or chrono would have to update to Rust 2021 and raise the MSRV to 1.57, which supports [panic in const context](https://blog.rust-lang.org/2021/12/02/Rust-1.57.0.html). See https://github.com/chronotope/chrono/issues/995.

src/weekday.rs
```diff
@@ -50,14 +50,29 @@ pub enum Weekday {
 }
 
 impl Weekday {
+    /// Create a `Weekday` from an `u32`, with Monday = 0.
+    /// Infallible, takes `n % 7`.
+    #[inline]
+    pub(crate) const fn from_u32_mod7(n: u32) -> Weekday {
```
jtmoon79, line 56: It's a trivial function but for due diligence, can you add test cases for this in the `mod test` section below?

e.g.
```rust
    #[test]
    fn test_from_u32_mod7() {
        //...
    }
```

jtmoon79 (approved):
LGTM

This is a safe and useful improvement!

Just one request for a new test case.

src/naive/internals.rs
pitdicker: Ai. I was wrong, this code is just ancient. In various places it is returning `0` instead of `Option`, and you are supposed to call `valid()` afterwards 😞. Let me see if I can clean that up a little.

src/naive/internals.rs
```diff
@@ -266,56 +267,66 @@ const OL_TO_MDL: &[u8; MAX_OL as usize + 1] = &[
 /// The whole bits except for the least 3 bits are referred as `Ol` (ordinal and leap flag),
 /// which is an index to the `OL_TO_MDL` lookup table.
 #[derive(PartialEq, PartialOrd, Copy, Clone)]
-pub(super) struct Of(pub(crate) u32);
+pub(super) struct Of(u32);
```
pitdicker, line 270: The value of `Of` is now private and always valid.

src/naive/internals.rs
```diff
@@ -266,56 +267,66 @@ const OL_TO_MDL: &[u8; MAX_OL as usize + 1] = &[
...
-        match ordinal <= 366 {
-            true => Some(Of((ordinal << 4) | flags as u32)),
-            false => None,
-        }
+        let of = Of((ordinal << 4) | flags as u32);
+        of.validate()
+    }
+
+    pub(super) const fn from_date_impl(date_impl: DateImpl) -> Of {
+        // We assume the value in the `DateImpl` is valid.
```
pitdicker, line 280: Had to add this to make `Of` private, and prevent unnecessary validation when converting from `DateImpl`. But `DateImpl` is not private... Had to stop somewhere :smile:

pitdicker (commented):
While making parts of `NaiveDate` const, I noticed there where `Option`s created in places where I would not expect them, and not created where they should.

`Of` was written to allow the creation of invalid values, and you were supposed to call `valid()` afterwards and handle the error. That didn't seem very Rusty, so I changed `Of` is to always be valid and return `Option`s earlier.

There is not any extra validation happening. It is just happening in the functions that would create the invalid state, not in the functions calling them.

esheppa:
Thanks @pitdicker - this looks great. Looks like there were a few critical parts I missed that led me down the path of an internals rewrite (that said, some of it was to try to simplify the code and avoid some conversions, but in the end the benchmarks were much worse). The good news is this looks like a great solution, and I can potentially look into doing a new PR with some of the potentially useful parts of #882 once this is merged

src/naive/internals.rs
esheppa: we may even have some correctness gains by separating these values out to separate modules, with the modules only exposing APIs that allow valid values (in cases where a const API still requires things like .valid())

esheppa:
With regards to:

> Or chrono would have to update to Rust 2021 and raise the MSRV to 1.57, which supports [panic in const context](https://blog.rust-lang.org/2021/12/02/Rust-1.57.0.html). See https://github.com/chronotope/chrono/issues/995.

What we can do here is have a non-default feature that allows this (eg `const-validation`), that said, for `Duration` another potentially better option is moving to using the `Duration` type from the stdlib. I've got some PRs (which need a little updating) which I'll update shortly that could provide a good solution here

RagibHasin:
> With regards to:
> 
> > Or chrono would have to update to Rust 2021 and raise the MSRV to 1.57, which supports [panic in const context](https://blog.rust-lang.org/2021/12/02/Rust-1.57.0.html). See #995.
> 
> What we can do here is have a non-default feature that allows this (eg `const-validation`), that said, for `Duration` another potentially better option is moving to using the `Duration` type from the stdlib. I've got some PRs (which need a little updating) which I'll update shortly that could provide a good solution here

`std::time::Duration` cannot represent negative durations. But negative durations feel awkward anyway IMO. If the decision is to move towards stdlib then would there be some new type (eg. Stride) which would be a signed `Duration` to be returned by subtraction operations?

esheppa:
> std::time::Duration cannot represent negative durations. But negative durations feel awkward anyway IMO. 

Yes I think the best solution is to avoid negative durations - however I'm also keen to recieve any feedback on where they would be useful.

The current design is implemented in this PR: https://github.com/chronotope/chrono/pull/895 (a continuation of a long series of designs, but this one is able to be used in the `0.4.x` series which is a large benefit)

Originally we had thought to use a new enum, eg:

```rust
enum TimeDelta { Forwards(Duration), Backwards(Duration) }
```

However in the current iteration I'm just reusing the `Result` type, with `Ok` containing a forwards duration, and `Err` containing the absolute value of the negative duration if applicable, as there didn't seem to be enough justification to add a new type. Importantly, because `Duration` doesn't implement `Error`, this is unlikely to be accidentally handled with `?`.

jtmoon79 (commented):
At risk of sounding like a broken record...

Could some 1:1 tests be added, e.g. each `function` gets it's own `test_function`?

On top of that, preferably added in two or more commits:
- A bunch of tests in one (or more) a _before changes_ commit
- and then one (or more) commit with changes + test adjustments.

The additional tests provide clarity to developers about function expectations (esp. code reviewers 😊) and give more confidence about the changes.

Just my request. I'm requesting this now and not previously because a new commit has added a few behavior changes (459defde7e422aa31bd7d11329cfcca1e7d353bb).

Also, just having more 1:1 function testing would be good for `chrono`.

pitdicker:
> At risk of sounding like a broken record...

Please do! I'll learn 😄 

For these changes though: this code is already tested really, really thoroughly.

- In this module there are tests for the year flags, ordinal + flag-format, and month-dag-flag formats. These test that one form can be converted to another and back correctly. And there are tests that create all possible year flags and ordinals, or year flags and dates, and check some methods.
- Although this is another module than `NaiveDate`, they are very closely integrated. `Of::succ` and `Of::pred` are not tested here, but `NaiveDate::succ` and `NaiveDate::pred` have tests for the interesting cases.

I don't see much to improve here w.r.t the existing tests.

Or did you mean: add tests to verify the functions keep working in const context?

pitdicker:
> I'm requesting this now and not previously because a new commit has added a few behavior changes

Oops, replied to soon. I'll add tests that verify the changed functions return `None` on invalid cases. We can check those now directly in this module 👍.

pitdicker:
> Thanks @pitdicker - this looks great. Looks like there were a few critical parts I missed that led me down the path of an internals rewrite (that said, some of it was to try to simplify the code and avoid some conversions, but in the end the benchmarks were much worse). The good news is this looks like a great solution, and I can potentially look into doing a new PR with some of the potentially useful parts of #882 once this is merged

Nice! Seems like most of the code necessary to make functions in chrono const already exists, it just has to come together :+1:.

pitdicker:
> What we can do here is have a non-default feature that allows this (eg `const-validation`), that said, for `Duration` another potentially better option is moving to using the `Duration` type from the stdlib. I've got some PRs (which need a little updating) which I'll update shortly that could provide a good solution here

Is that something that can be done on the 0.4.x branch, or would it require breaking changes? I'll reply on your https://github.com/chronotope/chrono/pull/895 when I get to know that area a little better.

Maybe for the somewhat simple work of making methods const we can make something work for the current implementation.

> What we can do here is have a non-default feature that allows this (eg `const-validation`)

I remember that removing a feature can only be done when going to a new major version. I.e. if we introduce it in 0.4.x, it can't be removed until 0.5. That doesn't seem too bad... Maybe a more general `msrv_workarounds` that potentially stays relevant?

This feels like a ok-ish solution to me. A macro that maps to regular `unwrap` without the feature, and a hack with implicit panic on older rust versions. But if we can avoid the mess and raise the MSRV that would be so much nicer.

Maybe just work on that assumption for now, and make a PR with the workarounds if the MSRV has to stay below 1.57?

jtmoon79 (approved):
LGTM

src/naive/internals.rs
```diff
@@ -368,33 +371,37 @@ impl fmt::Debug for Of {
...
     }
 
     #[inline]
-    pub(super) fn from_of(Of(of): Of) -> Mdf {
+    pub(super) const fn from_of(Of(of): Of) -> Mdf {
         let ol = of >> 3;
-        match OL_TO_MDL.get(ol as usize) {
-            Some(&v) => Mdf(of + (u32::from(v) << 3)),
-            None => Mdf(0),
+        if ol <= MAX_OL {
```
djc, line 408: Should be `<`? Please add some tests for the edge cases here...

src/naive/internals.rs
```diff
@@ -368,33 +371,37 @@ impl fmt::Debug for Of {
...
 
     #[cfg(test)]
-    pub(super) fn valid(&self) -> bool {
+    pub(super) const fn valid(&self) -> bool {
         let Mdf(mdf) = *self;
         let mdl = mdf >> 3;
-        match MDL_TO_OL.get(mdl as usize) {
-            Some(&v) => v >= 0,
-            None => false,
+        if mdl <= MAX_MDL {
```
djc, line 420: `<=` -> `<`?

src/weekday.rs
djc: Given the `mod7` requirement which is not enforced in the type system and the fact that this is used only in the `internals` module, I think this should be a free function in the `internals` module rather than a method here.

src/naive/date.rs
```diff
@@ -242,19 +242,29 @@ impl NaiveDate {
...
-    /// Makes a new `NaiveDate` from year and packed ordinal-flags, with a verification.
-    fn from_of(year: i32, of: Of) -> Option<NaiveDate> {
-        if (MIN_YEAR..=MAX_YEAR).contains(&year) && of.valid() {
-            let Of(of) = of;
-            Some(NaiveDate { ymdf: (year << 13) | (of as DateImpl) })
-        } else {
-            None
+
+    /// Makes a new `NaiveDate` from year, ordinal and flags.
+    /// Does not check whether the flags are correct for the provided year.
```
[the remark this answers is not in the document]
djc: Yes, let's do some `debug_assert!()`ing here.

src/naive/date.rs
```diff
@@ -242,19 +242,29 @@ impl NaiveDate {
...
+    const fn from_of(year: i32, ordinal: u32, flags: YearFlags) -> Option<NaiveDate> {
```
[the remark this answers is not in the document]
djc: This combination should be a separate commit with no other changes.

src/weekday.rs
```diff
@@ -260,6 +260,14 @@ mod tests {
             );
         }
     }
+
+    #[test]
+    fn test_from_u32_mod7() {
```
djc, line 265: Please squash this in the commit where the function is added.

src/naive/internals.rs
```diff
@@ -266,6 +266,8 @@ const OL_TO_MDL: &[u8; MAX_OL as usize + 1] = &[
 ///
 /// The whole bits except for the least 3 bits are referred as `Ol` (ordinal and leap flag),
 /// which is an index to the `OL_TO_MDL` lookup table.
+///
+/// The methods implemented on `Of` always return a valid value.
```
djc, line 270: Please squash this change into the commit that originated the described changes.

src/naive/internals.rs
```diff
@@ -271,27 +271,30 @@ pub(super) struct Of(pub(crate) u32);
...
     }
 
     #[inline]
-    pub(super) fn from_mdf(Mdf(mdf): Mdf) -> Of {
+    pub(super) const fn from_mdf(Mdf(mdf): Mdf) -> Of {
         let mdl = mdf >> 3;
-        match MDL_TO_OL.get(mdl as usize) {
-            Some(&v) => Of(mdf.wrapping_sub((i32::from(v) as u32 & 0x3ff) << 3)),
-            None => Of(0),
+        if mdl <= MAX_MDL {
```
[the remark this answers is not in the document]
pitdicker: The arrays in this file are a bit strange: index `0` has an invalid value, and the length is `MAX + 1`.
It makes some sense, as we encode ordinals and day of month starting from 1. This allows indexing without the need to subtract -1.

src/naive/internals.rs
pitdicker: Same array with an unused zero index. There is a test with the edge cases.

src/naive/internals.rs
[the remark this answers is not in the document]
djc: Okay, can you add a comment to each comparison where this is relevant?

src/naive/date.rs
```diff
@@ -241,19 +241,30 @@ impl NaiveDate {
...
+        if year < MIN_YEAR || year > MAX_YEAR {
+            return None; // Out-of-range
+        }
+        debug_assert!(YearFlags::from_year(year).0 == flags.0);
```
pitdicker, line 251: This needs 1.57. Shall I comment it out for now, and add a note to enable it on a raise of the MSRV?

pitdicker: Updated.

src/naive/internals.rs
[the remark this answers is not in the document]
djc: Would like the comment on the same line as the `if`, can be shorter like `// 1-based indexing`.

src/naive/internals.rs
```diff
@@ -467,11 +466,28 @@ impl fmt::Debug for Mdf {
     }
 }
 
+/// Create a `Weekday` from an `u32`, with Monday = 0.
+/// Infallible, takes `n % 7`.
```
djc, line 491: The `takes n % 7` comment is no longer correct, right?

src/naive/internals.rs
```diff
@@ -824,4 +840,12 @@ mod tests {
             }
         }
     }
+
+    #[test]
+    fn test_weekday_from_u32_mod7() {
+        for i in 0..=1000 {
+            assert_eq!(weekday_from_u32_mod7(i), Weekday::from_u32(i % 7).unwrap());
+        }
+        assert_eq!(weekday_from_u32_mod7(u32::MAX), Weekday::Thu);
```
djc, line 893: I don't think there's any point to testing with values larger than 7.

src/naive/internals.rs
```diff
@@ -339,16 +352,20 @@ impl Of {
...
+    /// Returns an `Of` with the previous day, or `None` if this is the first day of the year.
     #[inline]
-    pub(super) const fn pred(&self) -> Of {
-        let Of(of) = *self;
-        Of(of - (1 << 4))
+    pub(super) const fn pred(&self) -> Option<Of> {
+        if self.ordinal() == 1 {
+            return None;
+        }
+        Some(Of(self.0 - (1 << 4)))
```
djc, line 368: Style: I prefer writing this as

```rust
match self.ordinal() {
    1 =>  None,
    _ => Some(Of(self.0 - (1 << 4)))
}
```

src/naive/internals.rs
```diff
@@ -370,13 +387,17 @@ impl fmt::Debug for Of {
 /// The whole bits except for the least 3 bits are referred as `Mdl`
 /// (month, day of month and leap flag),
 /// which is an index to the `MDL_TO_OL` lookup table.
+///
+/// The methods implemented on `Mdf` do not always return a valid value.
+/// Dates than can't exist, like February 30, can still be represented.
```
djc, line 392: "than" -> "that"

Why do we need to allow `Mdf` instances with an invalid value?

src/naive/internals.rs
```diff
@@ -662,8 +683,7 @@ mod tests {
     fn test_of_fields() {
         for &flags in FLAGS.iter() {
             for ordinal in range_inclusive(1u32, 366) {
-                let of = Of::new(ordinal, flags).unwrap();
-                if of.valid() {
+                if let Some(of) = Of::new(ordinal, flags) {
```
djc, line 686: For tests, let's just `unwrap()` cases like these?

src/naive/date.rs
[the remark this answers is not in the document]
djc: Should we also rename this method? The current name doesn't really fit anymore.

pitdicker:
> Would like the comment on the same line as the `if`, can be shorter like `// 1-based indexing`.

I would also like to put it on the same line, but rustfmt disagrees with me.

src/naive/internals.rs
pitdicker: The intention was to say it is infallible, because it takes any `n` and does modulus 7. I'll see if I can word it better.

src/naive/internals.rs
pitdicker: That looks very clean, thank you 👍

src/naive/internals.rs
pitdicker: Testing whether an ordinal is valid is easy. Testing a month, day, leap year flag combination is valid, and doing it fast, is more involved.

The table lookup to convert `Mdf` to `Of` is pretty much the ideal place to put the validity check.

src/naive/internals.rs
pitdicker: We can't. The test is creating all possible combinations of year flags and ordinals. But some of these combinations are just invalid. It tests if those that are let through, have an ordinal that round-trips.

pitdicker:
> Should we also rename this method? The current name doesn't really fit anymore.

(Some of your comments are missing a reply field for some reason.)

I suppose it can be called `from_ordinal_and_flags`. I quite liked the short name, but agree it can be confusing.

pitdicker:
@djc Just to make sure, you are not waiting on me?

pitdicker:
# :tada:
Time to get my branch that makes most of `NaiveDate` and `NaiveTime` in shape.

Issue #960, opened 2023-02-13
%Z can't print the timezone abbreviation


howjmay:
Hi I am having a use case that I need to generate formatted time, and the text should contain the abbreviation of the timezone.
I am using `format_with_items()`, and a format string `"%a %b %d %I:%M:%S  '%Z' %p %Y"`. However, `%Z` doesn't return the abbreviation. What should I do?

djc:
This issue doesn't come with a reproducible example, nor does it show the actual output. Please extend your bug report.

howjmay:
@djc An example is like following 
```rust
use chrono::{format::StrftimeItems, Local};

fn main() {
    let now = Local::now();
    now.with_timezone(now.offset());
    let formatted = now
        .format_with_items(StrftimeItems::new("%a %b %d %I:%M:%S '%Z' %Y"))
        .to_string();
    println!("{formatted}");
}
```

expected result
```
Tue Feb 14 00:55:01 'CET' 2023
```

actual result
```
Tue Feb 14 12:56:24 '+08:00' 2023
```

The expected abbreviation `CET` is replaced by `+08:00`. Or I should use other escape character then I can get the abbreviation of timezone?

djc:
chrono itself doesn't know timezones beyond their offsets, so I don't think it can do what you want. Maybe look at the chrono-tz crate and see if it can help with this?

esheppa:
@howjmay - please see also this issue: https://github.com/chronotope/chrono/issues/749.

`chrono-tz` should do this correctly as per [here](https://github.com/chronotope/chrono-tz/blob/b1104001236cecc51123f857d91519f2afc2b08e/chrono-tz/src/timezone_impl.rs#L29) as the display implementation of the `TimeZone` type is called. You can use `chrono-tz` and `iana-time-zone` (a dependency of `chrono`) together to have behaviour equivalent to local. First use `iana-time-zone` to get the name of the locally set timezone, then you can parse that into a `chrono_tz::Tz`, as per the [example](https://docs.rs/iana-time-zone/latest/iana_time_zone/):

```rust
let tz_str = iana_time_zone::get_timezone()?;
let tz: chrono_tz::Tz = tz_str.parse()?;
```

There is also a partially complete PR to solve this https://github.com/chronotope/chrono/pull/750 and we would appreciate any thoughts or comments you have on it.

howjmay:
Hi @esheppa the method you provided is good. 
[...]

PR #1051, opened 2023-05-07
Add documentation for %Z quirk

campbellcole:
Adds documentation to `strftime` to clearly state what `%Z` actually does when formatting.

Related to #960.

### DIFF
--- a/src/format/strftime.rs
+++ b/src/format/strftime.rs
@@ -68,7 +68,7 @@ The following specifiers are available both to formatting and parsing.
 | `%r`  | `12:34:60 AM` | Hour-minute-second format in 12-hour clocks. Same as `%I:%M:%S %p`.   |
 |       |          |                                                                            |
 |       |          | **TIME ZONE SPECIFIERS:**                                                  |
-| `%Z`  | `ACST`   | Local time zone name. Skips all non-whitespace characters during parsing. [^8] |
+| `%Z`  | `ACST`   | Local time zone name. Skips all non-whitespace characters during parsing. Identical to `%:z` when formatting. [^8] |
 | `%z`  | `+0930`  | Offset from the local time to UTC (with UTC being `+0000`).                |
 | `%:z` | `+09:30` | Same as `%z` but with a colon.                                             |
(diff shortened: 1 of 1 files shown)


djc:
Thanks!

pitdicker:
Sorry I am late to the party. And I am still learning many details myself.

Chrono depends on the `Display` implementation of the type that implements `TimeZone` to format `%Z`. But that implementation is currently lacking for `Local`.
- `FixedOffset` prints the offset
- `Utc` prints "UTC"
- `Local` could make it available on Unix with the work of @esheppa in https://github.com/chronotope/chrono/pull/750
- `Local` on Windows does not seem easily possible.
- Do the timezone types in `chrono-tz` and `tzfile` not just work?

src/format/strftime.rs
```diff
@@ -68,7 +68,7 @@ The following specifiers are available both to formatting and parsing.
 | `%r`  | `12:34:60 AM` | Hour-minute-second format in 12-hour clocks. Same as `%I:%M:%S %p`.   |
 |       |          |                                                                            |
 |       |          | **TIME ZONE SPECIFIERS:**                                                  |
-| `%Z`  | `ACST`   | Local time zone name. Skips all non-whitespace characters during parsing. [^8] |
+| `%Z`  | `ACST`   | Local time zone name. Skips all non-whitespace characters during parsing. Identical to `%:z` when formatting. [^8] |
```
pitdicker, line 71: Maybe this could say something like:
"Time zone name. Depends on the `Display` implementation of the type implementing `TimeZone`. Skips all non-whitespace characters during parsing.

campbellcole:
>     * Do the timezone types in `chrono-tz` and `tzfile` not just work?

This is a very good question. I will try this out right now and perhaps submit a new pull request with documentation for this. I completely forgot all `Tz` variants implement `TimeZone`, and I haven't looked at their `Display` impls yet.

src/format/strftime.rs
djc: Yeah, that would probably be an improvement.

PR #1052, opened 2023-05-07
Fix panic in from_num_days_from_ce_opt

pitdicker:
The author in https://github.com/chronotope/chrono/pull/527 does not seem to be available anymore (no blame after two years)


djc (approved):

djc:
Thanks!

PR #1053, opened 2023-05-08
Bump the MSRV to 1.56

djc:
@pitdicker want to review?


pitdicker:
Yes 😄. I'll look at it today.

Would jumping to 1.57 be possible? That would be very useful because it allows panicking in const, which spares messy workarounds. And it would 'sell' the higher MSRV requirement.

src/naive/date.rs
```diff
@@ -2249,7 +2249,7 @@ mod tests {
         assert_eq!(
             NaiveDate::from_ymd_opt(2022, 8, 3)
                 .unwrap()
-                .checked_sub_months(Months::new((i32::MIN as i64).abs() as u32 + 1)),
+                .checked_sub_months(Months::new((i32::MIN as i64).unsigned_abs() as u32 + 1)),
```
pitdicker, line 2252:
```suggestion
                .checked_sub_months(Months::new(i32::MIN.unsigned_abs() + 1)),
```

This can now work without the casts.

src/format/scan.rs
```diff
@@ -316,10 +316,7 @@ pub(super) fn timezone_offset_2822(s: &str) -> ParseResult<(&str, Option<i32>)>
     let upto = s
         .as_bytes()
         .iter()
-        .position(|&c| match c {
-            b'a'..=b'z' | b'A'..=b'Z' => false,
-            _ => true,
-        })
+        .position(|&c| !matches!(c, b'a'..=b'z' | b'A'..=b'Z'))
```
pitdicker, line 319:
```suggestion
        .position(|&c| !c.is_ascii_alphabetic())
```

Suggested by clippy

Cargo.toml
```diff
@@ -11,6 +11,7 @@ readme = "README.md"
 license = "MIT/Apache-2.0"
 exclude = ["/ci/*"]
 edition = "2018"
```
pitdicker, line 13: Maybe update to edition 2021 while at it?

pitdicker:
The `matches` macro in `lib.rs` has a comment that in can be removed with the higher MSRV.

pitdicker (approved):
What do you think about going to 1.57?

djc:
> What do you think about going to 1.57?

I think we can do that as part of a PR that actually takes advantage of the newly stable API/language surface.

PR #1061, opened 2023-05-12
Gate test on `clock` feature

pitdicker:
Both https://github.com/chronotope/chrono/pull/1041 and https://github.com/chronotope/chrono/pull/1059 had a green CI, but combined 0.4.x has a test that fails to compile.


djc (approved):
Thanks for fixing that up.

PR #1062, opened 2023-05-12
[CI] Lint Windows target

pitdicker:


djc (approved):

pitdicker:
Interesting: only the first line of the `run` actions is executed. The following where not run on CI before:
```
cargo fmt --check --manifest-path fuzz/Cargo.toml
cargo clippy --manifest-path fuzz/Cargo.toml --color=always -- -D warnings
```

pitdicker:
Added cargo-deadlinks to the CI.

It found a couple of URLs that were broken. They all relative links on types/implementations such as:
```rust
impl Add<OldDuration> for NaiveDate
```

This implementation will be placed on the page of `NaiveDate`, and on the page of `OldDuration`. It is easy for a relative URL to break in such cases.

jtmoon79 (approved):
LGTM

esheppa (approved):

.github/workflows/lint.yml
```diff
@@ -17,8 +17,8 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - run: cargo fmt --check -- --color=always
       - run: cargo fmt --check --manifest-path fuzz/Cargo.toml
-      - run: cargo clippy --color=always -- -D warnings -A clippy::manual-non-exhaustive
-      - run: cargo clippy --color=always --target x86_64-pc-windows-msvc -- -D warnings -A clippy::manual-non-exhaustive
+      - run: cargo clippy --color=always -- -D warnings
```
djc, line 20: Would prefer to move this commit before the commit that duplicates this stuff for Windows.

PR #1000, opened 2023-03-24
Constrain timezone parsing

jtmoon79:
Constrain timezone parsing further. Only allow optional colon char `":"` between timezone hour offset and timezone minute offset. Previously these patterns were optionally allowed: `":"`, `": "`, `" :"`, `" : "`, `" "`.

Issue #660

Follow-up to discussion https://github.com/chronotope/chrono/pull/807#issuecomment-1465342408

PR #1063, opened 2023-05-12
Backport #983 and #1000

pitdicker:
Backports https://github.com/chronotope/chrono/pull/1000 and https://github.com/chronotope/chrono/pull/983, which are both backwards-compatible.
It would be helpful for me if there were available on the 0.4.x branch.


jtmoon79 (approved):
LGTM

esheppa:
Thanks @pitdicker - this is an interesting point as these were only merged into `main` first due to the previous MSRV, so there are potentially a few other PRs that should be given this treatment also. 

What are your thoughts on us finding all of those and bringing them into this PR to make sure none are forgotten?

Also @djc and I worked through a few different options when originally setting up the strategy that allows commits in `0.4.x` to get through to `main`, and I'd like to investigate that this doesn't cause any problems for that (eg essentially the same code showing up twice due to it already being in `main`)

djc:
I think git will be able to deal with this when merging.

pitdicker:
@djc Thanks for merging :smile:

> What are your thoughts on us finding all of those and bringing them into this PR to make sure none are forgotten?

I wanted to port the `TryFrom` implementations from main to 0.4.x, but that was part of a commit that also removed the `FromPrimitive` implementations. I'll part that as half a commit in another PR.

PR #1065, opened 2023-05-16
Merge 0.4.x

pitdicker:
It seems like the merged PRs caused more merge conflicts with main than changed lines :smile: (exaggeration of course).

I did not intend to make this PR, but was curious how bad it would turn out to be. And now the work was done anyway.


djc:
Thanks!

PR #1066, opened 2023-05-16
Backports

pitdicker:
Three more backports. One to port the `TryFrom` implementations for `Weekday` and `Month`, and two to reduce the difference with main.

@esheppa There are now only a handful of breaking changes that make main different from 0.4.x (and two places where a change in 0.4.x didn't make the merge to main, I'll have a look at those).

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -552,11 +552,13 @@ where
 }
 
 impl DateTime<FixedOffset> {
-    /// Parses an RFC 2822 date and time string such as `Tue, 1 Jul 2003 10:52:37 +0200`,
-    /// then returns a new [`DateTime`] with a parsed [`FixedOffset`].
+    /// Parses an RFC 2822 date-and-time string into a `DateTime<FixedOffset>` value.
     ///
-    /// RFC 2822 is the internet message standard that specifies the
-    /// representation of times in HTTP and email headers.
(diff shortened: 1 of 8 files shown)


djc:
IIRC the num-traits trait implementations are part of the public API, so we can't remove them in 0.4.x.

djc:
Also please link to the original PRs so I can page in discussion from the original merges.

pitdicker:
You are right, I should have put more effort into the PR description.

The `TryFrom` implementations were part of the commit "Remove dependency on num-traits" in https://github.com/chronotope/chrono/pull/794, which was merged in https://github.com/chronotope/chrono/pull/794. I have removed the part that drops `num-traits` as a dependency from the public API from that commit.

The two documentation changes are from https://github.com/chronotope/chrono/pull/887 and https://github.com/chronotope/chrono/pull/806 (only taken the documentation commit).

The serde commit is from https://github.com/chronotope/chrono/pull/669. I now see that is a tricky commit, maybe best to remove it:

> This is technically a breaking change but these structs were hidden from the docs.

djc:
> The serde commit is from #669. I now see that is a tricky commit, maybe best to remove it:

Yes, let's drop that one, too.

pitdicker:
:+1: Done. It is better to be careful.

djc (approved):

pitdicker:
@djc Thank you for all your review work.

Issue #196, opened 2017-12-11
Deserializing using serde deserialize does not work for String timestamp


Gricha:
Hello!

I was curious whether it's a bug, a design choice, or my incorrect usage of the framework and would definitely appreciate any help and guidance on this one!

I'm using chrono to support deserialization of time formats in my parsing layer. One of the APIs I am currently using sends me timestamps as string. Example
```json
{
  "timestamp": "1512954499"
}
```

I'd like to represent said JSON in my struct as:
```rust
#[derive(Debug, Deserialize)]
pub struct Data {
    timestamp: DateTime<Utc>,
}
```

I noticed that there is an attempt to support such situation by using the following function:
https://github.com/chronotope/chrono/blob/master/src/datetime.rs#L750

Unfortunately I was not able to get that working with strings. The following implementation of struct does not end up parsing the JSON properly:
```rust
#[derive(Debug, Deserialize)]
pub struct Data {
    #[serde(deserialize_with = "from_ts")]
    timestamp: DateTime<Utc>,
}
```

I was curious whether it's a bug or maybe my setup is plain wrong?

The one thing that I've noticed is that this Deserialize function directly uses `SecondsTimestampVisitor` to parse out the timestamps. That Visitor though, if I'm reading the code correctly, does not implement either `visit_str` or `visit_string`. I'm not sure if that's intended.

Sorry if it turns out to be a silly question, I'm still quite new to Rust! :)

quodlibetor:
The good news: you are correct about the `ts_seconds` module only working with integers, and you're also correct about all the reasons!

The tl;dr: I recommend that you copy/paste the `serde::ts_seconds` module out of chrono and into your own code, with your required modifications, but if you have evidence of this being a common pattern I'm happy to reconsider.

I *think* that string timestamps are far enough away from a common occurrence that I'm hesitant to add code to support them to `chrono`.

If you copy/paste the `ts_seconds` module (from what you've said you really only need the deserialize portions of it, I think) into your own code and add a `visit_str` method that looks sort of like:

```rust
// untested!
fn visit_str<E>(self, value: &str) -> Result<DateTime<FixedOffset>, E>
    where E: de::Error
{
    let ts: i64 = value.parse()
        .map_err(|e| E::custom(format!("Unable to parse timestamp as int: {}", e)))?;
    self.visit_i64(ts)
}
```

And then call that from your copied in deserialize function, you should be golden. Or at least a lighter shade than "burnt".

[...]

PR #1067, opened 2023-05-16
(WIP) Add support for serialising and deserialising string timestamps

sinking-point:
Work in progress. Do not merge yet.

This PR aims to address https://github.com/chronotope/chrono/issues/196#issuecomment-1549430697

- [x] Implement ts_seconds_string (in DateTime)
- [x] Implement ts_milliseconds_string
- [x] Implement ts_microseconds_string
- [x] Implement ts_nanoseconds_string
- [ ] Add optionals (e.g. ts_seconds_string_option)
- [ ] Support NaiveDateTime as well


sinking-point:
@djc Is my work so far on the right track? Would you like to request any changes to my approach?

djc (commented):
The approach looks okay to me, please clean up the commit history a bit by squashing (I usually use an interactive rebase).

sinking-point:
Thanks for the feedback @djc . There might be some more messy commits as I continue working on this, but I'll ensure to clean them up when I'm done.

djc:
I've changed this to draft status for now, please switch it to ready once you think it's ready (or just comment if you want feedback on something sooner).

pitdicker:
@sinking-point I have only glanced at the code, apologies for that.
The amount of lines needed for (de)serializing is massive! Is there a way to collapse that into something manageable with a macro?

sinking-point:
@pitdicker A lot of those lines are documentation and tests. I could potentially try to define a macro, but the tests and docs would remain.

sinking-point:
The reason there's so much repetition is that there are 4 dimensions: int/string, normal/optional, DateTime/NaiveDateTime, nano/micro/milli/seconds. I agree that this isn't very DRY and it would be good to find a DRYer way of doing it.

sinking-point:
TBH this is difficult without extending serde's [field attributes](https://serde.rs/field-attrs.html). It would be nice if there was a `via` attribute that takes another type (e.g. `TimestampNanoseconds`). Serde would deserialise into that first, then `into()` it to the target type (e.g. `DateTime`).

sinking-point:
There is already a similar feature in the [container attributes](https://serde.rs/container-attrs.html): `from` and `into`.

sinking-point:
Here's the usage I'm envisioning:

```rust
use chrono::{TimeZone, DateTime, Utc, NaiveDate};
use serde_derive::{Deserialize, Serialize};
use chrono::serde::{Timestamp, Nanoseconds};
#[derive(Deserialize, Serialize)]
struct S {
    #[serde(via = "Timestamp<Nanoseconds, i64>")] // could equally be Timestamp<Seconds, String>
    time: DateTime<Utc> // could equally be NaiveDateTime - both would implement From<Timestamp>
}
```

Maybe Timestamp could even take a timezone parameter as well, which would allow automatic conversion to the timezone of the DateTime.

To reiterate, this `via` attribute doesn't exist yet. We'd need to get it into `serde` for this idea to be possible.

sinking-point:
We could, but if we still want to have docs and tests for each variant then it wouldn't reduce the line count by all that much.

In my opinion, tests should be concrete. I feel that tests are best when they're a self-contained, readable specification of what a module should do. Also, one function of the tests would be to test the assumptions made by the programmer in defining the macros. However, if the tests are also defined by a macro then they will be influenced by those same assumptions, which might cancel out.

Have you seen my [comment on the issue](https://github.com/chronotope/chrono/issues/196#issuecomment-1554134522)? It seems that implementing my proposal would almost just be reimplementing part of the `serde_with` crate.

djc:
I did see it. I'm fine with closing this PR or maybe substituting it with a documentation pointer to the serde_with crate.

src/datetime/serde.rs
```diff
@@ -889,70 +991,1066 @@ pub mod ts_milliseconds_option {
...
+        fn test_deserialize_ts_microseconds_string() -> Result<(), serde_json::Error> {
+            const JSON: &'static str = r#"{"time":"1526522699918355"}"#;
+
+            let my_s: S = serde_json::from_str(JSON)?;
+
+            assert_eq!(my_s.time.timestamp_micros(), 1526522699918355);
+
+            Ok(())
+        }
+    }
```
jtmoon79, line 1157: Same question about maximum and minimum values, and the value immediately outside those boundaries.

src/datetime/serde.rs
```diff
@@ -889,70 +991,1066 @@ pub mod ts_milliseconds_option {
...
+        fn test_deserialize_ts_microseconds_string_option_none() -> Result<(), serde_json::Error> {
+            const JSON: &'static str = r#"{"time":null}"#;
+
+            let my_s: S = serde_json::from_str(JSON)?;
+
+            assert_eq!(my_s.time, None);
+
+            Ok(())
+        }
+    }
```
jtmoon79, line 1357: Same question about maximum and minimum values, and the value immediately outside those boundaries.

src/datetime/serde.rs
```diff
@@ -889,70 +991,1066 @@ pub mod ts_milliseconds_option {
...
+        fn test_deserialize_ts_milliseconds_string() -> Result<(), serde_json::Error> {
+            const JSON: &'static str = r#"{"time":"1526522699918"}"#;
+
+            let my_s: S = serde_json::from_str(JSON)?;
+
+            assert_eq!(my_s.time.timestamp_millis(), 1526522699918);
+
+            Ok(())
+        }
+    }
```
jtmoon79, line 1787: Same question about maximum and minimum values, and the value immediately outside those boundaries.

src/datetime/serde.rs
```diff
@@ -889,70 +991,1066 @@ pub mod ts_milliseconds_option {
...
+        fn test_deserialize_ts_milliseconds_string_option_none() -> Result<(), serde_json::Error> {
+            const JSON: &'static str = r#"{"time":null}"#;
+
+            let my_s: S = serde_json::from_str(JSON)?;
+
+            assert_eq!(my_s.time, None);
+
+            Ok(())
+        }
+    }
```
jtmoon79, line 1987: Same question about maximum and minimum values, and the value immediately outside those boundaries.

src/datetime/serde.rs
```diff
@@ -1137,6 +2235,353 @@ pub mod ts_seconds_option {
...
+        fn test_deserialize_ts_seconds_string() -> Result<(), serde_json::Error> {
+            const JSON: &'static str = r#"{"time":"1431684000"}"#;
+
+            let my_s: S = serde_json::from_str(JSON)?;
+
+            assert_eq!(my_s.time.timestamp(), 1431684000);
+
+            Ok(())
+        }
+    }
```
jtmoon79, line 2391: Same question about maximum and minimum values, and the value immediately outside those boundaries.

src/naive/datetime/serde.rs
```diff
@@ -1020,39 +2422,100 @@ pub mod ts_seconds_option {
...
+        fn test_deserialize_naive_ts_seconds_string_option_none() -> Result<(), serde_json::Error> {
+            const JSON: &'static str = r#"{"time":null}"#;
+
+            let my_s: S = serde_json::from_str(JSON)?;
+
+            assert_eq!(my_s.time, None);
+
+            Ok(())
+        }
+    }
```
jtmoon79, line 2518: Same question about testing maximum and minimum boundaries, and values immediately outside those boundaries.

src/naive/datetime/serde.rs
```diff
@@ -904,62 +2266,102 @@ pub mod ts_seconds {
...
+
+        #[test]
+        fn test_deserialize_naive_ts_seconds_string() -> Result<(), serde_json::Error> {
+            const JSON: &'static str = r#"{"time":"1431684000"}"#;
+
+            let my_s: S = serde_json::from_str(JSON)?;
+
+            assert_eq!(my_s.time.timestamp(), 1431684000);
+
+            Ok(())
```
jtmoon79, line 2318: Same question about testing maximum and minimum boundaries, and values immediately outside those boundaries.

src/naive/datetime/serde.rs
```diff
@@ -505,92 +1794,160 @@ pub mod ts_microseconds_option {
...
+        ) -> Result<(), serde_json::Error> {
+            const JSON: &'static str = r#"{"time":null}"#;
+
+            let my_s: S = serde_json::from_str(JSON)?;
+
+            assert_eq!(my_s.time, None);
+
+            Ok(())
+        }
+    }
```
jtmoon79, line 1915: Same question about testing maximum and minimum boundaries, and values immediately outside those boundaries.

src/naive/datetime/serde.rs
```diff
@@ -428,41 +797,961 @@ pub mod ts_microseconds {
...
+        fn test_deserialize_naive_ts_milliseconds_string() -> Result<(), serde_json::Error> {
+            const JSON: &'static str = r#"{"time":"1526522699918"}"#;
+
+            let my_s: S = serde_json::from_str(JSON)?;
+
+            assert_eq!(my_s.time.timestamp_millis(), 1526522699918);
+
+            Ok(())
+        }
+    }
```
jtmoon79, line 1710: Same question about testing maximum and minimum boundaries, and values immediately outside those boundaries.

jtmoon79 (changes requested):
This should have tests for maximum and minimum values, and more importantly, values just outside those boundaries.

PR #1068, opened 2023-05-17
Add more documentation to `TimeZone`

pitdicker:
Split out from https://github.com/chronotope/chrono/pull/1040.

### DIFF
--- a/src/offset/mod.rs
+++ b/src/offset/mod.rs
@@ -209,10 +209,118 @@ pub trait Offset: Sized + Clone + fmt::Debug {
     fn fix(&self) -> FixedOffset;
 }
 
-/// The time zone.
+/// The [`TimeZone`] trait is the backbone of chrono, allowing it to work with local time.
 ///
-/// The methods here are the primarily constructors for [`Date`](../struct.Date.html) and
-/// [`DateTime`](../struct.DateTime.html) types.
+/// # Using `TimeZone`
+///
+/// Most methods of this trait are deprecated, it is usually better or easier to work with the
+/// methods on [`DateTime`] and [`NaiveDateTime`] instead.
+///
+/// A type implementing [`TimeZone`] can be used to create a [`DateTime`]:
+///
+/// Creating a [`DateTime`] from a [`NaiveDateTime`]:
+/// - [`from_local_datetime`](#method.from_local_datetime) if the datetime is in local time zone.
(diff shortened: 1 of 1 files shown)


src/offset/mod.rs
```diff
@@ -209,10 +209,119 @@ pub trait Offset: Sized + Clone + fmt::Debug {
...
-/// [`DateTime`](../struct.DateTime.html) types.
+/// # Using `TimeZone`
+///
+/// Most methods of this trait are deprecated, it is usually better or easier to work with the
+/// methods on [`DateTime`] and [`NaiveDateTime`] instead.
+///
+/// A type implementing [`TimeZone`] can be used to create a [`DateTime`]:
+///
+/// Creating a [`DateTime`] from a [`NaiveDateTime`]:
+/// - [`from_local_datetime`](#method.from_local_datetime) if the datetime is local.
```
jtmoon79, line 222: ... is in local timezone.

src/offset/mod.rs
```diff
@@ -209,10 +209,119 @@ pub trait Offset: Sized + Clone + fmt::Debug {
...
+/// - [`from_utc_datetime`](#method.from_utc_datetime) if the datetime is in utc.
```
jtmoon79, line 223: ... is in UTC timezone.

src/offset/mod.rs
```diff
@@ -209,10 +209,119 @@ pub trait Offset: Sized + Clone + fmt::Debug {
...
+/// be given as either local or UTC.
+///
+/// [`from_offset`](#tymethod.from_offset) allows the type to be reconstructed from the associated
+/// [`Offset`] type, which gets stored in a [`DateTime`].
+///
+/// ## Example
+///
+#[cfg_attr(not(feature = "clock"), doc = "```ignore")]
+#[cfg_attr(feature = "clock", doc = "```rust")]
+/// use chrono::{LocalResult, Offset, TimeZone};
```
jtmoon79, line 251: Why not one `use` statement?

jtmoon79 (approved):
Approve with nitpick comments.

src/offset/mod.rs
pitdicker: I had some reasoning... But if I can't remember it must not have been that good :smile:.

pitdicker:
I'll give this another try in the future.

PR #1069, opened 2023-05-18
`(overflowing|checked)_(add|sub)_offset` implementations

pitdicker:
Split out from https://github.com/chronotope/chrono/pull/1048.

The main goal is to add `NaiveDateTime::checked_(add|sub)_offset` methods. To quote from https://github.com/chronotope/chrono/pull/1048:

Instead of panicking in the `Add` or `Sub` implementation of `NaiveDateTime` with `FixedOffset`, we need a way to be informed of out-of-range result.

I added the following methods ~(not public for now)~:
- `NaiveTime::overflowing_add_offset` and `NaiveTime::overflowing_sub_offset`
- `NaiveDateTime::checked_add_offset` and `NaiveDateTime::checked_sub_offset`

The `Add` and `Sub` implementations of `FixedOffset` with `NaiveTime`, `NaiveDateTime` and `DateTime` are reimplemented using of these methods. This fixes a code comment:
> this should be implemented more efficiently, but for the time being, this is generic right now.

The best place to put the `Add` and `Sub` implementations for `DateTime` is in the module of `DateTime`, because there they have access to its private fields. I have moved all implementations to the module of their output type for consistency.

-------

Adding an offset works differently from adding a duration. Adding a duration tries to do something smart—but still rudimentary—with leap seconds. Adding an offset should not touch the leap seconds at all. So the methods that operate on `NaiveTime` should be different.

I extracted the part that operates on the `NaiveDate` that could be shared into an `add_days` methods. Previously `NaiveDate::checked_add_days` would convert the days to a `Duration` in seconds, and later devide them to get the value in days again. This now works in a less roundabout way. ~Might~ This would also help with the const implementation later.

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -1247,22 +1247,13 @@ impl<Tz: TimeZone> AddAssign<Duration> for DateTime<Tz> {
     }
 }
 
-fn add_with_leapsecond<T>(lhs: &T, rhs: i32) -> T
-where
-    T: Timelike + Add<OldDuration, Output = T>,
-{
-    // extract and temporarily remove the fractional part and later recover it
-    let nanos = lhs.nanosecond();
-    let lhs = lhs.with_nanosecond(0).unwrap();
-    (lhs + OldDuration::seconds(i64::from(rhs))).with_nanosecond(nanos).unwrap()
-}
-
 impl<Tz: TimeZone> Add<FixedOffset> for DateTime<Tz> {
     type Output = DateTime<Tz>;
 
     #[inline]
-    fn add(self, rhs: FixedOffset) -> DateTime<Tz> {
-        add_with_leapsecond(&self, rhs.local_minus_utc())
+    fn add(mut self, rhs: FixedOffset) -> DateTime<Tz> {
+        self.datetime = self.naive_utc().checked_add_offset(rhs).unwrap();
+        self
     }
 }
 
@@ -1317,8 +1308,9 @@ impl<Tz: TimeZone> Sub<FixedOffset> for DateTime<Tz> {
     type Output = DateTime<Tz>;
 
     #[inline]
-    fn sub(self, rhs: FixedOffset) -> DateTime<Tz> {
-        add_with_leapsecond(&self, -rhs.local_minus_utc())
+    fn sub(mut self, rhs: FixedOffset) -> DateTime<Tz> {
+        self.datetime = self.naive_utc().checked_sub_offset(rhs).unwrap();
+        self
     }
 }
 
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -1247,6 +1247,25 @@ impl<Tz: TimeZone> AddAssign<Duration> for DateTime<Tz> {
     }
 }
 
+fn add_with_leapsecond<T>(lhs: &T, rhs: i32) -> T
+where
+    T: Timelike + Add<OldDuration, Output = T>,
+{
+    // extract and temporarily remove the fractional part and later recover it
+    let nanos = lhs.nanosecond();
+    let lhs = lhs.with_nanosecond(0).unwrap();
+    (lhs + OldDuration::seconds(i64::from(rhs))).with_nanosecond(nanos).unwrap()
+}
+
+impl<Tz: TimeZone> Add<FixedOffset> for DateTime<Tz> {
+    type Output = DateTime<Tz>;
+
+    #[inline]
+    fn add(self, rhs: FixedOffset) -> DateTime<Tz> {
+        add_with_leapsecond(&self, rhs.local_minus_utc())
+    }
+}
+
 impl<Tz: TimeZone> Add<Months> for DateTime<Tz> {
     type Output = DateTime<Tz>;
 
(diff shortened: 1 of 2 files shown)


src/offset/fixed.rs
```diff
@@ -169,75 +165,6 @@ impl arbitrary::Arbitrary<'_> for FixedOffset {
     }
 }
 
-// addition or subtraction of FixedOffset to/from Timelike values is the same as
```
djc, line 172: Please don't move code and change it in the same commit, this makes it very hard to review.

pitdicker: I have split it up into multiple commits, including the parts from your comment in https://github.com/chronotope/chrono/pull/1071#discussion_r1198850204.

src/naive/datetime/mod.rs
```diff
@@ -732,6 +741,16 @@ impl NaiveDateTime {
         Some(NaiveDateTime { date, time })
     }
 
+    /// Subtracts given [`FixedOffset`] from the current datetime.
```
jtmoon79, line 744: /// Subtracts given `rhs` from the current datetime.

The docstring will generate a link to the type `FixedOffset` in the function signature.

(If I ever contradicted myself in prior reviews with this advice; sorry, I realized this a few weeks ago that I was unnecessarily "overlinking" to types in docstrings. `cargo doc` will always generate links to the crate types in the function signature).

src/naive/datetime/mod.rs
```diff
@@ -658,6 +657,16 @@ impl NaiveDateTime {
         Some(Self { date: self.date.checked_add_months(rhs)?, time: self.time })
     }
 
+    /// Adds given [`FixedOffset`] to the current datetime.
```
jtmoon79, line 660: /// Adds given `rhs` to the current datetime.

jtmoon79 (approved):
Approve with docstring nitpicks

src/naive/datetime/mod.rs
pitdicker: On the other hand mentioning the type matches the documentation of the other methods.

pitdicker:
@djc I believe you were halfway done with reviewing this PR in https://github.com/chronotope/chrono/pull/1071#discussion_r1198850204? Three other PRs depend on this one.

Not meant to hurry you. Just wanting to highlight this is a good first one of my open PRs.

djc:
Yup, this is on my list (perhaps counterintuitively, the lack of a waiting-on-review label means I expect to review this sooner rather than the ones labeled as such -- if you have a suggestion for a better name, I'm all ears).

pitdicker:
@djc this PR is now quite small and the new `NaiveDateTime::checked_add_offset` and `NaiveDateTime::checked_sub_offset` make it easier to write code that avoids panics in other parts of chrono.

djc (commented):
Old review comment that's been stuck in pending mode...

src/naive/time/mod.rs
```diff
@@ -730,6 +730,30 @@ impl NaiveTime {
         OldDuration::seconds(secs + adjust) + OldDuration::nanoseconds(frac)
     }
 
+    /// Adds given `FixedOffset` to the current time, and returns the number of days that should be
+    /// added to a date as a result of the offset (either `-1`, `0`, or `1` because the offset is
+    /// always less than 24h).
+    ///
+    /// This method is similar to [`overflowing_add_signed`], but preserves leap seconds.
```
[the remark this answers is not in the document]
pitdicker: `overflowing_add_signed()` is not incorrect, or at least it works with leap seconds as documented.
Adding one hour to 13:59:60 with `overflowing_add_signed` should result in 14:59:59 (3600 seconds later).

But when adding a `FixedOffset` to convert a value from one time zone to another, it should do nothing special with leap seconds.
Adding one hour to 13:59:60 with `overflowing_add_offset` should result in 14:59:60 (the same value, but shifted by 3600 seconds).

pitdicker:
@djc This is a nice PR for us to push over the finish line.
#1071 to fix a panic in `TimeZone::from_local_datetime` depends on it, #1048 to fix other panics, #1270 for const initialization macros, and also a PR I want to open to make methods on `DateTime` const.

djc:
I'm aware, but you keep submitting other stuff...

pitdicker:
Haha, yes, that is my fault. But if we want to take some time with little new work to get the existing PRs ready that is fine with me.

pitdicker:
Added a small optimization to `NaiveDateTime::checked_(add|sub)_offset`: use `succ_opt` or `pred_opt` instead of `add_days`.
This method is used a lot in the `DateTime::with_*` methods to convert the value to local time and back to UTC.

Before the `add_days` optimization https://github.com/chronotope/chrono/pull/1214:
```
bench_datetime_with     time:   [52.090 ns 52.417 ns 52.773 ns]
  6 (6.00%) high mild
  1 (1.00%) high severe
```
After the `add_days` optimization:
```
bench_datetime_with     time:   [46.121 ns 46.263 ns 46.423 ns]
                        change: [-11.281% -10.739% -10.207%] (p = 0.00 < 0.05)
                        Performance has improved.
  4 (4.00%) high mild
  4 (4.00%) high severe
```
Now:
```
bench_datetime_with     time:   [34.281 ns 34.365 ns 34.458 ns]
                        change: [-26.354% -25.942% -25.467%] (p = 0.00 < 0.05)
                        Performance has improved.
  5 (5.00%) high mild
  5 (5.00%) high severe
```

In total we are ca. 35% faster in a good number of methods on `DateTime`.

djc:
Let's split this in 3 parts?

pitdicker:
This PR now contains only a tiny part 4.

src/datetime/mod.rs
```diff
@@ -1247,6 +1247,16 @@ impl<Tz: TimeZone> AddAssign<Duration> for DateTime<Tz> {
     }
 }
 
+impl<Tz: TimeZone> Add<FixedOffset> for DateTime<Tz> {
```
pitdicker, line 1250: In my opinion this impl doesn't make much sense. Should the offset be added to the datetime, or to the offset? And why? I would like to remove it on main.

djc: Removing it on main sounds good to me.

djc (approved):

src/datetime/mod.rs
pitdicker: I'll make a PR soon-ish.

PR #1070, opened 2023-05-18
Fix out-of-range panic in `NaiveWeek::last_day`

pitdicker:
Last commit from https://github.com/chronotope/chrono/pull/1048.

`NaiveWeek::last_day` creates an intermediate date with `NaiveWeek::first_day`, and adds 7. For weeks near `NaiveDate::MIN` the intermediate date could be out of range, while the final value is representable.

### DIFF
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -79,10 +79,13 @@ impl NaiveWeek {
     #[inline]
     #[must_use]
     pub fn first_day(&self) -> NaiveDate {
-        let start = self.start.num_days_from_monday();
-        let end = self.date.weekday().num_days_from_monday();
-        let days = if start > end { 7 - start + end } else { end - start };
-        self.date - Duration::days(days.into())
+        let start = self.start.num_days_from_monday() as i32;
+        let ref_day = self.date.weekday().num_days_from_monday() as i32;
+        // Calculate the number of days to subtract from `self.date`.
+        // Do not construct an intermediate date beyond `self.date`, because that may be out of
+        // range if `date` is close to `NaiveDate::MAX`.
+        let days = start - ref_day - if start > ref_day { 7 } else { 0 };
+        self.date.diff_days(days as i64).unwrap()
     }
 
     /// Returns a date representing the last day of the week.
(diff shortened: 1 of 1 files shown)


src/naive/date.rs
```diff
@@ -77,10 +77,10 @@ impl NaiveWeek {
     #[inline]
     #[must_use]
     pub fn first_day(&self) -> NaiveDate {
-        let start = self.start.num_days_from_monday();
-        let end = self.date.weekday().num_days_from_monday();
-        let days = if start > end { 7 - start + end } else { end - start };
-        self.date - Duration::days(days.into())
+        let start = self.start.num_days_from_monday() as i32;
+        let ref_day = self.date.weekday().num_days_from_monday() as i32;
+        let days = if start > ref_day { start - ref_day - 7 } else { start - ref_day };
```
djc, line 82: Please add a comment here and in `last_day()` about the overflow/underflow risks.

Also, please extract the common `start - ref_day` expression for both methods.

src/naive/date.rs
```diff
@@ -2995,23 +2998,31 @@ mod tests {
...
+            (Weekday::Fri, "Fri 2022-05-13", "Thu 2022-05-19"),
+            (Weekday::Sat, "Sat 2022-05-14", "Fri 2022-05-20"),
+            (Weekday::Sun, "Sun 2022-05-15", "Sat 2022-05-21"),
         ];
         for (start, first_day, last_day) in asserts {
             let week = date.week(start);
             let days = week.days();
-            assert_eq!(Ok(week.first_day()), NaiveDate::parse_from_str(first_day, "%Y-%m-%d"));
-            assert_eq!(Ok(week.last_day()), NaiveDate::parse_from_str(last_day, "%Y-%m-%d"));
+            assert_eq!(Ok(week.first_day()), NaiveDate::parse_from_str(first_day, "%a %Y-%m-%d"));
```
djc, line 3012: The change to add weekdays here should really be a separate commit.

src/naive/date.rs
pitdicker: > Also, please extract the common `start - ref_day` expression for both methods.

I don't think extracting it to a common methods it going to make it more readable. To compare:
```rust
let days = if start > ref_day { start - ref_day - 7 } else { start - ref_day };
```
And `last_day()`:
```rust
let days = if end < ref_day { end - ref_day + 7 } else { end - ref_day };
```

pitdicker: Added a comment.

djc: Sorry, I just meant the expression:

```rust
let days = end - ref_day - if end < ref_day { 7 } else { 0 };
```

pitdicker: Ah! That is indeed better.

djc (approved):

PR #1072, opened 2023-05-18
Refactor Local

pitdicker:
This extracts two commits from https://github.com/chronotope/chrono/pull/1017.

The platform-specific implementations of `Local` can get quite a bit simpler if they just return the offset for a local or UTC `NaiveDateTime`, instead of constructing a `DateTime` themselves. This also allows them to pick up the fix in https://github.com/chronotope/chrono/pull/1071.

The second commit was for a strange difference: `Utc::now` uses the standard library to get the current date, while for `Local::now` we call a Windows API ourselves. Because on both Unix and Windows we (should) get an UTC timestamp from the OS and then convert it to local time, we may just as well always use `Utc::now` inside `Local::now` and be consistent.


src/offset/local/unix.rs
```diff
@@ -11,10 +11,18 @@
...
-pub(super) fn naive_to_local(d: &NaiveDateTime, local: bool) -> LocalResult<DateTime<Local>> {
+pub(super) fn offset_from_utc_datetime(utc: &NaiveDateTime) -> LocalResult<FixedOffset> {
+    get_offset(utc, false)
+}
+
+pub(super) fn offset_from_local_datetime(local: &NaiveDateTime) -> LocalResult<FixedOffset> {
+    get_offset(local, true)
+}
+
+fn get_offset(d: &NaiveDateTime, local: bool) -> LocalResult<FixedOffset> {
```
djc, line 25: `get_` prefixes are mostly considered unidiomatic in Rust, just call this `offset()`?

src/offset/local/windows.rs
```diff
@@ -39,54 +39,43 @@ macro_rules! windows_sys_call {
...
+    get_offset(utc, false)
+}
+
+pub(super) fn offset_from_local_datetime(local: &NaiveDateTime) -> LocalResult<FixedOffset> {
+    get_offset(local, true)
+}
+
 /// Converts a local `NaiveDateTime` to the `time::Timespec`.
-pub(super) fn naive_to_local(d: &NaiveDateTime, local: bool) -> LocalResult<DateTime<Local>> {
+pub(super) fn get_offset(d: &NaiveDateTime, local: bool) -> LocalResult<FixedOffset> {
```
djc, line 51: Same here.

src/offset/local/stub.rs
```diff
@@ -8,218 +8,12 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use std::time::{SystemTime, UNIX_EPOCH};
+use crate::{FixedOffset, LocalResult, NaiveDateTime};
```
djc, line 11: I think I'd prefer to keep the non-WASM and `wasmbind` code in the same module. Or maybe both `stub` and `wasmbind` can become inline modules? In either case, please think about the optimally reviewable commit sequence for making changes like this.

djc (commented):
Great PR!

src/offset/local/stub.rs
pitdicker: Inline modules are a good idea 👍.

> In either case, please think about the optimally reviewable commit sequence for making changes like this.

I made an effort.

djc (approved):

pitdicker:
One builder seemed stuck. I have rebased on the current 0.4.x for another run.

src/offset/local/stub.rs
djc: Looks good, thanks!

PR #1078, opened 2023-05-22
Reformat `test_datetime_parse_from_str`

pitdicker:
This test had become quite unreadable.

This PR only adds `let parse = DateTime::<FixedOffset>::parse_from_str;`, does a find-and-replace, and `cargo fmt`.
(And it fixes a typo in a comment).


pitdicker:
You are fast!

djc:
Only for the trivial stuff, though. 😜

PR #1080, opened 2023-05-23
Make methods on `NaiveTime` const, update MSRV to 1.57

pitdicker:
This PR is pretty trivial, except for raising the MSRV to 1.57.

The reason is to use be able to panic in const context.
`.expect()` and `.unwrap()` are not yet available as const methods because there is some complexity with types that implement `Drop`. But we can get the next best thing with an `expect` macro.

In `NaiveTime` the use of panicking is limited, only deprecated methods need it. But I would really like to use it in more places.

I have a branch ready to make `NaiveDate` const, but that would be cleaner if it can be based on https://github.com/chronotope/chrono/pull/1069. (And it would need the MSRV bump and macro's from this PR).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,8 +32,8 @@ jobs:
       - run: cargo test --doc --all-features --color=always -- --color=always
 
   # later this may be able to be included with the below
-  # kept separate for now as the following don't compile on 1.56.1
-  # * arbitrary
+  # kept separate for now as the following don't compile on 1.57
+  # * arbitrary (requires 1.63 as of v1.3.0)
   rust_msrv:
     strategy:
       matrix:
@@ -43,7 +43,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: 1.56.1
+          toolchain: 1.57
       - uses: Swatinem/rust-cache@v2
       # run --lib and --doc to avoid the long running integration tests
       # which are run elsewhere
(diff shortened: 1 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -43,7 +43,7 @@ jobs:
       - uses: actions/checkout@v3
```
jtmoon79, line 43: Is the comment above on line 35 still relevant?

pitdicker: Good point, I updated the comment. Arbitrary requires 1.63.

djc (approved):

djc (commented):
So this makes a whole bunch of deprecated methods const. Do you think we should undeprecate them? If we don't, there seems little point in doing this...

pitdicker:
> So this makes a whole bunch of `deprecated` methods `const`. Do you think we should undeprecate them?

I think that would be better for the methods on `NaiveTime`.

> If we don't, there seems little point in doing this...

I have a branch ready to make similar methods on `NaiveDate` `const`, although some of them are deprecated and about which I agree they should stay that way.
And I have another branch to make the methods on `Duration` `const`.

Also I would like to use `NonZeroU32` in `NaiveDate`, which we can do without `unsafe` if the MSRV is increased to 1.57.

With this PR I hoped to unblock the other three :smile:.

pitdicker:
@djc Sorry to bother with this. What is your current opinion on this PR? You approved it?
I would like to rely on 1.57, and open other PRs that make more const.

pitdicker:
Very nice :tada:. And undeprecating or not is another issue, you might never hear me about it again.

PR #1081, opened 2023-05-23
Fix doctest warnings, remove mention of deprecated methods from main doc

pitdicker:
Currently a lot of the doctests would give warnings: 74 out of 240 use deprecated methods, have unused imports, or don't check a result that has `#[must_use]`.

These warnings are invisible unless you add:
```rust
#![doc(test(attr(deny(warnings))))]
```

Besides fixing the warnings I removed the prominent mention of the deprecated `Date` type from the main documentation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -382,7 +382,7 @@
 //! Advanced time zone handling is not yet supported.
 //! For now you can try the [Chrono-tz](https://github.com/chronotope/chrono-tz/) crate instead.
 
-#![doc(html_root_url = "https://docs.rs/chrono/latest/")]
+#![doc(html_root_url = "https://docs.rs/chrono/latest/", test(attr(deny(warnings))))]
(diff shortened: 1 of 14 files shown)


pitdicker:
Ai, 28 more doctests with warnings behind the 'serde' feature.

src/lib.rs
```diff
@@ -374,7 +374,7 @@
 //! Advanced time zone handling is not yet supported.
 //! For now you can try the [Chrono-tz](https://github.com/chronotope/chrono-tz/) crate instead.
 
-#![doc(html_root_url = "https://docs.rs/chrono/latest/")]
+#![doc(html_root_url = "https://docs.rs/chrono/latest/", test(attr(deny(warnings))))]
```
djc, line 385: In order to allow (at least in theory) each commit to pass CI, this should be added somewhere near the end, not near the beginning.

djc (approved):

PR #1082, opened 2023-05-24
Flexible offset formatting [1/3]

pitdicker:
There are a lot of issues with our current code to parse time zone offsets. For example `%z`, `%:z`, `%::z` and `%:::z` are all treated identical in the parser. In some cases `Z` is accepted, and there is a hack to accept `UTC`. There is no way to parse seconds.

Formatting is in a better state, but it would be nice if it could be more flexible. For example to be able to print seconds optionally, support padding, or print `Z`.

---

This is part 1 of a series of PRs to improve offset formatting and parsing. (1) adds a flexible formatter, (2) adds a flexible parser and fixes current parsing bugs, and (3) exposes the functionality in the format string.

With this PR it is now possible to specify the format of an offset:
- Is the precision in hours, minutes, seconds, or are minutes and/or seconds optional?
- Are the values seperated by a colon, nothing, or (when parsing) are both acceptable?
- Is `Z` used for `+00:00`?
- Should the hour value be zero-padded, space-padded (before the sign), or not be padded?

PR #1083, opened 2023-05-24
Flexible offset parsing [2/3]

pitdicker:
The first commit is shared with https://github.com/chronotope/chrono/pull/1082.
Then a couple of commits do some cleanup and add a new offset parser.
All the test keep passing, except some of those for `#z`. The difference is in the error causes, which have improved a bit.

Next are commits to make `%:z`, `%::z` and `%:::z` work as documented:
- `%:z` should require a colon (this is what makes it different from `%z`).
- `%::z` should also be able to parse, and even require, seconds.
- `%:::z` should only accept hours.

Finally the `FromStr` implementation of `DateTime<FixedOffset>` now accepts optional seconds in the offset, so that it can parse the `Debug` format of `DateTime`.

---

TODO: I still need to figure out some way to test all the code paths of the new offset parser.

### DIFF
--- a/src/format/parse.rs
+++ b/src/format/parse.rs
@@ -9,6 +9,7 @@ use core::str;
 use core::usize;
 
 use super::scan;
+use super::{Colons, OffsetFormat, OffsetPrecision};
 use super::{Fixed, InternalFixed, InternalInternal, Item, Numeric, Pad, Parsed};
 use super::{ParseError, ParseErrorKind, ParseResult};
 use super::{BAD_FORMAT, INVALID, NOT_ENOUGH, OUT_OF_RANGE, TOO_LONG, TOO_SHORT};
@@ -215,7 +216,13 @@ pub(crate) fn parse_rfc3339<'a>(parsed: &mut Parsed, mut s: &'a str) -> ParseRes
         parsed.set_nanosecond(nanosecond)?;
     }
 
-    let offset = try_consume!(scan::timezone_offset(s, |s| scan::char(s, b':'), true, false, true));
+    let offset_format = OffsetFormat {
+        precision: OffsetPrecision::Minutes,
+        colons: Colons::Colon,
+        allow_zulu: true,
+        padding: Pad::Zero,
+    };
+    let offset = try_consume!(scan::utc_offset(s, offset_format));
     if offset <= -86_400 || offset >= 86_400 {
         return Err(OUT_OF_RANGE);
     }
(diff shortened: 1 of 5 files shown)


src/format/parse.rs
```diff
@@ -219,7 +220,18 @@ fn parse_rfc3339<'a>(parsed: &mut Parsed, mut s: &'a str) -> ParseResult<(&'a st
         parsed.set_nanosecond(nanosecond)?;
     }
 
-    let offset = try_consume!(scan::timezone_offset_zulu(s, |s| scan::char(s, b':')));
+    let offset = if s.starts_with("UTC") || s.starts_with("utc") {
```
jtmoon79, line 223: 1) The quote from RFC 3339 in the code comment above suggests the special string to search for would be `"Z"`. Why is that not the special case here?

<br/> 

2) Why not search the entire remainder of the string `s`? e.g. why not

    if matches!(s, "UTC" | "utc") {

   (also, `matches` is more succinct).  Would there be a case where `s` should have trailing data after having matched the timezone in an RFC3339?
   
It seems like this code expression should be:

    let offset = if matches!(s, "Z" | "z") {
        s = &[1..];
        0
    } else {
        ...
    }
    if ! s.empty() {
        return Err(TOO_LONG);
    }

jtmoon79 (changes requested):

src/format/parse.rs
pitdicker: > 1. The quote from RFC 3339 in the code comment above suggests the special string to search for would be `"Z"`. Why is that not the special case here?

"Z" is handled by the parser. But this item does not only parse valid RFC 3339, but also supports "UTC" as offset. This is documented in https://docs.rs/chrono/latest/chrono/format/strftime/index.html#fn7. I suppose it makes some sense to support parsing this.
I'll add a comment.

> 2. Why not search the entire remainder of the string `s`? e.g. why not
>         if matches!(s, "UTC" | "utc") {
>         (also, `matches` is more succinct).  Would there be a case where `s` should have trailing data after having matched the timezone in an RFC3339?

I agree it is not likely to have any data after this.
But all parsing items only consume as much as they need to. There can be other items coming after it. And otherwise it is up to the calling function decide what to do with the trailing data.

jtmoon79: > "Z" is handled by the parser
Oh, it's handled by `scan::utc_offset(s, offset_format))` below. Gotcha.

> also supports "UTC" as offset. 
It is convenient but it's not strict RFC 3339, AFAICT. I'd prefer not allowing it. I won't block on this.

jtmoon79: > But all parsing items only consume as much as they need to.... And otherwise it is up to the calling function decide what to do with the trailing data.

Ah right, I recall now
https://github.com/chronotope/chrono/blob/c01e3a723883fc148c7e6052ddf4f6b42ff41e14/src/format/parse.rs#L519-L524

pitdicker: > It is convenient but it's not strict RFC 3339, AFAICT. I'd prefer not allowing it.

The commit tile says what I think about it :smile::
> Remove hack to accept "UTC" in `timezone_offset_zulu`

- One place it is used is for the `%+` formatting item, with this description:
  > Same as `%Y-%m-%dT%H:%M:%S%.f%:z`, i.e. 0, 3, 6 or 9 fractional digits for seconds and colons in the time zone offset.
  >
  > This format also supports having a `Z` or `UTC` in place of `%:z`. They are equivalent to `+00:00`.
  >
  > Note that all `T`, `Z`, and `UTC` are parsed case-insensitively.
  >
  > The typical `strftime` implementations have different (and locale-dependent) formats for this specifier. While Chrono's format for `%+` is far more stable, it is best to avoid this specifier if you want to control the exact output.
  So being less strict here seems useful to me.

- `DateTime<FixedOffset>::parse_from_rfc_3339` also uses it, and should be very strict however.

- A second RFC3339 parser is in `impl str::FromStr for DateTime<FixedOffset>`.
  That one is very forgiving, which is useful there.
  Maybe parsing `%+` should use that implementation?

pitdicker:
I did not realize `%::z` and  `%:::z` are fairly recent additions, added in https://github.com/chronotope/chrono/pull/759. Looking at that PR I understand they never properly added to the parser. That seems like an extra argument to fix them.

I'll keep this PR as a draft for a little longer. https://github.com/chronotope/chrono/pull/1145 has the proper fix for some changes I had to make here to RFC 3339 parsing code, and it is better if that is merged first.

KannarFr:
@pitdicker the two things this was waiting on have both landed — #1082 (June 2023) and #1145 (August 2023). It's still marked draft, and it now conflicts with `main`.

Meanwhile two smaller PRs have taken runs at the `%::z` half of it, #1639 and #1801, both deferring here — so the area keeps drawing duplicate effort at the partial fix while the complete one sits.

Would a rebase onto `main` be welcome? I'm happy to do the mechanical work, either pushed to a branch you can pull from or opened as a separate PR crediting you, whichever you prefer. I'd leave the design decisions alone. The TODO in the description about covering all the code paths of the new offset parser is something I could take a separate run at if that would help.

For context on why I care: I maintain a service that has to reproduce Java's `DateTimeFormatter.ISO_OFFSET_DATE_TIME` on the wire, and `%::z` not round-tripping — `format` emits `+01:30:45`, parsing that back is `TooLong` — is one of the reasons it hand-rolls an offset parser today.

djc:
chrono is deprecated and will not be accepting most kinds of changes:

PR #1085, opened 2023-05-26
Merge 0.4.x into main

pitdicker:
It has only been 1½ week, but plenty of merge conflicts again. Most of them because the unused imports in doctests are on the same line as the import of `TimeDelta` (instead of `Duration` on 0.4.x).


djc:
Thank you for taking this on!

pitdicker:
Thank you for all the time to put in as maintainer

PR #1087, opened 2023-05-28
Issue 835 timezone offset allow MINUS SIGN (U+2212) (branch 0.4.x)

jtmoon79:
PR for branch `0.4.x`.

The original PR for branch `main` is at #1001

Issue #835

### DIFF
--- a/src/format/scan.rs
+++ b/src/format/scan.rs
@@ -211,13 +211,28 @@ pub(super) fn timezone_offset<F>(s: &str, consume_colon: F) -> ParseResult<(&str
 where
     F: FnMut(&str) -> ParseResult<&str>,
 {
-    timezone_offset_internal(s, consume_colon, false)
+    timezone_offset_internal(s, consume_colon, false, true)
 }
 
+/// Parse a timezone from `s` and return the offset in seconds.
+///
+/// The `consume_colon` function is used to parse a mandatory or optional `:`
+/// separator between hours offset and minutes offset.
+///
+/// The `allow_missing_minutes` flag allows the timezone minutes offset to be
+/// missing from `s`.
+///
+/// The `allow_tz_minus_sign` flag allows the timezone offset negative character
+/// to also be `−` MINUS SIGN (U+2212) in addition to the typical
+/// ASCII-compatible `-` HYPHEN-MINUS (U+2D).
+/// This is part of [RFC 3339 & ISO 8601].
+///
+/// [RFC 3339 & ISO 8601]: https://en.wikipedia.org/w/index.php?title=ISO_8601&oldid=1114309368#Time_offsets_from_UTC
 fn timezone_offset_internal<F>(
     mut s: &str,
     mut consume_colon: F,
     allow_missing_minutes: bool,
+    allow_tz_minus_sign: bool,
 ) -> ParseResult<(&str, i32)>
 where
     F: FnMut(&str) -> ParseResult<&str>,
@@ -232,12 +247,23 @@ where
     }
     let negative = match s.chars().next() {
         Some('+') => {
+            // PLUS SIGN (U+2B)
             s = &s['+'.len_utf8()..];
 
             false
         }
         Some('-') => {
+            // HYPHEN-MINUS (U+2D)
             s = &s['-'.len_utf8()..];
+
+            true
+        }
+        Some('−') => {
+            // MINUS SIGN (U+2212)
+            if !allow_tz_minus_sign {
+                return Err(INVALID);
+            }
+            s = &s['−'.len_utf8()..];
 
             true
         }
@@ -309,7 +335,7 @@ where
 {
     match s.as_bytes().first() {
         Some(&b'z') | Some(&b'Z') => Ok((&s[1..], 0)),
-        _ => timezone_offset_internal(s, colon, true),
+        _ => timezone_offset_internal(s, colon, true, true),
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/format/scan.rs
```diff
@@ -262,13 +275,30 @@ where
             Ok((b[0], b[1]))
         }
     }
-    let negative = match s.as_bytes().first() {
-        Some(&b'+') => false,
-        Some(&b'-') => true,
+    let negative = match s.chars().next() {
```
djc, line 278: I'd like a separate commit first that switches from iterating over bytes to characters.

src/format/scan.rs
```diff
@@ -243,13 +243,26 @@ pub(super) fn timezone_offset<F>(s: &str, consume_colon: F) -> ParseResult<(&str
 where
     F: FnMut(&str) -> ParseResult<&str>,
 {
-    timezone_offset_internal(s, consume_colon, false)
+    timezone_offset_internal(s, consume_colon, false, true)
 }
 
+/// The `consume_colon` function is used to parse a mandatory or optional `:`
```
djc, line 249: Thanks for adding comments here.

I'd like to have the first line be a concise (one line) explanation of what this function does, before the comment goes into detail on what arguments mean.

jtmoon79:
I broke up the one large commit into:
- commit add tests for signage
- commit add tests for RFC 850
- commit add tests for TimeZoneName::new
- commit change parsing iteration over bytes to over chars
  this commit also touches on some tests added in prior commmits
- commit introduce parsing MINUS SIGN (U+2212) and tests
  this commit also touches on some tests added in prior commmits

src/format/scan.rs
```diff
@@ -379,7 +405,7 @@ pub(super) fn timezone_offset_2822(s: &str) -> ParseResult<(&str, Option<i32>)>
             Ok((s, None))
         }
     } else {
-        let (s_, offset) = timezone_offset(s, |s| Ok(s))?;
+        let (s_, offset) = timezone_offset_internal(s, |s| Ok(s), false, false)?;
```
djc, line 408: Is this making it stricter than it was before?

jtmoon79: No.

`timezone_offset(.., ..)` calls `timezone_offset_internal(.., .., false)`.

The only effective difference is the newly introduced last function parameter `allow_tz_minus_sign` set to `false` which disallows MINUS-SIGN (U+2212) as a timezone signage. Previous behavior also did not do any handling of MINUS-SIGN (U+2212).

src/offset/local/tz_info/timezone.rs
```diff
@@ -487,28 +487,36 @@ struct TimeZoneName {
 impl TimeZoneName {
```
pitdicker, line 487: Some quotes about this value from https://www.man7.org/linux/man-pages/man5/tzfile.5.html:

> Time zone designations should consist of at least three (3) and no more than six (6) ASCII characters from the set of alphanumerics, “-”, and “+”.  This is for compatibility with POSIX requirements for time zone abbreviations.

> * Some readers mishandle POSIX-style TZ strings that contain “<” or “>”. As a partial workaround, a writer can avoid using “<” or “>” for time zone abbreviations containing only alphabetic characters.
> * Many readers mishandle time zone abbreviations that contain non-ASCII characters. These characters are not recommended.
> * Some readers may mishandle time zone abbreviations that contain fewer than 3 or more than 6 characters, or that contain ASCII characters other than alphanumerics, “-”, and “+”. These abbreviations are not recommended.

pitdicker: The existing implementation seems to already strictly follow this specification.

jtmoon79: Interesting reference.  I'm not sure what's being asked for this PR 🤔

pitdicker: Ah, sorry. The method you changed is part of the tzfile parser. So I don't understand why you modify it, as it is not related to the normal parsing methods.

jtmoon79: > Ah, sorry. The method you changed is part of the tzfile parser. So I don't understand why you modify it, as it is not related to the normal parsing methods.

I'm pretty sure this is what needs to change to allow parsing HYPHEN-MINUS (U+2212) characters as timezone signage. Did I make the change at the wrong place?

jtmoon79: I think I understand; you're saying this change is unnecessary for parsing outside of parsing a timezone file. Good point.

IMO, it would be better to have both approaches be consistent about handling of timezone HYPHEN-MINUS (U+2212) as that's more user-friendly.
However, I can remove this parsing code if necessary.

pitdicker: This struct is for the abbreviation of a timezone, it is not an offset. I would drop it to be honest.

jtmoon79: > The method you changed is part of the tzfile parser. 

btw, if you're certain of which methods are for only parsing from timezone files, maybe you could include a PR with some code comments on that? Put that valuable hard-won knowledge into print! :-)

I didn't realize the files under `src/offset/local/tz_info` only dealt with parsing from a timezone file. Looking though that code, there are variables for handling timezone files, but AFAICT, no declarative comments that this was exclusively meant for that particular task.

pitdicker: Good point. I would like to try and integrate these files a bit better into chrono once https://github.com/chronotope/chrono/pull/1017 is merged.

jtmoon79: > Good point. I would like to try and integrate these files a bit better into chrono once #1017 is merged.

Are you suggesting you'd prefer these changes to `src/offset/local/tz_info/timezone.rs` to be removed from this PR?

djc (approved):

PR #1091, opened 2023-05-28
add tests for panics found by @Koral77 (branch 0.4.x)

jtmoon79:
Issue #1010

### DIFF
--- a/tests/dateutils.rs
+++ b/tests/dateutils.rs
@@ -136,3 +136,80 @@ fn try_verify_against_date_command_format() {
         date += chrono::Duration::days(55);
     }
 }
+
+// The following panic tests should remove `#[should_panic]` after
+// Issue #1010 is fixed.
+
+#[test]
+#[should_panic]
+/// Derived from https://github.com/Koral77/replay_files/blob/ce5b5c88c2a45a089d7d5ce19c60a5e50283ba60/replays/chrono/replay_chrono1/src/main.rs
+fn issue_1010_panic1() {
+    let _local0 = chrono::naive::NaiveDateTime::from_timestamp_opt(-4227854320, 1678774288);
+    let _local1 = chrono::Duration::microseconds(-7019067213869040);
+    let _local2_param0_helper1 = _local0.unwrap();
+    // panics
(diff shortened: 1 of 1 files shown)


pitdicker:
Why do some of these tests panic while they are fixed?

jtmoon79:
> Why do some of these tests panic while they are fixed?

🤷

I just wanted to establish some Issue-specific tests so the issue is acknowledged. As the code is fixed, the `#[should_panic]` should be removed.

pitdicker:
- `test_issue_1010_panic7` should panic (it returns `None`), a value for the month of `4294967295` doesn't make sense.
- `test_issue_1010_panic6` and `test_issue_1010_panic5` have the same issue, a value for the day or ordinal of `4294967295` doesn't make sense.
- `test_issue_1010_panic8` can't do anything except panic on out of range years.
- `test_issue_1010_panic4` used to panic because the input contains Unicode characters, but now just fails because the format string doesn't match.

I have added tests for these issues when fixing the panics. Maybe best to not add these invalid cases?

pitdicker:
With the panics fixed in https://github.com/chronotope/chrono/pull/1093, `test_issue_1010_panic1`, `test_issue_1010_panic2`, `test_issue_1010_panic3` still panic because of the `is_ok()` check, while they are trying to do something that should result in an error.

jtmoon79:
Good review @pitdicker . Indeed, I was hasty in throwing in the code from the examples. I updated the cases to be more sensible.

These tests are somewhat redundant. However, I'd like to move toward a project habit of adding useful test cases _per issue_. It gives the project higher "quality confidence". Hopefully @djc agrees.
See my rationale explained in PR https://github.com/chronotope/chrono/pull/1097

djc:
I'm not interested in adding "negative" tests (that test that some behavior still panics).

Honestly, I'm not a big fan of the current (proposed) practice for doing issue tests. For one thing, it's annoying to name tests by their issue number instead of giving them a somewhat descriptive (while hopefully still concise) name that describes what behavior their testing, because it means it's harder to determine what the test is trying to cover. Instead, add a comment linking to an issue. Also, we should keep tests closer to the code they cover (typically in a unit test, not in an integration test) rather than keeping a separate suite of "issues" test.

Issue #528, opened 2021-01-20
Issue parsing `DateTime` with "reduced precision"


jakevossen5:
I see that in the limitations there is a "Chrono inherently does not support an inaccurate or partial date and time representation" statement, but, I am having to deal with 8601 dates of the form "2020-01-20T05Z", which should be at hour five of January 5th 2021.

I realize that there isn't a great way to represent these in Chrono, but I would like to propose a couple solutions.

1. Return a`DateTime` representing the earliest possible moment that it could be, and a `DateTime` representing the latest possible moment (or a `Duration`).
2. Add some parameter to the parse functions (or create new ones) that allow for some rounding (as in, 05 becomes 05:00:00, 05:30:00, or 05:59:59 depending on what you select).

Would this be something that this project would be implement? If I am given some guidance on what direction to go in, I could probably work on this and try and open up a PR.

quodlibetor:
Yes this is very much doable, but a bit ugly, and something I would love!

The main thing that we need to do is to update [`Parsed`](https://github.com/chronotope/chrono/blob/3467172c31188006147585f6ed3727629d642fed/src/format/parsed.rs#L18-L27) to have optional default arguments.

The main issue is that we need to construct a DateTime from a non-overlapping set of fields, [`to_naive_datetime_with_offset`](https://github.com/chronotope/chrono/blob/3467172c31188006147585f6ed3727629d642fed/src/format/parsed.rs#L571) is the most-inclusive method, and it is a bit of a beast.

That said, I think that many of the places where we return errors in the various `Parsed::to_*` methods could check for a default before actually erroring.

If we add some `default_*` fields to `Parsed` then we could expose them through some public APIs and make the whole thing more pleasant.

Any step on this path would be immensely appreciated!

jakevossen5:
Thanks for the guidance! I will take a hack at this, and let you know if I run into any issues.

quodlibetor:
Sounds great!

PR #1094, opened 2023-05-28
Allow default values when parsing, and error on unused fields

pitdicker:
### Default values when parsing

`NaiveTime`:
When parsing seconds and nanoseconds may already be omitted. This is now extended to minutes.

`NaiveDate`:
When a day field is missing, default to the first day of the month when parsing year-month-day, or default to the first day of the week when parsing year-week-weekday.
If there are multiple choices, such as when there is no day field but both a month and week field, we don't arbitrarily pick a default.

`NaiveDateTime`:
Uses the combination of fallbacks from `NaiveTime` and `NaiveDate`. And if `hour` is missing, defaults that to zero also.

`DateTime`:
No changes. Only seconds and nanoseconds may be omitted.

### Error on unused fields
- `NaiveTime` errors if there are any date fields or an offset present.
- `NaiveDate` errors if there are any time fields or an offset present.
- `NaiveDateTime` errors if there is an offset present.

The various choices here are relatively easy to change. We could for example make `NaiveDateTime` require a full date and at least an hour.

Fixes https://github.com/chronotope/chrono/issues/528, https://github.com/chronotope/chrono/issues/1075.
cc @MarcoGorelli


src/naive/date.rs
```diff
@@ -2976,6 +2989,12 @@ mod tests {
             NaiveDate::parse_from_str("2019-01-0", "%Y-%W-%w").ok(),
             NaiveDate::from_ymd_opt(2019, 1, 13),
         );
+
+        assert!(NaiveDate::parse_from_str("2014-5-7T12:34:56", "%Y-%m-%dT%H:%M:%S").is_err());
+        assert!(NaiveDate::parse_from_str("2014-5-7+09:30", "%Y-%m-%dT%z").is_err());
+        assert!(
+            NaiveDate::parse_from_str("2014-5-7T12:34:56+09:30", "%Y-%m-%dT%H:%M:%S%z").is_err()
+        );
```
jtmoon79, line 2997: IIUC, shouldn't this include `assert!(...is_ok())` for _Only Year_, _Only Year and Month_?
Similar values should be tested in above function `test_date_from_str`.

jtmoon79 (changes requested):
It's possible I'm misunderstanding, but I think some necessary test asserts are missing.

src/naive/time/tests.rs
```diff
@@ -343,6 +339,10 @@ fn test_time_parse_from_str() {
     assert!(NaiveTime::parse_from_str("\t\t12:59 PM\t", "\t\t%H:%M\t%P\t").is_err());
     assert!(NaiveTime::parse_from_str("12:59  PM", "%H:%M %P").is_err());
     assert!(NaiveTime::parse_from_str("12:3456", "%H:%M:%S").is_err());
+
+    assert!(NaiveTime::parse_from_str("2014-5-7T12:34:56", "%Y-%m-%dT%H:%M:%S").is_err());
+    assert!(NaiveTime::parse_from_str("12:34:56+09:30", "%%H:%M:%S%z").is_err());
+    assert!(NaiveTime::parse_from_str("2014-5-7T12:34:56+09:30", "%Y-%m-%dT%H:%M:%S%z").is_err());
 }
```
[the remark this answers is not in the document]
pitdicker: I added the test.

pitdicker:
I have removed the part from this PR to error on unused fields. Now it should accept strictly more strings as valid when parsing, so it should be good for the 0.4.x branch.

I will open a follow up PR against main if this is merged.

src/format/parsed.rs
```diff
@@ -514,10 +515,17 @@ impl Parsed {
...
             Some(v @ 0..=59) => v,
             Some(_) => return Err(OUT_OF_RANGE),
-            None => return Err(NOT_ENOUGH),
+            None => {
+                if self.second.is_none() && self.nanosecond.is_none() {
+                    0
+                } else {
+                    return Err(NOT_ENOUGH);
+                }
+            }
```
djc: Style: writing this as a `match` expression will be slightly more concise and IMO easier to read.

src/format/parsed.rs
```diff
@@ -553,7 +553,16 @@ impl Parsed {
     /// Either way those fields have to be consistent to each other.
     pub fn to_naive_datetime_with_offset(&self, offset: i32) -> ParseResult<NaiveDateTime> {
         let date = self.to_naive_date();
-        let time = self.to_naive_time();
+        let time = if self.hour_div_12.is_none()
```
djc: What's your consideration for doing this here? Why not do this inside `to_naive_time()`?

src/format/parsed.rs
```diff
@@ -349,36 +352,50 @@ impl Parsed {
...
-
             (_, Some(isoyear), &Parsed { isoweek: Some(isoweek), weekday: Some(weekday), .. }) => {
-                // ISO year, week, day of the week
-                let date = NaiveDate::from_isoywd_opt(isoyear, isoweek, weekday);
-                let date = date.ok_or(OUT_OF_RANGE)?;
+                let date =
+                    NaiveDate::from_isoywd_opt(isoyear, isoweek, weekday).ok_or(OUT_OF_RANGE)?;
+                (verify_ymd(date) && verify_ordinal(date), date)
+            }
+            #[rustfmt::skip]
```
djc, line 378: Why skip formatting there?

src/format/parsed.rs
```diff
@@ -299,6 +299,10 @@ impl Parsed {
     /// Gregorian year and ISO week date year can have their century number (`*_div_100`) omitted,
     /// the two-digit year is used to guess the century number then.
     pub fn to_naive_date(&self) -> ParseResult<NaiveDate> {
+        self.to_naive_date_inner(true)
+    }
+
+    fn to_naive_date_inner(&self, use_default: bool) -> ParseResult<NaiveDate> {
```
djc, line 305: What's the use case for this? I'd prefer not to do this.

src/format/parsed.rs
```diff
@@ -580,7 +580,7 @@ impl Parsed {
             (None, Some(_)) => 0, // UNIX timestamp may assume 0 offset
             (None, None) => return Err(NOT_ENOUGH),
         };
-        let datetime = self.to_naive_datetime_with_offset(offset)?;
+        let datetime = self.to_naive_datetime_with_offset_inner(offset, false)?;
```
djc, line 583: Why not?

src/format/parsed.rs
pitdicker: I added a `use_default` flag, so that we can require that a date which includes an offset does not accept a string without minutes, hours or days.

But `to_naive_date` is a public function, so I could not change the signature of that one and made an 'inner' method. It is also not my preference...

src/format/parsed.rs
pitdicker: If a date includes an offset in hours and maybe minutes, I think we should require the date and time to also have a precision of hours and minutes without fallback.

src/format/parsed.rs
jtmoon79: FWIW I've learned to abide with rustfmt's choices.

src/naive/date.rs
```diff
@@ -552,12 +552,24 @@ impl NaiveDate {
     ///            Ok(NaiveDate::from_ymd_opt(2014, 5, 17).unwrap()));
     /// ```
     ///
-    /// Out-of-bound dates or insufficient fields are errors.
+    /// If the format string doesn't contain a day field, it defaults to the first day.
+    /// If year and month are given it defaults to the first day of the month.
+    /// If year and week are given it defaults to the first day of the week.
+    ///
```
jtmoon79, line 558: I really like this detailed explanation in the docstring.

However, it's somewhat confusing.

My suggestion includes referring specifically to `fmt`, use the phrase _specifier_ (that is used elsewhere), and being more explicit about the inferred subject:

    /// If `fmt` does not contain a day specifier then the day value defaults to the first day of the week.
    /// If `fmt` only has year and month specifiers then the day value defaults to the first day of the month.
    /// If `fmt` only has year and week specifier then the day value value defaults to the first day of the week.

(I'm not sure if my proposed sentences are technically correct, you'll have to review that)

src/naive/datetime/mod.rs
```diff
@@ -272,8 +272,10 @@ impl NaiveDateTime {
     ///            Ok(NaiveDate::from_ymd_opt(2015, 7, 1).unwrap().and_hms_milli_opt(8, 59, 59, 1_123).unwrap()));
     /// ```
     ///
-    /// Missing seconds are assumed to be zero,
-    /// but out-of-bound times or insufficient fields are errors otherwise.
+    /// Missing hours, minutes, seconds and nanoseconds are assumed to be zero.
+    /// A missing day will, depending on the other available fields, default to the first day of
+    /// the month or the first day of the week.
+    /// Out-of-bound times or insufficient fields are errors otherwise.
     ///
```
jtmoon79, line 279: I really like this detailed explanation in the docstring.

However, it's somewhat confusing.

My suggestion includes referring specifically to `fmt`, using the formal phase _specifier_ (as in "strftime specifier" which is how it's referred elsewhere) and being more explicit about the inferred subject:

    /// If `fmt` does not contain an hours, minutes, seconds or nanoseconds specifier than that is assumed to be zero.
    //// If `fmt` does not contain a day specifier then it will first day of the month or first day of the week, depending upon other specifiers.
    /// Otherwise, out-of-bound times or insufficient fields are errors.

src/format/parsed.rs
```diff
@@ -703,6 +647,74 @@ impl Parsed {
...
+    let ndays = firstweek as i32 + (week as i32 - 1) * 7 + weekday_nr as i32;
+    let date =
+        newyear.checked_add_signed(OldDuration::days(i64::from(ndays))).ok_or(OUT_OF_RANGE)?;
+    if date.year() != year {
+        return Err(OUT_OF_RANGE);
+    }
+    Ok(date)
+}
+
+fn resolve_year(y: Option<i32>, q: Option<i32>, r: Option<i32>) -> ParseResult<Option<i32>> {
```
jtmoon79, line 675: Please add a docstring for `resolve_year`.

Please use full variables names, e.g. `y` should be `year`, etc.
I don't know what `q` or `r` mean (I have some guesses, but it'd be better to use full words for those variables).

Tangential: a PR that type-aliased years, months, etc. would greatly increase code readability, i.e. `type year = i32;`, etc.

jtmoon79 (changes requested):

src/format/parsed.rs
pitdicker: Maybe I have to learn that yet :laugh:. In some cases rustfmt makes the code less readable, and I try to rewrite it in a way that formats nice again.
Using an extra pair of braces here seems to give pretty nice formatting:
```rust
            (
                (true, Some(year), None),
                &Parsed { month: Some(month), week_from_sun: None, week_from_mon: None, .. },
            ) => {
                let date = NaiveDate::from_ymd_opt(year, month, 1).ok_or(OUT_OF_RANGE)?;
                (verify_isoweekdate(date) && verify_ordinal(date), date)
            }
            (
                (true, Some(year), None),
                &Parsed { week_from_sun: Some(week), week_from_mon: None, month: None, .. },
            ) => {
                let date = resolve_week_number(year, week, Weekday::Sun, Weekday::Sun)?;
                (verify_ymd(date) && verify_isoweekdate(date) && verify_ordinal(date), date)
            }
```

src/naive/datetime/mod.rs
pitdicker: Thank you for the suggestion. I have updated it to:
```rust
    /// If `fmt` does not contain enough specifiers to get a complete date, parsing inserts a
    /// default value:
    /// - If `fmt` has year and month specifiers parsing assumes it is the first day of the month.
    /// - If `fmt` has year and week specifiers parsing assumes it is the first day of the week.
```

src/naive/date.rs
pitdicker: Thank you for the suggestion. I have updated it to:
```rust
    /// If `fmt` does not contain enough specifiers to get a complete date, parsing inserts a
    /// default value:
    /// - If `fmt` has year and month specifiers parsing assumes it is the first day of the month.
    /// - If `fmt` has year and week specifiers parsing assumes it is the first day of the week.
```

src/format/parsed.rs
pitdicker: > Tangential: a PR that type-aliased years, months, etc. would greatly increase code readability, i.e. `type year = i32;`, etc.

I quite like bare integers. But are you interested in trying out how it looks?

Issue #1096, opened 2023-05-28
formatting `NaiveDateTime` as `RFC3339` gives a bad-explaining error


hasezoey:
Currently when trying to format a `NaiveDateTime` with `RFC3339` (or to be more specific `.format("%+")`), there a very generic error:
```txt
thread 'main' panicked at 'a Display implementation returned an error unexpectedly: Error', /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/alloc/src/string.rs:2533:14
stack backtrace:
   0: rust_begin_unwind
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/std/src/panicking.rs:579:5
   1: core::panicking::panic_fmt
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/core/src/panicking.rs:64:14
   2: core::result::unwrap_failed
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/core/src/result.rs:1750:5
   3: core::result::Result<T,E>::expect
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/core/src/result.rs:1047:23
   4: <T as alloc::string::ToString>::to_string
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library/alloc/src/string.rs:2532:9
   5: bin::main
             at ./crates/bin/src/main.rs:10:10
```
no mention of that it cannot format because no timezone is provided

from my testing the error comes from:
https://github.com/chronotope/chrono/blob/083d3bf196537dfdbce6c29543734022f68e61a4/src/format/mod.rs#L687
because `off` is `None` and results in the following error:
https://github.com/chronotope/chrono/blob/083d3bf196537dfdbce6c29543734022f68e61a4/src/format/mod.rs#L697

versions:
rustc rustc 1.69.0 (84c898d65 2023-04-16)
chrono 0.4.24

reproduction code:

```rust
let test = NaiveDateTime::from_timestamp_millis(0).unwrap();
println!("test {}", test.format("%+"));
```

possible fixes:
- add a implicit zero timezone
- change the current error into something that has a message
- return a error before it even gets to this point explaining that `RFC3339` is not supported for `NaiveDateTime`

jtmoon79:
I was able to reproduce this issue on branch `0.4.x` and branch `main`.

[...]

PR #1097, opened 2023-05-28
Add tests/issues.rs, test Issue #1096 (branch 0.4.x)

jtmoon79:
- Add new file `tests/issues.rs` for regression tests specific to issues.
- add test for Issue #1096

---

If this PR is accepted, I will add another PR that moves all issue-specific tests into `tests/issues.rs`.

IMO, the issue-specific tests just make better sense in their own file. Since they cover niche topics it's often hard to know where to put issue-specific tests or what to title them. This code organization, placing them in `tests/issues.rs`, makes it easier to quickly add useful user-provided regression tests in a more organized manner.
User-provided issue-specific tests provide valuable certainty that a particular issue has been fixed.
They also exercise often niche behaviors that is too specific for "regular tests" to cover. Users spend a fair amount of time discovering then isolating an issue. Why not put that hard-won knowledge to use as another assurance of quality?
Issue-specific tests also demonstrate code-use for other users running into similar problems or with related questions.

I don't plan to add a test for all issues from the Issue backlog. But maybe going forward it can be made into a project practice that every fixed Issue gets a new test case in `tests/issues.rs` when that issue has reasonably simple reproduction steps provided,
e.g. a commit pattern like
1. add test to `tests/issues.rs` demonstrating failure (add test attribute `#[should_panic]` if the issue causes a panic)
2. code change with fix, adjust the test case for the fix

Issue-specific tests are not meant to replace "regular tests" or "general tests". Those should also be updated and augmented as necessary.

### DIFF
--- a/tests/issues.rs
+++ b/tests/issues.rs
@@ -1,0 +1,17 @@
+// issues.rs
+//
+// Regressions tests of specific issues.
+
+#[cfg(test)]
+use chrono::NaiveDateTime;
+
+#[test]
+#[should_panic]
+/// When Issue #1096 is fixed then `#[should_panic]` should be removed.
+/// See https://github.com/chronotope/chrono/issues/1096
+fn issue_1096() {
+    let test = NaiveDateTime::from_timestamp_millis(0).unwrap();
+    let tf = test.format("%+");
+    let output = tf.to_string();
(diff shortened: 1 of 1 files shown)


djc:
As explained in https://github.com/chronotope/chrono/pull/1091#issuecomment-1567929272, I'm not a fan.

PR #1099, opened 2023-05-29
README badge github commits-since

jtmoon79:
The github label `latest` is currently pointing to release `0.4.23` so shields.io badge generator believes that is the latest (the latest is `0.4.24`).

Example at https://github.com/jtmoon79/super-speedy-syslog-searcher/#super-speedy-syslog-searcher-s4-

Why? Badges are grokkable and just fun eye candy.


djc:
I don't think we should add more badges right now, just adds clutter and takes focus away from the existing ones.

PR #1101, opened 2023-05-29
README badge downloads

jtmoon79:
Example at https://github.com/jtmoon79/super-speedy-syslog-searcher/#super-speedy-syslog-searcher-s4-

Why? Badges are are grokkable and just fun eye candy.


djc:
I don't think we should add more badges right now, just adds clutter and takes focus away from the existing ones.

PR #1103, opened 2023-05-29
GitHub yml shell `set -eux`, use bash

jtmoon79:
Previously, a failure from `curl` or `sh` would be ignored (any failure before the final script statement would be ignored).

Use `bash`. Previously `sh` was emitting an error `sh: 139: [: x86_64: unexpected operator`
For example https://github.com/chronotope/chrono/actions/runs/5107217021/jobs/9179925191?pr=1087#step:6:12

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -121,7 +121,8 @@ jobs:
       - run: |
           set -euxo pipefail
           export RUST_BACKTRACE=1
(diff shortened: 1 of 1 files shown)


pitdicker:
Is this a common CI trick?

jtmoon79:
> Is this a common CI trick?

This is more of a "best practice for shell scripting" trick.

djc (approved):

PR #1095, opened 2023-05-28
Remove time 0.1 as optional dependency

pitdicker:
See https://github.com/chronotope/chrono/issues/1073.

Fixes https://github.com/chronotope/chrono/issues/876, https://github.com/chronotope/chrono/issues/285.

PR #1105, opened 2023-05-29
Bump version to 0.4.25

djc:


pitdicker:
Is https://github.com/chronotope/chrono/pull/1095 something you would like to do at the same time? That was my reason to open it now.

PR #1106, opened 2023-05-29
Update CITATION.cff

kevinmatthes:
Thank you for your new version v0.4.25!

After updating the Rust manifest, I also tried to grab the updated citation data but noticed that the new version information is not entered there, yet.  Hence, I would like to suggest you these changes to the CITATION.cff to update the version and release date.


pitdicker:
Just a comment out of curiosity: what is this file used for? It is new to me.

djc (approved):

PR #1107, opened 2023-05-29
Remove num-iter dependency

tottoto:
We might be able to use `core::ops::RangeInclusive` instead of `num_iter::RangeInclusive`.


djc (approved):
Definitely, thanks!

PR #1111, opened 2023-05-30
Update README

pitdicker:
There are plenty of things you can put in the readme or leave out of it. This is my take on it :smile:.

- I have tried to highlight what I consider to be the most attractive feature of chrono: that it helps to handle dates and times correctly, which means working with timezone information by default.
- The readme no longer describes chrono by comparing it to an old version of `time` or by mentioning ISO 8601, but lists a number of features instead.
- I don't think the precursors to chrono need to be called out in the readme.
- I have cut down the section about limitations, it should not span half the readme. And I would like to honestly say leap seconds are just not fully supported.
- Crate features, MSRV policy and license are pretty common parts of the readme of other popular crates, so I added those.
- Many crates have a code snipped to put into `Cargo.toml`, but this seems unnecessary to me. Especially now that there is `cargo add`.
- A direct reference to the documentation seems useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-[Chrono][docsrs]: Date and Time for Rust
+[Chrono][docsrs]: Timezone-aware date and time handling
 ========================================
 
 [![Chrono GitHub Actions][gh-image]][gh-checks]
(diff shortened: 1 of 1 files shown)


djc (approved):
Good improvement, thanks!

PR #1115, opened 2023-05-31
Backport and update CITATION.cff

pitdicker:
I didn't backport it before because I didn't understand the file...

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -43,3 +43,13 @@ jobs:
       - run: cargo doc --all-features --no-deps
         env:
           RUSTDOCFLAGS: -Dwarnings
+
+  cffconvert:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


pitdicker:
There seems to be a bug in my git client where two commits right after each other with the same description but different authors have the same author assigned :shrug:.

I don't care about attribution personally, but fixed it anyway.

djc (approved):

djc:
I wonder if there's an easy bash thing that'll check that the `Cargo.toml` version number and the `CITATION.cff` version number match...

pitdicker:
At least now they live on the same branch when you increase the version number.

I'll have a look, but no promises...

pitdicker:
With AWK it should not be that hard, I'll try to add it to this PR.

djc:
Thanks for looking into it!

PR #1117, opened 2023-05-31
Format toml with taplo

tottoto:
This pull request proposes to format `toml` files with [`taplo`](https://github.com/tamasfe/taplo). I believe mechanical linting and formatting improve formatting consistency and reduces costs to maintain them by hand.

The followings will be introduced:
- Uses `taplo` [default settings](https://taplo.tamasfe.dev/configuration/formatter-options.html) except of not collapsing list automatically when the line length is longer than the max length (which is defaultly set to `80`).
- Uses two spaces indent, which is used in [rust repository](https://github.com/rust-lang/rust) and is the default by `taplo`.
- Leaves the comment indentations to `taplo`.
- Adds a ci to check the `toml` files with `taplo`.

When having more preferred settings for this project, please feel free to request them.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -27,6 +27,14 @@ jobs:
         env:
           RUSTFLAGS: "-Dwarnings"
 
+  toml:
+    runs-on: ubuntu-latest
+    container:
+      image: tamasfe/taplo:0.8.0
+    steps:
+      - run: taplo lint
+      - run: taplo fmt --check --diff
+
   cargo-deny:
     runs-on: ubuntu-latest
     steps:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,12 +35,12 @@ rustc-serialize = { version = "0.3.20", optional = true }
 serde = { version = "1.0.99", default-features = false, optional = true }
 pure-rust-locales = { version = "0.5.2", optional = true }
 criterion = { version = "0.4.0", optional = true }
-rkyv = {version = "0.7", optional = true}
+rkyv = { version = "0.7", optional = true }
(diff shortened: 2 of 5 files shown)


djc (approved):

djc:
Thanks! I'm happy with this PR, but please be aware that I'm constrained on review time and so PRs with relatively minor benefits sometimes make me grumpy. For future changes in such a direction, please consider asking for feedback on the idea before executing on it. For example, I am uninterested in reviewing a PR that enables a bunch of disabled-by-default clippy lints.

tottoto:
Thank you for taking your time to review, and thanks for your great work to maintain and develop chrono, a very fundamental project. Here was my background for proposing this pull request: Because I know that there are various assessments of the cost of keeping code consistent in any ways, in order not to bother the maintainers, I searched for similar precedents before submitting this pull request and found #1102, so I thought it would not be problem to propose it. 

Regarding clippy, I have understood that it is intended because there is already a setting in the ci.

djc:
@tottoto yup, #1102 was also around the limit of my threshold for things that are worth a review. Anyway, just want to make sure you don't end up doing work that I don't want to review.

tottoto:
Sure. I think I could understand your criteria by following your clippy example. When I propose something, I'll ask for feedback before create a pull request with the actual patch. And I'm afraid whether you had reviewed and merged changes you didn't want. Please feel free to revert them if you want (, or if it is bothersome for you, I'm willing to do it).

pitdicker:
@tottoto I don't think there is anything to worry about, thank you for the contributions.

Is there something you would want to pick up next?

tottoto:
@pitdicker Thank you for your following. Currently I do not have something I could propose. When I find something, I would like to ask for your help.

Issue #573, opened 2021-07-03
DateTime#signed_duration_since should borrow instead of consume self and other


rkofman:
I'm very new to Rust, so this could be misunderstanding the language and/or the library, but it seems that `DateTime#signed_duration_since` could borrow the value as read-only since it is a non-mutating action (or `DateTime` could implement the `Copy` trait).

As is, clients are forced to `clone()` both the left and right side of the expression -- which seems like an unnecessary requirement for a method that calculates their difference and returns an `OldDuration`.

I'm happy to create a PR that adds a `borrowed_signed_duration_since` interface (to avoid a breaking change in the existing interface), but would appreciate somebody with more experience to first chime in and let me know that it's a reasonable change. This would be my first Rust PR.

:o)

PR #1119, opened 2023-06-01
Let `DateTime::signed_duration_since` take argument with `Borrow`

pitdicker:
Fixes https://github.com/chronotope/chrono/issues/573.

It is unfortunate that `DateTime::signed_duration_since` currently consumes its argument.

I am not sure if this would be a breaking change. Making an argument generic can change type inference. But I don't yet see how you could end up with a `DateTime` as argument to this function in a way where the type is not yet known.

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -1048,6 +1048,15 @@ impl<Tz: TimeZone> Sub<DateTime<Tz>> for DateTime<Tz> {
     }
 }
 
+impl<Tz: TimeZone> Sub<&DateTime<Tz>> for DateTime<Tz> {
+    type Output = OldDuration;
+
+    #[inline]
+    fn sub(self, rhs: &DateTime<Tz>) -> OldDuration {
+        self.signed_duration_since(rhs)
+    }
+}
+
 impl<Tz: TimeZone> Add<Days> for DateTime<Tz> {
     type Output = DateTime<Tz>;
 
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -8,7 +8,6 @@ extern crate alloc;
 
 #[cfg(all(not(feature = "std"), feature = "alloc"))]
 use alloc::string::{String, ToString};
-#[cfg(any(feature = "alloc", feature = "std", test))]
 use core::borrow::Borrow;
 use core::cmp::Ordering;
 use core::fmt::Write;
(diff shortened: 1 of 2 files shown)


djc:
> It is unfortunate that `DateTime::signed_duration_since` currently consumes its argument.

Why is it unfortunate?

> I am not sure if this would be a breaking change. Making an argument generic can change type inference. But I don't yet see how you could end up with a `DateTime` as argument to this function in a way where the type is not yet known.

If the argument's previous, concrete type is still accepted, I think it's generally not considered semver-breaking.

src/datetime/mod.rs
```diff
@@ -409,8 +408,12 @@ impl<Tz: TimeZone> DateTime<Tz> {
     /// This does not overflow or underflow at all.
     #[inline]
     #[must_use]
-    pub fn signed_duration_since<Tz2: TimeZone>(self, rhs: DateTime<Tz2>) -> OldDuration {
-        self.datetime.signed_duration_since(rhs.datetime)
+    pub fn signed_duration_since<B, Tz2>(self, rhs: B) -> OldDuration
```
djc, line 411: Would prefer to spell this as `rhs: impl Borrow<DateTime<Tz2>>` -- the reduced indirection makes this easier to understand IMO. It might even be possible to replace the `Tz2` here with `impl TimeZone` as well. (Since we're not using it, do we even need the `Tz2: TimeZone` bound?)

djc (commented):
As for the commit history, I guess you're making small commits because of feedback I gave earlier, but these are actually smaller than what I would do. The idea I'm chasing here is one of semantically independent changes -- in this case, it seems your change to `signed_duration_since()` is perhaps motivated by the addition of the `Sub` impl, and I usually prefer to add tests in the same commit as the code that they test.

(I recently missed a regression in commits that I reviewed because the commit that used newly introduced code was separate from the commit that introduced the new code, and the replacement was not equivalent -- so I think it's worth being careful about this stuff.)

pitdicker:
> As for the commit history, I guess you're making small commits because of feedback I gave earlier, but these are actually smaller than what I would do. The idea I'm chasing here is one of semantically independent changes -- in this case, it seems your change to `signed_duration_since()` is perhaps motivated by the addition of the `Sub` impl, and I usually prefer to add tests in the same commit as the code that they test.

The `Sub` implementation was an afterthought. I'll get this down to two commits, with the test combined with the change.

djc:
> Tt requires cloning the `DateTime` that is given as an argument if it is not a trow-away value.

I guess most of my usage of chrono is with a `Tz` that is `Copy`, such that `DateTime<Tz>` is also copy. What `Tz` are you using here?

pitdicker:
> I guess most of my usage of chrono is with a `Tz` that is `Copy`, such that `DateTime<Tz>` is also `Copy`. What `Tz` are you using here?

Sorry, my assumptions for this PR where just wrong, and I didn't check it well.

Still it might be useful? The type in `tzfile` is not `Copy`: https://docs.rs/tzfile/0.1.3/tzfile/struct.Offset.html.

src/datetime/mod.rs
pitdicker: Nice, new syntax! Removing the `Tz2: TimeZone` didn't work.

djc (approved):

pitdicker:
As a bit of communication: I have had some extra time this week because I am home with my first round of corona. But I don't really do anything that requires long concentration :smile:.

djc:
Ah, sorry to hear that! Hope your symptoms don't get too bad and it clears up soon.

PR #1120, opened 2023-06-02
Add "Errors" and "Panics" sections to API docs

pitdicker:
It is [recommended by the API guidelines](https://rust-lang.github.io/api-guidelines/documentation.html#function-docs-include-error-panic-and-safety-considerations-c-failure) to have "Errors" and/or "Panics" sections in the API documentation.

Many methods already documented under what circumstances they would panic or return `None`. In that case I only added a header.

I made sure the list of failure causes for each method are complete. Only the panics in https://github.com/chronotope/chrono/pull/1048 have been left out because I consider them bugs that are going to be fixed soonish.

In a couple of cases I made a drive-by fix to the surrounding documentation.

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -121,6 +121,13 @@ impl<Tz: TimeZone> DateTime<Tz> {
     /// Makes a new `DateTime` with given **local** datetime and offset that
     /// presents local timezone.
     ///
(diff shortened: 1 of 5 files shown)


djc (approved):

djc:
Thanks!

Issue #1112, opened 2023-05-30
Parsing explicitly space-padded format specifiers appears broken


Joey9801:
Hi,

It looks like a recent change to Chrono has broken parsing of the `%e`/`%_d` format specifiers (`"Same as %d but space-padded"`). Rather than expecting a space padding character, the various strptime-like methods now return a `Chrono(ParseError(Invalid))` error.

The following repro works as I would expect when built against chrono 0.4.24, and fails against chrono 0.4.25.
```rust
use chrono::{FixedOffset, DateTime, TimeZone};

fn main() {
    let s = "Mon Apr  3 22:32:00 BST 2023";
    let fmt = "%a %b %e %T %Z %Y";
    let offset = FixedOffset::east_opt(3600).unwrap();

    let parsed = offset.datetime_from_str(s, fmt).unwrap();
    let expected = offset.with_ymd_and_hms(2023, 4, 3, 22, 32, 0).unwrap();

    assert_eq!(expected, parsed);
}
```

The changes in #807 look suspiciously related to this issue, though I am not familiar enough with this codebase to know exactly what the issue is.

Many thanks,
Joe

djc:
#807 was not merged into the 0.4.x branch directly, but it's possible some of those changes made it into 0.4.25 via a backport.

PR #1122, opened 2023-06-03
Add more tests for `parse_from_str` and specifiers

jtmoon79:
Add exhaustive set of tests for `DateTime<FixedOffset>::parse_from_str` function and all strftime specifiers.

Issue #1112


jtmoon79:
@djc this PR is not complete. Let me know if you'd consider accepting this and then I'll complete it (add tests for all `strftime` specifiers).

Before submitting a fix for PR #807 which caused Bug #1112, I'd like to create more exhaustive testing for parsing. There is already test `test_strftime_items` for parsing strftime specifiers. But ``test_strftime_items` is focused on internal parsing APIs and did not catch #1112.  I'd like to add exhaustive tests that focuses on the user-facing `parse_from_str` APIs. I may also add more asserts within `test_strftime_items`.

pitdicker:
I have to admit all the test failures when i was working on the parsing of `%:z`, `%::z` etc were highly annoying…. but also quite useful and I was thankful for them. But added these tests is a large undertaking!

pitdicker:
@jtmoon79 The last commit in https://github.com/pitdicker/chrono/commits/parse_padding is what I was thinking it would take to work with whitespace in combination with padding specifiers. Just putting it in here to spare some double work.

pitdicker:
I would like to see at least the tests from https://github.com/chronotope/chrono/pull/807 back on the 0.4.x branch. I don't think that part would be any problem, and it would go a long way in fixing the merge conflicts of https://github.com/chronotope/chrono/pull/1083 and https://github.com/chronotope/chrono/pull/1094. Since it was your work, do you want to make a PR?

jtmoon79:
> I would like to see at least the tests from #807 back on the 0.4.x branch. Since it was your work, do you want to make a PR?

I'll submit something in a few days or so.

pitdicker:
I don't feel like we have huge gaps in our coverage of parsing. More tests is not per se better, as we have to maintain and fix merge conflicts in tests like this (which currently takes more effort than I like and would have expected).

@jtmoon79 Do you want to create more targeted tests for parts of the code where we have low coverage? Either because of actual issues that point to such or the coverage date?

jtmoon79:
> @jtmoon79 Do you want to create more targeted tests for parts of the code where we have low coverage? Either because of actual issues that point to such or the coverage date?

I entirely forgot about this PR! 🙀
Let me get back to you, hopefully within a few weeks rather than months.

> I don't feel like we have huge gaps in our coverage of parsing. 

IMO #1112 is evidence this is not the case.  I'd like to submit something for that. But the code has changed quite a lot so I need to fixup this PR.

pitdicker:
> I entirely forgot about this PR! 🙀
> Let me get back to you, hopefully within a few weeks rather than months.

No worries :smile:.

> IMO #1112 is evidence this is not the case. I'd like to submit something for that.

Well that's a good argument.
I think for that issue the problem was that we don't have a test for our (lack of) support for padding specifiers. Just yesterday I was writing my notes on that into an issue #1425.

May I suggest a ca. 20-line test for the four cases of padding specifiers (default, none, zero or space) and their interaction with preceding whitespace or a trailing zero?

PR #1090, opened 2023-05-28
tests: date path consolidate (branch 0.4.x)

jtmoon79:
- Consolidate `date_path`.
- Sanity check `date` can run `date --version` before attempting to use it.
- Print test progress messages.

PR #1125, opened 2023-06-05
Fix dead code error when running dateutils test on Windows

pitdicker:
The tests fail to run on Windows since https://github.com/chronotope/chrono/pull/1090.
The CI didn't catch it because it runs `cargo test --lib` and `cargo test --doc`, not plain `cargo test`.


jtmoon79 (approved):
LGTM.  Thanks @pitdicker !

PR #1126, opened 2023-06-05
Make formatting available without `alloc`

pitdicker:
It is possible to make the formatting methods available without the `alloc` feature with relatively little changes.

1. Format to `fmt::Formatter` instead of `String`. This means replacing `push_str()` with `write_str()?`, and replacing `push()` with `write_chr()?`.
2. When creating a `DelayedFormat` the timezone name is rendered to a `String`. I have replaced this with a fixed-size buffer of 63 bytes. It was slightly tricky to get the `Display` method of `Offset` to format into a byte buffer, that is what `TzName` is for.
3. `Display` can take some tricky formatting specifiers to justify/pad/truncate our formatting result. This is very difficult to do without rendering the intermediate result to a `String`. As solution I choose to render to an intermediate string in such a case in `Display::fmt`, and to ignore the specifiers in `no_std`.

Part 2 and 3 seem like a small price to pay for making all these methods available to `no_std` targets.

Previously `DelayedFormat` gave the impression it would work without allocations (also in the documentation), because 'why else would we have this type'? But the first thing it did was create a temporary string to write into :laughing:.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ See [docs.rs](https://docs.rs/chrono/latest/chrono/) for the API reference.
 
 Default features:
 
-* `alloc`: Enable features that depend on allocation (primarily string formatting)
+* `alloc`: Enable features that depend on allocation (some formatting methods)
 * `std`: Enables functionality that depends on the standard library. This is a superset of `alloc`
   and adds interoperation with standard library types and traits.
 * `clock`: Enables reading the system time (`now`) and local timezone (`Local`).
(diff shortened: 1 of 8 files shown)


pitdicker:
Rustfmt alternates with how it wants to format the large match in `format_inner`. It also did this in https://github.com/chronotope/chrono/pull/1082. I have included the reformat as a separate commit.

djc:
Let's discuss in #1127 and see whether it still makes sense to also do this.

pitdicker:
When we add a new formatting API I think we can do better than writing the timezone name to a fixed-size string on every use.

Issue #214, opened 2018-01-25
No test for Unix timestamp; failing new test case?


sanmai-NL:
I didn’t discover a unit test for parsing a Unix timestamp ("%s"). This test case fails, why?
```
let time_zero = DateTime::parse_from_str("0", "%s").unwrap();
```
```
panicked at 'called `Result::unwrap()` on an `Err` value: ParseError(NotEnough)', libcore/result.rs:916:5
```

quodlibetor:
You need to use a `NaiveDateTime` if you don't have a timezone (which timestamps don't). Also I don't believe that this will work correctly until 0.4.1 is released, which will be soon.

smallsweet:
Hi, I am also confused.
Even if using a `DateTime<FixedOffset>` I would assume all timestamps converted using `%s` to be in UTC, since that's part of the definition itself.

`
%s | 994518299 | UNIX timestamp, the number of seconds since 1970-01-01 00:00 UTC. [7]
`

pitdicker:
Fixed in https://github.com/chronotope/chrono/pull/1136.

PR #1136, opened 2023-06-07
Consider `%s` to be a timestamp in UTC

pitdicker:
Everywhere in chrono we treat a Unix timestamp to be in UTC.
But when trying to parse a timestamp with `%s` to a `DateTime` it used to fail because there was no offset supplied.

We now fall back to an offset of `0` if there is a timestamp and no other offset.

Fixes https://github.com/chronotope/chrono/issues/214.

### DIFF
--- a/src/format/parsed.rs
+++ b/src/format/parsed.rs
@@ -627,7 +627,12 @@ impl Parsed {
     /// plus a time zone offset.
     /// Either way those fields have to be consistent to each other.
     pub fn to_datetime(&self) -> ParseResult<DateTime<FixedOffset>> {
-        let offset = self.offset.ok_or(NOT_ENOUGH)?;
+        // If there is no explicit offset, consider a timestamp value as indication of a UTC value.
+        let offset = match (self.offset, self.timestamp) {
+            (Some(off), _) => off,
+            (None, Some(_)) => 0, // UNIX timestamp may assume 0 offset
+            (None, None) => return Err(NOT_ENOUGH),
+        };
         let datetime = self.to_naive_datetime_with_offset(offset)?;
         let offset = FixedOffset::east_opt(offset).ok_or(OUT_OF_RANGE)?;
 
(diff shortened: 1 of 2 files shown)


src/format/parsed.rs
```diff
@@ -627,7 +627,8 @@ impl Parsed {
     /// plus a time zone offset.
     /// Either way those fields have to be consistent to each other.
     pub fn to_datetime(&self) -> ParseResult<DateTime<FixedOffset>> {
-        let offset = self.offset.ok_or(NOT_ENOUGH)?;
+        // If there is no explicit offset, consider a timestamp value as indication of a UTC value.
+        let offset = self.offset.or(self.timestamp.map(|_| 0)).ok_or(NOT_ENOUGH)?;
```
djc, line 631: I don't think this formulation is very clear. What about something more explicit?

```rust
let offset = match self.timestamp {
    Some(_) => 0,
    None => self.offset.ok_or(NOT_ENOUGH)?,
};
```

pitdicker: Agreed that is better to have clear code than a comment.

The order is different though:
```rust
let offset = match self.offset {
    Some(off) => off,
    None => match self.timestamp {
        Some(_) => 0,
        None => return Err(NOT_ENOUGH),
    }
};
```

Do you want to make it this explicit?

djc: So is that actually correct? By the reasoning of the original issue, timestamps are supposed to always be in UTC, right?

pitdicker: If some input contains both a timestamp and a nonzero offset, I see no reason to start calling that invalid.
I imagine not everyone takes the UTC part of timestamps that seriously.

djc: Okay, then I would write it like this:

```rust
let offset = match (self.offset, self.timestamp) {
    (Some(off), _) => off,
    (None, Some(_)) => 0, // UNIX timestamp may assume 0 offset
    (None, None) => return Err(NOT_ENOUGH),
};
```

pitdicker: That is better, sorry I didn't come up with it.

djc (approved):

Issue #1054, opened 2023-05-09
[Feature Request]  `AddAssign`/`SubAssign` implementation to `chrono::Duration`


eval-exec:
We have  `AddAssign`/`SubAssign` implementation for `DateTime`/`Date` this PR: https://github.com/chronotope/chrono/pull/698

And I think it would be more convinient if there are `AddAssign`/`SubAssign` implementation for `chrono::Duration`

djc:
We currently maintain compatibility with the `time::Duration` type depending on features (from time 0.1, which is no longer being maintained), so I think adding this today would be a bit of a compatibility hazard.

pitdicker:
Added in https://github.com/chronotope/chrono/pull/1327.

Issue #717, opened 2022-06-22
Consider adding a "Default" implementation to Duration


eminence:
Would you consider having `chrono::Duration` implement `Default`, which would produce a zero-duration `Duration` ?

This would be consistent with [`std::time::Duration`](https://doc.rust-lang.org/std/time/struct.Duration.html#impl-Default)

eminence:
Ah, I had forgotten about `oldtime`.   `chrono::Duration` already had a slightly different API from `time::Duration`.   For example `chrono::Duration` implements `Hash` but `time::Duration` does not.   (Though otherwise their APIs are nearly identical).  

Maybe chrono 0.5 could add a `impl Default for Duration` ?

djc:
I'd be happy to add that once we straighten out the compatibility situation.

delightfulagony:
What's the current state of this issue?

As far as I understood from @djc, the reason for not implementing Default was because `std::time::Duration` didn't implement it in `0.1.x`.

As of `0.3.17` default is implemented, does that mean this is no longer "API-Incompatible" and could be implemented?

djc:
On the main branch (which will eventually become 0.5) we can add a `Default` implementation to what's now called `TimeDelta`. On the 0.4.x branch due to compatibility concerns that will not be viable.

You seem to be confusing the `std::time::Duration` type with the `Duration` type from the time crate. Due to compatibility, chrono 0.4.x will always (conditionally) depend on time 0.1, while chrono 0.5 does not depend on time at all.

delightfulagony:
You're right, I mistook `std::time` with the time crate.

But if this time I checked right, as of `0.1.3` for `std::time::Duration`, [default is also implemented](https://doc.rust-lang.org/std/time/struct.Duration.html#impl-Default-for-Duration)

djc:
Yes, but the `Duration` we expose in chrono is actually the one from the time crate, so `std::time::Duration` has no bearing on this issue.

esheppa:
@delightfulagony - if this https://github.com/chronotope/chrono/pull/860 or a linked PR gets merged, then you will be able to use `std::time::Duration` directly, including its `Default` implementation, would that satisfy your use case?

delightfulagony:
> Yes, but the `Duration` we expose in chrono is actually the one from the time crate, so `std::time::Duration` has no bearing on this issue.

Oh ok, so I misunderstood you, alright, thank you

[...]

PR #1137, opened 2023-06-08
Duration features

pitdicker:
If we drop time 0.1 compatibility in https://github.com/chronotope/chrono/pull/1095 it would be nice if we could also take care of some of the wishlist items for `Duration`.

- Derive `Default`: https://github.com/chronotope/chrono/issues/717
- Add a `subsec_nanos` method: https://github.com/chronotope/chrono/issues/154
- Add `try_*` builders: https://github.com/chronotope/chrono/pull/941
- Implement `AddAssign` and `SubAssign`: https://github.com/chronotope/chrono/issues/1054

Three features that would be more involved and not part of this PR are:
- Parsing: https://github.com/chronotope/chrono/issues/579
- Format `Duration`: https://github.com/chronotope/chrono/issues/197
- Serde: https://github.com/chronotope/chrono/issues/117

Making the remaining methods const (https://github.com/chronotope/chrono/issues/309, https://github.com/chronotope/chrono/pull/638) would depend on raising the MSRV as in https://github.com/chronotope/chrono/pull/1080.

Based on https://github.com/chronotope/chrono/pull/1095.

### DIFF
--- a/src/duration.rs
+++ b/src/duration.rs
@@ -16,6 +16,8 @@ use core::{fmt, i64};
 #[cfg(feature = "std")]
 use std::error::Error;
 
+use crate::try_opt;
+
 #[cfg(feature = "rkyv")]
 use rkyv::{Archive, Deserialize, Serialize};
 
@@ -37,15 +39,6 @@ const SECS_PER_HOUR: i64 = 3600;
 const SECS_PER_DAY: i64 = 86_400;
 /// The number of (non-leap) seconds in a week.
 const SECS_PER_WEEK: i64 = 604_800;
-
-macro_rules! try_opt {
-    ($e:expr) => {
-        match $e {
-            Some(v) => v,
-            None => return None,
-        }
-    };
-}
 
 /// ISO 8601 time duration with nanosecond precision.
 ///
--- a/src/duration.rs
+++ b/src/duration.rs
@@ -16,7 +16,7 @@ use core::{fmt, i64};
 #[cfg(feature = "std")]
 use std::error::Error;
 
-use crate::try_opt;
+use crate::{expect, try_opt};
 
 #[cfg(feature = "rkyv")]
 use rkyv::{Archive, Deserialize, Serialize};
@@ -89,16 +89,16 @@ impl Duration {
     /// Panics when the duration is out of bounds.
     #[inline]
     #[must_use]
-    pub fn weeks(weeks: i64) -> Duration {
-        Duration::try_weeks(weeks).expect("Duration::weeks out of bounds")
+    pub const fn weeks(weeks: i64) -> Duration {
+        expect!(Duration::try_weeks(weeks), "Duration::weeks out of bounds")
     }
 
     /// Makes a new `Duration` with given number of weeks.
     /// Equivalent to `Duration::seconds(weeks * 7 * 24 * 60 * 60)` with overflow checks.
     /// Returns `None` when the duration is out of bounds.
     #[inline]
-    pub fn try_weeks(weeks: i64) -> Option<Duration> {
-        weeks.checked_mul(SECS_PER_WEEK).and_then(Duration::try_seconds)
+    pub const fn try_weeks(weeks: i64) -> Option<Duration> {
+        Duration::try_seconds(try_opt!(weeks.checked_mul(SECS_PER_WEEK)))
     }
 
     /// Makes a new `Duration` with given number of days.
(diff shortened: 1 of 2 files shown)


pitdicker:
Maybe we can fix the confusion with `std::Duration` and our current `OldDuration` name in the documentation?
Rename `Duration` to `TimeDelta`, and make `Duration` a type alias.

jtmoon79:
> Maybe we can fix the confusion with `std::Duration` and our current `OldDuration` name in the documentation? Rename `Duration` to `TimeDelta`, and make `Duration` a type alias.

Might be too late to chime in on this, but `ChronoDuration` is readily self-explanatory.

pitdicker:
If a rename/alias to avoid confusion with the standard library is up for discussion:
- `ChonoDuration`
- `TimeDelta` (currently on main)
- `TimeSpan`
- `Interval`

@jtmoon79 I may not get the subtleties of the words completely. Is there one that is most appropriate for a value that can be positive and negative?

jtmoon79:
> If a rename/alias to avoid confusion with the standard library is up for discussion:
> 
>     * `ChonoDuration`
> 
>     * `TimeDelta` (currently on main)
> 
>     * `TimeSpan`
> 
>     * `Interval`
> 
> 
> @jtmoon79 I may not get the subtleties of the words completely. Is there one that is most appropriate for a value that can be positive and negative?

I think all the terms proposed there are sensible for positive and negative values.

I strongly prefer `ChronoDuration` because it readily explains itself as:
1. a `Duration`, similar to `std::Duration`
2. from the `chrono` crate

Synonyms like `TimeDelta`, `TimeSpan`, etc. imply the object is very different from `std::Duration`, and may add difficulty for non-native English readers.

fsktom:
I'm very excited about these quality of life improvements to the duration type! (esp. finally having the `iter::Sum` implementation)

I have to agree with @jtmoon79 on the new name for `Duration`.
`ChronoDuration` really seems like the most sensible option. The other two sound too scientific/abstract in my opinion and may really make it seem like it's a lot different from `std::time::Duration`.
I think it's a better name, even if it makes it a little bit more verbose and the following would look pretty weird (consecutive _chrono_):
```rust
let dur = chrono::ChronoDuration::zero();
```
Though I suppose most people would do ```use chrono::ChronoDuration``` if they're dealing with time spans anyway.

src/duration.rs
```diff
@@ -76,75 +78,111 @@ impl Duration {
...
     /// or less than `i64::MIN` seconds.
     #[inline]
     #[must_use]
-    pub fn seconds(seconds: i64) -> Duration {
+    pub const fn seconds(seconds: i64) -> Duration {
+        expect!(Duration::try_seconds(seconds), "Duration::seconds out of bounds")
+    }
+
+    /// Makes a new `Duration` with given number of seconds.
+    /// Returns None when the duration is more than `i64::MAX` milliseconds
```
jtmoon79, line 154: Nitpick: docstring referring to specific value None (`None`) should format it as code, i.e. add backticks

    /// Returns `None` when the duration is ...

src/duration.rs
```diff
@@ -309,13 +345,22 @@ impl Duration {
...
+    // TODO: make this method const once our MSRV is 1.58+
     #[cfg(feature = "std")]
     pub fn to_std(&self) -> Result<StdDuration, OutOfRangeError> {
         if self.secs < 0 {
             return Err(OutOfRangeError(()));
         }
         Ok(StdDuration::new(self.secs as u64, self.nanos as u32))
     }
+
+    /// Returns true if and only if d < MIN || d > MAX
```
jtmoon79, line 357: nitpick: why use the mathematical proof phrase "if and only if" here? I suggest wording "if", i.e.

    /// Returns `true` if d < MIN || d > MAX

jtmoon79 (changes requested):
I strongly prefer using alias `ChronoDuration`, e.g.

    use crate::duration::Duration as ChronoDuration;

However, using just `Duration` is acceptable so long as it is consistent everywhere, e.g.

    use crate::duration::Duration;

src/date.rs
```diff
@@ -576,7 +576,7 @@ where
 mod tests {
     use super::Date;
 
-    use crate::oldtime::Duration;
+    use crate::duration::Duration;
```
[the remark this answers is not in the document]
pitdicker: See https://github.com/chronotope/chrono/pull/1095#issuecomment-1666550830. This is a concern unrelated to this PR in my opinion.

djc:
Going to hold off on this for the next week, let's discuss again around Sep 18.

pitdicker:
Would you like me to split up this PR?

djc:
Yes, let's try it. Please rate limit your submissions!

pitdicker:
Haha. Okay.

pitdicker:
With https://github.com/chronotope/chrono/pull/1327 merged this now only has the changes to make methods on `Duration` const. Partially based on https://github.com/chronotope/chrono/pull/638.

src/duration.rs
```diff
@@ -347,12 +347,21 @@ impl Duration {
...
+    // TODO: make this method const once our MSRV is 1.58+
     pub fn to_std(&self) -> Result<StdDuration, OutOfRangeError> {
         if self.secs < 0 {
             return Err(OutOfRangeError(()));
         }
         Ok(StdDuration::new(self.secs as u64, self.nanos as u32))
     }
+
+    /// Returns `true` if d < MIN || d > MAX
+    const fn out_of_bounds(&self) -> bool {
```
djc, line 359: This should be a constructor, not a method on an already constructed type.

pitdicker: You are completely right!

Actually I wanted to have a `Duration::new` in https://github.com/chronotope/chrono/pull/1310 and https://github.com/chronotope/chrono/pull/1290.

pitdicker:
The remainder of this PR has become part of https://github.com/chronotope/chrono/pull/1337.

PR #1138, opened 2023-06-08
Move all tests into modules, fix clippy warnings

pitdicker:
Most of this PR is just moving code around so that all tests are inside a `tests` module.
Then it adds `#![deny(clippy::tests_outside_test_module)]` to make sure it stays this way.

Clippy didn't check tests on CI, and also didn't check code behind feature flags.
I enabled them and had to fix a couple of warnings. They are in the first 4 commits.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,9 +17,8 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - run: cargo fmt --check -- --color=always
       - run: cargo fmt --check --manifest-path fuzz/Cargo.toml
-      - run: cargo clippy --color=always -- -D warnings
       - run: |
-          cargo clippy --color=always --target x86_64-pc-windows-msvc \
+          cargo clippy --all-features --all-targets --color=always \
             -- -D warnings
       - run: |
           cargo clippy --manifest-path fuzz/Cargo.toml --color=always \
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -1423,34 +1423,6 @@ where
     }
 }
 
-#[test]
-fn test_add_sub_months() {
-    let utc_dt = Utc.with_ymd_and_hms(2018, 9, 5, 23, 58, 0).unwrap();
-    assert_eq!(utc_dt + Months::new(15), Utc.with_ymd_and_hms(2019, 12, 5, 23, 58, 0).unwrap());
-
-    let utc_dt = Utc.with_ymd_and_hms(2020, 1, 31, 23, 58, 0).unwrap();
-    assert_eq!(utc_dt + Months::new(1), Utc.with_ymd_and_hms(2020, 2, 29, 23, 58, 0).unwrap());
-    assert_eq!(utc_dt + Months::new(2), Utc.with_ymd_and_hms(2020, 3, 31, 23, 58, 0).unwrap());
-
-    let utc_dt = Utc.with_ymd_and_hms(2018, 9, 5, 23, 58, 0).unwrap();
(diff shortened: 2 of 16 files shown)


src/datetime/serde.rs
```diff
@@ -174,7 +174,6 @@ pub mod ts_nanoseconds {
     /// assert_eq!(as_string, r#"{"time":1526522699918355733}"#);
     /// # Ok::<(), serde_json::Error>(())
     /// ```
-    #[must_use]
```
djc, line 177: Why?

src/naive/date.rs
```diff
@@ -2389,13 +2338,60 @@ mod serde {
...
+    // = ((MIN_YEAR+400n+1)-01-01 minus 0001-01-01) - 146097n days
+    //
+    // n is set to 1000 for convenience.
+    const MIN_DAYS_FROM_YEAR_0: i32 = (MIN_YEAR + 400_000) * 365 + (MIN_YEAR + 400_000) / 4
+        - (MIN_YEAR + 400_000) / 100
+        + (MIN_YEAR + 400_000) / 400
+        - 146_097_000;
+
+    // only used for testing, but duplicated in naive::datetime
+    const MAX_BITS: usize = 44;
```
djc, line 2374: Please move the `const`s below the tests.

djc (approved):

src/datetime/mod.rs
```diff
@@ -1423,34 +1423,6 @@ where
...
-    let cdt_dt = FixedOffset::west_opt(5 * 60 * 60)
-        .unwrap()
-        .with_ymd_and_hms(2018, 9, 5, 18, 58, 0)
-        .unwrap();
-    let utc_dt2: DateTime<Utc> = cdt_dt.into();
-    assert_eq!(utc_dt, utc_dt2);
-}
-
 #[cfg(all(test, any(feature = "rustc-serialize", feature = "serde")))]
 fn test_encodable_json<FUtc, FFixed, E>(to_string_utc: FUtc, to_string_fixed: FFixed)
```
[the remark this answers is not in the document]
pitdicker: They are shared by the `serde` and `rustc-serialize` modules. I moved them in an earlier attempt, but it was clearer to keep them here with the imports.

src/naive/date.rs
pitdicker: No problem. Why?

.github/workflows/lint.yml
```diff
@@ -17,9 +17,11 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - run: cargo fmt --check -- --color=always
       - run: cargo fmt --check --manifest-path fuzz/Cargo.toml
-      - run: cargo clippy --color=always -- -D warnings
       - run: |
-          cargo clippy --color=always --target x86_64-pc-windows-msvc \
+          cargo clippy --features=serde,rkyv,arbitrary,unstable-locales, \
+            --tests --color=always -- -D warnings
+      - run: |
+          cargo clippy --tests --color=always --target x86_64-pc-windows-msvc \
```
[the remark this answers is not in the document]
pitdicker: I didn't know that argument, :+1:

src/naive/date.rs
djc: I generally prefer to order code in a top-down fashion -- public/high-level/high-complexity/dependent code before private/lower-level/low-complexity/depended on code. I find that this generally makes stuff easier to find (in this case, will want to look at/modify the test more often than the consts).

src/naive/date.rs
```diff
@@ -2237,16 +2237,19 @@ mod rustc_serialize {
     }
 
     #[cfg(test)]
-    use rustc_serialize::json;
+    mod tests {
```
djc, line 2240: Can you fix up the commit message on this one? Seems to be okay on the other one.

PR #1142, opened 2023-06-09
Parse module cleanup

pitdicker:
Another mostly useless PR, but I am working on something better :smile:.

For some reason clippy doesn't find the standalone tests in `format::scan` and `format::parse` yesterday. Maybe because they had a `#[cfg(test)]` attribute?

Also these modules had `#![allow(deprecated)]` on top, and used the deprecated methods `trim_left` and `trim_left_matches`.


djc (approved):

djc:
Happy to merge this after a rebase.

PR #1143, opened 2023-06-10
ISO 8601 parsers

pitdicker:
My main motivation for adding an ISO 8601 parser is that chrono is currently unable to parse is own `Debug` output if the year is outside the `0..=9999` range supported by RFC 3339. Edit: I forgot the `FromString` implementation of `DateTime` is very forgiving and accepts this.

A second motivation is that this crate frequently mentions ISO 8601. Then it is strange it does not even have a parser for that format :laughing:.

ISO 8601 date/time strings have a pretty complex format. I see it as a combination of 12 different formats for the date, 12 formats for the time, and 4 for the offset.

I have made the parser public as a `parse_from_iso8601` method on `NaiveDate`, `NaiveTime`, `NaiveDateTime` and `DateTime::<FixedOffset>`.

Sorry that this is such a large PR.
The only thing I can say is that about half of it are tests and documentation.

Fixes https://github.com/chronotope/chrono/issues/587.

PR #1144, opened 2023-06-10
Various fixes for formatting dates with a year outside of 0..=9999

pitdicker:
We have a couple of problems when formatting dates with a year outside of 0..=9999.

### RFC 2822
Formatting in `DateTime::to_rfc2822` panicked with "writing rfc2822 datetime to string should never fail".

- The panic message is now "date outside of defined range for rfc2822".
- I added a new method `DateTime::try_to_rfc2822` that can return `None` when formatting fails.
- `DateTime::to_rfc2822` is deprecated with the notice: "Can panic on years outside of the range 0..=9999. Use `try_to_rfc2822()` instead."

Formatting with the `RFC2822` formatting item could return a panic in the `Display` method of `DelayedFormat`.

- The formatting item can be localized, making it already not fully correct according to RFC 2822. But it is convenient, and we have a dedicated method, so I kept it that way.
- Formatting is extended to support years outside the range defined in RFC 2822 to prevent an error. It is now up to the calling function to ensure the date is in range.

### RFC 3339 and ISO 8601

RFC 3339 is not defined for dates with a year outside of 0..=9999. ISO 8601 does support it.

`DateTime::to_rfc3339` was a hybrid method, claiming to format a datetime as both valid RFC 3339 and ISO 8601. It would write invalid RFC 3339 strings if the year is out of range, which could not be parsed by `DateTime::parse_from_rfc3339`.

- `to_rfc3339` now panics on out-of-range dates with "date outside of defined range for rfc3339", just like `to_rfc2822`.
- A new `try_to_rfc3339` can return `None` when formatting fails.
- A new method `to_iso8601` functions exactly as `to_rfc3339` did, because dates that are out of range for RFC 3339 are still valid ISO 8601.
- `to_rfc3339` is deprecated with the notice: "Can panic on years outside of the range 0..=9999. Use `try_to_rfc3339()` or `to_iso8601` instead.". This way when users update chrono to a new version a panic is not silently introduced.

Other RFC 3339 formatting methods:
- `DateTime::to_rfc3339_opts` is redefined to format to an ISO 8601 representation when the year is out of range.
- The `RFC3339` formatting item is also redefined to format to an ISO 8601 representation when the year is out of range (for the same reasons as the `RFC2822` formatting item: to not panic inside the `Display` implementation of `DelayedFormat`).
- This is the formatting item for `"%+"`.

ISO 8601 and usually valid RFC 3339 is also the `Debug` output of `DateTime`.

We have no parser that can read these out-of-range dates, which is what https://github.com/chronotope/chrono/pull/1143 is for.

### Year formatting items

I have updated the documentation of the formatting items `YearDiv100`, `YearMod100`, `IsoYearDiv100` and `IsoYearMod100`. The implementation supports formatting dates outside the 0..=9999 range, which is now reflected in the documentation.


jtmoon79 (approved):
LGTM

src/datetime/mod.rs
```diff
@@ -492,35 +492,143 @@ impl<Tz: TimeZone> DateTime<Tz> {
...
+    )]
     pub fn to_rfc2822(&self) -> String {
+        self.try_to_rfc2822().expect("date outside of defined range for rfc2822")
+    }
+
+    /// Returns an RFC 2822 date and time string such as `Tue, 1 Jul 2003 10:52:37 +0200`.
+    ///
+    /// # Errors
+    ///
+    /// RFC 2822 is only defined on years 0 through 9999, and returns an error on dates of outside
```
jtmoon79, line 520: nitpick: move word `of` in `error on dates of outside this range.` to `error on dates outside of this range.`

jtmoon79 (changes requested):
I strongly recommend some tests to confirm the panic. It's an important behavior presumption to verify as it has a large affect on the important parts of chrono. e.g.

```rust
#[should_panic]
fn test_rfc_2822_year_range_panic() {
    use chrono::{TimeZone, Utc};
    Utc.with_ymd_and_hms(10000, 1, 2, 3, 4, 5).unwrap().to_rfc2822();
}
```

Same test case for RFC 3339.

It'd also be good to add global constants for that range, e.g.

```rust
const RFC_2822_YEAR_MAX = 9999;
const RFC_2822_YEAR_MIN = 0;
```

Same for RFC 3339.

Then I'd improve the prior test case and add two more test cases to test those boundaries.

```rust
fn test_rfc_2822_year_range_max() {
    use chrono::{TimeZone, Utc};
    Utc.with_ymd_and_hms(RFC_2822_YEAR_MAX, 1, 2, 3, 4, 5).unwrap();
    assert_eq!(dt.to_rfc2822(), "Sat, 2 Jan 9999 03:04:05 +0000");
}

fn test_rfc_2822_year_range_min() {
    use chrono::{TimeZone, Utc};
    Utc.with_ymd_and_hms(RFC_2822_YEAR_MIN, 1, 2, 3, 4, 5).unwrap();
    assert_eq!(dt.to_rfc2822(), "Sat, 2 Jan 0000 03:04:05 +0000");
}

#[should_panic]
fn test_rfc_2822_year_range_panic_over_max() {
    use chrono::{TimeZone, Utc};
    Utc.with_ymd_and_hms(RFC_2822_YEAR_MAX + 1, 1, 2, 3, 4, 5).unwrap().to_rfc2822();
}

#[should_panic]
fn test_rfc_2822_year_range_panic_under_min() {
    use chrono::{TimeZone, Utc};
    Utc.with_ymd_and_hms(RFC_2822_YEAR_MIN - 1, 1, 2, 3, 4, 5).unwrap().to_rfc2822();
}

```

(I don't know of `Sat` is the correct day of week for my code examples)

Then do the same for RFC 3339.

pitdicker:
> I strongly recommend some tests to confirm the panic. It's an important behavior presumption to verify as it has a large affect on the important parts of chrono. e.g.

Thank you for even writing out some test cases.

src/datetime/tests.rs
```diff
@@ -548,10 +549,36 @@ fn test_datetime_rfc2822() {
...
+    let dt = Utc.with_ymd_and_hms(RFC_2822_YEAR_MIN, 1, 2, 3, 4, 5).unwrap();
+    assert_eq!(dt.to_rfc2822(), "Sun, 02 Jan 0000 03:04:05 +0000");
+}
+
+#[test]
+#[should_panic]
+#[cfg(any(feature = "alloc", feature = "std"))]
+#[allow(deprecated)]
+fn test_rfc_2822_year_range_panic_high() {
+    let _ = Utc.with_ymd_and_hms(10000, 1, 2, 3, 4, 5).unwrap().to_rfc2822();
```
jtmoon79, line 568: This should test the boundaries around the constants, e.g.

```rust
let _ = Utc.with_ymd_and_hms(RFC_2822_YEAR_MAX + 1, 1, 2, 3, 4, 5).unwrap().to_rfc2822();
```

jtmoon79 (changes requested):

src/format/mod.rs
```diff
@@ -244,8 +250,18 @@ pub enum Fixed {
...
+    /// When formatting the output will be localized according to `locale`.
+    ///
+    /// This does not output a strictly valid RFC 2822 string, even when not localized. RFC 2822 is
+    /// only defined on years 0 through 9999. We format a date outside these ranges anyway to
+    /// prevent a panic when formatting with `"%+"`.
     RFC2822,
     /// RFC 3339 & ISO 8601 date and time syntax.
+    ///
+    /// Note that if the year of the `DateTime` is outside of the range 0 through 9999 the date
+    /// while be formatted as an expanded representation according to ISO 8601. These dates are not
```
jtmoon79, line 263: grammar? I think you meant:

    /// Note that if the year of the `DateTime` is outside of the range 0 through 9999 then 
    /// the date will be formatted as an expanded representation according to ISO 8601.

src/format/formatting.rs
```diff
@@ -611,38 +615,54 @@ pub(crate) fn write_rfc3339(
 }
 
 #[cfg(any(feature = "alloc", feature = "std"))]
-/// write datetimes like `Tue, 1 Jul 2003 10:52:37 +0200`, same as `%a, %d %b %Y %H:%M:%S %z`
+/// Write datetimes like `Tue, 1 Jul 2003 10:52:37 +0200`, similar to `%a, %d %b %Y %H:%M:%S %z`.
+///
+/// # Errors
+///
+/// RFC 2822 is only defined on years 0 through 9999, and returns an error on dates outside this
+/// range.
```
jtmoon79, line 623: Perhaps mention

    /// Uses `default_locale()` for locale information.

src/datetime/mod.rs
```diff
@@ -516,12 +623,16 @@ impl<Tz: TimeZone> DateTime<Tz> {
...
+    /// Return an RFC 3339 and ISO 8601 date and time string with subseconds formatted as per
+    /// `SecondsFormat`.
     ///
     /// If `use_z` is true and the timezone is UTC (offset 0), uses `Z` as
     /// per [`Fixed::TimezoneOffsetColonZ`]. If `use_z` is false, uses
-    /// [`Fixed::TimezoneOffsetColon`]
+    /// [`Fixed::TimezoneOffsetColon`].
+    ///
+    /// Note that if the year of the `DateTime` is outside of the range 0 through 9999 the date
+    /// while be formatted as an expanded representation according to ISO 8601. This makes the
```
jtmoon79, line 634: grammar. I suggest:

    /// Note that if the year of the `DateTime` is outside of the range 0 through 9999 then
    /// the date will be formatted as an expanded representation according to ISO 8601.

src/datetime/mod.rs
```diff
@@ -490,24 +490,131 @@ impl<Tz: TimeZone> DateTime<Tz> {
...
+    #[must_use]
+    pub fn try_to_rfc3339(&self) -> Option<String> {
+        let year = self.year();
+        if !(0..=9999).contains(&year) {
+            return None;
+        }
+        Some(self.to_iso8601())
+    }
+
+    /// Returns an ISO 8601 date and time string such as `1996-12-19T16:39:57-08:00`.
```
jtmoon79, line 594: nitpick: I suggest:

     /// Returns an ISO 8601 date and time `String` such as `"1996-12-19T16:39:57-08:00"`.

src/datetime/mod.rs
```diff
@@ -490,24 +490,131 @@ impl<Tz: TimeZone> DateTime<Tz> {
...
+    ///
+    /// Note that standard supports many different format, we choose the one that is compatible
```
jtmoon79, line 596: grammar is odd. I suggest:

     /// Note that though standard supports many different formats,
     /// we choose the a format that is compatible with RFC 3339 for most common cases.

src/datetime/mod.rs
```diff
@@ -490,24 +490,131 @@ impl<Tz: TimeZone> DateTime<Tz> {
...
+    /// with RFC 3339 for most common cases.
+    ///
```
jtmoon79, line 598: nitpick: Remove the blank line as the sentence `This format supports...` is extending the idea of the prior sentence.

src/datetime/mod.rs
```diff
@@ -490,24 +490,131 @@ impl<Tz: TimeZone> DateTime<Tz> {
...
+    #[deprecated(
+        since = "0.4.27",
+        note = "Can panic on years outside of the range 0..=9999. Use `try_to_rfc3339()` or `to_iso8601` instead."
+    )]
     pub fn to_rfc3339(&self) -> String {
-        // For some reason a string with a capacity less than 32 is ca 20% slower when benchmarking.
+        self.try_to_rfc3339().expect("date outside of defined range for rfc3339")
+    }
+
+    /// Returns an RFC 3339 date and time string such as `1996-12-19T16:39:57-08:00`.
```
jtmoon79, line 565: I suggest:

      /// Returns an RFC 3339 date and time `String` such as `"1996-12-19T16:39:57-08:00"`.

src/datetime/mod.rs
```diff
@@ -490,24 +490,131 @@ impl<Tz: TimeZone> DateTime<Tz> {
...
+    ///
+    /// let dt = Utc.with_ymd_and_hms(10_000, 1, 1, 0, 0, 0).unwrap();
+    /// assert_eq!(dt.try_to_rfc3339(), None);
+    /// ```
+    #[cfg(any(feature = "alloc", feature = "std"))]
+    #[cfg_attr(docsrs, doc(cfg(any(feature = "alloc", feature = "std"))))]
+    #[must_use]
+    pub fn try_to_rfc3339(&self) -> Option<String> {
+        let year = self.year();
+        if !(0..=9999).contains(&year) {
```
jtmoon79, line 588: Interesting: `if !(0..=9999).contains(&year)` compiles to the same assembly as `if year < &0 || &9999 < year`.

👍 

https://godbolt.org/z/rb5vvos88

jtmoon79 (approved):
Approved with feedback about various docstring oddities and suggestions.

src/datetime/mod.rs
pitdicker: Interesting, thank you. Although I know these optimizations are possible and sometimes rely on them, I still find it amazing.

src/format/formatting.rs
pitdicker: I am a bit uncomfortable with the fact that `write_rfc2822_inner` localizes the day and month.
Even while using `format_localized` I think the `RFC2822` item should use English names as defined in the standard.

@djc Shall I remove localization from `write_rfc2822`? It lives behind the `unstable-locales` feature, so making a change in behaviour seems ok.

src/datetime/mod.rs
pitdicker: The intention was not really to talk about the return type here, only the format of the returned data.

src/datetime/mod.rs
pitdicker: As above.

src/datetime/mod.rs
pitdicker: Thank you for reading closely :+1:

src/datetime/mod.rs
```diff
@@ -516,12 +623,16 @@ impl<Tz: TimeZone> DateTime<Tz> {
...
 
-    /// Return an RFC 3339 and ISO 8601 date and time string with subseconds
-    /// formatted as per `SecondsFormat`.
+    /// Return an RFC 3339 and ISO 8601 date and time string with subseconds formatted as per
+    /// `SecondsFormat`.
     ///
     /// If `use_z` is true and the timezone is UTC (offset 0), uses `Z` as
     /// per [`Fixed::TimezoneOffsetColonZ`]. If `use_z` is false, uses
-    /// [`Fixed::TimezoneOffsetColon`]
+    /// [`Fixed::TimezoneOffsetColon`].
```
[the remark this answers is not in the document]
pitdicker: Reading it again I think it is better to not mention `Fixed::TimezoneOffsetColon` and `Fixed::TimezoneOffsetColonZ`, but to only describe the behaviour.

pitdicker:
@jtmoon79 Sorry it took so long, thank you for the review!

src/format/formatting.rs
jtmoon79: > write_rfc2822_inner localizes the day and month.
Even while using format_localized I think the RFC2822 item should use English names as defined in the standard.

I agree: follow the standards.

djc: Agreed.

PR #1146, opened 2023-06-15
Fix NaiveTime doc typo

zachs18:
`3:00:59 + 60s = 3:01:59`, not `3:02:00`.

This is not an API change, so it is trivially semver-compatible, so I'm PR'ing to `0.4.0`. Please let me know if it should be targeted to `main` instead.


djc (approved):
Thanks!

Issue #1129, opened 2023-06-06
Local time unsupport openwrt


cppcoffee:
`chrono::Local` does not support the openwrt system and prints out utc.

```rust
use chrono::{Utc, Local};

fn main() {
    println!("{}", Local::now());
    println!("{}", Utc::now());
}
```

on ubuntu output:
```
2023-06-06 11:37:45.784123237 +08:00
2023-06-06 03:37:45.784374037 UTC
```

on openwrt output:

```
2023-06-06 03:39:08.963993795 +00:00
2023-06-06 03:39:08.964326294 UTC
```

pitdicker:
@cppcoffee Is there a not-too-difficult way for me to test chrono on openwrt?

cppcoffee:
Hi @pitdicker , OpenWrt has virtualization image that you can try

https://openwrt.org/docs/guide-user/virtualization/start

pitdicker:
The docker image has no timezone files and `/etc/TZ` is a dangling pointer. Maybe timezone support has to be installed first?

cppcoffee:
On my openwrt environment, the `/etc/TZ` file content as follows:

```
# cat /etc/TZ
CST-8
```

cppcoffee:
OpenWrt commands for setting the time:

```shell
$ uci get system.@system[0].timezone
$ uci set system.@system[0].timezone='Asia/Shanghai'
$ uci commit system
$ /etc/init.d/system reload
$ date
```

cppcoffee:
iana-time-zone fixed:

https://github.com/strawlab/iana-time-zone/pull/109

Use chrono + iana-time-zone new version, `chrono::Local` prints out UTC. 

https://github.com/strawlab/iana-time-zone/pull/109#issuecomment-1579975176

djc:
I suppose that would be because the time zone info files aren't found in the places where they are stored on conventional Linux systems.

cppcoffee:
Where chrono to find the time zone info files?

djc:
https://github.com/chronotope/chrono/blob/main/src/offset/local/unix.rs#L78

cppcoffee:
On OpenWrt print `chrono::Local::now()`, without `TZ` env, output UTC time.

```shell
$ ./chrono-demo-uclibc
2023-06-17T03:44:47.615484252+00:00
```

If use `TZ=CST-8` env, output local time.

```
$ cat /etc/TZ
CST-8
[...]

PR #1148, opened 2023-06-19
Use libc rather than iana-time-zone on Android.

qwandor:
As discussed in #1018, this switches back to using `localtime_r` and `mktime` from libc to get the local timezone on Android, rather than depending on `iana-time-zone` and trying to parse Android's tzdata files directly. This is better because the tzdata file format is not stable, and we don't want to use `iana-time-zone` in the Android platform.

I'll send a separate PR to check for the availability of the new `localtime_rz` and friends and use them instead where possible.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,9 +19,8 @@ name = "chrono"
 [features]
 default = ["clock", "std", "oldtime", "wasmbind"]
 alloc = []
-libc = []
 std = []
-clock = ["std", "winapi", "iana-time-zone"]
+clock = ["std", "winapi", "iana-time-zone", "libc"]
 oldtime = ["time"]
 wasmbind = ["wasm-bindgen", "js-sys"]
 unstable-locales = ["pure-rust-locales", "alloc"]
@@ -46,11 +45,11 @@ js-sys = { version = "0.3", optional = true }       # contains FFI bindings for 
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3.0", features = ["std", "minwinbase", "minwindef", "timezoneapi", "sysinfoapi"], optional = true }
 
-[target.'cfg(unix)'.dependencies]
+[target.'cfg(all(unix, not(target_os = "android")))'.dependencies]
 iana-time-zone = { version = "0.1.45", optional = true, features = ["fallback"] }
 
 [target.'cfg(target_os = "android")'.dependencies]
-android-tzdata = "0.1.1"
+libc = { version = "0.2.146", optional = true }
 
 [dev-dependencies]
 serde_json = { version = "1" }
--- a/src/offset/local/android.rs
+++ b/src/offset/local/android.rs
@@ -1,0 +1,76 @@
+use super::{FixedOffset, NaiveDateTime};
+use crate::traits::{Datelike, Timelike};
+use crate::LocalResult;
+use libc::{localtime_r, mktime, time_t};
+use std::io;
+use std::mem;
+use std::ptr::null;
+
+pub(super) fn offset_from_utc_datetime(utc: &NaiveDateTime) -> LocalResult<FixedOffset> {
+    let time = utc.timestamp() as time_t;
+    // Safe because an all-zero `struct tm` is valid.
+    let mut result = unsafe { mem::zeroed() };
+    // Safe because localtime_r only accesses the pointers passed to it during the call. It does
+    // also try to read the `TZ` environment variable, but we can assume it's not set on Android as
+    // the timezone is read from a system property instead.
+    unsafe {
+        if localtime_r(&time, &mut result).is_null() {
+            panic!("localtime_r failed: {}", io::Error::last_os_error());
+        }
+    }
+    LocalResult::Single(
+        FixedOffset::east_opt(
+            result.tm_gmtoff.try_into().expect("localtime_r returned invalid UTC offset"),
+        )
+        .expect("localtime_r returned invalid UTC offset"),
+    )
+}
+
+pub(super) fn offset_from_local_datetime(local: &NaiveDateTime) -> LocalResult<FixedOffset> {
+    // Calling mktime with different isdst values allows us to detect the ambiguous case where DST
+    // ends.
+    let (utc_timestamp, isdst) = mktime_with_dst(local, -1);
+    let (utc_timestamp0, isdst0) = mktime_with_dst(local, 0);
+    let (utc_timestamp1, isdst1) = mktime_with_dst(local, 1);
+    if utc_timestamp == -1 {
+        LocalResult::None
+    } else if isdst0 == isdst1 {
+        LocalResult::Single(offset_from_timestamp(local, utc_timestamp))
+    } else {
+        LocalResult::Ambiguous(
+            offset_from_timestamp(local, utc_timestamp0),
+            offset_from_timestamp(local, utc_timestamp1),
+        )
+    }
+}
+
+fn offset_from_timestamp(local: &NaiveDateTime, utc_timestamp: time_t) -> FixedOffset {
+    FixedOffset::east_opt(
+        (local.timestamp() - utc_timestamp).try_into().expect("Invalid UTC offset"),
+    )
+    .expect("Invalid UTC offset")
+}
+
+/// Calls `mktime` from libc with the given local date-time and isdst value.
+///
+/// Returns the timestamp and new isdst value set by `mktime`.
+fn mktime_with_dst(local: &NaiveDateTime, isdst: i32) -> (time_t, i32) {
+    let mut tm = libc::tm {
+        tm_sec: local.second() as i32,
+        tm_min: local.minute() as i32,
+        tm_hour: local.hour() as i32,
+        tm_mday: local.day() as i32,
+        tm_mon: local.month0() as i32,
+        tm_year: local.year() - 1900,
+        tm_wday: 0,
+        tm_yday: 0,
+        tm_isdst: isdst,
+        tm_gmtoff: 0,
+        tm_zone: null(),
+    };
+    // Safe because mktime only accesses struct it is passed during the call, and doesn't store the
+    // pointer to access later. It does also try to read the `TZ` environment variable, but we can
+    // assume it's not set on Android as the timezone is read from a system property instead.
+    let timestamp = unsafe { mktime(&mut tm) };
+    (timestamp, tm.tm_isdst)
+}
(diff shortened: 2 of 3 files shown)


djc:
So I'm honestly not sure after the latest discussion in #1018 whether I still want to go with using `localtime_r()` and `mktime()`, given that they're still not really safe.

pitdicker:
Is this summary correct?

Tthe current solution in chrono is fine for most users, and avoids `localtime_r` for which we got a security advisory in the past. This PR wants to reintroduce using `localtime_r`.

Reason is the use of chrono  in the android project. The dependency `iana-time-zone` is not imported there, because it in turn depends on `android_system_properties`. That one is not acceptable because it uses an unstable API, and duplicates the functionality of `rustutils`. But `iana-time-zone` can't switch to `rustutils` as it is not available on crates.io.

A second reason is that the file format used for the timezone database on android is not guaranteed to be stable. It seems to me the benefits of being able to make use of the OS timezone database instead of shipping a database with an application, which can get outdated, outweigh the potention breakage in the future. It should not be too hard to update the parser to the new format.

Maybe the use of `localtime_r` should be behind an off-by-default feature with a scary name, like `unsafe_android_libc`?

djc:
Yeah, I think your summary is correct. On top of that, future versions of Android might come with a safer libc API to get at the timezone offset data.

> Maybe the use of `localtime_r` should be behind an off-by-default feature with a scary name, like `unsafe_android_libc`?

Seems like a reasonable solution.

qwandor:
I've updated this PR to include the code to try at runtime to find the new `localtime_rz` and friends and use them if they are available. If not, it will fall back to `localtime_r` and `mktime`. I've tested this both on a nightly build of Android which has the new functions available, and an older build which doesn't. Both work as expected.

I disagree that "the current solution in chrono is fine for most users". Use of the system property as an API is not great, but depending on the format of the timezone database is worse (especially for applications outside of the platform) as it could change any time the database is updated, which happens fairly regularly. Suppose that a bunch of users have installed an application using chrono to handle timezones. The application works fine, until at some point in the future some country suddenly changes their daylight savings time rules in a way which requires new code to handle some new logic to handle. Android ships an update to the system timezone database, along with new parser code to handle it. All the applications using the supported APIs continue to work fine, but the application using chrono suddenly starts failing. To the users, it looks like Android just suddenly broke the application, but really the problem is that the application was using an unstable internal implementation detail rather than a supported API.

As far as I know it's not possible with cargo to make adding a feature remove a dependency, like the suggested `unsafe_android_libc` feature above. I guess we could do it the other way around, and have a feature like `android_use_unsupported_internal_api` to go back to the current approach of parsing the tzdata files, but the more different code paths we have the harder this is to test, and I don't think there's any situation where that is the right thing to do.

djc:
I agree that the current situation is not great. But, Android is the platform here that has chosen to deviate from standard ways of exposing the current timezone and system timezone database. You seem to make an argument that the platform can ship a new system timezone database and new parser code to handle potential format changes in it -- if you can ship that, why can't you also ship APIs that expose that data to userspace with a better API at the same time? That would also solve this problem AFAICT.

qwandor:
> > I disagree that "the current solution in chrono is fine for most users". Use of the system property as an API is not great, but depending on the format of the timezone database is worse (especially for applications outside of the platform) as it could change any time the database is updated, which happens fairly regularly.
> 
> I agree that the current situation is not great. But, Android is the platform here that has chosen to deviate from standard ways of exposing the current timezone and system timezone database. You seem to make an argument that the platform can ship a new system timezone database and new parser code to handle potential format changes in it -- if you can ship that, why can't you also ship APIs that expose that data to userspace with a better API at the same time? That would also solve this problem AFAICT.

On talking more to people internally about this, it sounds like I was partly wrong. Android doesn't generally make changes to the format and parser in mainline updates, usually just between releases. Still, the format is not guaranteed to be stable. And developers of Android applications probably want them to work correctly on devices running a range of different Android releases without having to rebuild them.

Android does expose APIs to userspace to access the timezone data. The main ones are the Java APIs, which are not very helpful here unless we add a bunch of JNI code to chrono, which seems like something it would be best to avoid. The supported native APIs up to API level 34 are `localtime_r` and friends. Adding a new API requires a new release, so we've added `tzalloc` and `localtime_rz` and so on in API level 35, which will be sometime after Android 14. I'd be happy for chrono to just use these functions, and that would be fine for Rust code in the platform, but application developers probably want their apps to work on older versions of Android too, without having to have different code paths for each version, which is why I've kept the fallback to `localtime_r` and `mktime`.

> (Shipping the timezone database in a format that regularly has to change to accomodate updates seems problematic to me -- why is this format better than the format which all other Linux distributions use, whose format has changed ~never?)

From what I understand there are changes in other Linux distributions too. The syntax of the format is stable, but the semantics (i.e. what gets filled in and the meaning of what's there) is less stable.

djc:
If the format is usually not changed between mainline updates, I think that means we're safe if we (a) keep the current strategy for current/old Android versions and (b) use the new `localtime_rz` stuff for newer platforms, right? Anecdotally the current strategy is working for Android versions that people are using chrono on, and as the new libc APIs roll out we'll transparently upgrade to a stable supported libc API.

qwandor:
I've added a feature, PTAL.

Kijewski:
I'm a maintainer of `iana-time-zone`. What I would like to know is: Why aren't you reaching out  to [`android_system_properties`](https://github.com/nical/android_system_properties/)? The library is used in [44.5k](https://github.com/nical/android_system_properties/network/dependents) projects, so if you found some usage of restricted and/or unstable API, then why are you so vague in calling out the problems? The library is [really small and well documented](https://github.com/nical/android_system_properties/blob/b988701d228920111502e4f0b0270472c49f76f3/src/lib.rs).

The library could be much smaller if Rust had weak symbols or if it dropped support for Android < 5. The former won't happen anytime soon. The latter would be a downgrade. (I guess you guys want to sell more new phones, so maybe it would be an upgrade to you.) ;) Your [rustutils](https://cs.android.com/android/platform/superproject/+/master:system/librustutils/system_properties.rs) in fact does not seem to support Android < 5, does it?

djc:
> I'm a maintainer of `iana-time-zone`. What I would like to know is: Why aren't you reaching out to [`android_system_properties`](https://github.com/nical/android_system_properties/)? The library is used in [44.5k](https://github.com/nical/android_system_properties/network/dependents) projects, so if you found some usage of restricted and/or unstable API, then why are you so vague in calling out the problems? The library is [really small and well documented](https://github.com/nical/android_system_properties/blob/b988701d228920111502e4f0b0270472c49f76f3/src/lib.rs).

Did you already read all of the preceding discussion in #1018?

Kijewski:
I did. I realize that the discussion was about the format of Android's tzdata, too, not only `android_system_properties`, but the usage of an unstable API function was mentioned why `android_system_properties` could not be used in their projects. Yes, `__system_property_get()` should not be used anymore, but it is only used as a fallback if `__system_property_read_callback()` was not found.

I found #1018 quite odd. In my book "TL;TR: It uses unstable API" does not count as an answer. Multiple questions about the unstable API weren't (properly) answered. In my not-so-humble opinion it's a bad style from the googlers, that they didn't try to contact the downstream libraries to maybe work with them, because I think it would be in their best interest, too.

(Actually, I was only looking if #1129 was resolved, and was only looking through the issues because of that. Then the title of the PR just caught my eye, whether there was something that could be fixed/enhanced in `iana-time-zone`, etc.)

Kijewski:
@qwandor, thank you for the explanation!

pitdicker:
> Use of the system property as an API is not great, but depending on the format of the timezone database is worse (especially for applications outside of the platform) as it could change any time the database is updated, which happens fairly regularly. … The application works fine, until at some point in the future some country suddenly changes their daylight savings time rules in a way which requires new code to handle some new logic to handle. Android ships an update to the system timezone database, along with new parser code to handle it.

Android is just using TZif files, just like Linux, macOS and the BSDs. There is RFC 8536 to describe the format. If worse comes to worse and there is some rule for future dates that can't be expressed, current practice is to just fall back to a long list of transition dates. This already happens for countries that adjust the clock around ramadan, which depends on the moon. I don't see the format breaking anytime soon.

pitdicker:
The nice thing about doing the calculations in chrono is that we can improve over what libc offers:
- We can return the offset abbreviation, as in https://github.com/chronotope/chrono/pull/750 (to fix issue https://github.com/chronotope/chrono/issues/749).
- We can return `LocalResult::Ambiguous` if a datetime falls in a transition fold.
- We can return `LocalResult::None` if a datetime falls in a transition gap, instead of a bogus value.
- We may (in a future release) tell what the surrounding valid dates are if a datetime falls in a transition gap.

pitdicker:
> From what I understand there are changes in other Linux distributions too. The syntax of the format is stable, but the semantics (i.e. what gets filled in and the meaning of what's there) is less stable.

As far as I know they can choose whether to include `backzone`, and the range of dates to include a fast lookup for, before depending on `TzRule`.

pitdicker:
> I work on the platform so my primary concern is with that, and the immediate issue I'm trying to solve is that we are blocked on updating the version of chrono we vendor into the platform. I would also like Android applications to be using the correct APIs, so I'm trying to help with that where I can.

@qwandor Just want to say that although I am doing difficult and don't like the direction, I appreciate your work.

pitdicker:
That is the current status of this PR?

qwandor:
I added a feature flag in 3923d5c to revert to the old behaviour, this is awaiting review.

pitdicker:
Reintroducing [CVE-2020-26235](https://nvd.nist.gov/vuln/detail/CVE-2020-26235)/[RUSTSEC-2020-0159](https://rustsec.org/advisories/RUSTSEC-2020-0159.html), which caused chrono a lot of pain (and moved a good number of users to time-rs which was earlier with a kind of fix), is not something to quickly step over. I still think the motivation 'we don't want to use `iana-time-zone` in the Android platform' is not strong enough.

With the changes to the documentation and the following discussion, keep in mind that for us it reads like: reintroduce CVE on android to avoid using an unstable API.

> If the format is usually not changed between mainline updates, I think that means we're safe if we (a) keep the current strategy for current/old Android versions and (b) use the new `localtime_rz` stuff for newer platforms, right? Anecdotally the current strategy is working for Android versions that people are using chrono on, and as the new libc APIs roll out we'll transparently upgrade to a stable supported libc API.

If we go with this approach this PR could make easier progress.

pitdicker:
I've opened an issue for using `localtime_rz` on Android if available.
Closing this PR.

Issue #1127, opened 2023-06-05
RFC: a new formatting API


djc:
I don't think #1121 is the right approach. The fallibility here is with (a) parsing the format string, which might be invalid, and (b) whether the type we're formatting from contains enough information to fill out the fields the format string has.

Making something work on 0.4.x is "trivial" in a sense because we can just invent new API that does whatever we want. Postponing the failures of parsing from the parsing stage to the formatting stage doesn't sound like a great option, in particular a reasonable approach is to use a long-lived formatter (parsed from a string) for many values. For getting enough data out of the input type, that seems like something we could assert at compile time using generics.

As such, I think the high-level structure here should be:

1. A fallible function/method that converts a format string to `Vec` of items (reusing the existing item type if we can)
2. A fallible function/method that binds a `Vec` or slice (for the non-`alloc` case) to a particular input type, failing if the type does not contain enough data for the input type
3. The outcome of step 2 would be something like a `struct Widget<'a, I> { items: Cow<'a, [Items]>, input: PhantomData<I> }` (replacing the `Cow` with just a slice if `alloc` is not enabled, I think?) which implements `Display` and will only fail (as you explained in https://github.com/chronotope/chrono/pull/902#issuecomment-1575519183) if the inner `Formatter` fails
4. Probably some high level API that wraps over step 2 and 3 above, particularly for the allocating case

To the extent that we can't implement this (efficiently) within the current API (from what I've seen, this will just get ugly with the approach chosen in #902), we should invent a new entry point. The new entry point should be able to satisfy all the use cases for the existing entry points so that we can, at some point in the near future, start deprecating the existing entry points. In particular, there should be affordances for the non-`alloc` use case and there should be a clean way (that is, with minimal code duplication) to factor in the localization stuff.

cc @jaggededgejustice @pitdicker

pitdicker:
> Postponing the failures of parsing from the parsing stage to the formatting stage doesn't sound like a great option, in particular a reasonable approach is to use a long-lived formatter (parsed from a string) for many values.

That is a good usecase that is not yet covered by the solution in https://github.com/chronotope/chrono/pull/902.

> For getting enough data out of the input type, that seems like something we could assert at compile time using generics.

While far less common, I think there is some use to keeping the formatting string dynamic. But we can get pretty much the same with const fns?

> 2. A fallible function/method that binds a `Vec` or slice (for the non-`alloc` case) to a particular input type, failing if the type does not contain enough data for the input type

[...]

PR #1152, opened 2023-06-23
Make `StrftimeItems` with `unstable-locales` work without allocating

pitdicker:
For https://github.com/chronotope/chrono/issues/1127 I am slowly exploring some options around `StrftimeItems`. Nothing I am happy with yet though :smile:.

This seems like a simple and useful change by itself. This makes `StrftimeItems` quite a bit faster with the `unstable-locales` feature, work without allocating, and shrinks the size of `StrftimeItems`.

Previously with the `unstable_locales` feature, on every `StrftimeItems::new_localized` we would lookup and parse three localized formatting strings and store them in `Vec`'s. Just in case they would be needed.

I changed it to delay the lookup until there is a specifier that uses one.
And instead of pre-parsing the localized format string into a `Vec` and returning the items one by one, we now store only a reference to the localized string and switch to parsing that, and when done continue with parsing the user-supplied format string.

The size of `StrftimeItems` goes from 80 bytes to 32 without the `unstable-locales` feature, and from 112 bytes to 56 with the feature.

### DIFF
--- a/src/format/strftime.rs
+++ b/src/format/strftime.rs
@@ -742,4 +742,21 @@ mod tests {
         let mut buf = String::new();
         let _ = write!(buf, "{}", dt.format("%#z")).expect_err("parse-only formatter should fail");
     }
+
+    #[test]
+    #[cfg(not(feature = "unstable-locales"))]
+    fn test_type_sizes() {
+        use core::mem::size_of;
+        assert_eq!(size_of::<Item>(), 24);
+        assert_eq!(size_of::<StrftimeItems>(), 80);
+    }
+
+    #[test]
+    #[cfg(feature = "unstable-locales")]
+    fn test_type_sizes() {
+        use core::mem::size_of;
+        assert_eq!(size_of::<Item>(), 24);
+        assert_eq!(size_of::<StrftimeItems>(), 112);
+        assert_eq!(size_of::<Locale>(), 2);
+    }
 }
--- a/src/format/strftime.rs
+++ b/src/format/strftime.rs
@@ -181,19 +181,8 @@ Notes:
 */
 
 #[cfg(feature = "unstable-locales")]
-extern crate alloc;
-
-#[cfg(feature = "unstable-locales")]
-use alloc::vec::Vec;
-
-#[cfg(feature = "unstable-locales")]
 use super::{locales, Locale};
 use super::{Fixed, InternalFixed, InternalInternal, Item, Numeric, Pad};
-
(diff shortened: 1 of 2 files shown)


pitdicker:
Benchmark before (`cargo bench --features=__internal_bench,unstable-locales bench_parse_strftime`):
```
bench_parse_strftime    time:   [341.12 ns 343.02 ns 344.99 ns]

bench_parse_strftime_localized
                        time:   [1.1265 µs 1.1274 µs 1.1284 µs]
  5 (5.00%) high mild
  1 (1.00%) high severe
```

After:
```
     Running benches\chrono.rs (target\release\deps\chrono-6c281f2c5c0fa9af.exe)
bench_parse_strftime    time:   [317.97 ns 318.33 ns 318.72 ns]
                        change: [-7.8388% -7.3233% -6.8142%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low severe
  2 (2.00%) low mild
  6 (6.00%) high mild
  1 (1.00%) high severe

bench_parse_strftime_localized
                        time:   [319.14 ns 320.01 ns 320.96 ns]
                        change: [-71.850% -71.765% -71.669%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
```

src/format/strftime.rs
```diff
@@ -387,12 +335,36 @@ impl<'a> StrftimeItems<'a> {
                     'U' => num0!(WeekFromSun),
                     'V' => num0!(IsoWeek),
                     'W' => num0!(WeekFromMon),
-                    'X' => queue_from_slice!(self.t_fmt),
+                    #[cfg(not(feature = "unstable-locales"))]
+                    'X' => queue_from_slice!(T_FMT),
+                    #[cfg(feature = "unstable-locales")]
+                    'X' => {
+                        if let Some(locale) = self.locale {
```
djc, line 342: Can you add a helper function for this? The duplication is a bit unfortunate.

djc (commented):
Looks nice!

src/format/strftime.rs
pitdicker: I am not great with macro's but gave it a try with a method.

djc (approved):
Nice! I prefer a method over a macro, anyway.

pitdicker:
Thank you.

cecton:
I'm really happy to see this finally happening. I made a few serious attempts back then but I couldn't find a good solution

PR #1156, opened 2023-06-27
Factor out formatting to `format/formatting.rs`

pitdicker:
I am going to regret this because of the other PRs that will have to be rebased :smile:.


src/format/parse.rs
```diff
@@ -574,6 +574,41 @@ mod tests {
                 assert_eq!(parse_all($fmt, &$items), Ok(expected))
             });
         }
+        macro_rules! lit {
```
djc, line 577: Having two copies of this does not seem like an improvement.

djc (approved):

src/format/parse.rs
pitdicker: They were the reason the submodule declaration where low in the file: only macros declared on the line above are available to submodules.

> Having two copies of this does not seem like an improvement.

I kind of agree. They were used only in one method `StrftimeItems` until a large test in `format::parse` was added.
Because we duplicate macro's in many places to make tests shorter it seemed acceptable to me.

pitdicker: I don't mind changing it if you see a better way.

djc: I think my solution would be to make these normal functions (maybe with `#[inline(always)]` instead of macros) and/or import enum variants into local namespaces instead. IMO none of these really need to be macros.

pitdicker:
I forgot to mark the functions with `#[inline]`. But the benchmark is unaffected, so I prefer not to add them.

src/format/mod.rs
```diff
@@ -295,40 +295,20 @@ pub enum Item<'a> {
...
-        Item::Literal($x)
-    };
+const fn num(numeric: Numeric) -> Item<'static> {
+    Item::Numeric(numeric, Pad::None)
 }
-macro_rules! sp {
-    ($x:expr) => {
-        Item::Space($x)
-    };
+const fn num0(numeric: Numeric) -> Item<'static> {
```
djc, line 310: Please keep empty lines between these functions.

src/format/parse.rs
```diff
@@ -561,6 +561,9 @@ impl str::FromStr for DateTime<FixedOffset> {
 
 #[cfg(test)]
 mod tests {
+    use crate::format::InternalInternal as II;
```
djc, line 564: I think we should probably just import the `InternalInternal::*` variants too, as well as the `Fixed::*` ones. Are these imports in the smallest enclosing scope? It looks like they're all being used in `test_parse()` only.

src/format/strftime.rs
```diff
@@ -180,28 +180,31 @@ Notes:
    China Daylight Time.
 */
 
+use super::Item::{self, Literal, Space};
```
djc, line 183: At the module level I'd prefer not to import variants.

src/format/strftime.rs
```diff
@@ -318,118 +321,126 @@ impl<'a> StrftimeItems<'a> {
                 }
 
                 let item = match spec {
-                    'A' => fix!(LongWeekdayName),
-                    'B' => fix!(LongMonthName),
-                    'C' => num0!(YearDiv100),
+                    'A' => fixed(Fixed::LongWeekdayName),
```
djc, line 324: But let's import all the variants right before this `match`?

src/format/strftime.rs
```diff
@@ -492,9 +503,12 @@ impl<'a> StrftimeItems<'a> {
 
 #[cfg(test)]
 mod tests {
+    use super::StrftimeItems;
```
djc, line 506: (As elsewhere, except I'd be okay with importing some often-used variants inside `tests` if they're used by multiple tests.)

src/format/parse.rs
pitdicker: We can't import the variants of both `Fixed` and `Numeric`, because some of the names overlap.

pitdicker:
👍 Good comments, much better to limit the scope where the variants are directly available.

I moved the definitions of `D_FMT`, `D_T_FMT` and `T_FMT` into `parse_next_item` so they can make use of the import.

src/format/parse.rs
djc: Okay, that's unfortunate.

djc (approved):
Go ahead and merge this, looking good!

pitdicker:
Thank you for the extra permissions! First merge here :smile:

djc:
> Thank you for the extra permissions! First merge here 😄

You've definitely earned them. 😄

PR #1157, opened 2023-06-28
Add FromStr for FixedOffset

mcronce:
#1092 appears to have stalled; I addressed the feedback from that original PR and added a few quick test cases


djc (approved):

djc:
Thanks!

PR #1158, opened 2023-06-29
Offset parsing cleanup

djc:
Start with some cleanups from #1157, simplify some of the surrounding code.

### DIFF
--- a/src/format/parse.rs
+++ b/src/format/parse.rs
@@ -445,7 +445,7 @@ where
                     }
 
                     &TimezoneName => {
-                        try_consume!(scan::timezone_name_skip(s));
+                        try_consume!(Ok((s.trim_start_matches(|c: char| !c.is_whitespace()), ())));
                     }
 
                     &TimezoneOffsetColon
--- a/src/format/parse.rs
+++ b/src/format/parse.rs
@@ -454,7 +454,9 @@ where
                     | &TimezoneOffset => {
                         let offset = try_consume!(scan::timezone_offset(
                             s.trim_start(),
-                            scan::colon_or_space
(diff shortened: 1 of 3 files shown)


pitdicker (approved):
All good improvements :+1:.

I would like to switch to a better offset parser (https://github.com/chronotope/chrono/pull/1083) at some point though.

djc:
I've removed part 1 of your offset series. Rewriting the parser in one go assumes that we have good test coverage for this code. I've been trying to set up codecov.io but it's not working so far, meanwhile it might make sense to just use `cargo-llvm-cov` to get manual coverage reports. As long as I don't trust code coverage I'd prefer incrementally rewriting code (one small commit at a time, as I did here) instead of replacing it outright with a new implementation (which makes it harder to review how the old implementation influenced the new one).

pitdicker:
> I've removed part 1 of your offset series. Rewriting the parser in one go assumes that we have good test coverage for this code.

Thank you.

Thanks to @jtmoon79 we have very exhaustive coverage of time offset parsing code.

I don't really see a way to reach the result in https://github.com/chronotope/chrono/pull/1083 with only incremental changes.
Does it help that the coverage of the new `scan::utc_offset` and `scan:parse_offset_minute` are 100% according to `cargo-llvm-cov`.?

djc:
Where do you think you'd have trouble reaching the result incrementally? It seems sort of straightforward to me.

> Does it help that the coverage of the new `scan::utc_offset` and `scan:parse_offset_minute` are 100% according to `cargo-llvm-cov`.?

Yes, that definitely helps.

PR #1159, opened 2023-06-30
Feature gate tests instead of methods

pitdicker:
This gets rid of the `#[cfg(any(feature = "alloc", feature = "std", test))]` pattern on methods.

I adjusted parsing tests `test_rfc3339` and `test_rfc3339` to work without formatting. That seemed better to me than to make them conditional.

### DIFF
--- a/src/date.rs
+++ b/src/date.rs
@@ -4,7 +4,7 @@
 //! ISO 8601 calendar date with time zone.
 #![allow(deprecated)]
 
-#[cfg(any(feature = "alloc", feature = "std", test))]
+#[cfg(any(feature = "alloc", feature = "std"))]
 use core::borrow::Borrow;
 use core::cmp::Ordering;
 use core::ops::{Add, AddAssign, Sub, SubAssign};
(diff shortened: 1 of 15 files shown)


src/date.rs
```diff
@@ -4,7 +4,7 @@
 //! ISO 8601 calendar date with time zone.
 #![allow(deprecated)]
 
-#[cfg(any(feature = "alloc", feature = "std", test))]
+#[cfg(any(feature = "alloc", feature = "std"))]
```
djc, line 7: The commit message on this one seems confusing -- AFAICT it was already only available on `alloc`/`std`, but now is no longer available on `test`?

djc (commented):
Did you check that the feature gates work at each intermediate commit (I suppose in this case, that we can run tests in `no_std` mode)? That's usually my goal with these kinds of things.

pitdicker:
> Did you check that the feature gates work at each intermediate commit (I suppose in this case, that we can run tests in `no_std` mode)? That's usually my goal with these kinds of things.

That was my goal also. The commits first adjust the tests, than change the cfg.

pitdicker:
`test_from_system_time` was duplicated for `windows` and non-`windows` targets. The only difference was the precision of the nanoseconds. I don't think we have to test this detail of `SystemTime` from the standard library.

djc (approved):

PR #1162, opened 2023-07-01
Add feature gates

pitdicker:
In https://github.com/chronotope/chrono/pull/1159 I added feature gates to unit tests, but forgot about the integration tests.
Probably because I was working on Windows at the time, where they don't run.

### DIFF
--- a/tests/dateutils.rs
+++ b/tests/dateutils.rs
@@ -1,14 +1,8 @@
-#[cfg(unix)]
-use chrono::offset::TimeZone;
-#[cfg(unix)]
-use chrono::Local;
-#[cfg(unix)]
-use chrono::{Datelike, NaiveDate, NaiveDateTime, NaiveTime, Timelike};
+#![cfg(all(unix, feature = "clock", feature = "std"))]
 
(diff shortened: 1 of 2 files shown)


djc:
Please never merge without getting a review -- I've enforced that now with branch permissions, and will in principle also always wait for your review.

djc:
Don't we run these integration tests in CI? How did we miss this in the original PR?

pitdicker:
> Please never merge without getting a review -- I've enforced that now with branch permissions, and will in principle also always wait for your review.

Sorry, won't happen again. I was finding my way.

> Don't we run these integration tests in CI? How did we miss this in the original PR?

No, we only run `no_std` tests [here](https://github.com/chronotope/chrono/blob/0.4.x/.github/workflows/test.yml#L79-L97) with:
```
cargo test --lib --no-default-features
cargo test --doc --no-default-features
```

Do you want me to add it?

djc:
Yes, probably a good idea.

djc:
Thanks for noticing -- added branch protection for 0.4.x to mirror main now.

Issue #94, opened 2016-09-09
Getting time and  formatting it seems slow.


dpc:
Working on `slog-rs` - a logger, I've been doing some benchmarking, and getting a current timestamp (formatted) from chrono is taking a lot of time.

Fist I've discovered that just taking time (without formatting) takes around 60ns in my benchmarks. Does it require doing two syscalls (one for time, one for timezone)? I've filled https://github.com/rust-lang/rust/issues/36358 to inquire why it isn't faster.

Second formatting is taking around 500ns, which seems to much. I wish formatting provided something like `fn fmt_rfc3339<W : io::Write>(&self, io : W) -> io::Result<()>` which does not require any allocation. Also, when investigating the code, it seems there's a lot of `cloned()` calls, and the whole thing is based on `StrftimeItems` which like `printf` in C: scans the format every time.

lifthrasiir:
I'm right on the trip and the reply may be not quick. That said...

> Does it require doing two syscalls (one for time, one for timezone)?

Time zone is a local concept to the current process, normally implemented as a part of libc (glibc has *tons* of the code dedicated for this, and includes a complete reader for tzfile). Right now Chrono depends on libtime for this task, that is AFAIK just a call to `gettimeofday` (a system call) and `localtime_r` (may do a bunch of things at the first time, and probably locks the internal data structure).

I'm actually thinking of replacing them with an independent implementation (I've met some crazy bugs from libtime), and I guess `localtime_r` can be completely eliminated---Chrono's `Local` can be designed to make a common case extremely faster. Per `gettimeofday`, I do believe that it *should* be replaced with `clock_gettime` anyway (that has nsec precision) and completely agree that vDSO can be used if possible. I will take a note for that :) (By the way, I was currently somehow blocked on getting a nsec-level timestamp in macOS, which seems to be too complex...)

> Second formatting is taking around 500ns, which seems too much. I wish formatting provided something like `fn fmt_rfc3339<W : io::Write>(&self, io : W) -> io::Result<()>` which does not require any allocation.

The following code would be equivalent (untested):

```rust
use chrono::format::{Item, Fixed};
datetime.format_with_items([Item::Fixed(Fixed::RFC3339)].iter().cloned())
```

I think this should be reasonably fast, but it still depends on the entire formatting architecture of Rust that is not very fast. (It does not allocate, but has tons of indirections.) I believe Chrono excels at breaking the time components apart, though.

[...]

PR #1163, opened 2023-07-01
New `Formatter` to replace `DelayedFormat`

pitdicker:
Add a new formatter that can work without allocating and has less performance overhead, as mentioned in https://github.com/chronotope/chrono/issues/1127#issuecomment-1612471735.

I have made an effort to make the formatting code more readable, and split it up into reasonable commits. ~My first PR to chrono that adds functionality but ends up with less lines :smile:.~

~I would like to depend on this PR to get https://github.com/chronotope/chrono/pull/1035 over the finish line. Serialization worked without allocating before, and the new offset formatter should work without allocating in order to use it there.~

This doesn't yet include extra benchmarks or the rest of my proposal for the new formatting API. But I have enough in another branch to test it all works.

Four standalone `format*` functions [`chrono::format`](https://docs.rs/chrono/latest/chrono/format/index.html#functions) are deprecated: they were near unusable, taking `&mut Formatter<'_>` as an argument, which can't be constructed outside of a `Display` implementation.

----

#### Why a new formatter?

This adds a new type `Formatter<I, O>`, that can replace `DelayedFormat<I>` in a new formatting API.

Our current formatting code has noticeable overhead (see https://github.com/chronotope/chrono/issues/94). This is mostly caused by one design choice: if `DelayedFormat::new` is given an offset, it first formats a timezone name into a `String`. In case of `DateTime<FixedOffset>` and `DateTime<Local>` this involves formatting the offset, as there is no name available. This is responsible for 20~25% of the ca. 40% overhead in the `format_with_items` benchmark.

`Formatter<I, O>` takes the offset as a generic, so it can delay formatting the timezone name until it is needed, which often is never.

I made a couple of other changes that all reduce the overhead a tiny bit or help readablility: split the formatting function into smaller methods on `Formatter`, make better use of `match`, format using a smaller integer type `i32`, and delay getting locale strings until use.

We also format to `impl Write` instead of `String` like in https://github.com/chronotope/chrono/pull/1126 to work without `alloc`.

All combined we have only about ~10% overhead left compared to the `format_manual` benchmark, which seems good enough to me.

### DIFF
--- a/src/format/formatting.rs
+++ b/src/format/formatting.rs
@@ -90,6 +90,240 @@ where
     ) -> Formatter<I, Off> {
         Formatter { date, time, offset, items, locale }
     }
+
+    #[inline]
+    fn format(&self, w: &mut impl Write) -> fmt::Result {
+        for item in self.items.clone() {
+            match *item.borrow() {
+                Item::Literal(s) | Item::Space(s) => w.write_str(s),
+                Item::OwnedLiteral(ref s) | Item::OwnedSpace(ref s) => w.write_str(s),
+                Item::Numeric(ref spec, pad) => self.format_numeric(w, spec, pad),
+                Item::Fixed(ref spec) => self.format_fixed(w, spec),
+                Item::Error => Err(fmt::Error),
+            }?;
+        }
+        Ok(())
+    }
+
+    fn format_numeric(&self, w: &mut impl Write, spec: &Numeric, pad: Pad) -> fmt::Result {
+        use self::Numeric::*;
+
+        let week_from_sun = |d: NaiveDate| d.weeks_from(Weekday::Sun);
+        let week_from_mon = |d: NaiveDate| d.weeks_from(Weekday::Mon);
+
+        let (width, v) = match *spec {
+            Year => (4, self.date.map(|d| i64::from(d.year()))),
+            YearDiv100 => (2, self.date.map(|d| i64::from(d.year()).div_euclid(100))),
+            YearMod100 => (2, self.date.map(|d| i64::from(d.year()).rem_euclid(100))),
+            IsoYear => (4, self.date.map(|d| i64::from(d.iso_week().year()))),
+            IsoYearDiv100 => (2, self.date.map(|d| i64::from(d.iso_week().year()).div_euclid(100))),
+            IsoYearMod100 => (2, self.date.map(|d| i64::from(d.iso_week().year()).rem_euclid(100))),
+            Month => (2, self.date.map(|d| i64::from(d.month()))),
+            Day => (2, self.date.map(|d| i64::from(d.day()))),
+            WeekFromSun => (2, self.date.map(|d| i64::from(week_from_sun(d)))),
+            WeekFromMon => (2, self.date.map(|d| i64::from(week_from_mon(d)))),
+            IsoWeek => (2, self.date.map(|d| i64::from(d.iso_week().week()))),
+            NumDaysFromSun => (1, self.date.map(|d| i64::from(d.weekday().num_days_from_sunday()))),
+            WeekdayFromMon => (1, self.date.map(|d| i64::from(d.weekday().number_from_monday()))),
+            Ordinal => (3, self.date.map(|d| i64::from(d.ordinal()))),
+            Hour => (2, self.time.map(|t| i64::from(t.hour()))),
+            Hour12 => (2, self.time.map(|t| i64::from(t.hour12().1))),
+            Minute => (2, self.time.map(|t| i64::from(t.minute()))),
+            Second => {
+                (2, self.time.map(|t| i64::from(t.second() + t.nanosecond() / 1_000_000_000)))
+            }
+            Nanosecond => (9, self.time.map(|t| i64::from(t.nanosecond() % 1_000_000_000))),
+            Timestamp => (
+                1,
+                match (self.date, self.time, self.offset.as_ref()) {
+                    (Some(d), Some(t), None) => Some(d.and_time(t).timestamp()),
+                    (Some(d), Some(t), Some(off)) => {
+                        Some(d.and_time(t).timestamp() - i64::from(off.fix().local_minus_utc()))
+                    }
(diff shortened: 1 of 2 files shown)


pitdicker:
I added the ability to format with [formatting parameters](https://doc.rust-lang.org/std/fmt/index.html#formatting-parameters) that can work without allocations in `no_std`.

We need two adapters for `fmt::Write`. One that can count the number of characters written, so we know how much padding is required. And another that can truncate if we write more than the maximum allowed characters.

This does add some complexity, but not too much in my opinion. And it makes formatting with `no_std` support the same features as with `std`.

jtmoon79:
I really like this PR for the performance benefits.

A request for your PRs @pitdicker , are you able to set your PRs to Draft until you feel done with additions?  While I don't mind doing PR reviews, it might be more efficient for both of us if I know that a PR is "still cooking" before deciding to dive in.  Just a suggestion.

pitdicker:
Thank for the reviews :smile:.

I intended this to be done. This PR and two others (maybe more are coming) are split of from my branch to add a new formatting API which I keep working at, sorry.... Now at 50 commits :disappointed: 

In this PR I adjusted the benchmarks to be comparable to the `to_rfc3339` benchmark, to get a second baseline besides `format_manual`.

Before:
```
bench_format            time:   [875.81 ns 877.40 ns 879.11 ns]
Found 9 outliers among 100 measurements (9.00%)
  8 (8.00%) high mild
  1 (1.00%) high severe

bench_format_with_items time:   [705.81 ns 707.47 ns 709.34 ns]
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high mild

bench_format_manual     time:   [589.20 ns 591.01 ns 593.20 ns]
Found 11 outliers among 100 measurements (11.00%)
  4 (4.00%) high mild
  7 (7.00%) high severe

bench_datetime_to_rfc3339
                        time:   [232.58 ns 233.11 ns 233.65 ns]
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
```

The interesting numbers are:
- `format_manual` using the formatting machinery of the standard library is ca. 2,5× slower `to_rfc3339`.
- `format_with items` and `format` are 1,2× and 1,5× slower respectively than `format_manual`.

With my WIP branch that relies on this PR the numbers are:
```
bench_format            time:   [727.49 ns 731.78 ns 736.64 ns]
                        change: [-13.988% -12.839% -11.527%] (p = 0.00 < 0.05)
                        Performance has improved.

bench_format_to_string  time:   [574.73 ns 577.34 ns 579.79 ns]
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild

bench_format_with       time:   [441.08 ns 442.00 ns 443.02 ns]
  6 (6.00%) high mild

bench_format_with_items time:   [564.46 ns 566.93 ns 569.68 ns]
                        change: [-20.521% -20.049% -19.585%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high mild

bench_format_manual     time:   [590.28 ns 592.68 ns 595.31 ns]
                        change: [+0.1228% +1.0523% +1.9683%] (p = 0.03 < 0.05)
                        Change within noise threshold.
Found 9 outliers among 100 measurements (9.00%)
  6 (6.00%) high mild
  3 (3.00%) high severe

bench_datetime_to_rfc3339
                        time:   [169.69 ns 170.18 ns 170.66 ns]
                        change: [-27.113% -26.795% -26.469%] (p = 0.00 < 0.05)
                        Performance has improved.
  1 (1.00%) low mild
  4 (4.00%) high mild
  1 (1.00%) high severe

bench_datetime_to_rfc3339_opts
                        time:   [138.26 ns 138.67 ns 139.11 ns]
  3 (3.00%) low mild
  2 (2.00%) high mild
  2 (2.00%) high severe
```

The new benchmarks `format_to_string` and `format_with` use the new formatter directly without wrapping it in the old `DelayedFormat`.
- The new `format_to_string` has comparable performance to `format_manual`
- The new `format_with` takes only 0,75× the time of `format_manual`, and ca. 60% of the old `format_with_items`.

But I also found a way to optimize `to_rfc3339` methods a little more. `format_with` is 2,6× times slower than that (no PR open for that yet).

---

Most of our time with formatting is spend converting numbers to strings of decimals. Printing the 9 digits of a nanosecond takes for example ca. 40% of the time spend in `to_rfc3339`.

The trick is to copy number formatting for small numbers from `to_rfc3339`: special-case numbers with only one or two digits to not go trough the number formatting of the standard library.

pitdicker:
Disclaimer: Of course a lot of the performance depends on which formatting items are used. The benchmarked RFC 3339 format only uses numbers for example. So these numbers are only an indication.

pitdicker:
Included a commit from my WIP branch here: collect all formatting tests in the `formatting` module. So we can see a bit better which parts are covered.

src/format/formatting.rs
```diff
@@ -5,58 +5,395 @@
...
+    pub fn new(
+        date: Option<NaiveDate>,
+        time: Option<NaiveTime>,
+        offset: Option<Off>,
+        items: I,
+    ) -> Formatter<I, Off> {
+        Formatter { date, time, offset, items, locale: default_locale() }
+    }
+
+    /// Makes a new `DelayedFormat` value out of local date and time, UTC offset and locale.
```
jtmoon79, line 62: I think you meant:

      /// Makes a new `Formatter` value out of ...

src/format/formatting.rs
```diff
@@ -5,58 +5,395 @@
...
+}
+
+impl<'a, I, B, Off> Formatter<I, Off>
+where
+    I: Iterator<Item = B> + Clone,
+    B: Borrow<Item<'a>>,
+    Off: Offset + Display,
+{
+    /// Makes a new `Formatter` value out of local date and time and UTC offset.
+    ///
```
jtmoon79, line 46: I suggest mentioning

     /// Uses the `default_locale()` as the locale.

PR #1058, opened 2023-05-11
 Fix locale formatting for `%c` and `%r`

scarf005:
- fixes #985
changed to target `0.4.x` instead of main in #988

PR #1165, opened 2023-07-03
Fix locale formatting for `%c` and `%r`

pitdicker:
Just opening for the CI for now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ time = { version = "0.1.43", optional = true }
 num-traits = { version = "0.2", default-features = false }
 rustc-serialize = { version = "0.3.20", optional = true }
 serde = { version = "1.0.99", default-features = false, optional = true }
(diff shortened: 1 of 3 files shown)


pitdicker:
So... Basically I felt bad for @scarf005 for breaking their approach in https://github.com/chronotope/chrono/pull/1058 with https://github.com/chronotope/chrono/pull/1152. And I wondered if we couldn't make the fix in `pure-rust-locales`. Making an alternative PR was not the intention.

This branch fails to build with my branch of `pure-rust-locales` because it seems cargo doesn't correctly checkout a repro with large file storage (I think?). But the tests succeed with a local checkout and a relative path in `cargo.toml`.

The corresponding PR to `pure-rust-locales` is https://github.com/cecton/pure-rust-locales/pull/5.

@djc Would you be willing to review this in combination with https://github.com/cecton/pure-rust-locales/pull/5?

cc @cecton

djc (approved):
Nice work!

PR #1166, opened 2023-07-03
Also run integration tests with `no_std` on CI

pitdicker:
See https://github.com/chronotope/chrono/pull/1162#issuecomment-1617653089.


pitdicker:
The CI tried to run the test, see https://github.com/chronotope/chrono/actions/runs/5446717690/jobs/9907851402?pr=1166. But of course didn't run it because of the feature flag.

pitdicker:
It doesn't run the test, but does verify that the we can just run `cargo test --no-default-features` without compilation errors.

PR #1168, opened 2023-07-03
Localize decimal point

pitdicker:
Under the `unstable-locales` feature we did not yet localize the decimal point for fractional seconds.


jtmoon79 (approved):
LGTM

This is a straightforward addition and improvement for locales.

djc (approved):

PR #1170, opened 2023-07-05
Split `test_parse`

pitdicker:
Combined with the extra tests on the main branch this test got so large it did not even compile on all targets (WASM).

### DIFF
--- a/src/format/parse.rs
+++ b/src/format/parse.rs
@@ -570,7 +570,6 @@ mod tests {
         use crate::format::Item::{Literal, Space};
         use crate::format::Numeric::*;
 
-        // workaround for Rust issue #22255
         fn parse_all(s: &str, items: &[Item]) -> ParseResult<Parsed> {
             let mut parsed = Parsed::new();
             parse(&mut parsed, s, items.iter())?;
--- a/src/format/parse.rs
+++ b/src/format/parse.rs
@@ -720,84 +720,84 @@ mod tests {
 
     #[test]
     fn test_parse_fixed() {
-        use crate::format::Fixed;
+        use crate::format::Fixed::*;
         use crate::format::Item::Literal;
 
         // fixed: month and weekday names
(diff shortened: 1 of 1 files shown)


src/format/parse.rs
```diff
@@ -564,28 +564,26 @@ mod tests {
     use crate::format::*;
     use crate::{DateTime, FixedOffset, TimeZone, Timelike, Utc};
 
-    #[test]
-    fn test_parse() {
-        use crate::format::InternalInternal::*;
-        use crate::format::Item::{Literal, Space};
-        use crate::format::Numeric::*;
+    fn parse_all(s: &str, items: &[Item]) -> ParseResult<Parsed> {
```
djc, line 567: Style nit: helper below the tests, please.

src/format/parse.rs
```diff
@@ -564,28 +564,26 @@ mod tests {
...
+        parse(&mut parsed, s, items.iter())?;
+        Ok(parsed)
+    }
 
-        fn parse_all(s: &str, items: &[Item]) -> ParseResult<Parsed> {
-            let mut parsed = Parsed::new();
-            parse(&mut parsed, s, items.iter())?;
-            Ok(parsed)
-        }
+    macro_rules! check {
```
djc, line 573: Unrelated to this particular change, but since you're cleaning this area up: does this need to be a macro?

djc (approved):
Nice!

src/format/parse.rs
pitdicker: I'll learn at some point :smile:

src/format/parse.rs
pitdicker: You thought: let's ask something easy? :laughing:
I still use a macro to created the `Parsed` structs to compare against, but the test now uses more regular rust code.

djc: I didn't ask you to change anything about it. 😋

PR #1171, opened 2023-07-05
chore: conversion to prost_types::Timestamp

mustakimali:
## Overview

We often need to convert `prost_types::Timestamp` into `DateTime<Utc>` and the other way around. So we end up defining extension traits for this in multiple places. Instead of making another crate for this (eg. `prost-types-ext`) I think it makes sense to add this here behind a feature flag.

A similar PR for `prost_type` crate (by another author) https://github.com/tokio-rs/prost/pull/414 was rejected.

## Usage
```toml
chrono = { version = "0.4", features = ["prost"] }
```

```rs
let now = Utc::now();
let timestamp = Timestamp::from(now.clone());
let now_converted = DateTime::try_from(timestamp).expect("prost_types --> chrono");

assert_eq!(now, now_converted);
```


mustakimali:
> I would consider prost to be downstream of chrono, so these impls should probably live there.

This attempt was rejected there before, unfortunately,
https://github.com/tokio-rs/prost/pull/414#issuecomment-937096372

djc:
I commented there: https://github.com/tokio-rs/prost/pull/414#issuecomment-1647601658. It definitely doesn't make sense for chrono to have prost in its public dependencies.

Issue #238, opened 2018-04-04
Right nanoseconds parsing


gukoff:
Consider `2018-01-01T20:00:00.123`.
`Utc.datetime_from_str` parses it as a datetime with 123 nanoseconds.
[Every Python package](https://nbsoftsolutions.com/blog/iso-8601-and-nanosecond-precision-across-languages) and [Go](https://play.golang.org/p/Dqm2yVlo8U9) parse it like 123000000 nanoseconds.

I haven't found an ISO specification for nanoseconds. But shouldn't chrono better align with libraries in other languages?

dekellum:
What format string are you passing in the call to `datetime_from_str`?  What version of chrono?

gukoff:
Format: `%Y-%m-%dT%H:%M:%S.%f`
Version: 0.4.1

dekellum:
I _think_ you want this instead:

~~~ diff
+++ Format: %Y-%m-%dT%H:%M:%S%.f
--- Format: %Y-%m-%dT%H:%M:%S.%f
~~~
(Swap order of final '.' and '%')

gukoff:
Looks bizarre. This fails to parse: "microsecond must be in 0..999999".
The error itself says it's microseconds, not nanoseconds.

Why `%.f`? What's the difference between `%f` and `%.f`?

Enet4:
@gukoff That is documented [here](https://docs.rs/chrono/0.4.1/chrono/format/strftime/index.html). `%.f` is left-aligned, and so achieves exactly what you originally intended.

quodlibetor:
I'm not sure what the docs _should_ look like for the `%f` format specifier, but the fact that it is right-aligned is probably too late to change and is also probably hard to document. PRs making this less of a footgun are welcome!

pitdicker:
Updated the documentation in https://github.com/chronotope/chrono/pull/1173.

PR #1173, opened 2023-07-06
Clarify nanosecond formatting specifier doc

pitdicker:
https://github.com/chronotope/chrono/issues/238 suggests the documentation here could use some work.
And a description like the current "The default `%f` is right-aligned and always zero-padded to 9 digits for the compatibility with glibc and others" is something I have to read a couple of times to make some sense of.

This is my attempt to simplify the documentation.


djc (approved):
👍

PR #1175, opened 2023-07-07
Deprecate `DateTime::{from_local, from_utc}`

pitdicker:
I propose to deprecate `DateTime::from_local` and `DateTime::from_utc` because they are unergonomic and are easy to misuse.

`DateTime::from_local` and `DateTime::from_utc` don't take a type implementing `TimeZone` as an argument, but an `Offset`.
So you first somehow need to get the correct offset for that datetime.

This also makes these methods unergonomic: they must be used with type annotations because we can't tell from the offset what the `TimeZone` type should be.

### Example of using `DateTime::from_local` correctly

```rust
let naive = NaiveDate::from_ymd_opt(2000, 1, 12).unwrap().and_hms_opt(2, 0, 0).unwrap();
// Get the offset for this local datetime (`FixedOffset`)
let offset = Local.offset_from_local_datetime(&naive).single().unwrap();
// Construct a `DateTime<Local>`
let datetime = DateTime::<Local>::from_local(naive, offset);
```

The methods `TimeZone::from_local_datetime` or `NaiveDateTime::and_local_timezone` are easier to use.

### Can be misused

These methods make it easy to create a `DateTime` that is invalid. Example:
```rust
use chrono::{DateTime, FixedOffset, Local, NaiveDate, TimeZone};
let naive = NaiveDate::from_ymd_opt(2000, 1, 12).unwrap().and_hms_opt(2, 0, 0).unwrap();

// I am pretty sure +18:00 is not your local offset
let offset = FixedOffset::east_opt(18 * 60 * 60).unwrap();
let datetime = DateTime::<Local>::from_local(naive, offset);

let datetime_correct = Local.from_local_datetime(&naive).unwrap();
assert_ne!(datetime, datetime_correct);
```

Now a `DateTime` where the offset does not match what the `TimeZone` would calculate as the offset is not absolutely wrong. A user could change the timezone of `Local` for example, and any `DateTime`'s created before that would have a different offset.

### Are there situations where it makes more sense to take an `Offset` argument?

Only few.

If a type implementing `OffSet` belongs to a `TimeZone` that has a fixed offset, `Utc` and `FixedOffset`, it will be just as performant to use the `TimeZone::from_local_datetime` and  `TimeZone::from_utc_datetime` methods to create a `DateTime`.

If the `TimeZone` has an offset that can vary, it is always wrong to reuse an `Offset` when the naive datetime changes. You should use the the `TimeZone::*` methods.

The two valid uses I can think of is when an implementer of `TimeZone` wants to override the provided methods `from_local_datetime` and `from_utc_datetime` to create a `DateTime` (which it should have no reason for), and to recreate a `DateTime` in something like a deserializer or from FFI.

For those cases I added a new `DateTime::from_naive_utc_and_offset` method.
It is new instead of keeping `from_utc` so that users of `from_utc` get a warning they are probably using a method for something it is not intended for.

### DIFF
--- a/src/datetime/tests.rs
+++ b/src/datetime/tests.rs
@@ -1481,3 +1481,16 @@ fn test_auto_conversion() {
     let utc_dt2: DateTime<Utc> = cdt_dt.into();
     assert_eq!(utc_dt, utc_dt2);
 }
+
+#[test]
+#[cfg(feature = "clock")]
+#[allow(deprecated)]
+fn test_test_deprecated_from_offset() {
+    let now = Local::now();
+    let naive = now.naive_local();
+    let utc = now.naive_utc();
+    let offset: FixedOffset = *now.offset();
+
+    assert_eq!(DateTime::<Local>::from_local(naive, offset), now);
+    assert_eq!(DateTime::<Local>::from_utc(utc, offset), now);
+}
--- a/src/datetime/tests.rs
+++ b/src/datetime/tests.rs
@@ -674,7 +674,7 @@ fn test_rfc3339_opts() {
     assert_eq!(dt.to_rfc3339_opts(Nanos, false), "2018-01-11T10:05:13.084660000+08:00");
     assert_eq!(dt.to_rfc3339_opts(AutoSi, false), "2018-01-11T10:05:13.084660+08:00");
 
-    let ut = DateTime::<Utc>::from_utc(dt.naive_utc(), Utc);
+    let ut = dt.naive_utc().and_utc();
     assert_eq!(ut.to_rfc3339_opts(Secs, false), "2018-01-11T02:05:13+00:00");
     assert_eq!(ut.to_rfc3339_opts(Secs, true), "2018-01-11T02:05:13Z");
     assert_eq!(ut.to_rfc3339_opts(Millis, false), "2018-01-11T02:05:13.084+00:00");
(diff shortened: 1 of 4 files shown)


src/datetime/tests.rs
```diff
@@ -469,7 +469,7 @@ fn test_rfc3339_opts() {
     assert_eq!(dt.to_rfc3339_opts(Nanos, false), "2018-01-11T10:05:13.084660000+08:00");
     assert_eq!(dt.to_rfc3339_opts(AutoSi, false), "2018-01-11T10:05:13.084660+08:00");
 
-    let ut = DateTime::<Utc>::from_utc(dt.naive_utc(), Utc);
+    let ut = dt.naive_utc().and_utc();
```
jtmoon79, line 472: The old behavior is deprecated but still supported.  Can this test case code be modified to test the deprecated behavior and the new behavior (and all other test case code in this PR)?

jtmoon79 (changes requested):

jtmoon79:
Overall, I like this, as I agree those functions are unergonomic. Marking them `deprecated` with recommendations is a good choice.

src/datetime/tests.rs
pitdicker: Hmm. We had three tests that happened to use it, but not in a meaning ful way.

Do we want to add new tests for deprecated methods?

jtmoon79: > Hmm. We had three tests that happened to use it, but not in a meaningful way.
> 
> Do we want to add new tests for deprecated methods?

I think it is worthwhile to still test formally deprecated code.

pitdicker: I added a basic test.

djc (approved):

PR #1177, opened 2023-07-08
Various fixed to allow testing with `no_std`

pitdicker:
`lib.rs` has an interesting line: `#![cfg_attr(not(any(feature = "std", test)), no_std)]`.
`std` is always available while testing, even with `--no-default-features`.

Some fixes are just a matter of correcting imports or using feature gates.

The more interesting ones are:
- Compare the parsing of a format string with the expected items while iterating in `test_strftime_items`.
- I am quite pleased with the alternative to using `assert_eq!(value.to_string(), "expected")` or `assert_eq!(format!("{}", value), "expected")`.
  We add a type that implements `fmt::Write` and knows `expected`. If it receives a byte that does not match `expected` it returns an error. This is used in the `assert_debug_eq` and `assert_display_eq` functions.
- We were missing feature gates in `src/oldtime.rs` that should have been added in https://github.com/chronotope/chrono/pull/1159. We should have a good look at the checks the CI runs at some point.

I have verified that all the tests we skip with `--no-default-features` really require allocations or another feature from `std`.


pitdicker:
cc @jtmoon79 I know no one who is as interested in tests :smile:

src/utils.rs
```diff
@@ -0,0 +1,54 @@
...
+#[cfg(test)]
+impl<'a> Write for WriteCompare<'a> {
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        if let Some(remainder) = self.remainder.strip_prefix(s) {
+            self.remainder = remainder;
+            Ok(())
+        } else {
+            #[cfg(feature = "std")]
+            eprintln!(
+                "formatting difference: `(left == right)`\n  left: `\"{}{}(...)\"`\n right: `\"{}\"`",
```
jtmoon79, line 28: I suggest using `{:?}` here so formatting characters like newline and tab are escaped. You can remove the explicit `"` in the string.

src/datetime/tests.rs
```diff
@@ -639,26 +632,10 @@ fn test_datetime_format_with_local() {
...
-    use std::thread;
-
-    // UTC is known to be `Send`.
-    let a = Utc::now();
-    thread::spawn(move || {
-        let _ = a;
-    })
-    .join()
-    .unwrap();
+fn test_datetime_is_send_and_copy() {
```
jtmoon79, line 635: hmm... this is beyond the edge of my rust knowledge.
Does this test `test_datetime_is_send_and_copy` invoke trait `Send` on a `Utc`? And trait `Copy`?
I can see function `_assert_send_copy` requires `T` to have those traits, but AFAICT, those traits are not exercised here.

src/utils.rs
```diff
@@ -0,0 +1,54 @@
...
+                self.expected
+            );
+            Err(fmt::Error)
+        }
+    }
+}
+
+#[cfg(test)]
+#[track_caller]
+pub(crate) fn assert_display_eq<D>(value: D, expected: &str)
```
jtmoon79, line 40: Add a docstring comment here describing why `assert_display_eq` was implemented. As a typical code reader, I would be confused why chrono goes through so much work to implement this.

```rust
/// Implement the rough equivalent of `assert_eq!` for values with the `Display` trait that
/// does not use `std` functions.
/// This allows chrono tests to ensure chrono behavior when `no_std` is in effect.
```

I don't know if my made-up explanation is accurate but you get the idea.

src/utils.rs
```diff
@@ -0,0 +1,54 @@
...
+where
+    D: fmt::Display,
+{
+    write!(&mut WriteCompare::new(expected), "{}", value).unwrap();
+}
+
+#[cfg(test)]
+#[track_caller]
+pub(crate) fn assert_debug_eq<D>(value: D, expected: &str)
```
jtmoon79, line 49: This also needs a docstring explanation as discussed above for `assert_display_eq`.

src/utils.rs
```diff
@@ -0,0 +1,54 @@
...
+
+#[cfg(test)]
+impl<'a> WriteCompare<'a> {
+    pub(crate) fn new(expected: &'a str) -> Self {
+        Self { expected, remainder: expected }
+    }
+}
+
+#[cfg(test)]
+impl<'a> Write for WriteCompare<'a> {
```
jtmoon79, line 20: This also needs a docstring explanation as discussed below for `assert_display_eq`.

jtmoon79 (changes requested):

src/datetime/tests.rs
pitdicker: Learned it here https://github.com/rust-lang/rust/pull/30872#issuecomment-173014586.

jtmoon79: I created a little example of the old code here that tested the `Copy` trait. Suprisingly, I couldn't get the underlying `clone()` to print anything (I'm pretty sure `clone()` is invoked by doing a copy) so I'm not sure what's happening.

```rust
fn main() {
    println!("Hello, world!");
}

#[derive(Debug, PartialEq, Eq)]
pub struct Object2 {}
impl Object2 {
    pub fn new() -> Self {
        Object2 {}
    }
}
impl Clone for Object2 {
    fn clone(&self) -> Object2 {
        eprintln!("Clone");
        *self
    }
}
impl Copy for Object2 {}

#[test]
fn test_copy() {
    eprintln!("test_copy start...");
    let a = Object2::new();
    let b = a;
    assert_eq!(a, b);
}
```

Run with `cargo test -v -- --nocapture`

I guess your change is probably okay... 🤔

pitdicker: It works the other way around. `Clone` may use a simple `memcpy`, but `Copy` is guaranteed to be just a copy andnot call any functions.

src/utils.rs
pitdicker: I didn't know that! 👍
The `"`'s are to imitate the format of the `assert_eq!` macro.

jtmoon79 (approved):
LGTM

Overridding various common `assert*` macros does add some mental overhead to tests. However, this is thorough about testing `no_std` which is good to do.

pitdicker:
If I look at this PR honestly, it is better to close it.

The test harness relies on the standard library, so we have that available anyway.
I still think it is cool to be able to run the tests without needing any allocations :smile:.
But it means needlessly restricting our test suite

I'll open an other Pr with the 'good parts'.

pitdicker:
> Based on this does this mean this crate is not no_std? @pitdicker

Chrono works with `no_std` just fine when the default `std` feature is not enabled.
In this PR I just got lost in some experiments around the test suite that didn't turn out to be useful.

Issue #899, opened 2022-11-26
You should have bumped to 0.5 when you made the change that caused Alex-PK/chrono-locale#5


ctrlcctrlv:
FYI. Patched by Alex-PK/chrono-locale#6 but @alex-pk hasn't merged anything since 2018 so maybe you want to review this.

djc:
Hmm, that's a tricky one. There's an `Internatl` variant that tries to make it clear that this is intended to be non-exhaustive, but clearly it's not entirely effective. Given that this hasn't been much of an issue, I don't think we're going to go back and fix the semver issue here now. We should start by making some types explicitly non-exhaustive now that we've bumped the MSRV on main.

Are you aware with the built-in localization support that chrono offers?

ctrlcctrlv:
Well, it broke one of my font editor's core libraries, `ipc.rlib`. From its name, you can assume that that library is included in every binary in the MFEK org pretty much.

Regarding `unstable-locales`, no, was not aware of it, and `ipc.rlib` far predates it. But I can look into changing to it when it's stabilized, my patch Alex-PK/chrono_locale#6 fixes the problem for me and the MFEK org can just maintain that library for as long as it takes to stabilize.

PR #1179, opened 2023-07-10
fix #332: stricter %Y

### LINKED ISSUE #332: %Y should return an error when presented with a non-4 digit year
Currently parsing a year like 19 with %Y will return 0019.

It should return an error because the year doesn't have 4 digits.

Users should use %y (lowercase) for 2 digit years and %Y for 4 digit years.

----------
Use case:

I hit this bug while porting a program from python, where it tries to parse the date in different formats in order (to clean up dirty data).

The input is: 30-09-19 (and we expect the format %Y-%m-%d to fail, and later on the format: %d-%m-%y to catch it) (It should September 30th 2019).

In the rust version I'll have to either check that the year is not too far in the future, or re-arrange the order that we check the different formats in.

kamadorueda:
The main idea behind the fix is refactoring a little bit the parser so that:
- (same as before) When %Y or %G is used (which allows for a sign), and there is a `+` or `-` preceding the number, it parses from 1 to _infinity_ digits
- (changed) When %Y is used, the min_width was increased from 1 to 4
- (same as before) When %G is used, the min_width was kept at 1 (but happy to change it to 4 as well for consistency if the maintainers agree)
- (same as before) The remaining numeric directives parse as before

Closes #332

### DIFF
--- a/src/format/parse.rs
+++ b/src/format/parse.rs
@@ -361,47 +361,71 @@ where
 
             Item::Numeric(ref spec, ref _pad) => {
                 use super::Numeric::*;
-                type Setter = fn(&mut Parsed, i64) -> ParseResult<()>;
 
-                let (width, signed, set): (usize, bool, Setter) = match *spec {
-                    Year => (4, true, Parsed::set_year),
-                    YearDiv100 => (2, false, Parsed::set_year_div_100),
-                    YearMod100 => (2, false, Parsed::set_year_mod_100),
-                    IsoYear => (4, true, Parsed::set_isoyear),
-                    IsoYearDiv100 => (2, false, Parsed::set_isoyear_div_100),
-                    IsoYearMod100 => (2, false, Parsed::set_isoyear_mod_100),
-                    Month => (2, false, Parsed::set_month),
-                    Day => (2, false, Parsed::set_day),
-                    WeekFromSun => (2, false, Parsed::set_week_from_sun),
-                    WeekFromMon => (2, false, Parsed::set_week_from_mon),
-                    IsoWeek => (2, false, Parsed::set_isoweek),
-                    NumDaysFromSun => (1, false, set_weekday_with_num_days_from_sunday),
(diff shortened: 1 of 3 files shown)


pitdicker:
I think instead of this it would be cool if parsing would respect padding. That is something I would really like to see.
Requiring zero digits to pad out the year would then be the `%0Y` specifier.
@kamadorueda I started an experiment to do this, do you want to take over? https://github.com/chronotope/chrono/compare/main...pitdicker:chrono:parse_padding

The tricky part is that the formatting specifiers have 4 padding variants https://docs.rs/chrono/latest/chrono/format/strftime/index.html:
- `%Y` default (which is currently forgiving when parsing, and in my opinion too big of a change to break)
- `%-Y` no padding (define this to not allow space or zero padding when parsing)
- `%_?` spaces as padding (define this is *allow* an arbitrary amount of unicode whitespace when parsing)
- `%0?` zero's as padding (define this to *require* spaces when parsing)

Put the `Pad` enum only has three types:
```rust
pub enum Pad {
    None,
    Zero,
    Space,
}
```

It will need at least a fourth variant: `Default` or `ZeroRequired`, or maybe even a fifth...
Currently padding is ignored when parsing, so we have to freedom to define these however we please.

But whatever we parse every formatting specifier to in `strftime` should match what we want for both formatting and parsing that specifier. And I think it is best if formatting doesn't change at all, and parsing without a padding specifier doesn't change.
So you'll have to think through each specifier.

---
Strictly speaking both this PR and adding variants to the `Pad` enum would be breaking changes. But I don't think the last one will cause breakage in practice. Well, only for crates that exhaustively match on our `Item` enums, which only the unmaintained `chrono-locale` seems to do. https://github.com/chronotope/chrono/issues/899

pitdicker:
In my opinion we would need to do something about parsing padding first, even if we would go with the direction of this PR. Otherwise we would break users that rely on the current behavior of `%Y`, without having any workaround available.

PR #1181, opened 2023-07-10
Allow missing seconds in `NaiveTime::from_str`

pitdicker:
Alternative to https://github.com/chronotope/chrono/pull/619.

### DIFF
--- a/src/naive/time/mod.rs
+++ b/src/naive/time/mod.rs
@@ -1324,26 +1324,36 @@ impl fmt::Display for NaiveTime {
 /// let t = NaiveTime::from_hms_nano_opt(23, 59, 59, 1_234_567_890).unwrap(); // leap second
 /// assert_eq!("23:59:60.23456789".parse::<NaiveTime>(), Ok(t));
 ///
+/// // Seconds are optional
+/// let t = NaiveTime::from_hms_opt(23, 56, 0).unwrap();
+/// assert_eq!("23:56".parse::<NaiveTime>(), Ok(t));
+///
 /// assert!("foo".parse::<NaiveTime>().is_err());
 /// ```
 impl str::FromStr for NaiveTime {
     type Err = ParseError;
 
     fn from_str(s: &str) -> ParseResult<NaiveTime> {
-        const ITEMS: &[Item<'static>] = &[
+        const HOUR_AND_MINUTE: &[Item<'static>] = &[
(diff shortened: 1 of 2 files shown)


src/naive/time/tests.rs
```diff
@@ -265,7 +266,7 @@ fn test_date_from_str() {
     assert!("".parse::<NaiveTime>().is_err());
     assert!("x".parse::<NaiveTime>().is_err());
     assert!("15".parse::<NaiveTime>().is_err());
-    assert!("15:8".parse::<NaiveTime>().is_err());
+    assert!("15:".parse::<NaiveTime>().is_err());
```
jtmoon79, line 269: Why just add test assert `"15:"` ? (I think assert `"15:8"` was doing something useful here)

pitdicker: Instead of `15:8` in the invalid cases we have the similar ` 9:08` in the valid cases.
`15:` is new, to check we don't allow a trailig colon.

jtmoon79: The `15:8` exercises combination double-digit hour, single-digit minute
whereas ` 9:08` is exercising single-digit hour leading space, double-digit minute leading zero.
I recommend keeping `15:8`.

jtmoon79 (approved):

src/naive/time/mod.rs
```diff
@@ -1324,26 +1324,36 @@ impl fmt::Display for NaiveTime {
...
             Item::Space(""),
         ];
+        const TRAILING_WHITESPACE: [Item<'static>; 1] = [Item::Space("")];
 
         let mut parsed = Parsed::new();
-        parse(&mut parsed, s, ITEMS.iter())?;
+        let s = parse_and_remainder(&mut parsed, s, HOUR_AND_MINUTE.iter())?;
+        // Seconds are optional, don't fail if parsing them doesn't succeed.
+        let s = parse_and_remainder(&mut parsed, s, SECOND_AND_NANOS.iter()).unwrap_or(s);
+        parse(&mut parsed, s, TRAILING_WHITESPACE.iter())?;
```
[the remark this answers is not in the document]
pitdicker: I don't know. The `FromStr` implementations of `NaiveTime`, `NaiveDate`, `NaiveDateTime` and `DateTime` all allow trailing whitespace.

djc (approved):

PR #1184, opened 2023-07-11
Add `StrftimeItems::{parse, parse_to_owned}` and more documentation

pitdicker:
Adds `StrftimeItems::parse` and `StrftimeItems::parse_to_owned`.

`parse` returns `Vec<Item<'_>>` and can return an error if the format string is invalid.

With the new formatting API the `Vec` hopefully gets more use, and then it may become difficult that is contains references to the format string. That is what the `OwnedLiteral` and `OwnedSpace` variants of `Item` are for, but we had no easy way to make those.
`parse_to_owned` returns `Vec<Item<'static>>`.

And I added some documentation and doctests.

### DIFF
--- a/src/format/strftime.rs
+++ b/src/format/strftime.rs
@@ -156,10 +156,17 @@ Notes:
    China Daylight Time.
 */
 
+#[cfg(feature = "alloc")]
+extern crate alloc;
+
 use super::{fixed, internal_fixed, num, num0, nums};
 #[cfg(feature = "unstable-locales")]
 use super::{locales, Locale};
 use super::{Fixed, InternalInternal, Item, Numeric, Pad};
+#[cfg(any(feature = "alloc", feature = "std"))]
+use super::{ParseError, BAD_FORMAT};
+#[cfg(feature = "alloc")]
+use alloc::vec::Vec;
 
 /// Parsing iterator for `strftime`-like format strings.
 ///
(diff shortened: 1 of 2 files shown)


jtmoon79 (approved):
Nice.

src/format/strftime.rs
```diff
@@ -279,6 +286,58 @@ impl<'a> StrftimeItems<'a> {
...
+    ///
+    /// // Parsing
+    /// let mut parsed = Parsed::new();
+    /// parse(&mut parsed, "11 Jul 2023  9.00", fmt_items.as_slice().iter())?;
+    /// let parsed_dt = parsed.to_naive_datetime_with_offset(0)?;
+    /// assert_eq!(parsed_dt, datetime);
+    /// # Ok::<(), chrono::ParseError>(())
+    /// ```
+    #[cfg(any(feature = "alloc", feature = "std"))]
+    pub fn parse(self) -> Result<Vec<Item<'a>>, ParseError> {
```
djc, line 334: Why do you want to add this to the public API?

src/format/strftime.rs
```diff
@@ -279,6 +286,58 @@ impl<'a> StrftimeItems<'a> {
...
+        let result: Vec<_> = self.collect();
+        if result.iter().any(|i| i == &Item::Error) {
```
djc, line 336: Instead of first collecting and then iterating again to find `Error` values, we can do both in one iteration by relying on the `FromIter` implementation for `Result`. Something like this:

```rust
self.into_iter().map(|item| match item == Item::Error {
    false => Ok(item),
    true => Err(BAD_FORMAT),
}).collect()
```

src/format/strftime.rs
```diff
@@ -338,6 +338,48 @@ impl<'a> StrftimeItems<'a> {
...
+    /// let datetime = NaiveDate::from_ymd_opt(2023, 7, 11).unwrap().and_hms_opt(9, 0, 0).unwrap();
+    ///
+    /// assert_eq!(
+    ///     datetime.format_with_items(fmt_items.as_slice().iter()).to_string(),
+    ///     "11 Jul 2023  9.00"
+    /// );
+    /// # Ok::<(), ParseError>(())
+    /// ```
+    #[cfg(any(feature = "alloc", feature = "std"))]
+    pub fn parse_to_owned(self) -> Result<Vec<Item<'static>>, ParseError> {
```
djc, line 376: Here, too: why should this be added to the public API? What use case does it address that is currently underserved?

src/format/strftime.rs
pitdicker: It comes from step 1 of in your post https://github.com/chronotope/chrono/issues/1127#issue-1741210542.
> 1. A fallible function/method that converts a format string to `Vec` of items (reusing the existing item type if we can)

So to make it easier to re-use a parsed format string with `format_with_items` and `format::parse` (I would like to move that function to a method on `Parsed`.

From the documentation example:
```rust
use chrono::format::{parse, Parsed, StrftimeItems};
use chrono::NaiveDate;

let fmt_items = StrftimeItems::new("%e %b %Y %k.%M").parse()?;
let datetime = NaiveDate::from_ymd_opt(2023, 7, 11).unwrap().and_hms_opt(9, 0, 0).unwrap();

// Formatting
assert_eq!(
    datetime.format_with_items(fmt_items.as_slice().iter()).to_string(),
    "11 Jul 2023  9.00"
);

// Parsing
let mut parsed = Parsed::new();
parse(&mut parsed, "11 Jul 2023  9.00", fmt_items.as_slice().iter())?;
let parsed_dt = parsed.to_naive_datetime_with_offset(0)?;
assert_eq!(parsed_dt, datetime);
# Ok::<(), chrono::ParseError>(())
```

pitdicker:
This is a smallish PR that improves the documentation of `StrftimeItems` and adds two methods to make it easier to re-use its result: `parse` and `parse_to_owned`.

@djc you already reviewed a lot the past few days, but how to you feel about getting this one over the finish line? Also because you already gave it one round of review.

src/format/strftime.rs
```diff
@@ -176,7 +192,30 @@ pub struct StrftimeItems<'a> {
 }
 
 impl<'a> StrftimeItems<'a> {
-    /// Creates a new parsing iterator from the `strftime`-like format string.
+    /// Creates a new parsing iterator from a `strftime`-like format string.
+    ///
+    /// # Errors
+    ///
+    /// While iterating [`Item::Error`] will be returned if the format string contains an invalid
+    /// or unrecognised formatting specifier.
```
djc, line 207: "unrecognised" -> "unrecognized".

src/format/strftime.rs
```diff
@@ -228,7 +228,51 @@ impl<'a> StrftimeItems<'a> {
...
+    ///
+    /// See the [`format::strftime` module](crate::format::strftime) for all supported formatting
+    /// specifiers.
+    ///
+    ///  [`DateTime::format_localized_with_items`]: crate::DateTime::format_localized_with_items
+    ///
+    /// # Errors
+    ///
+    /// While iterating [`Item::Error`] will be returned if the format string contains an invalid
+    /// or unrecognised formatting specifier.
```
djc, line 258: Here, too.

djc (approved):

PR #1185, opened 2023-07-14
Zonne docs timestamp

z-o-n-n-e:


pitdicker:
`DateTime::timestamp` works exactly as described: "Returns the number of non-leap seconds since January 1, 1970 0:00:00 UTC".
The number of seconds returned by `DateTime::timestamp` is independent of the timezone.

Maybe I can clear it up a bit:
A `DateTime` value always specifies an exact point in time, independent of a timezone. You could call it date and time in UTC. The timezone information of `DateTime<Tz>` allows it to know a second value: the local time in that timezone.

If you change the timezone, for example with `DateTime::with_timezone`, only the local time will change, the time in UTC stays the same.

`DateTime::timestamp` gives the difference between some `DateTime` in UTC and January 1, 1970 0:00:00 UTC.

pitdicker:
Is there an issue that I can help with?

z-o-n-n-e:
Thanks for the clarification, I understand

PR #1187, opened 2023-07-15
Remove `#![deny(dead_code)]`

pitdicker:
This would make development for more a bit easier.
And the CI fails on warnings anyway.


pitdicker:
Proof we still fail on CI if there is dead code: https://github.com/chronotope/chrono/actions/runs/5560750590/jobs/10157947719?pr=1187#step:7:192

pitdicker:
The CI fails because of the new clippy lints in Rust 1.71.

djc (approved):

PR #1188, opened 2023-07-16
add github job yamllint

jtmoon79:
Add github job "yamllint" that runs python yamllint against project .yml files.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -34,6 +34,19 @@ jobs:
       - run: taplo lint
       - run: taplo fmt --check --diff
 
+  yamllint:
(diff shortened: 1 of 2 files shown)


jtmoon79:
See https://github.com/chronotope/chrono/actions/runs/5570334997/jobs/10174304379?pr=1188

pitdicker:
In my opinion the toml lint, cff lint, and now yaml lint don't do much to increase the quality of chrono as a library compared to clippy, rustfmt or the doc lints. But there is also no harm in having them.

Can the toml (4s), cff (8s) and yaml (15s) lints be combined into one CI run?

Issue #577, opened 2021-07-25
Parsing negative NaiveDateTime timestamps with serde(with = "ts_milliseconds") fails


ousado:
Parsing negative NaiveDateTime timestamps with #[serde(with = "ts_milliseconds")] (and other implementations) fails.

```rust
use chrono::{NaiveDateTime}; // 0.4.19
use chrono::naive::serde::ts_milliseconds::deserialize as from_milli_ts;
use serde::Deserialize;

#[derive(Deserialize)]
struct S {
  #[serde(deserialize_with = "from_milli_ts")]
  time: NaiveDateTime
}
        
fn example() -> Result<S, serde_json::Error> {
  let my_s: S = serde_json::from_str(r#"{ "time": -1 }"#)?;
  Ok(my_s)
}
        

fn main() {
    let value = -1;
    let r0 = value / 1000;
    let r1 = (value % 1000);
    // this is the implementation in chrono::naive::serde::ts_milliseconds
    let r = NaiveDateTime::from_timestamp_opt(   
        value / 1000, 
        ((value % 1000) * 1_000_000) as u32);  // conversion into u32 fails for negative values 
    println!("{:?} {} {}", r, r0, r1);
    example().unwrap();
}
```

pitdicker:
Fixed in https://github.com/chronotope/chrono/pull/1194.

PR #1194, opened 2023-07-21
Fix deserialization of negative timestamps

pitdicker:
Fixes https://github.com/chronotope/chrono/issues/577.

Deserialization should use `div_euclid` en `rem_euclid` like in the `DateTime::from_timestamp_millis`.
Where available I used the existing methods `timestamp_millis_opt`, `NaiveDateTime::from_timestamp_micros` and `NaiveDateTime::from_timestamp_millis`.

For deserializing `u64` values it doesn't matter, we can just use regular div and mod.

I also fixed two cases where the `u64` input was cast to an `i64` without checking the range.


jtmoon79 (approved):

djc (approved):
Please switch the order of the commits around to avoid failing tests in our commit history (which is annoying if you want to use `git bisect`).

PR #1198, opened 2023-07-24
Format refactorings

pitdicker:
This are the 11 commits of https://github.com/chronotope/chrono/pull/1163 that are mostly just refactorings. Or at least don't add new functionality.

Does splitting them out so they can be reviewed and merge separately help?

### DIFF
--- a/src/format/locales.rs
+++ b/src/format/locales.rs
@@ -1,5 +1,9 @@
 use pure_rust_locales::{locale_match, Locale};
 
+pub(crate) const fn default_locale() -> Locale {
+    Locale::POSIX
+}
+
 pub(crate) const fn short_months(locale: Locale) -> &'static [&'static str] {
     locale_match!(locale => LC_TIME::ABMON)
 }
--- a/src/format/locales.rs
+++ b/src/format/locales.rs
@@ -1,41 +1,95 @@
-use pure_rust_locales::{locale_match, Locale};
+#[cfg(feature = "unstable-locales")]
+mod localized {
+    use pure_rust_locales::{locale_match, Locale};
 
-pub(crate) const fn default_locale() -> Locale {
-    Locale::POSIX
+    pub(crate) const fn default_locale() -> Locale {
+        Locale::POSIX
+    }
+
+    pub(crate) const fn short_months(locale: Locale) -> &'static [&'static str] {
+        locale_match!(locale => LC_TIME::ABMON)
+    }
+
+    pub(crate) const fn long_months(locale: Locale) -> &'static [&'static str] {
+        locale_match!(locale => LC_TIME::MON)
+    }
+
+    pub(crate) const fn short_weekdays(locale: Locale) -> &'static [&'static str] {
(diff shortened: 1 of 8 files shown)


src/format/locales.rs
```diff
@@ -1,41 +1,93 @@
...
-    locale_match!(locale => LC_TIME::T_FMT)
+    pub(crate) const fn t_fmt(locale: Locale) -> &'static str {
+        locale_match!(locale => LC_TIME::T_FMT)
+    }
+
+    pub(crate) const fn t_fmt_ampm(locale: Locale) -> &'static str {
+        locale_match!(locale => LC_TIME::T_FMT_AMPM)
+    }
 }
+#[cfg(feature = "unstable-locales")]
```
djc, line 45: Style nit: please add an empty line after inline modules, here and below.

src/format/locales.rs
```diff
@@ -1,41 +1,93 @@
...
+pub(crate) use localized::*;
+#[cfg(feature = "unstable-locales")]
+pub use pure_rust_locales::Locale;
```
djc, line 48: Maybe make this `pub use` inside the module? Seems cleaner.

djc (approved):
Looks great!

src/format/locales.rs
pitdicker: The module is only `pub(crate)`, while this re-export should be `pub`.

djc: Have you tried it? I would expect that you can have more-public items inside a less-public module, as long as there's a more-public export.

pitdicker: This was the error:
```
error[E0365]: `Locale` is only public within the crate, and cannot be re-exported outside
  --> src\format\mod.rs:61:9
   |
61 | pub use locales::Locale;
   |         ^^^^^^^^^^^^^^^ re-export of crate public `Locale`
   |
   = note: consider declaring type or module `Locale` with `pub`

error[E0365]: `Locale` is only public within the crate, and cannot be re-exported outside
   --> src\lib.rs:432:13
    |
432 |     pub use crate::Locale;
    |             ^^^^^^^^^^^^^ re-export of crate public `Locale`
    |
    = note: consider declaring type or module `Locale` with `pub`

error[E0365]: `Locale` is only public within the crate, and cannot be re-exported outside
   --> src\lib.rs:462:9
    |
462 | pub use format::Locale;
    |         ^^^^^^^^^^^^^^ re-export of crate public `Locale`
    |
    = note: consider declaring type or module `Locale` with `pub`

For more information about this error, try `rustc --explain E0365`.
error: could not compile `chrono` (lib) due to 3 previous errors
```

pitdicker:
Shall I merge this one first and then rebase https://github.com/chronotope/chrono/pull/1168 onto it? They are going to conflict.

src/format/locales.rs
djc: Okay, fair enough!

djc:
> Shall I merge this one first and then rebase #1168 onto it? They are going to conflict.

Whichever you prefer!

Issue #549, opened 2021-03-22
Set year/month/day all at once


zyxw59:
With the `Datelike` trait, there are methods to return a new date with a different year/month/day. However, these do not provide the ability to easily set more than one of those at once, since the intermediate value might not be valid.

e.g.

```rust
let a = NaiveDate::from_ymd(2019, 1, 29);
let b = NaiveDate::from_ymd(2020, 2, 29);

fn with_year_month_1<D: Datelike>(date: &D, year: i32, month: u32) -> Option<D> {
  date.with_year(year)?.with_month(month)?
}

fn with_year_month_2<D: Datelike><date: &D, year: i32, month: u32) -> Option<D> {
  date.with_month(month)?.with_year(year)?
}

with_year_month_1(a, 2020, 2); // -> Some(2020-02-29)
with_year_month_2(a, 2020, 2); // -> None (2019-02-29 is invalid)
with_year_month_1(b, 2019, 1); // -> None (2019-02-29 is invalid)
with_year_month_2(b, 2019, 1); // -> Some(2020-01-29)
```

It would be useful to have a method which applied multiple of these methods at once, without checking for validity of intermediate dates.

pitdicker:
If you want to replace two of the three parts of a date, is it not just as easy to create a new date?
Example:
```rust
use chrono::{NaiveDate, Datelike};

let a = NaiveDate::from_ymd(2019, 1, 29);
let b = NaiveDate::from_ymd(2020, 2, 29);

NaiveDate::from_ymd_opt(2020, 2, a.day()); // Some(2020-02-29)
NaiveDate::from_ymd_opt(2019, 1, b.day()); // Some(2019-01-29)
```

This is less tricky to get right.

jtmoon79:
At first glance, I'm hesitant to add more API "surface area" for something that is fairly trivial for the user to do.

However, the OP makes a good point that there is this non-obvious "gotcha" occurring here. (My two cents)

pitdicker:
Maybe we should add some documentation to `Datelike::with_month` or even to the `Datelike` trait itself on how to change multiple values at once.

jtmoon79:
> Maybe we should add some documentation to `Datelike::with_month` or even to the `Datelike` trait itself on how to change multiple values at once.

My preference is the OPs suggestion but documentation is better than nothing.

pitdicker:
Added some documentation in https://github.com/chronotope/chrono/pull/1199.

PR #1199, opened 2023-07-24
Add warning against combining multiple `Datelike::with_*`

pitdicker:
This documents the various error causes for the `Datelike::with_*` methods and adds a warning against combining multiple `Datelike_with_*`. See some discussion in https://github.com/chronotope/chrono/issues/549.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -1,6 +1,33 @@
 use crate::{IsoWeek, Weekday};
 
 /// The common set of methods for date component.
+///
+/// Methods such as [`year`], [`month`], [`day`] and [`weekday`] can be used to get basic
+/// information about the date.
+///
+/// The `with_*` methods can change the date.
+///
+/// # Warning
+///
+/// The `with_*` methods can be convenient to change a single component of a date, but they must be
+/// used with some care. Examples to watch out for:
+///
+/// - [`with_year`] changes the year component of a year-month-day value. Don't use this method if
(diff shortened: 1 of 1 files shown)


src/traits.rs
```diff
@@ -1,6 +1,32 @@
 use crate::{IsoWeek, Weekday};
 
 /// The common set of methods for date component.
+///
+/// Methods such as [`year`], [`month`], [`day`] and [`weekday`] can be used to get basic
+/// information about the date.
+///
+/// The `with-*` methods can change the date.
```
djc, line 8: Let's use `with_` consistently instead of sometimes using `with-`.

src/traits.rs
```diff
@@ -1,6 +1,32 @@
...
+///
+/// # Warning
+///
+/// The `with-*` methods can be convenient to change a single component of a date. But use them
```
djc, line 12: Sentences starting with "But" don't read that well... Maybe something like this?

"The `with-*` methods provide a convenient way to change a single component of a date, but they must be used with care. Examples to watch out for:"

I find bullet lists easier to review in code with a blank line before the first bullet (between the "Examples to watch out for:" and the first bullet).

djc (approved):
Nice!

src/traits.rs
pitdicker: Oops, that was a `-` by mistake.

djc (approved):

Issue #841, opened 2022-10-11
Panic on macOS 10.12


fralonra:
Hi, all.

The following minimum code panicked and I don't know why:

```rust
// main.rs

use chrono::prelude::*;

fn main() {
    let dt = Local::now();
}
```

panic message:
```
thread 'main' panicked at 'unable to select local time type: FindLocalTimeType("no local time type is available for the specified timestamp")'
```

from this line: https://github.com/chronotope/chrono/blob/45142764ab6e1f2957437aa187970b94d3aaeadc/src/offset/local/unix.rs#L132

**env**
- rustc    1.62.1 (e092d0b6b 2022-07-16)
- rustup  1.25.1 (bb60b1e89 2022-07-12)
- chrono 0.4.22
- macOS 10.12.1

djc:
So yes, I guess we might not "support" macOS 10.12 -- looks like that's been out of supported service for about 3 years at this point. If you can come up with a patch for it (and it doesn't add too much complexity) we'll consider fixing it. You can have a look at the file you referenced to get a feeling for what's going on.

esheppa:
It would also be useful to understand what is in the /etc/localtime file - a debug output of the parsed zone file should be sufficient to figure out what is going on.  @fralonra if you clone down the `chrono` source code you can use [`tz_info::TimeZone::from_file`](https://github.com/chronotope/chrono/blob/45142764ab6e1f2957437aa187970b94d3aaeadc/src/offset/local/tz_info/timezone.rs#L83) for this

fralonra:
Thanks for all!

I cloned the source code and found that the error was generated from:
https://github.com/chronotope/chrono/blob/d147e9a2152d1d961372bf5a6a261c644746efdf/src/offset/local/tz_info/timezone.rs#L174

I've also tried [`time`](https://github.com/time-rs/time) crate, and that crate worked as expected on the target machine.

esheppa:
Thanks @fralonra - could you show the `dbg!` output of the `self` type at the start of that function (`TimeZoneRef::find_local_time_type`)

djc:
@test9753 would you be able to figure out a PR for this? I probably won't have time to look into it (and can't easily test), would be happy to answer questions you have and/or review your code.

jfro:
looks like we've run into this with our users as well. we recently updating some our dependencies and that included our use of flexi_logger switching from `time` to `chrono`. We still support macOS 10.11+ as long as apple doesn't blow that up for us.

[...]

PR #1201, opened 2023-07-25
Tzif without TZ string

pitdicker:
[RFC 8536](https://datatracker.ietf.org/doc/html/rfc8536#section-3.3.1), which specifies the TZif format, says:
"Local time for timestamps on or after the last transition is specified by the TZ string in the footer (Section 3) if present and nonempty; otherwise, it is unspecified."

https://github.com/chronotope/chrono/issues/841 shows that older versions of macOS ship TZif files without a TZ string. They expect us to use the offset that the last transition switches to for dates after the last offset. I think that is quite reasonable.

Fixes https://github.com/chronotope/chrono/issues/841.

I have included the test file provided by @jfro as a test.

### DIFF
--- a/src/offset/local/tz_info/timezone.rs
+++ b/src/offset/local/tz_info/timezone.rs
@@ -756,6 +756,38 @@ mod tests {
     }
 
     #[test]
+    fn test_no_tz_string() -> Result<(), Error> {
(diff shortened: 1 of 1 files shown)


djc (approved):
Thanks for taking care of this!

Issue #1190, opened 2023-07-17
Switch to `windows-sys` on 0.4.x branch


pitdicker:
https://github.com/chronotope/chrono/pull/1003#issuecomment-1546648645 mentioned we can switch from `winapi` to `windows-sys` on the 0.4.x branch now that our MSRV exceeds 1.48.

But I don't know if there are other concerns.

pitdicker:
See https://github.com/chronotope/chrono/pull/712 for the original PR to switch to `windows-sys`.

djc:
I'm open to this. An alternative approach would be to use the windows-bindgen crate to generate more minimal API surface in an integration test instead of using windows-sys. The problem with windows-sys is that it adds lots of download size, so it's not a great fit for low-level crates like chrono.

pitdicker:
Documentation on `windows-bindgen`:
https://github.com/microsoft/windows-rs/blob/0.48.0/docs/readme.md#windows-bindgen
https://kennykerr.ca/rust-getting-started/standalone-code-generation.html

It does come with a note:

> **Warning:** Standalone code generation should only be used as a last resort for the most demanding scenarios. It is much simpler to use the [windows-sys](https://crates.io/crates/windows-sys) crate and let Cargo manage this dependency. This windows-sys crate provides raw bindings, is heavily tested and widely used, and should not meaningfully impact your build time.

I'll have a look at it.

djc:
I think chrono would qualify as a fairly demanding scenario, and I think the warning is maybe a bit overblown.

See more context in https://github.com/microsoft/windows-rs/issues/1720.

PR #1202, opened 2023-07-25
Switch to windows-bindgen

pitdicker:
As discussed in https://github.com/chronotope/chrono/issues/1190.
I have to admit `windows-bindgen` was surprisingly easy to set up.

Edit: The CI is not happy 😆. That is when you declare something as easy.


pitdicker:
No clue why the CI fails:
```
     Running tests\win_bindings.rs (target\debug\deps\win_bindings-961cd5ef61c9bddd.exe)

running 1 test
test gen_bindings ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.56s

   Doc-tests chrono
error: unexpected closing delimiter: `}`
  --> D:\a\chrono\chrono\src\win_bindings.rs:56:1
   |
53 |  clone(&self) -> Self {
   |                       - this delimiter might not be properly closed...
54 |         *self
55 |     }
   |     - ...as it matches this but it has different indentation
56 | }
   | ^ unexpected closing delimiter

error: doctest failed, to rerun pass `--doc`
```

Maybe cargo starts compiling the doctests while the file is not entirely written yet?

src/win_bindings.rs
```diff
@@ -0,0 +1,51 @@
+#![allow(unreachable_pub)]
+// Bindings generated by `windows-bindgen` 0.49.0
```
pitdicker, line 2: Windows-bindgen inserts an empty line below the comment, that rustfmt removes. Maybe also something to open an issue for?

pitdicker:
> No clue why the CI fails:

This was an interesting problem. I should have set `OpenOptions::truncate`. But why did the code look somewhat valid? And why couldn't I replicate the problem?

The CI checks out the repo with CR LF line endings, while I check it out with LF.
The test overwrites the file, but always writes LF line endings. So the new file is shorter, and not truncated because of the missing open option.

It just so happened that the ca. 50 line endings that the new file is shorter left a valid-looking fragment of code of 50 characters at the end.

tests/win_bindings.rs
```diff
@@ -0,0 +1,20 @@
...
+fn gen_bindings() -> Result<()> {
+    let apis = [
+        "Windows.Win32.System.Time.SystemTimeToFileTime",
+        "Windows.Win32.System.Time.SystemTimeToTzSpecificLocalTime",
+        "Windows.Win32.System.Time.TzSpecificLocalTimeToSystemTime",
+    ];
+
+    let bindings = windows_bindgen::standalone(&apis);
+    let mut file = OpenOptions::new().write(true).create(true).open("src/win_bindings.rs")?;
+    file.write("#![allow(unreachable_pub)]\n".as_bytes())?;
```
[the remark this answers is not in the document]
djc: I guess I'd go for `fs::write("src/win_bindings.rs", format!("#![allow(unreachable_pub)]\n{bindings}"))?;` instead, what do you think?

I'd like this test to fail if the existing contents of `src/win_bindings.rs` are different from the newly generated code.

Also, let's put this in `src/offset/local/win_bindings.rs`?

tests/win_bindings.rs
```diff
@@ -0,0 +1,20 @@
+#![cfg(all(windows, feature = "clock", feature = "std"))]
+
+use std::fs::OpenOptions;
```
djc, line 3: I prefer not to import `io::Result`, instead importing `io::self` and using `io::Result<()>`.

src/offset/local/windows.rs
```diff
@@ -13,11 +13,7 @@ use std::io::Error;
 use std::ptr;
 use std::result::Result;
 
-use winapi::shared::minwindef::FILETIME;
-use winapi::um::minwinbase::SYSTEMTIME;
-use winapi::um::timezoneapi::{
-    SystemTimeToFileTime, SystemTimeToTzSpecificLocalTime, TzSpecificLocalTimeToSystemTime,
-};
+use crate::win_bindings::*;
```
djc, line 16: I'd still prefer not to use `*` imports.

djc (commented):
This is looking great! Would make sense to file bugs for the blank line and `unreachable_pub` stuff. I'm guessing the latter might be tricky for them...

tests/win_bindings.rs
```diff
@@ -11,5 +11,15 @@ fn gen_bindings() -> io::Result<()> {
     ];
 
     let bindings = windows_bindgen::standalone(&apis);
-    fs::write("src/offset/local/win_bindings.rs", format!("#![allow(unreachable_pub)]\n{bindings}"))
+
+    let path = "src/offset/local/win_bindings.rs";
+    let existing = fs::read_to_string(path)?;
+    // Skip the first two lines when comparing:
+    // - We add `#![allow(unreachable_pub)]` as a first line.
+    // - Rustfmt removes the empty line after the comment.
```
djc, line 19: I think we should make sure the code is formatted before it's written, so we can compare directly (and don't need to format as a separate stuff when we need to commit a new version).

Can you run `rustfmt` against the generated file? Or use the `prettyplease` crate, if it yields the same formatting for this code?

pitdicker:
We may want to hold off on this. I could not find the repro for `windows-bindgen`. It seems to get replaced by `riddle` https://github.com/microsoft/windows-rs/pull/2544, https://crates.io/crates/riddle.

djc:
Ah, you meant repo instead of repro. It does look like riddle is the hot new thing...

MarijnS95:
Riddle is indeed the new thing, and works a treat for generating these bindings. You could have a simple script or (documented) minimal cmd-line to call `riddle` with one argument, it no longer needs to be hardcoded in a `#[test]` or `build.rs`.

MarijnS95:
This is how I employed it in one of the crates I maintain: https://github.com/Traverse-Research/amd-ext-d3d-rs/commit/d5ba12272795a7364de43646f0d1bc1edd65bd08

pitdicker:
> This is how I employed it in one of the crates I maintain: [Traverse-Research/amd-ext-d3d-rs@d5ba122](https://github.com/Traverse-Research/amd-ext-d3d-rs/commit/d5ba12272795a7364de43646f0d1bc1edd65bd08)

Thank you, this is very helpful!

pitdicker:
`windows-bindgen` is back as a library version of `riddle` https://github.com/microsoft/windows-rs/pull/2609.
It's API seems to still be unstable, and now requires an input file 🤷.

Cargo.toml
```diff
@@ -20,6 +20,7 @@ name = "chrono"
 default = ["clock", "std", "oldtime", "wasmbind"]
 alloc = []
 libc = []
+winapi = ["windows-targets"]
```
MarijnS95, line 23: Should we really call this feature "winapi" or is it for backwards compatibility?

src/offset/local/win_bindings.txt
```diff
@@ -0,0 +1,6 @@
+--out src/offset/local/win_bindings.rs
+--config flatten sys
```
MarijnS95, line 2: Have you considered the non-sys bindings that `windows-bindgen` can now generate, with a cleaner API and helpers? The overhead in `windows-core` and the extra functions/types that may need to be generated might not be worth it though, but it's good to have that documented.

tests/win_bindings.rs
```diff
@@ -0,0 +1,19 @@
...
+fn gen_bindings() {
+    let input = "src/offset/local/win_bindings.txt";
+    let output = "src/offset/local/win_bindings.rs";
+    let existing = fs::read_to_string(output).unwrap();
+
+    let log = bindgen(["--etc", input]).unwrap();
+    eprintln!("{}", log);
+
+    let new = fs::read_to_string(output).unwrap();
+    if !new.lines().eq(existing.lines()) {
```
djc, line 16: Why are we comparing this on a line-by-line basis? To get around the `\r\n` vs `\n` differences? If so, please add a comment to that effect.

djc (commented):
This is looking really nice!

djc:
@MarijnS95 thanks for the reviews!

Cargo.toml
pitdicker: It is for backwards compatibility. But I wonder if that is not overly cautious.

tests/win_bindings.rs
MarijnS95: Pretty sure it's that, all files should be checked in to the repo with `\n` but because the test runs on Windows (`windows-bindgen` runs just fine on Linux, fwiw!) the output will be `\r\n`.

Personally I'd just run this test in the CI with a helpful "how to regenerate" message and a `git status --porcelain` check for files that may not be checked in, as linked above:

https://github.com/Traverse-Research/amd-ext-d3d-rs/commit/d5ba12272795a7364de43646f0d1bc1edd65bd08#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR62

(Note that this line [changed a bit on main](https://github.com/Traverse-Research/amd-ext-d3d-rs/blob/2adfbb50cf08e7836d06e927c7e1b2b2226cc182/.github/workflows/ci.yml#L60) because we're back to `cargo r -p api_gen`.)

pitdicker: > (windows-bindgen runs just fine on Linux, fwiw!)

I just rebooted from Linux to Windows to work on this :rofl:

pitdicker: @MarijnS95 I have to admit I am pretty bad at all this :smile:. Is it easy for you to change this PR to something better?
And thanks for helping out already!

Cargo.toml
djc: Per [Hyrum's law](https://www.hyrumslaw.com/), let's err on the side of backwards compat unless there are strong reasons not to.

src/offset/local/win_bindings.txt
djc: Yes, that plan makes sense to me.

djc:
I think it would be great to just merge this in its current state, we can get it into 0.4.27.

PR #1205, opened 2023-07-27
Make methods on `NaiveDate` const

pitdicker:
Depends on the first couple of commits from https://github.com/chronotope/chrono/pull/1069.

### DIFF
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -25,6 +25,7 @@ use crate::format::{
 use crate::month::Months;
 use crate::naive::{IsoWeek, NaiveDateTime, NaiveTime};
 use crate::oldtime::Duration as OldDuration;
+use crate::{expect, try_opt};
 use crate::{Datelike, Weekday};
 
 use super::internals::{self, DateImpl, Mdf, Of, YearFlags};
@@ -255,8 +256,8 @@ impl NaiveDate {
     /// or if `year` is out of range for `NaiveDate`.
     #[deprecated(since = "0.4.23", note = "use `from_ymd_opt()` instead")]
     #[must_use]
-    pub fn from_ymd(year: i32, month: u32, day: u32) -> NaiveDate {
-        NaiveDate::from_ymd_opt(year, month, day).expect("invalid or out-of-range date")
+    pub const fn from_ymd(year: i32, month: u32, day: u32) -> NaiveDate {
+        expect!(NaiveDate::from_ymd_opt(year, month, day), "invalid or out-of-range date")
     }
 
     /// Makes a new `NaiveDate` from the [calendar date](#calendar-date)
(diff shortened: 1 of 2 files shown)


src/naive/date.rs
```diff
@@ -1389,6 +1412,30 @@ impl NaiveDate {
         NaiveWeek { date: *self, start }
     }
 
+    // This duplicates `Datelike::year()`, because trait methods can't be const yet.
+    #[inline]
+    const fn _year(&self) -> i32 {
```
djc, line 1417: What happens if we skip the `_` prefix here?

src/naive/date.rs
```diff
@@ -768,11 +768,11 @@ impl NaiveDate {
             return Some(self);
         }
 
-        i64::try_from(days.0).ok().and_then(|d| self.add_days(-d))
+        i32::try_from(days.0).ok().and_then(|d| self.add_days(-d))
     }
 
-    fn add_days(self, days: i64) -> Option<Self> {
-        let secs = days.checked_mul(86400)?; // 86400 seconds in one day
+    fn add_days(self, days: i32) -> Option<Self> {
```
djc: Why is this useful/necessary?

src/naive/date.rs
pitdicker: To my surprise nothing blows up :smile:

src/naive/date.rs
pitdicker: The type of `NaiveDate` is an `i32`, and we don't even use all the bits of it for storing the date.
So any value larger than `i32` doesn't make sense.

To me the question is more: why did it accept an `i64` in the first place? Because it was part of adding a 64-bit duration.

src/naive/date.rs
pitdicker: Rustdoc was not happy about it, but that is fixable.
https://github.com/chronotope/chrono/actions/runs/5694596475/job/15436073978?pr=1205

src/naive/date.rs
pitdicker: Or do you mean: why include these commits? It avoids `Duration::second`, which is not const in time 0.1.
And they were just a nice simplification in my opinion.

djc: No, I was asking the former question. Sounds good.

src/naive/date.rs
```diff
@@ -992,19 +1012,20 @@ impl NaiveDate {
...
-    pub fn and_hms_nano_opt(
+    pub const fn and_hms_nano_opt(
         &self,
         hour: u32,
         min: u32,
         sec: u32,
         nano: u32,
     ) -> Option<NaiveDateTime> {
-        NaiveTime::from_hms_nano_opt(hour, min, sec, nano).map(|time| self.and_time(time))
+        let time = try_opt!(NaiveTime::from_hms_nano_opt(hour, min, sec, nano));
```
djc, line 1022: Can we just `?` instead of using `try_opt!()`?

src/naive/date.rs
```diff
@@ -735,8 +746,11 @@ impl NaiveDate {
     /// );
     /// ```
     #[must_use]
-    pub fn checked_add_days(self, days: Days) -> Option<Self> {
-        i32::try_from(days.0).ok().and_then(|d| self.add_days(d))
+    pub const fn checked_add_days(self, days: Days) -> Option<Self> {
+        if days.0 > i32::MAX as u64 {
```
djc, line 750: Style nit: I prefer keeping the return expressions symmetric:

```rust
match days.0 <= i32::MAX as u64 {
    true => self.add_days(days.0 as i32),
    false => None,
}
```

src/naive/date.rs
pitdicker: `?` is not available in const, that is the reason to have this macro.

djc: Ok.

djc (approved):

PR #1207, opened 2023-07-27
Use `NonZeroI32` inside `NaiveDate`

pitdicker:
`NonZeroI32` is ugly to work with. But I really like what it brings: `NaiveDate`, `NaiveDateTime` and `DateTime` take up te same size when wrapped in `Option`.

We can easily make it a `NonZeroI32` because a `NaiveDate` consists of a year, ordinal and flags. The ordinal is always in the range `1..=366`, and the flags are also never `0`.

I considered if switching the `Of` type to `NonZero*32` would make things easier. But we do most calculations on that type, and it quickly gets unergonomic (I tried).

### DIFF
--- a/src/naive/date/mod.rs
+++ b/src/naive/date/mod.rs
@@ -16,6 +16,7 @@
 #[cfg(feature = "alloc")]
 use core::borrow::Borrow;
 use core::iter::FusedIterator;
+use core::num::NonZeroI32;
 use core::ops::{Add, AddAssign, Sub, SubAssign};
 use core::{fmt, str};
 
@@ -99,7 +100,7 @@ mod tests;
 )]
 #[cfg_attr(feature = "rkyv-validation", archive(check_bytes))]
 pub struct NaiveDate {
-    yof: i32, // (year << 13) | of
+    yof: NonZeroI32, // (year << 13) | of
 }
 
 /// The minimum possible `NaiveDate` (January 1, 262145 BCE).
@@ -140,7 +141,7 @@ impl NaiveDate {
         debug_assert!(YearFlags::from_year(year).0 == flags.0);
         let yof = (year << 13) | (ordinal << 4) as i32 | flags.0 as i32;
         match yof & OL_MASK <= MAX_OL {
-            true => Some(NaiveDate { yof }),
+            true => Some(NaiveDate::from_yof(yof)),
             false => None, // Does not exist: Ordinal 366 in a common year.
         }
     }
@@ -691,10 +692,10 @@ impl NaiveDate {
     pub(crate) const fn add_days(self, days: i32) -> Option<Self> {
         // fast path if the result is within the same year
         const ORDINAL_MASK: i32 = 0b1_1111_1111_0000;
-        if let Some(ordinal) = ((self.yof & ORDINAL_MASK) >> 4).checked_add(days) {
+        if let Some(ordinal) = ((self.yof() & ORDINAL_MASK) >> 4).checked_add(days) {
             if ordinal > 0 && ordinal <= 365 {
-                let year_and_flags = self.yof & !ORDINAL_MASK;
-                return Some(NaiveDate { yof: year_and_flags | (ordinal << 4) });
+                let year_and_flags = self.yof() & !ORDINAL_MASK;
+                return Some(NaiveDate::from_yof(year_and_flags | (ordinal << 4)));
             }
         }
         // do the full check
@@ -939,7 +940,7 @@ impl NaiveDate {
     /// Returns the packed month-day-flags.
     #[inline]
     const fn mdf(&self) -> Mdf {
-        Mdf::from_ol((self.yof & OL_MASK) >> 3, self.year_flags())
+        Mdf::from_ol((self.yof() & OL_MASK) >> 3, self.year_flags())
     }
 
     /// Makes a new `NaiveDate` with the packed month-day-flags changed.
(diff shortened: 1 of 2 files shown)


src/naive/internals.rs
```diff
@@ -15,14 +15,17 @@
...
+use crate::{expect, Weekday};
 use core::fmt;
+use core::num::NonZeroI32;
 
-/// The internal date representation: `year << 13 | Of`
-pub(super) type DateImpl = i32;
+/// The internal date representation: `year << 13 | Of`.
+/// Because the ordinal in `Of` is always in the range 1..=366, `DateImpl` will never be `0`.
+/// The year flags can also never be `0`.
+pub(super) type DateImpl = NonZeroI32;
```
djc, line 25: I'm open to this change, but I'd like another commit before this that changes `DateImpl` to be a newtype `struct` rather than a type alias. Ideally we can then add some methods (like `new()`) that avoids a bunch of this repetition.

pitdicker: Much better, I should have gone with this in the first place :+1:.

src/naive/internals.rs
```diff
@@ -18,11 +18,27 @@
...
 use core::fmt;
 
+#[cfg(feature = "rkyv")]
+use rkyv::{Archive, Deserialize, Serialize};
+
 /// The internal date representation: `year << 13 | Of`
-pub(super) type DateImpl = i32;
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, PartialOrd, Ord)]
+#[cfg_attr(feature = "rkyv", derive(Archive, Deserialize, Serialize))]
+#[allow(unreachable_pub)] // must be pub for rkyv
```
djc, line 28: `// must be pub for rkyv` seems surprising, why is this? There should really be a way for private types to be `Archive` etc.

src/naive/isoweek.rs
```diff
@@ -47,7 +47,7 @@ pub(super) fn iso_week_from_yof(year: i32, of: Of) -> IsoWeek {
         }
     };
     let flags = YearFlags::from_year(year);
-    IsoWeek { ywf: (year << 10) | (week << 4) as DateImpl | DateImpl::from(flags.0) }
+    IsoWeek { ywf: DateImpl::new((year << 10) | (week << 4) as i32 | flags.0 as i32) }
```
djc, line 50: Maybe we need `DateImpl::from_yw()` or `DateImpl::from_yof()`? Looks like we have a bunch of repetition of similar patterns.

src/naive/internals.rs
```diff
@@ -18,11 +18,27 @@
...
+pub struct DateImpl(i32);
+
+pub(super) const MAX_YEAR: DateImpl = DateImpl::new(i32::MAX >> 13);
+pub(super) const MIN_YEAR: DateImpl = DateImpl::new(i32::MIN >> 13);
+
+impl DateImpl {
+    pub(super) const fn new(val: i32) -> Self {
```
djc, line 35: Maybe also make these methods `#[inline]`.

pitdicker:
While working on https://github.com/chronotope/chrono/pull/1043 and now I feel we have abstractions in the wrong place here.

The modules `naive::date` and `naive::internals` don't have a clear separation. In several cases methods on `NaiveDate` directly work on the internal `i32`. And this is entirely reasonable to implement some operations efficiently.

@djc What do you think of the following?
- Split `naive::date` into multiple files like `naive::time` and `naive::datetime`.
- Merge `naive::internals` into `naive::date`
- Remove the `DateImpl` type (use `NaiveDate` instead)
- Maybe: move `Of` and `Mdf` into a separate modules to enforce a clean abstraction
- Simplify whatever this makes easier.

djc:
Splitting up `naive::date` similar to `datetime` and `time` is great, let's do that first in a separate PR?

I think your plan makes sense conceptually, I'd like to see how module sizes work out before we (mostly you 😄 ) work through all of the proposed changes. From a quick look, moving `NaiveDate` itself and its trait implementations into a separate module results in a module that's well over 2000 lines of code, so I don't think merging (currently at ~900 lines) into it would be a great idea.

pitdicker:
:+1: I'll give it a try and and we will know soon enough if it works well.

pitdicker:
> I'd like to see how module sizes work out before we (mostly you 😄 ) work through all of the proposed changes.

Sorry, I missed this sentence.

The `naive::date` module did not really increase in size in https://github.com/chronotope/chrono/pull/1212. Folding the functionality of `Of` into `NaiveDates` takes only a couple of lines extra. And in the end I moved the `Mdf`, `YearFlags` and `NaiveWeek` types to modules of their own.

pitdicker:
Rewritten on top of #1212. Still keeping as draft until that refactor is completed..

src/naive/date/mod.rs
```diff
@@ -99,7 +100,7 @@ mod tests;
 )]
 #[cfg_attr(feature = "rkyv-validation", archive(check_bytes))]
 pub struct NaiveDate {
-    yof: i32, // (year << 13) | of
+    yof: NonZeroI32, // (year << 13) | of
```
djc, line 103: Does this change the `NaiveDate` `MAX` or `MIN`? If not, why not?

pitdicker: This doesn't change that much
Our internal representation uses an ordinal that is always in the range 1..366, and year flags that are never 0. So the value was already never 0; we now tell the type system as much.

src/naive/date/mod.rs
```diff
@@ -1394,17 +1395,30 @@ impl NaiveDate {
...
+    /// Create a new `NaiveDate` from a raw year-ordinal-flags `i32`.
+    ///
+    /// In a valid value an ordinal is never `0`, and neither are the year flags. This method
+    /// doesn't do any validation; it only panics if the value is `0`.
+    const fn from_yof(yof: i32) -> NaiveDate {
+        NaiveDate { yof: expect!(NonZeroI32::new(yof), "invalid internal value") }
+    }
+
+    /// Get the raw year-ordinal-flags `i32`.
+    const fn yof(&self) -> i32 {
```
djc, line 1407: I think we should `#[inline]` this.

djc (approved):

pitdicker:
# :tada:

Thank you! That were a lot of PRs to go through.

PR #1210, opened 2023-07-30
Parse Zulu suffix with `%z`

### LINKED ISSUE #945: Can't parse "2000-01-01T01:23+01:00"?
The following runs fine:
```rust
use chrono::{DateTime};

fn main() {
    let parse_from_str = DateTime::parse_from_str;
    let ts = "2000-01-01T01:23+01:00";
    let fmt = "%Y-%m-%dT%H:%M%z";
    let parsed = parse_from_str(ts, fmt);
    println!("{:?}", parsed);
}
```
runs fine and returns
```
Ok(2000-01-01T01:23:00+01:00)
```

However, if I instead put
```rust
let ts = "2000-01-01T01:23Z";
```
then it fails:
```
Err(ParseError(Invalid))
```

By contrast, in Python, I can parse that string using this format:
```python
In [1]: import datetime as dt

In [2]: dt.datetime.strptime("2000-01-01T01:23Z", "%Y-%m-%dT%H:%M%z")
Out[2]: datetime.datetime(2000, 1, 1, 1, 23, 0, tzinfo=datetime.timezone.utc)
```

Should `parse_from_str` be able to parse the Zulu suffix `'Z'` with `'%z'`?

(note: this question was originally posted on [StackOverflow](https://stackoverflow.com/q/75209990/4451315), and it was suggested to report it here)

simon04:
Fixes #945.

### DIFF
--- a/src/datetime/tests.rs
+++ b/src/datetime/tests.rs
@@ -942,6 +942,10 @@ fn test_utc_datetime_from_str() {
         "2015-02-18T23:16:9.15UTC".parse::<DateTime<Utc>>(),
         Ok(ymdhms_milli_utc(2015, 2, 18, 23, 16, 9, 150))
     );
+    assert_eq!(
+        DateTime::<Utc>::parse_from_str("2015-02-18T23:16Z", "%Y-%m-%dT%H:%M%#z"),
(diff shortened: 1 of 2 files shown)


pitdicker:
Closing for the reason mentioned above. Thank you @simon04 for your work here!

PR #1212, opened 2023-08-02
Refactor `internals` module

pitdicker:
As discussed in https://github.com/chronotope/chrono/pull/1207.

The main motivation was removing the `Of` type, because it is an abstraction that is more in the way than helping.
I also moved things around to make the files smaller.
And the construction of `IsoWeek` and to/from `Mdf` is changed to not access private fields.

A lot of stuff here, but I am happy with the clean separation this brings.

### DIFF
--- a/src/naive/internals.rs
+++ b/src/naive/internals.rs
@@ -303,11 +303,6 @@ impl Of {
         }
     }
 
-    #[inline]
-    pub(super) const fn flags(&self) -> YearFlags {
-        YearFlags((self.0 & 0b1111) as u8)
-    }
-
     #[cfg_attr(feature = "cargo-clippy", allow(clippy::wrong_self_convention))]
     #[inline]
     pub(super) const fn to_mdf(&self) -> Mdf {
--- a/src/naive/internals.rs
+++ b/src/naive/internals.rs
@@ -346,13 +346,6 @@ impl Of {
         Mdf::from_of(*self)
     }
 
-    /// Returns an `Of` with the next day, or `None` if this is the last day of the year.
-    #[inline]
-    pub(super) const fn succ(&self) -> Option<Of> {
-        let of = Of(self.0 + (1 << 4));
-        of.validate()
-    }
-
     /// Returns an `Of` with the previous day, or `None` if this is the first day of the year.
     #[inline]
     pub(super) const fn pred(&self) -> Option<Of> {
@@ -866,10 +859,6 @@ mod tests {
         assert!(Of::from_mdf(Mdf::new(2, 29, regular_year).unwrap()).is_none());
         assert!(Of::from_mdf(Mdf::new(2, 29, leap_year).unwrap()).is_some());
         assert!(Of::from_mdf(Mdf::new(2, 28, regular_year).unwrap()).is_some());
-
-        assert!(Of::new(365, regular_year).unwrap().succ().is_none());
-        assert!(Of::new(365, leap_year).unwrap().succ().is_some());
-        assert!(Of::new(366, leap_year).unwrap().succ().is_none());
 
         assert!(Of::new(1, regular_year).unwrap().pred().is_none());
         assert!(Of::new(1, leap_year).unwrap().pred().is_none());
--- a/src/naive/internals.rs
+++ b/src/naive/internals.rs
@@ -402,6 +402,20 @@ impl Mdf {
     pub(super) const fn to_of(&self) -> Option<Of> {
         Of::from_mdf(*self)
     }
+
+    #[inline]
+    pub(super) const fn ordinal(&self) -> Option<u32> {
+        let mdl = self.0 >> 3;
+        match MDL_TO_OL[mdl as usize] {
+            XX => None,
+            v => Some((mdl - v as i32 as u32) >> 1),
+        }
+    }
+
+    #[inline]
+    pub(super) const fn year_flags(&self) -> YearFlags {
+        YearFlags((self.0 & 0b1111) as u8)
+    }
 }
 
 impl fmt::Debug for Mdf {
--- a/src/naive/internals.rs
+++ b/src/naive/internals.rs
@@ -15,7 +15,6 @@
 
 #![cfg_attr(feature = "__internal_bench", allow(missing_docs))]
 
-use crate::Weekday;
 use core::fmt;
 
 /// The year flags (aka the dominical letter).
@@ -315,12 +314,6 @@ impl Of {
         YearFlags((self.0 & 0b1111) as u8)
     }
 
-    #[inline]
-    pub(super) const fn weekday(&self) -> Weekday {
-        let Of(of) = *self;
-        weekday_from_u32_mod7((of >> 4) + (of & 0b111))
-    }
-
     #[cfg_attr(feature = "cargo-clippy", allow(clippy::wrong_self_convention))]
     #[inline]
     pub(super) const fn to_mdf(&self) -> Mdf {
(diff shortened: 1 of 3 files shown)


src/naive/internals.rs
```diff
@@ -318,12 +319,6 @@ impl Of {
         self.0 >> 4
     }
 
-    #[inline]
-    pub(super) const fn with_ordinal(&self, ordinal: u32) -> Option<Of> {
-        let of = Of((ordinal << 4) | (self.0 & 0b1111));
-        of.validate()
```
pitdicker, line 327: Previously there was no validation of an ordinal beyond this check. Because it happened after the shift left, some invalid values where silently accepted. I added a test.

src/naive/internals.rs
```diff
@@ -687,33 +691,6 @@ mod tests {
         }
     }
 
-    #[test]
-    fn test_mdf_to_of() {
```
pitdicker, line 829: I ported a couple of tests for cases that would otherwise not be covered to `naive::date::tests`.
But not entirely 1:1. For example I see no added value for `test_mdf_to_of` when there also was `test_mdf_to_of_to_mdf` wich does the same thing and adds one extra step.

src/naive/date/mod.rs
```diff
@@ -68,6 +68,8 @@ const MAX_OL: i32 = 366 << 4;
 // Allows for quick day of week calculation from the 1-based ordinal.
 const WEEKDAY_FLAGS_MASK: i32 = 0b111;
 
+const YEAR_FLAGS_MASK: i32 = LEAP_YEAR_MASK | WEEKDAY_FLAGS_MASK;
```
djc, line 71: Bottom.

src/naive/date/mod.rs
```diff
@@ -155,6 +157,11 @@ impl NaiveDate {
         ((self.yof & ORDINAL_MASK) >> 4) as u32
     }
 
+    #[inline]
+    const fn year_flags(&self) -> YearFlags {
```
djc, line 200: It's not a move, because it just appears?

src/naive/internals.rs
```diff
@@ -314,19 +313,6 @@ impl Of {
         }
     }
 
-    #[inline]
```
djc, line 319: Seems to be unrelated to the main change in this commit?

src/naive/isoweek.rs
```diff
@@ -25,32 +25,32 @@ pub struct IsoWeek {
...
-            // next year
-            (year + 1, 1)
+impl IsoWeek {
+    /// Returns the corresponding `IsoWeek` from the year and the `Of` internal value.
+    //
+    // internal use only. we don't expose the public constructor for `IsoWeek` for now,
+    // because the year range for the week date and the calendar date do not match and
+    // it is confusing to have a date that is out of range in one and not in another.
+    // currently we sidestep this issue by making `IsoWeek` fully dependent of `Datelike`.
+    pub(super) fn from_yof(year: i32, ordinal: u32, year_flags: YearFlags) -> Self {
```
djc, line 35: I think these are two changes:

* Make it a method on `IsoWeek`
* Change from consuming `Of` to consuming `ordinal` and `year_flags`

Please split into two commits accordingly.

src/naive/date/mod.rs
```diff
@@ -218,11 +218,8 @@ impl NaiveDate {
     /// Makes a new `NaiveDate` from year and packed month-day-flags.
     /// Does not check whether the flags are correct for the provided year.
     const fn from_mdf(year: i32, mdf: Mdf) -> Option<NaiveDate> {
-        if year < MIN_YEAR || year > MAX_YEAR {
-            return None; // Out-of-range
-        }
-        match mdf.to_of() {
-            Some(of) => Some(NaiveDate { yof: (year << 13) | (of.inner() as DateImpl) }),
+        match mdf.ordinal() {
```
djc, line 261: Please change one method per commit where feasible.

src/naive/year_flags.rs
```diff
@@ -0,0 +1,167 @@
+// This is a part of Chrono.
```
djc, line 1: This module ends up too small.

djc (commented):
I tried to review this but the commit history is still pretty messy IMO. I also feel like I suggested starting small in https://github.com/chronotope/chrono/pull/1207#issuecomment-1656735505 (just splitting it up). When you serve up these large PRs I have to set apart a larger chunk of (concentrated) time to try get through it, which makes it less likely that I'll be able to get to it, so I recommend splitting this into a few smaller PRs which I can more easily digest (and course correct if I think we should go in a different direction).

I think the end result is good here, I like the idea of unwinding `Of` and creating a cleaner separation between `Mdf` and `NaiveDate`, but to get there requires a bunch more smaller steps. If you're not already doing so, I suggest self-reviewing individual commits to check if they are easy to review (that is, don't require state that you as an author have paged in).

src/naive/isoweek.rs
pitdicker: I have to admit it were two commits at some point :smile:. I'll split them up.

src/naive/date/mod.rs
pitdicker: Yes, there was one more use of `Of::flags` that made me unable to remove its counterpart. I'll try to reorder the commits.

src/naive/date/mod.rs
```diff
@@ -53,6 +53,17 @@ mod tests;
 pub(super) const MAX_YEAR: i32 = i32::MAX >> 13;
 pub(super) const MIN_YEAR: i32 = i32::MIN >> 13;
 
+const ORDINAL_MASK: i32 = 0b1_1111_1111_0000;
```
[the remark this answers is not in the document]
pitdicker: In my mind it helps understanding the code `const`s like `MIN-*`, `MAX`_*, `*_MASK`, etc. are close to the type definition, together with the format description and the methods that work with the constants.

But no problem, I can move them down.

pitdicker:
How would you like this to be split up?
First a PR to split up the files/modules, or first one to remove the `Of` type?

And you consider some modules too small. Before I start changing, what would be you preference for the modules to end up with?

And I think it is more useful to keep the same file structure for the `naive::time`, `naive::date` and `naive::datetime` modules than to have small files. Combining `serde` and `rustc-serialize` modules is also not going to do much for the 0.5 branch.

djc:
> I don't always have an organized approach to programming. It is more like doing a couple of prototypes. Once I have something with the right direction I start picking it apart into multiple smaller commits and keep polishing them to help with review.

I will also often end up with a larger commit than what I'd like for review, and I think that's pretty more normal. I've definitely also gone through and rebuilt a branch from scratch (maybe by cherry-picking some changes from the previous iteration). I think the main "trick" here is to re-review your changes regularly (even before committing) to see if the changes need to be together or if there is an axis along which to split.

> How would you like this to be split up?
> First a PR to split up the files/modules, or first one to remove the `Of` type?

I don't really care which one goes first, as long as it's in smaller chunks.

> And you consider some modules too small. Before I start changing, what would be you preference for the modules to end up with?

My usual guideline is that 400-800 lines of code is a decent module size. This can vary a bit -- large modules are sometimes hard to prevent because that is really a natural privacy/API boundary, but I generally dislike having lots of < 200 line modules, switching between them takes more effort IMO.

> And I think it is more useful to keep the same file structure for the `naive::time`, `naive::date` and `naive::datetime` modules than to have small files. Combining `serde` and `rustc-serialize` modules is also not going to do much for the 0.5 branch.

IMO mirroring the structure for `time`/`date`/`datetime` isn't so important if the relevant code size is very different. What's the issue with making merges to main harder?

pitdicker:
> What's the issue with making merges to main harder?

There is no issue. Just that if we combine the `serde` and `rustc-serialize` modules into one file to have a reasonable size, it will still be a small file on main which only has `serde`.

pitdicker:
I have spend some time in `git rebase` for this PR. It now only has the changes to remove the `Of` type and `DateImpl`.

I did not find a really clean way to split the last three commits.

pitdicker:
Hmm. Is a -0.05% decrease in code coverage a reason to fail a CI run?

djc:
> Hmm. Is a -0.05% decrease in code coverage a reason to fail a CI run?

I will not hesitate a PR like that. If you want to adjust the CodeCov config for it, go ahead.

pitdicker (commented):
I have gone over all commits again (it was all pretty paged out). I renamed one commit, but did not see something that felt like it should change.

The commits are ordered so we can move one method at a time to `NaiveDate`. Which was tricky because of all the interdependencies.
The moved code usually does not look the same. I find it is best to read the new code with the bit representation in mind to see why it makes sense.

The more a commit gets to the end op this PR, the more helper methods or now useless tests open up and can be removed with it.

djc:
Okay, can you split this up into 3-4 PRs somehow and do one at a time?

pitdicker:
Sure, https://github.com/chronotope/chrono/pull/1428 has the first four.

djc:
Thanks for accommodating me!

pitdicker:
And I want to thank you a lot for all the review work today!

PR #1214, opened 2023-08-06
Optimize `NaiveDate::add_days` for small values

pitdicker:
I expect this method to be used much more often with small values than with values that end up an *x* number of years into the past or future. So I would like to add a fast path.

In the common case where the result of `add_checked_signed` and similar ends up within the same year we can avoid a couple of divisions and lookups and just change the ordinal.

Before:
```
bench_naivedate_add_signed
                        time:   [32.797 ns 34.745 ns 36.560 ns]
Found 13 outliers among 100 measurements (13.00%)
  13 (13.00%) high severe
```

After:
```
bench_naivedate_add_signed
                        time:   [12.886 ns 13.422 ns 13.855 ns]
                        change: [-65.054% -63.390% -61.655%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -782,9 +782,15 @@ impl NaiveDate {
 
     /// Add a duration of `i32` days to the date.
     pub(crate) const fn add_days(self, days: i32) -> Option<Self> {
-        if days == 0 {
-            return Some(self);
+        // fast path if the result is within the same year
+        const ORDINAL_MASK: i32 = 0b1_1111_1111_0000;
+        if let Some(ordinal) = ((self.ymdf & ORDINAL_MASK) >> 4).checked_add(days) {
+            if ordinal > 0 && ordinal <= 365 {
+                let year_and_flags = self.ymdf & !ORDINAL_MASK;
+                return Some(NaiveDate { ymdf: year_and_flags | (ordinal << 4) });
+            }
         }
+        // do the full check
         let year = self.year();
         let (mut year_div_400, year_mod_400) = div_mod_floor(year, 400);
         let cycle = internals::yo_to_cycle(year_mod_400 as u32, self.of().ordinal());
--- a/bench/benches/chrono.rs
+++ b/bench/benches/chrono.rs
@@ -6,7 +6,7 @@ use chrono::format::StrftimeItems;
 use chrono::prelude::*;
 #[cfg(feature = "unstable-locales")]
 use chrono::Locale;
-use chrono::{DateTime, FixedOffset, Local, Utc, __BenchYearFlags};
+use chrono::{DateTime, Duration, FixedOffset, Local, Utc, __BenchYearFlags};
 
 fn bench_datetime_parse_from_rfc2822(c: &mut Criterion) {
     c.bench_function("bench_datetime_parse_from_rfc2822", |b| {
@@ -195,6 +195,15 @@ fn bench_format_manual(c: &mut Criterion) {
         })
     });
 }
+
+fn bench_naivedate_add_signed(c: &mut Criterion) {
+    let date = NaiveDate::from_ymd_opt(2023, 7, 29).unwrap();
+    let extra = Duration::days(25);
+    c.bench_function("bench_naivedate_add_signed", |b| {
+        b.iter(|| black_box(date).checked_add_signed(extra).unwrap())
+    });
+}
+
 criterion_group!(
     benches,
     bench_datetime_parse_from_rfc2822,
@@ -210,6 +219,7 @@ criterion_group!(
     bench_format,
     bench_format_with_items,
     bench_format_manual,
+    bench_naivedate_add_signed,
 );
 
 #[cfg(feature = "unstable-locales")]


src/naive/date.rs
```diff
@@ -782,9 +782,15 @@ impl NaiveDate {
 
     /// Add a duration of `i32` days to the date.
     pub(crate) const fn add_days(self, days: i32) -> Option<Self> {
-        if days == 0 {
-            return Some(self);
+        // fast path if the result is within the same year
+        const ORDINAL_MASK: i32 = 0b1_1111_1111_0000;
+        if let Some(ordinal) = ((self.ymdf & ORDINAL_MASK) >> 4).checked_add(days) {
```
djc, line 787: Does this have good test coverage?

djc (approved):

src/naive/date.rs
pitdicker: I'll have a look later today

jtmoon79: Here is the [current test of `add_days`](https://github.com/chronotope/chrono/blob/v0.4.26/src/naive/date.rs#L2785).

This is such a crucial function that I would take the time to test it thoroughly.

I recommend adding a `#[cfg(test)]`-only code path within the `add_days` to turn on and turn off the fast path. There's probably many tricks to do this. One way is to have a magic file that is checked. This file can be created or deleted by the test case code.

```rust
        // fast path if the result is within the same year
        const ORDINAL_MASK: i32 = 0b1_1111_1111_0000;

        #[cfg(test)]
        let mut ADD_DAYS_ENABLE_FAST_PATH: bool = std::path::Path("/tmp/ADD_DAYS_ENABLE_FAST_PATH").is_file();
        #[not(cfg(test))]
        const ADD_DAYS_ENABLE_FAST_PATH: bool = false;

        if ADD_DAYS_ENABLE_FAST_PATH && let Some(ordinal) = ((self.ymdf & ORDINAL_MASK) >> 4).checked_add(days) {
           // ...
```

The test case code would create and later delete the file `"/tmp/ADD_DAYS_ENABLE_FAST_PATH"` as needed.

I recommend test cases for `days` values
* `0` to `367` that compares `add_days` when `ADD_DAYS_ENABLE_FAST_PATH` is `true` and `false`

(this "turn on / turn off" trick could also be used in the bench so future bench runners can reprove to themselves that the function is still worthwhile).

<br/>

I recommend moving that one crucial operation `let Some(ordinal) = ((self.ymdf & ORDINAL_MASK) >> 4).checked_add(days)` out to it's own private _testable_ function and then test that function for all values from `0` to `366` inclusive. Prove beyond a doubt that statement is behaving as expected. It's fairly straightforward to do and adds needed confidence for this change.

jtmoon79 (changes requested):

pitdicker:
> Does this have good test coverage?

> I'll have a look later today

That must have been one of the longer days in history...

With a small change in `add_days` to make the result invalid we have 37 failing tests:
```
failures:
    datetime::tests::test_add_sub_months
    datetime::tests::test_datetime_add_assign_local
    datetime::tests::test_datetime_add_days
    datetime::tests::test_datetime_add_months
    datetime::tests::test_datetime_date_and_time
    datetime::tests::test_datetime_fixed_offset
    datetime::tests::test_datetime_from_local
    datetime::tests::test_datetime_offset
    datetime::tests::test_datetime_rfc2822
    datetime::tests::test_datetime_rfc3339
    datetime::tests::test_datetime_sub_assign_local
    datetime::tests::test_datetime_sub_days
    datetime::tests::test_datetime_sub_months
    datetime::tests::test_from_system_time
    datetime::tests::test_parse_datetime_utc
    datetime::tests::test_rfc3339_opts
    datetime::tests::test_utc_datetime_from_str
    format::formatting::tests::test_datetime_format_alignment
    format::parse::tests::test_rfc2822
    format::parse::tests::test_rfc3339
    format::parsed::tests::issue_551
    format::parsed::tests::test_parsed_to_naive_date
    format::parsed::tests::test_parsed_to_naive_datetime_with_offset
    month::tests::test_month_enum_primitive_parse
    month::tests::test_month_enum_try_from
    naive::date::tests::test_date_add
    naive::date::tests::test_date_add_days
    naive::date::tests::test_date_parse_from_str
    naive::date::tests::test_date_sub_days
    naive::date::tests::test_naiveweek
    naive::date::tests::test_week_iterator_limit
    naive::date::tests::test_weeks_from
    naive::datetime::tests::test_and_local_timezone
    naive::datetime::tests::test_and_utc
    naive::datetime::tests::test_datetime_add
    offset::fixed::tests::test_date_extreme_offset
    offset::local::tests::verify_correct_offsets_distant_past
```

More seriously: we have reasonable but not exhaustive coverage for values within a year, for values that cross the year boundary and have to account for leap years, and for values than can cause an overflow in `naive::date::test_date_add_days`.

Adding more tests does not seem that useful to me.

jtmoon79:
>> I recommend test cases for days values
>>    0 to 367 inclusive that compares add_days when `ADD_DAYS_ENABLE_FAST_PATH` is true and false

> Adding more tests does not seem that useful to me.

I think the confidence this provides is worth the relatively small amount of work. (admittedly, I'm not doing the work! though if you want me to then I can implement my suggestion)

pitdicker:
@djc Did you want to give this a second look? Or was https://github.com/chronotope/chrono/pull/1214#discussion_r1285741322 only a question and is this approved?

djc (approved):

PR #1215, opened 2023-08-11
Add `.github/codecov.yml` and set threshold to 1%

pitdicker:
According to https://github.com/codecov/codecov-action/issues/554 we need to put a `codecov.yml` file with a threshold in the top-level directory, `dev/`, or `.github`.

I have set it to 5%, which will basically never fail.

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,5 @@
+coverage:
+  status:
+    project:
+      default:
+        threshold: 1%


pitdicker:
In https://github.com/chronotope/chrono/pull/1212 I added it as an extra commit to confirm this works.

djc:
I would say 5% is overly loose, let's start at 1% and see how it goes?

pitdicker:
:+1: changed to 1%.

djc (approved):

PR #1216, opened 2023-08-11
CI: fail on warnings in `features-check`

pitdicker:
Edit:
In https://github.com/chronotope/chrono/pull/1198 and https://github.com/chronotope/chrono/pull/1200 I did not get the feature gates completely right, and they would give warnings for some combination of features. But the CI did not complain.

This changes the `features-check` job to deny warnings.

I also removed the unused `libc` feature from the combinations `cargo hack` checks. This reduced the combinations from 384 to 156, which should shave ca. 5 minutes off our CI time.


djc (approved):
Great, thanks!

PR #1219, opened 2023-08-19
De-blob serde

### LINKED ISSUE #1218: serde_derive brings binary blob and breaks builds
serde_derive has chosen interesting approach
- https://github.com/serde-rs/serde/issues/2538

I've sent PR to make the binary blob opt-in approach so the top-level binary chooses the desired build / trust model instead

This breaks Nix builds etc so I would recommend restricting the SemVer scope on serde_derive for dev-dep and serde for dep

It was labeled as "Experimental" to embed the binary blob but it went straight to prod :(

```toml
[dependencies]
serde = { version = "1.0.99, <= 1.0.171", optional = true, default-features = false }
```

The above should also restrict the serde_derive on dev-dependencies to that range.

pinkforest:
Fixes #1218 

I've been advocating to get serde to do the binary blobs opt-in basis but at best there will be opt-out at top bin level only :(

This will restrict the max version to ensure only serde without binary blobs gets pulled

Crates `time` and `rust-analyzer` have already done this and `rand` etc. may follow as well soon.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ __doctest = []
 time = { version = "0.1.43", optional = true }
 num-traits = { version = "0.2", default-features = false }
 rustc-serialize = { version = "0.3.20", optional = true }
-serde = { version = "1.0.99", default-features = false, optional = true }
+serde = { version = "1.0.99, <= 1.0.171", default-features = false, optional = true }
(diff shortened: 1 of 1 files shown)


pitdicker:
Downstream users can choose to restrict the maximum version of a dependency, this does not have to be done in chrono. Let's leave the choice up to them.

PR #1220, opened 2023-08-21
Do not depend on `android-tzdata` if the `clock` feature is not enabled

AlexTMjugador:
All usages of `android-tzdata` types are within the `offset::local` module, [but such module is not built if the `clock` feature is not enabled](https://github.com/chronotope/chrono/blob/0be4e7c1d67614683f1e9345ed7d7df2643f6f0b/src/offset/mod.rs#L32-L33). Therefore, `android-tzdata` is effectively an unused dependency when the `clock` feature is disabled.

To avoid downloading unnecessary crates on build time, let's make the dependency on `android-tzdata` conditional on enabling the `clock` feature, as it was done with e.g. `iana-time-zone`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ default = ["clock", "std", "oldtime", "wasmbind"]
 alloc = []
 libc = []
 std = []
-clock = ["std", "winapi", "iana-time-zone"]
+clock = ["std", "winapi", "iana-time-zone", "android-tzdata"]
 oldtime = ["time"]
 wasmbind = ["wasm-bindgen", "js-sys"]
 unstable-locales = ["pure-rust-locales", "alloc"]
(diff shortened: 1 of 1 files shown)


pitdicker (approved):
For reference: `android-tzdata` was added in https://github.com/chronotope/chrono/pull/978.
I see no reason it shouldn't be optional.

djc (approved):

djc:
Thanks!

AlexTMjugador:
Thanks for the review and merging!

I submitted this PR to the `0.4.x` branch because it proposed backwards-compatible changes, but after glossing over the `main` branch, [I see that `android-tzdata` is not an optional dependency there](https://github.com/chronotope/chrono/blob/bc410bb054932518822eb393147aad939862c7a5/Cargo.toml#L47-L48). Would a PR backporting this change to `main` be appropriate?

pitdicker:
Every couple of weeks the 0.4.x branch is merged into main, no need to do extra work :smile:.

PR #1221, opened 2023-08-21
Fix some comments and panic messages.

umanwizard:


djc:
These changes should probably target the 0.4.x branch, please adjust your PR accordingly?

pitdicker:
The `Duration` → `TimeDelta` fixes only apply to main.
We have another commit that should be backported from main to the 0.4.x branch. Maybe accept this one, and backport both later?

Issue #854, opened 2022-10-26
Investigate integration with some components from icu4x project


esheppa:
It would need to first be determined which project sits lower in the stack, and it may be better for any integration to be done as a separate crate

My initial thoughts of what could be useful are:
* icu_calendar - this could potentially be leveraged to implement alternative calendar systems for `NaiveDate`
* icu_timezone
* icu_datetime - could be an alternative to pure-rust-locales which we currently use

djc:
An alternative to datetime would be interesting since pure-rust-locales seems to be more or less a one-off which was built for chrono but isn't getting much maintenance. Otherwise, it seems mostly undesirable to implement primitives in chrono based on other crates?

PR #1227, opened 2023-08-26
Explore using icu4

bim9262:
Relates to #854, but doesn't replace datetime code as a mentioned possible approach.

Uses icu4 for:

* short_months
* long_months
* short_weekdays
* long_weekdays
* am_pm
* decimal_point

Still uses pure_rust_locales for (I couldn't figure out how to get this from icu4 yet...):

* d_fmt
* d_t_fmt
* t_fmt
* t_fmt_ampm

### DIFF
--- a/src/datetime/tests.rs
+++ b/src/datetime/tests.rs
@@ -1477,8 +1477,8 @@ fn locale_decimal_point() {
     assert_eq!(dt.format_localized("%T%.6f", nl_NL).to_string(), "18:58:00,123456");
     assert_eq!(dt.format_localized("%T%.9f", nl_NL).to_string(), "18:58:00,123456780");
 
-    assert_eq!(dt.format_localized("%T%.f", ar_SY).to_string(), "18:58:00.123456780");
-    assert_eq!(dt.format_localized("%T%.3f", ar_SY).to_string(), "18:58:00.123");
-    assert_eq!(dt.format_localized("%T%.6f", ar_SY).to_string(), "18:58:00.123456");
-    assert_eq!(dt.format_localized("%T%.9f", ar_SY).to_string(), "18:58:00.123456780");
+    assert_eq!(dt.format_localized("%T%.f", ar_SY).to_string(), "18:58:00٫123456780");
+    assert_eq!(dt.format_localized("%T%.3f", ar_SY).to_string(), "18:58:00٫123");
+    assert_eq!(dt.format_localized("%T%.6f", ar_SY).to_string(), "18:58:00٫123456");
+    assert_eq!(dt.format_localized("%T%.9f", ar_SY).to_string(), "18:58:00٫123456780");
 }
--- a/src/format/formatting.rs
+++ b/src/format/formatting.rs
@@ -309,154 +309,155 @@ fn format_inner(
         Item::Fixed(ref spec) => {
             use self::Fixed::*;
 
-            let ret = match *spec {
-                ShortMonthName => date.map(|d| {
-                    w.write_str(short_months(locale)[d.month0() as usize])?;
-                    Ok(())
-                }),
-                LongMonthName => date.map(|d| {
-                    w.write_str(long_months(locale)[d.month0() as usize])?;
-                    Ok(())
-                }),
-                ShortWeekdayName => date.map(|d| {
-                    w.write_str(
-                        short_weekdays(locale)[d.weekday().num_days_from_sunday() as usize],
-                    )?;
-                    Ok(())
-                }),
-                LongWeekdayName => date.map(|d| {
-                    w.write_str(
-                        long_weekdays(locale)[d.weekday().num_days_from_sunday() as usize],
-                    )?;
-                    Ok(())
-                }),
-                LowerAmPm => time.map(|t| {
-                    let ampm = if t.hour12().0 { am_pm(locale)[1] } else { am_pm(locale)[0] };
-                    for c in ampm.chars().flat_map(|c| c.to_lowercase()) {
-                        w.write_char(c)?
-                    }
-                    Ok(())
-                }),
-                UpperAmPm => time.map(|t| {
-                    w.write_str(if t.hour12().0 { am_pm(locale)[1] } else { am_pm(locale)[0] })?;
-                    Ok(())
-                }),
-                Nanosecond => time.map(|t| {
-                    let nano = t.nanosecond() % 1_000_000_000;
-                    if nano == 0 {
+            let ret =
+                match *spec {
+                    ShortMonthName => date.map(|d| {
+                        w.write_str(short_months(locale)[d.month0() as usize].as_ref())?;
                         Ok(())
-                    } else {
-                        w.write_str(decimal_point(locale))?;
-                        if nano % 1_000_000 == 0 {
-                            write!(w, "{:03}", nano / 1_000_000)
-                        } else if nano % 1_000 == 0 {
-                            write!(w, "{:06}", nano / 1_000)
+                    }),
+                    LongMonthName => date.map(|d| {
+                        w.write_str(long_months(locale)[d.month0() as usize].as_ref())?;
+                        Ok(())
+                    }),
+                    ShortWeekdayName => date.map(|d| {
+                        w.write_str(
+                            short_weekdays(locale)[d.weekday().num_days_from_sunday() as usize]
+                                .as_ref(),
+                        )?;
+                        Ok(())
+                    }),
+                    LongWeekdayName => date.map(|d| {
+                        w.write_str(
+                            long_weekdays(locale)[d.weekday().num_days_from_sunday() as usize]
+                                .as_ref(),
+                        )?;
+                        Ok(())
+                    }),
+                    LowerAmPm => time.map(|t| {
+                        let ampm = am_pm(locale);
+                        let cur_ampm = &ampm[if t.hour12().0 { 1 } else { 0 }];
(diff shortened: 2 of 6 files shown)


djc:
What's the benefit here? How does it impact compile sizes and binary sizes?

bim9262:
My end goal is to support #281 and to improve localization overall - although I suppose it is subjective if the data from icu is better than the data in `pure_rust_locales`). In the tests you can see that there's some discrepencies in the localization data between the two.

For the compile size are you referring to the size of `libchrono.rlib`?

I made a little test wrapper that depends on chrono ( `chrono = { version = "0.4.26", features = ["unstable-locales"] }`) and compared the size with and without the icu addition.

without icu the bin is 4.6M
with icu the bin is 7.5M.

djc:
So a 3M binary size penalty seems like a big downside...

pitdicker:
> So a 3M binary size penalty seems like a big downside...

We already don't include timezone data but leave it in chrono-tz because it would add 2+ Mb to the binary size (and there is the problem it can get outdated...). I would argue that that date is a lot closed to the core functionality of chrono.

src/datetime/tests.rs
```diff
@@ -1477,8 +1477,8 @@ fn locale_decimal_point() {
     assert_eq!(dt.format_localized("%T%.6f", nl_NL).to_string(), "18:58:00,123456");
     assert_eq!(dt.format_localized("%T%.9f", nl_NL).to_string(), "18:58:00,123456780");
 
-    assert_eq!(dt.format_localized("%T%.f", ar_SY).to_string(), "18:58:00.123456780");
-    assert_eq!(dt.format_localized("%T%.3f", ar_SY).to_string(), "18:58:00.123");
-    assert_eq!(dt.format_localized("%T%.6f", ar_SY).to_string(), "18:58:00.123456");
-    assert_eq!(dt.format_localized("%T%.9f", ar_SY).to_string(), "18:58:00.123456780");
+    assert_eq!(dt.format_localized("%T%.f", ar_SY).to_string(), "18:58:00٫123456780");
```
pitdicker, line 1480: An Arabic comma (or what is it called) in roman numerals doesn't seem right.

src/format/formatting.rs
```diff
@@ -309,154 +309,155 @@ fn format_inner(
...
-                }),
-                UpperAmPm => time.map(|t| {
-                    w.write_str(if t.hour12().0 { am_pm(locale)[1] } else { am_pm(locale)[0] })?;
-                    Ok(())
-                }),
-                Nanosecond => time.map(|t| {
-                    let nano = t.nanosecond() % 1_000_000_000;
-                    if nano == 0 {
+            let ret =
+                match *spec {
```
pitdicker, line 313: Rustfmt is at it again :laughing:. https://github.com/chronotope/chrono/pull/1082#discussion_r1246536469

pitdicker:
Some links for a bit of context/history of our localization support:
- https://github.com/chronotope/chrono/pull/453
- https://github.com/chronotope/chrono/issues/475
- https://github.com/chronotope/chrono/pull/1152
- https://github.com/chronotope/chrono/pull/1165

Some of the goals where to not generate the data at build time, to not increase the binary size with locale data that is not used, and to work without allocating.

src/datetime/tests.rs
bim9262: Agreed, some locales have `outdigits` defined as part of `LC_CTYPE`, but `pure-rust-locales` doesn't support these. I did check though and there's nothing defined for `ar_SY`. I beleive that ICU has the numbers, though I'm still researching.

bim9262:
I'll take a look at generating rust files from the icu data that only has the data we'd need. For example there was a lot of extra information included about non-gregorian calendars. Producing data in a format like `pure-rust-locales` should allow  us to have `&'static str`s, and not need to alloc. This should also cut down on the size of the additional allocation information. :crossed_fingers:

bim9262:
Upon further testing by dumping out some mods from the ICU data I see results like:

```rust
#[allow(non_snake_case,non_camel_case_types,dead_code,unused_imports)]
pub mod aa_DJ {
    pub const SHORT_MONTHS: &[&str] = &["M01", "M02", "M03", "M04", "M05", "M06", "M07", "M08", "M09", "M10", "M11", "M12"];
    pub const LONG_MONTHS: &[&str] = &["M01", "M02", "M03", "M04", "M05", "M06", "M07", "M08", "M09", "M10", "M11", "M12"];
    pub const SHORT_WEEKDAYS: &[&str] = &["Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"];
    pub const LONG_WEEKDAYS: &[&str] = &["Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"];
    pub const AM_PM: &[&str] = &["AM", "PM"];
    pub const DECIMAL_POINT: &str = ".";
    pub const DIGITS: &[char] = &['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'];
}
```

Since there the data for months doesn't look good. I don't think that it would be a good to use icu for months (and probably not weekdays either)

The data for decimal and digits looks promising (the digits are displayed 9-0 on github, even though they are 0-9 in my terminal. Probably due to left to right vs right to left. As a side note, although arabic is read right to left, the numbers are constructed left to right):

```rust
pub mod ar_SY {
    [...]
    pub const DECIMAL_POINT: &str = "٫";
    pub const DIGITS: &[char] = &['٠', '١', '٢', '٣', '٤', '٥', '٦', '٧', '٨', '٩'];
}
```

Still need to investigate the ordinal numbers...

bim9262:
This shows a comparison between the glibc and icu for am_pm and the decimal point

https://gist.github.com/bim9262/920d29bd6155e207239bf11f42fe5da6

There are a lot of interesting cases where glibc has something more specific than am/pm or icu will be more specific. 

Some examples to check out: 

- hsb_DE (glibc blank, icu localized)
- ig_NG (glibc am/pm, icu localized)
- hif_FJ (glibc localized, icu am/pm)
- ks_IN (both localized, but very different)

I'd be tempted to take use glibc unless it's blank or am/pm then try icu as the fallback.

autarch:
> But I don't think our strftime-format string is going to be flexible enough to provide optimal localized formatting. And I don't think we have the experience to come up with enough extensions to get this right either. If one project can design a good format and API it would be ICU.

ICU has a set of patterns for date & time components. See https://cldr.unicode.org/translation/date-time/date-time-patterns

It covers a lot more cases than the `strftime` formatting specifiers. You can use the ICU _data_ with the `strftime` specifiers as well. You just have to make some decisions about the mapping. I did this with my Perl module [DateTime](https://metacpan.org/dist/DateTime). I could enumerate the choices I made there if there's interest in this.

autarch:
> Upon further testing by dumping out some mods from the ICU data I see results like:
> 
> ```rust
> #[allow(non_snake_case,non_camel_case_types,dead_code,unused_imports)]
> pub mod aa_DJ {
>     pub const SHORT_MONTHS: &[&str] = &["M01", "M02", "M03", "M04", "M05", "M06", "M07", "M08", "M09", "M10", "M11", "M12"];
>     pub const LONG_MONTHS: &[&str] = &["M01", "M02", "M03", "M04", "M05", "M06", "M07", "M08", "M09", "M10", "M11", "M12"];
>     pub const SHORT_WEEKDAYS: &[&str] = &["Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"];
>     pub const LONG_WEEKDAYS: &[&str] = &["Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"];
>     pub const AM_PM: &[&str] = &["AM", "PM"];
>     pub const DECIMAL_POINT: &str = ".";
>     pub const DIGITS: &[char] = &['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'];
> }
> ```
> 
> Since there the data for months doesn't look good. I don't think that it would be a good to use icu for months (and probably not weekdays either)

This sort of thing happens when the ICU project's data for a locale is incomplete. You can see the underlying data for that locale at https://github.com/unicode-org/cldr/blob/main/common/main/aa_DJ.xml. It's nearly empty. Compare it to a more complete locale at https://github.com/unicode-org/cldr/blob/main/common/main/am.xml.

There's a whole system for locale "inheritance". In the case of `aa-DJ` it should fall back to `aa`, which _does_ have data (at least in the CLDR GitHub repo). But the output above looks like it's falling back to the `root` locale. I'm not sure how you dumped that data. Or maybe it's using an older version of the data where the `aa` locale was empty.

autarch:
One more thing re: CLDR formatting. If you want to support this I think it's best to come up with new functions that accept these formats, like `format_cldr` or something along those lines.

pitdicker:
> So a 3M binary size penalty seems like a big downside...

> I don't think our strftime-format string is going to be flexible enough to provide optimal localized formatting. And I don't think we have the experience to come up with enough extensions to get this right either. If one project can design a good format and API it would be ICU.
> 
> Do we want to combine some localization data from ICU, while lacking their formatting API, with the formatting string format from libc?
> 
> Given that we use a strftime formatting string like libc, it makes sense to me that we also use the localization data from libc. Our localization would be 'as good as', and match, libc. That is already a useful thing for the Rust ecosystem. And when you want more advanced formatting, ICU4X would be the way to go

In my opinion chrono should focus mostly on being a good date-and-time library. ICU can be better at localization. I propose to close this PR and instead make it easy to convert to/from `icu::calendar::{Date, DateTime}`. Maybe behind an `icu` feature?

PR #1229, opened 2023-08-28
Implement operations for core::time::Duration on DateTime types

djc:
@pitdicker what do you think? Any suggestions for what tests we should add?

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -9,6 +9,7 @@ use core::borrow::Borrow;
 use core::cmp::Ordering;
 use core::fmt::Write;
 use core::ops::{Add, AddAssign, Sub, SubAssign};
+use core::time::Duration;
 use core::{fmt, hash, str};
 #[cfg(feature = "std")]
 use std::time::{SystemTime, UNIX_EPOCH};
@@ -1131,6 +1132,17 @@ impl<Tz: TimeZone> Add<OldDuration> for DateTime<Tz> {
     }
 }
 
+impl<Tz: TimeZone> Add<Duration> for DateTime<Tz> {
+    type Output = DateTime<Tz>;
+
+    #[inline]
+    fn add(self, rhs: Duration) -> DateTime<Tz> {
+        let rhs = OldDuration::from_std(rhs)
+            .expect("overflow converting from core::time::Duration to chrono::Duration");
+        self.checked_add_signed(rhs).expect("`DateTime + Duration` overflowed")
+    }
+}
+
 impl<Tz: TimeZone> AddAssign<OldDuration> for DateTime<Tz> {
     #[inline]
     fn add_assign(&mut self, rhs: OldDuration) {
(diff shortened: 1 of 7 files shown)


src/datetime/mod.rs
```diff
@@ -1197,6 +1229,18 @@ impl<Tz: TimeZone> SubAssign<OldDuration> for DateTime<Tz> {
...
 
+impl<Tz: TimeZone> SubAssign<Duration> for DateTime<Tz> {
+    #[inline]
+    fn sub_assign(&mut self, rhs: Duration) {
+        let rhs = OldDuration::from_std(rhs)
+            .expect("overflow converting from core::time::Duration to chrono::Duration");
+        let datetime =
+            self.datetime.checked_sub_signed(rhs).expect("`DateTime - Duration` overflowed");
+        let tz = self.timezone();
+        *self = tz.from_utc_datetime(&datetime)
```
pitdicker, line 1240: Why can't this be `*self -= rhs;`?

pitdicker (commented):
Looks like a nice addition to me.
And I don't think it would cause too much confusion between `chrono::Duration` and `core::time::duration` because these methods are usually used as operators.

Do you also want to implement them for `NaiveTime`?

> Any suggestions for what tests we should add?

I would think a basic test with `Duration::ZERO`, `Duration::from_secs(3600)`, `chrono::Duration::max_value().to_std()` and `Duration::MAX` (which should fail) for all operators is enough, because we already do the correctness tests for `chrono::Duration`.

djc:
> Any idea how to get the CI running for PRs from within the repository? Maybe we should remove this line? https://github.com/chronotope/chrono/blob/main/.github/workflows/test.yml#L5

Well, the workflow should also run on `pull_request:`, right, so I don't think removing that line would help.

pitdicker:
I understand about a year ago you and @esheppa put some thoughts into replacing `chrono::Duration` with `std::time::Duration`, or making `chrono::Duration` a wrapper around it.

Do you see adding these impls as an alternative to that?

djc:
So the plan I came up with is that in 0.5 we should only allow `core::time::Duration` as input, and have something like `enum TimeDelta { Forward(core::time::Duration), Backward(core::time::Duration) }` for API where we need a "signed" duration as an output type. This is then a step forward to allow `core::time::Duration` as input additively in places where we already allow `chrono::Duration`.

Another step might be to add a opaque type `struct WrappedDuration` that holds an `enum DurationKind { Chrono(..), Core(..) }` with `impl From<chrono::Duration> for WrappedDuration` and `impl From<core::time::Duration> for WrappedDuration` and then replace the `chrono::Duration` argument in various APIs with `impl Into<WrappedDuration>`...

djc:
(As for the motivation, I also have a use case at work where we're exercising this API and I would like to avoid using `chrono::Duration` everywhere.)

pitdicker (approved):

pitdicker:
`Duration` can hold much larger values than we can work with, and that anyone should need. If the goals is to avoid a potentially fallible conversing between an `u64` and `i64` I am not a fan. This is what signed integers are for instead of an enum. But unrelated to this PR :smile:

PR #1231, opened 2023-08-29
Use CODECOV_TOKEN when uploading coverage data

djc:


pitdicker (approved):
Nice! We will get code coverage reporting right eventually :smile:

Issue #1236, opened 2023-08-31
test `test_issue_866` fails (`and_hms_milli_opt(1, 2, 59, 1000)` returns `None`)


jtmoon79:
On Ubuntu 22 x86_64 using rust 1.72.0

```
$ cargo test --release --all-features
...
failures:

---- offset::local::tests::test_issue_866 stdout ----
thread 'offset::local::tests::test_issue_866' panicked at 'called `Option::unwrap()` on a `None` value', src/offset/local/mod.rs:266:77


failures:
    offset::local::tests::test_issue_866

test result: FAILED. 232 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.02s

error: test failed, to rerun pass `--lib`

```

Relates to Issue #866 and PR #1076 (I added [this test](https://github.com/chronotope/chrono/blob/v0.4.28/src/offset/local/mod.rs#L266)).

jtmoon79:
Why wasn't this caught by the CI ?

pitdicker:
Removed the test, as it is pretty much guaranteed to cause trouble.

Any method that returns `LocalResult` can return something that is not `LocalResult::Single`. Testing that it only returns `LocalResult::Single` by calling `LocalResult::unwrap` does not make all that much sense.

pitdicker:
> Why wasn't this [caught by the CI](https://github.com/chronotope/chrono/blob/v0.4.28/.github/workflows/test.yml#L19) ?

Probably because the CI tests only a couple of time zones, and one that has a transition at the time in this test isn't among them.

PR #1237, opened 2023-08-31
CI: run tests on `i686` and `wasm32-wasi`

pitdicker:
Fixes https://github.com/chronotope/chrono/issues/352.

### DIFF
--- a/src/datetime/tests.rs
+++ b/src/datetime/tests.rs
@@ -1185,8 +1185,11 @@ fn test_subsecond_part() {
     assert_eq!(1234567, datetime.timestamp_subsec_nanos());
 }
 
+// Some targets, such as `wasm32-wasi`, have a problematic definition of `SystemTime`, such as an
+// `i32` (year 2035 problem), or an `u64` (no values before `UNIX-EPOCH`).
+// See https://github.com/rust-lang/rust/issues/44394.
 #[test]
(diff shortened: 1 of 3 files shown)


src/datetime/tests.rs
```diff
@@ -1186,7 +1186,10 @@ fn test_subsecond_part() {
 }
 
 #[test]
-#[cfg(feature = "std")]
+#[cfg(all(feature = "std", not(target = "wasm32-wasi")))]
+// Some targets, such as `wasm-wasi` have a problematic definition of `SystemTime`, such as an
+// `i32` (year 2035 problem), or an `u64` (no values before `UNIX-EPOCH`).
+// See https://github.com/rust-lang/rust/issues/44394.
```
djc, line 1192: Nit: I'd like the comments to go before the attributes, keeping the attributes right next to the code.

djc (approved):

djc:
Does this also fix #1236?

pitdicker:
I have no clue why codecov thinks our coverage in `local/unix/timezone.rs` has massively changed from 85.69% to 66.67%, while at the same time saying the change in that file is -0.14%.

For now codecov seems mostly trouble, but maybe we get some use out of it eventually...

djc (approved):

djc:
LGTM, thanks!

Issue #1223, opened 2023-08-22
Resolve 2 security issues 


martinfrances107:
There are 2 problems that I see .. I just want to report them 

There is a difference between the latest  version published on crates.io  ( 0.4.26 ) 
and the main branch here..

For example when I run ```cargo audit``` on a upstream package ( theta-chart ) 

I get the first warning 

```
Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
Updating crates.io index
Scanning Cargo.lock for vulnerabilities (64 crate dependencies)

Crate:     time
Version:   0.1.45
Title:     Potential segfault in the time crate
Date:      2020-11-18
ID:        RUSTSEC-2020-0071
URL:       https://rustsec.org/advisories/RUSTSEC-2020-0071
Severity:  6.2 (medium)
Solution:  Upgrade to >=0.2.23
Dependency tree:
time 0.1.45
└── chrono 0.4.26
    └── theta-chart 0.0.5
```

This problem is known .... when I run cargo audit on the "main" branch 
I see that it has fixed  **but**  that is I see a different security advisory 

```
Crate:     atty
Version:   0.2.14
Warning:   unsound
Title:     Potential unaligned read
Date:      2021-07-04
ID:        RUSTSEC-2021-0145
URL:       https://rustsec.org/advisories/RUSTSEC-2021-0145
Dependency tree:
atty 0.2.14
└── criterion 0.4.0
    └── chrono 0.5.0-alpha.1

```

djc:
For RUSTSEC-2020-0071, see #602. For RUSTSEC-2021-0145, note that the dependency is guarded by an `__internal-bench` Cargo feature, so this should not have any effect on your code.

martinfrances107:
Can I ask a Question, I can't find the answer  after looking 

What does  the release cycle look like -- when will the already commit patch,  that fixes the first advisory,  be published on crates.io?

If there is any manual testing to be done please nudge me and I will jump on it.

Secondly, this PR  https://github.com/chronotope/chrono/pull/1224 fixes the second advisory

pitdicker:
Thank you for looking into this. Unfortunately both are not just simple fixes.

The time 0.1 dependency is only there because one of its types is part of our public API. We don't use the functionality that is related to the security advisory.
Issues related to this advisory are opened somewhat frequently... This is the latest with a workaround: https://github.com/chronotope/chrono/issues/1189.

[...]

PR #1224, opened 2023-08-22
Bump criterion from 0.4.0 to 0.5.1.

martinfrances107:
As part of #1223 . This PR fixes this security advisory 

```
Crate:     atty
Version:   0.2.14
Warning:   unsound
Title:     Potential unaligned read
Date:      2021-07-04
ID:        RUSTSEC-2021-0145
URL:       https://rustsec.org/advisories/RUSTSEC-2021-0145
Dependency tree:
atty 0.2.14
└── criterion 0.4.0
    └── chrono 0.5.0-alpha.1
```

PR #1243, opened 2023-09-01
Move benchmarks to a separate crate

pitdicker:
As reported in https://github.com/chronotope/chrono/issues/1223 a security advisory for `atty` shows up when we resolve the one for time 0.1.

By moving the benchmarks to a separate crate criterion and its dependencies no longer affect our MSRV or can cause advisories, and criterion can be updated to 5.1.

~I needed a `lib.rs` file to have a valid crate, and we had a tiny one in `ci/core-test`. It seemed like a nice opportunity to merge them.~ Edit: doesn't work.

Fixes https://github.com/chronotope/chrono/issues/1223, alternative to https://github.com/chronotope/chrono/pull/1224.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,11 +17,15 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - run: cargo fmt --check -- --color=always
       - run: cargo fmt --check --manifest-path fuzz/Cargo.toml
+      - run: cargo fmt --check --manifest-path bench/Cargo.toml
       - run: |
           cargo clippy --all-features --all-targets --color=always \
             -- -D warnings
       - run: |
           cargo clippy --manifest-path fuzz/Cargo.toml --color=always \
+            -- -D warnings
+      - run: |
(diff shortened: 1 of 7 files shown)


djc:
I think we should move criterion to be a dev-dependency instead and make it a proper benchmark within the main crate. Then we can restrict our MSRV test to only check `cargo check --lib --features foo,bar` to avoid making dev-dependencies part of our MSRV tests.

pitdicker:
I gave that another try this morning, but it doesn't seem to work: https://github.com/chronotope/chrono/pull/1245

djc (approved):
Okay, let's do it. Can you rename the top-level crate to `bench` so it's slightly less confusing to have `benches/benches`?

benches/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "benches"
+version = "0.1.0"
+edition = "2021"
+
+# Even as a `dev-dependency` Criterion and its dependencies can affect the MSRV of chrono.
```
jtmoon79, line 6: This is a really informative comment. 👍

PR #1244, opened 2023-09-01
Add a feature and method to override `now`

pitdicker:
Updated version of https://github.com/chronotope/chrono/pull/580. A comment in the related issue https://github.com/chronotope/chrono/issues/105#issuecomment-633260201 has collected 28 thumb ups in three years. This may be our most popular feature request.

The only method added is `Local::override_now(datetime: Option<DateTime<FixedOffset>>)`, and it affects `Local::now()` and `Utc::now()`. I made it a method on `Local` because it takes a datetime and an offset, while for `Utc` only a datetime would make sense.

The functionality is only available behind the `test-override` feature while running tests.

### DIFF
--- a/src/offset/local/mod.rs
+++ b/src/offset/local/mod.rs
@@ -171,7 +171,6 @@ impl Local {
     /// use chrono::{Local, FixedOffset, TimeZone, Datelike};
     /// fn is_today_leap_day() -> bool {
     ///     let today = Local::now().date_naive();
-    /// dbg!(today);
     ///     today.month() == 2 && today.day() == 29
     /// }
     ///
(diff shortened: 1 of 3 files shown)


pitdicker:
I just wondered: `cfg(test)` makes this method available in unit tests, but is it also available in integration tests? Yes.

src/offset/local/mod.rs
```diff
@@ -145,8 +147,38 @@ impl Local {
...
         Utc::now().with_timezone(&Local)
     }
+
+    /// Override the value that will be returned by `Local::now()` and `Utc::now()`, for use in
+    /// tests.
+    ///
+    /// This method is only available behind the `test-override` feature, only works within the
+    /// current thread, and only in `cfg(test)`.
+    ///
+    /// ```no_run
```
pitdicker, line 163: Codecov correctly says the lines of this doctest are not covered. But I can't help it :smile:.

mgeisler:
Just a comment from a random bystander: instead of injecting this on a global level, people can opt-in code using a test double such as https://crates.io/crates/mockall_double. It does not "fix" existing code that uses chrono today, but it gives a more clean way to control the test setup.

pitdicker:
I'm not going to push for this issue. Others can feel free to pick it up and/or advocate for it though.

Issue #69, opened 2016-04-14
Provide `days_in_month()`


the-kenny:
For some calculations it's useful to get the number of days in the current month. 

For example, I want to get the first and last day in a given month in my current project. The former is possible via `.with_day(1)`, but the latter isn't easily possible (or at least I don't see it)

lifthrasiir:
Duplicate with #29. See [my comment](https://github.com/lifthrasiir/rust-chrono/issues/29#issuecomment-84492746) for the workaround; `.pred()` is a key method. I believe this is not a good fit for Chrono, as i) the calendar system is fixed and ii) other similar libraries don't seem to have such features (especially for those where the calendar system is fixed).

vsimonian:
Don't mean to jump into an old issue here, but I was going to ask the same question.

@lifthrasiir I understand and respect where you're coming from, but I'm not so sure I agree that chrono wouldn't be a good place to have such a function.

First of all, other libraries across other languages do indeed provide such utilities. To list a few ones I know of:

C++ (Boost library):
```cpp
date dt(y, m, d);
date eom = dt.end_of_month_day();
gregorian_calendar::is_leap_year(y);
```

C# (built-in):
```csharp
DateTime.DaysInMonth(y, m);
DateTime.IsLeapYear(y);
```

Ruby on Rails (built-in):
```ruby
Time.days_in_month(m, y)
Time.days_in_year(y)
```

Python (built-in, calendar module):
```python
calendar.monthrange(y, m)
calendar.isleap(y)
```

Java (built-in):
```java
Calendar.getInstance().getActualMaximum(Calendar.DAY_OF_MONTH);
// No equivalent for isLeapYear(y)
```

JavaScript (Datejs library):
```javascript
Date.getDaysInMonth(y, m)
Date.isLeapYear(y)
```

JavaScript (moment library):
```javascript
moment('yyyy-mm', 'YYYY-MM').daysInMonth()
moment([y]).isLeapYear()
```

However, to play devil's advocate, some languages indeed don't provide such a utility, namely Go, Swift, and Objective-C.

I did see the workaround, though you mentioned it may not perform well. I just don't know enough (still absorbing Rust!) to know how "badly" it performs. Is it really that much of a performance/memory issue?

All said, I really think that `get_days_in_month(month, year)` and `is_leap_year(year)` would be a good addition. But I have an open mind if you have a better idea. Maybe I could write my first library with only 2 functions? :stuck_out_tongue_winking_eye:

[...]

Issue #29, opened 2015-03-21
Leap year and last day of month


gwenn:
Would you mind adding something like:
```rust
fn is_leap_year(year: i32) -> bool {
    return year % 4 == 0 && (year % 100 != 0 || year % 400 == 0)
}

fn last_day_of_month(year: i32, month: u32) -> u32 {
    match month {
        1 | 3 | 5 | 7 | 8 | 10 | 12 => 31,
        4 | 6 | 9 | 11 => 30,
        2 => if is_leap_year(year) { 29 } else { 28 },
        _ => panic!("invalid month: {}" , month),
    }
}
```
?
Thanks.

lifthrasiir:
While Chrono is inherently tied to the proleptic Gregorian calendar, I don't think such specific functions are not a good fit for general purpose library. If you only care about the correct answer (and not the performance), the following can be used:

```
fn is_leap_year(year: i32) -> bool {
    NaiveDate::from_ymd_opt(year, 2, 29).is_some()
}

fn last_day_of_month(year: i32, month: 32) -> u32 {
    NaiveDate::from_ymd_opt(year, month + 1, 1).unwrap_or(NaiveDate::from_ymd(year + 1, 1, 1)).pred().day()
}
```

I think JodaTime does have methods like `is_leap`, but mostly for supporting different calendar systems with varying leap definitions.

westy92:
You can add this functionality for your project with:
```rust
trait NaiveDateExt {
    fn days_in_month(&self) -> i32;
    fn days_in_year(&self) -> i32;
    fn is_leap_year(&self) -> bool;
}

impl NaiveDateExt for chrono::NaiveDate {
    fn days_in_month(&self) -> i32 {
        let month = self.month();
        match month {
            1 | 3 | 5 | 7 | 8 | 10 | 12 => 31,
            4 | 6 | 9 | 11 => 30,
            2 => if self.is_leap_year() { 29 } else { 28 },
            _ => panic!("Invalid month: {}" , month),
        }
    }

    fn days_in_year(&self) -> i32 {
        if self.is_leap_year() { 366 } else { 365 }
    }

    fn is_leap_year(&self) -> bool {
        let year = self.year();
        return year % 4 == 0 && (year % 100 != 0 || year % 400 == 0)
    }
}
```

sshine:
> *such specific functions are not a good fit for general purpose library*

I apologize for necroposting, but could you explain why?

pitdicker:
[...]

Issue #579, opened 2021-08-03
[Feature Request] Support parsing ISO 8601 Duration strings


praveentiru:
The crate supports parsing of date strings but, there is no equivalent support for parsing duration strings. This is request to support parsing of duration strings. Details here -> https://en.wikipedia.org/wiki/ISO_8601#Durations

I am open to work on this issue but, would need direction/guidance to do the same.

praveentiru:
I was poking around the code base. I can do parsing using Parsed using similar approach as implementation of `FromStr` trait for `DateTime`. But, I am not sure what would be the right way to expose this functionality.

Look forward to feedback.

meh:
Is anyone working on this? I kind of need it.

esheppa:
I like the struct you suggest, but this part from wiki concerns me:
> The smallest value used may also have a decimal fraction,[[37]](https://en.wikipedia.org/wiki/ISO_8601#cite_note-37) as in "P0.5Y" to indicate half a year.

A pragmatic option may be that we only support integer values everywhere but in the seconds part, however it would be interesting to understand whether that is overly restrictive.

djc:
Oh, good catch. I can't think of any strategy for figuring out how restrictive that would be. Starting with integer values only seems like a decent approach to get at least some support in.

xfix:
ISO 8601 has a lot of features that aren't particularly useful. Personally I rather would look into supporting [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339), which is a practical subset of it.

djc:
Huh, Wikipedia has this:

> IETF RFC 3339[[45]](https://en.wikipedia.org/wiki/ISO_8601#cite_note-45) defines a profile of ISO 8601 for use in [Internet protocols](https://en.wikipedia.org/wiki/Internet_protocol_suite) and [standards](https://en.wikipedia.org/wiki/Internet_standard). It explicitly excludes durations and dates before the [common era](https://en.wikipedia.org/wiki/Common_era).

So it looks like RFC 3339 does in fact include durations? I'm happy to bill our support as only for the RFC 3339 subset of durations. It does look like I missed a `weeks` field in my proposed definition above.

xfix:
Oh wait, it doesn't provide durations, I just got confused by CTRL+F in RFC 3339 providing a result in "ISO 8601 Collected ABNF" section.

djc:
[...]

PR #1247, opened 2023-09-02
Add `NaiveDate::diff_months_days`

pitdicker:
This is my oldest branch of chrono, that never got turned into a PR. From ca. 1½ year ago.
I kept putting it off, because arguing for the utility of something is not my hobby :smile:.

### `NaiveDate::leap_year()` 

Sometimes you just need to know whether the current date falls in a leap year.

A naive way to determine this is to do it manually:
```rust
let year = date.year();
let is_leap = year % 4 == 0 && (year % 100 != 0 || year % 400 == 0);
```
But that kind of defeats the purpose of using chrono, and it involves 3 divisions.

Better to do:
```rust
let is_leap = date.with_ordinal(366).is_some();
```
But I would say that is pretty creative API use, that few will come up with.

It also involves some work, while the information is right there as a flag in `NaiveDate`. I propose to expose it with `NaiveDate::leap_year()`.

See https://github.com/chronotope/chrono/issues/29 and https://github.com/chronotope/chrono/issues/69 for related issues.

### `NaiveDate::diff_months_days()`
For a project I needed to know the number of whole months and remaining days between two dates.
This turns out to be a bit difficult, until you know how to look at the problem.

The correct answer is different if you are counting down towards a date, compared to counting the number of months and days since an event.
In both cases there is a *reference date*, and when using this method as `reference.diff_months_days(other)` you get the right answer without having to think about it much.

`NaiveDate::diff_months_days` provides a human way to talk about larger durations: "there are 4 months and 13 days between those dates". The `days` value can then of course be rounded to something like weeks without much effort: "there are 4 months and 2 weeks between those dates". Or "4½ months".
Also the value of `months` can be turned into `years` and `months` without much effort.

But the difficult part, getting `months` and `days`, is what `NaiveDate::diff_months_days` is for.

### DIFF
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -1427,6 +1427,70 @@ impl NaiveDate {
         self.ymdf & (0b1000) == 0
     }
 
+    /// Difference between two dates expressed as whole months and the remaining days.
+    /// This method takes into account the correct number of days of each passing month.
+    ///
+    /// The order of the arguments is (`self`, `other`), not (first, last).
+    /// `self` can be seen as the reference date.
+    /// - When *counting down* from the reference date to `other`, the number of remaining days may
+    /// depend on the number of days in the first month.
+    /// - When *counting the elapsed period* from `self` to `other`, the number of remaining days
+    ///   may depend on the number of days in the last month before `other`.
+    ///
+    /// Returns `(months, days)`.
+    ///
+    /// ```
+    /// # use chrono::NaiveDate;
+    /// let ymd = |y, m, d| NaiveDate::from_ymd_opt(y, m, d).unwrap();
+    ///
+    /// // Elapsed months and days since a reference date:
+    /// assert_eq!(ymd(2022, 2, 27).diff_months_days(ymd(2022, 5, 26)), (2, 29));
+    /// assert_eq!(ymd(2022, 2, 27).diff_months_days(ymd(2022, 5, 27)), (3, 0));
+    /// assert_eq!(ymd(2022, 2, 27).diff_months_days(ymd(2022, 5, 28)), (3, 1));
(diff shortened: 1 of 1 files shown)


pitdicker:
Code coverage is not happy, https://github.com/chronotope/chrono/pull/1248 should help.

pitdicker:
https://github.com/chronotope/chrono/issues/416 seems to be a related request and can be solved with `diff_months_days` as a building block.

djc:
I think I'm open to something like `leap_year()`. My question is whether `leap_year()` is the most ergonomic formulation or if we should have `year_days() -> u16` (name TBD) instead. I'm much less convinced of the benefit of `diff_month_days()`.

pitdicker:
> My question is whether `leap_year()` is the most ergonomic formulation or if we should have` year_days() -> u16` (name TBD) instead.

I prefer just `leap_year()`. I have a couple of cases were I use something like this were I do not need to know the number of days in the year, but to map an ordinal or to know the number of days in februari.

> I'm much less convinced of the benefit of `diff_month_days()`.

I'll split it out `leap_year()` into another PR, and we can discuss the other a bit more here (I hope).

pitdicker:
@djc What would you say if instead of this PR i work on adding
```rust
struct CalendarPeriod {
    months: u32,
    days: u32,
    time: core::time::Duration,
}
```
from your comment in https://github.com/chronotope/chrono/issues/579#issuecomment-1283593033?

In other words create a type with full ISO 8601 duration support (I think I can do something reasonable with fractions also).
We can implement most of the methods of https://crates.io/crates/date_component on it.

I would like to start with adding the type and basic operations first, and add parsing and formatting later.

djc:
That sounds much more interesting to me, particularly because I would like to excise https://github.com/PoiScript/iso8601-duration from my dependency trees.

pitdicker:
Closing. I'll be back 😄

PR #1248, opened 2023-09-02
Include doctests for code coverage

pitdicker:
This requires nightly, see https://github.com/taiki-e/cargo-llvm-cov/issues/2.

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -12,12 +12,13 @@ jobs:
       CARGO_TERM_COLOR: always
     steps:
       - uses: actions/checkout@v3
-      - name: Install Rust
-        run: rustup update stable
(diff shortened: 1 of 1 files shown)


pitdicker:
Code coverage increased from 85.82% to  91.41% :rocket:.

djc (approved):
Nice, that's a pretty cheap win!

Do we know why it needs nightly? Is there some tracking issue I could follow?

pitdicker:
I am not sure what would be the right issue to track.
https://doc.rust-lang.org/rustc/instrument-coverage.html#including-doc-tests mentions code coverage relies on the `-Z unstable-options --persist-doctests`. So either that must be stabilized (tracking issue https://github.com/rust-lang/rust/issues/56925), or code coverage must switch to something different.
https://github.com/taiki-e/cargo-llvm-cov/issues/2 seems easier to follow for updates.

Issue #93, opened 2016-08-26
DateTime::parse_from_str not equivalent to TimeZone::datetime_from_str


Mark-Simulacrum:
DateTime::parse_from_str will parse `Mon Mar 14 08:32:45 2016 -0700`, but TimeZone::datetime_from_str will not.

Relevant IRC conversation starts here: https://botbot.me/mozilla/rust-tools/2016-08-26/?msg=71963735&page=1

pitdicker:
The irc logs are no longer available. But I think the issue is in how `TimeZone::datetime_from_str` works. From the documentation:
"If the to-be-parsed string includes an offset, it must match the offset of the `TimeZone`, otherwise an error will be returned."

So with `TimeZone::datetime_from_str` before parsing a string that includes an offset, you are expected to know the time zone. I.e. you have to know the offset before parsing a string as `FixedOffset`.

Also in the case of time zones like provided by `chrono_tz` parsing may fail even when you know the time zone, because the offset may have changed with a new revision of the time zone database.

I don't really see any use case where you would want to use `TimeZone::datetime_from_str` over `DateTime::parse_from_str`.

djc:
> I don't really see any use case where you would want to use `TimeZone::datetime_from_str` over `DateTime::parse_from_str`.

Should we deprecate the shit out of it?

pitdicker:
:rofl: With pleasure https://github.com/chronotope/chrono/pull/1251.

PR #1251, opened 2023-09-03
Deprecate `TimeZone::datetime_from_str`

pitdicker:
To quote https://github.com/chronotope/chrono/issues/93#issuecomment-1703926194:
> From the documentation:
> "If the to-be-parsed string includes an offset, it must match the offset of the `TimeZone`, otherwise an error will be returned."
>
> So with `TimeZone::datetime_from_str` before parsing a string that includes an offset, you are expected to know the time zone. I.e. you have to know the offset before parsing a string as `FixedOffset`.
>
> Also in the case of time zones like provided by `chrono_tz` parsing may fail even when you know the time zone, because the offset may have changed with a new revision of the time zone database.
>
> I don't really see any use case where you would want to use `TimeZone::datetime_from_str` over `DateTime::parse_from_str`.

There is one case where it is currently used in tests and documentation that makes some sense: `Utc.datetime_from_str`. This allows parsing a string without offset values directly into a `DateTime<Utc>`.

Users would have now have to write `NaiveDateTime::parse_from_str(/**/).and_utc()`. I don't think this is much worse, it may even show more clearly what is happening.

### DIFF
--- a/src/naive/datetime/tests.rs
+++ b/src/naive/datetime/tests.rs
@@ -339,6 +339,40 @@ fn test_datetime_parse_from_str() {
 }
 
 #[test]
+fn test_datetime_parse_from_str_with_spaces() {
+    let parse_from_str = NaiveDateTime::parse_from_str;
+    let dt = NaiveDate::from_ymd_opt(2013, 8, 9).unwrap().and_hms_opt(23, 54, 35).unwrap();
+    // with varying spaces - should succeed
+    assert_eq!(parse_from_str(" Aug 09 2013 23:54:35", " %b %d %Y %H:%M:%S"), Ok(dt));
+    assert_eq!(parse_from_str("Aug 09 2013 23:54:35 ", "%b %d %Y %H:%M:%S "), Ok(dt));
+    assert_eq!(parse_from_str(" Aug 09 2013  23:54:35 ", " %b %d %Y  %H:%M:%S "), Ok(dt));
+    assert_eq!(parse_from_str("  Aug 09 2013 23:54:35", "  %b %d %Y %H:%M:%S"), Ok(dt));
+    assert_eq!(parse_from_str("   Aug 09 2013 23:54:35", "   %b %d %Y %H:%M:%S"), Ok(dt));
+    assert_eq!(parse_from_str("\n\tAug 09 2013 23:54:35  ", "\n\t%b %d %Y %H:%M:%S  "), Ok(dt));
+    assert_eq!(parse_from_str("\tAug 09 2013 23:54:35\t", "\t%b %d %Y %H:%M:%S\t"), Ok(dt));
+    assert_eq!(parse_from_str("Aug  09 2013 23:54:35", "%b  %d %Y %H:%M:%S"), Ok(dt));
+    assert_eq!(parse_from_str("Aug    09 2013 23:54:35", "%b    %d %Y %H:%M:%S"), Ok(dt));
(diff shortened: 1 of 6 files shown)


src/naive/datetime/tests.rs
```diff
@@ -338,6 +338,40 @@ fn test_datetime_parse_from_str() {
     );
 }
 
+#[test]
```
djc, line 341: Please split this into two commits. As a reminder, moving code and changing it in the same code makes it harder to review the changes.

djc (approved):

src/naive/datetime/tests.rs
pitdicker: Yes, of course.

kryptan:
Top-level documentation still [recommends `datetime_from_str`](https://github.com/chronotope/chrono/blob/7c44bde46fc04b6bebc9c53cee739aa775489861/src/lib.rs#L264).

Also, deprecation warning points to `DateTime::parse_from_str` while `NaiveDateTime::parse_from_str(...).and_utc()` should actually be used as the OP suggests (parsing UTC time is probably the most common use case of parsing with known offset).

tmccombs:
I have another use case:

I want to parse a string in the _local_ time zone.

So I currently use:

```
Local.datetime_from_str(s, "%F %T")
```

But moving away from the deprecated function that becomes:

```rust

NaiveDateTime::parse_from_str(s, "%F %T").ok().and_then(|d| d.and_local_timezone(Local).single())
```

or possibly

```rust
NaiveDateTime::parse_from_str(s, "%F %T").ok()?.and_local_timezone(Local).single()
```

djc:
@tmccombs sorry you had a bad experience here.

1. Would you be able to send a PR to improve the deprecation note?
2. I agree this sucks, but I'm afraid for the 0.4 API there's not much we can do this due to semver compatibility constraints. For 0.5 improving this is definitely on the list, but that'll probably take some time to materialize.

pitdicker:
I did not consider the use case of parsing a string without offset info to a known timezone, such as `Local`. Seems like an argument to un-deprecate and instead improve the documentation. @djc What do you think?

tmccombs:
If doing that, then it may make sense to have it return an error if the format string includes an offset placeholder in the next major release.

djc:
> I did not consider the use case of parsing a string without offset info to a known timezone, such as `Local`. Seems like an argument to un-deprecate and instead improve the documentation. @djc What do you think?

Maybe a `Local`-specific API for this would be better than bringing back this semi-generalized API.

seanlinsley:
Here's an example resolving the deprecation for iCal timestamp parsing, going from [this](https://github.com/shouya/malakal/blob/40ac166f7ae7009baac0f3503335fd647d57cfd8/src/ical.rs#L105-L121) to this:

```rs
fn from_timestamp(s: &str, tzid: Option<&str>) -> Result<DateTime<Utc>> {
  use chrono_tz::Tz;
  use std::str::FromStr;

  if let Ok(time) = NaiveDateTime::parse_from_str(s, "%Y%m%dT%H%M%S") {
    if let Some(tz) = tzid.map(|tz| Tz::from_str(tz).ok()).unwrap_or(Some(Tz::UTC)) {
      if let Some(time) = time.and_local_timezone(tz).single() {
        return Ok(time.with_timezone(&Utc));
      }
    }
  }

  bail!("failed to parse timestamp {}", s)
}
```

The documentation or deprecation warning could definitely be improved, at least.

djc:
@seanlinsley feel free to submit a PR implementing any improvements you have in mind. Unfortunately I won't have time to work on that myself.

PR #1255, opened 2023-09-04
Clarify `Timelike::num_seconds_from_midnight` is a simple mapping

pitdicker:


src/traits.rs
```diff
@@ -316,6 +316,11 @@ pub trait Timelike: Sized {
     fn with_nanosecond(&self, nano: u32) -> Option<Self>;
 
     /// Returns the number of non-leap seconds past the last midnight.
+    ///
+    /// Every value in 00:00:00-23:59:59 maps to an integer in 0-86399.
+    ///
+    /// This method is not intended to provide the real number of seconds since midnight on a given
+    /// day. It does not take things into account such as a DST transition.
```
djc, line 323: The last sentence feels a bit awkward? I'd rewrite this last sentence as "It does not take things like DST transitions into account.", what do you think?

pitdicker: Yes, that is better :+1:

djc (approved):

PR #1259, opened 2023-09-06
Add note about the `TZ` environment variable and soundness

pitdicker:
I think we should say something about this in the docs because it has been a security issue that received a lot of attention, and affected a lot of dependencies.

Alternative to https://github.com/chronotope/chrono/pull/626.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -369,6 +369,29 @@
 //!
 //! The MSRV is explicitly tested in CI. It may be bumped in minor releases, but this is not done
 //! lightly.
+//!
+//! ## `TZ` environment variable and soundness
+//!
+//! All versions of chrono before 0.4.20 are affected by [RUSTSEC-2020-0159], like [time] was by
+//! [CVE-2020-26235] and [RUSTSEC-2020-0071]. We used the POSIX method `localtime_r` to get the
+//! local time, which may segfault if the `TZ` environment variable gets modified in a
+//! multi-threaded process.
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -377,6 +377,28 @@
...
+//! [CVE-2020-26235] and [RUSTSEC-2020-0071]. We used the POSIX method `localtime_r` to get the
+//! local time, which may segfault if the `TZ` environment variable gets modified in a
+//! multi-threaded process.
+//!
+//! Rust assumes the environment to be writable and uses locks to access it from multiple threads.
+//! So do some other programming languages and libraries, but there is no shared locking mechanism.
+//! More importantly POSIX declares modifying the environment in a multi-threaded process as unsafe
+//! (see [rust#27970] for more discussion).
+//!
+//! Since version 4.20 chrono no longer uses `localtime_r` but has a pure-rust solution. So when
```
djc, line 393: Maybe we can expand on this a little bit? `uses Rust code to query the operating system timezone (via iana-time-zone) and timezone data (DST transitions).`

pitdicker: I gave it a second try. A bit different though, because `iana-time-zone` is only used as fallback and the time zone data is much more than DST transitions.

pitdicker:
Now that similar information is included in `lib.rs` in https://github.com/chronotope/chrono/pull/1246 I merged it with that one.

PR #1262, opened 2023-09-06
Expand `From` impls of `js_sys::Date`

pitdicker:
While updating https://github.com/chronotope/chrono/pull/1040 I noticed our `From` impls of `js_sys::Date` are not as flexible as they could be.
The only allow converting from `DateTime<Utc>` instead of a generic `DateTime`.
And we had nothing to convert to `DateTime<Local>` and preserve the offset.

This does change type inference. Does that make it a problematic change?


djc:
Do we have a use case for this? Adding the extra impls could indeed be an inference hazard, and I generally dislike adding API without someone at least asking for it -- it's still more code to maintain.

Issue #13, opened 2014-11-30
Initialization macros


lifthrasiir:
`date!(2014-11-30)`, `time!(12:34)`, `datetime!(2014-11-30 12:34)`? The current initialization syntax is somewhat verbose. It _can_ be theoretically done without a procedural macro, provided that we keep away from the strict syntax of ISO 8601 (no `datetime!(2014-11-30T12:34)` etc.).

PR #1270, opened 2023-09-08
Add easy initialization macros

pitdicker:
I am very happy to have something like this!
```rust
// NaiveDate
date!(2023-09-08); // calendar date
date!(2023-251); // ordinal date

// NaiveTime
time!(7:03);
time!(7:03:00);
time!(23:59:60); // leap second

// NaiveDateTime
datetime!(2023-09-08 7:03);
datetime!(2023-09-08 7:03:25);

// DateTime<FixedOffset>
datetime!(2023-09-08 7:03:25+02:00);
datetime!(2023-09-08 7:03:25-02:00);
```

My macro skills are not great though. I do not know how to avoid the code duplication. Ans I can't make the `time!` macro support fractional seconds yet. Something TT muncher something.

Depends on https://github.com/chronotope/chrono/pull/1069 for creating a const `DateTime<FixedOffset>`.

Fixes https://github.com/chronotope/chrono/issues/13.


pitdicker:
Another open problem: clippy warns about integers starting with `0`, which is common in date and time values.

djc:
> Another open problem: clippy warns about integers starting with `0`, which is common in date and time values.

Can probably add clippy `#[allow()]` attributes in the macro output?

pitdicker:
Fixed the clippy warning with help from @Alexendoo in https://github.com/rust-lang/rust-clippy/issues/11472.

pitdicker:
- The macro to create a `DateTime<FixedOffset>` hits the same problem as in https://github.com/chronotope/chrono/pull/1286#issuecomment-1716972437: Rust 1.57 doesn't support making `DateTime<Tz>::from_naive_utc_and_offset` const.
  Added a workaround.
- We can't easily see whether a macro is `const`, so I added a test for it.
- Added an `offset!` macro. `offset!(+05:43)` is a lot easier to read then `FixedOffset::east_opt(20_580).unwrap()`.

pitdicker:
I have separated this into smaller commits.

And by creating a temporary `const` inside the macros we can guarantee the values are always checked at compile time.

pitdicker:
Updated to remove the rust 1.57 workaround.

These macro's can make tests nicer to read. Example commit: https://github.com/chronotope/chrono/commit/8e4092f97acdbced0bd57b16d15e368331ef5af4.
They give us parity with one of the selling points of the time crate.

src/macros.rs
```diff
@@ -0,0 +1,316 @@
+//! Macro's for easy initialization of date and time values.
+
+/// Create a `NaiveDate` with a statically known value.
```
jtmoon79, line 3: Link the `NaiveDate`.

Since the macro does not have a return type the docs will not automatically link to `NaiveDate`.

So change to a link

     /// Create a [`NaiveDate`] with ...

at the bottom, add the link target

    /// [`NaiveDate`]: crate::naive::NaiveDate

This nitpick goes for the rest of the PR.

If you use VS Code, it will highlight the link to turquoise blue when it is correctly set.

Screenshot 2024-01-30 102402

src/macros.rs
```diff
@@ -0,0 +1,316 @@
...
+                    Some(t) => t,
+                    None => panic!("invalid offset"),
+                };
+            OFFSET
+        }
+    }};
+    (-$h:literal:$m:literal) => {{
+        #[allow(clippy::zero_prefixed_literal)]
+        {
+            assert!($h < 24 || $m < 60, "invalid offset");
```
jtmoon79, line 253: I recommend splitting these out to be more specific what value is wrong. Also, reprint the wrong value.  This helps the user quickly and easily identify the culprit.

     assert!($h < 24, "invalid hour offset {}", $h);
     assert!($m < 60, "invalid minute offset {}", $m);

Apply this comment to rest of the PR.

src/macros.rs
```diff
@@ -0,0 +1,316 @@
...
+/// assert_eq!(date!(2023-09-08), date!(2023-251));
+/// ```
+#[macro_export]
+macro_rules! date {
+    ($y:literal-$m:literal-$d:literal) => {{
+        #[allow(clippy::zero_prefixed_literal)]
+        {
+            const DATE: $crate::NaiveDate = match $crate::NaiveDate::from_ymd_opt($y, $m, $d) {
+                Some(d) => d,
+                None => panic!("invalid calendar date"),
```
jtmoon79, line 25: I prefer to repeat back to the user the invalid values.  More specific clues in error messages are always nice.

     panic!("invalid calendar date from values {:?}, {:?}, {:?}", $y, $m, $d)

jtmoon79 (commented):
These are very handy!

I have some suggestions to improve user-friendliness.

src/macros.rs
pitdicker: Good comments, thanks @jtmoon79!
I'm not sure if we can use a format string for panics in const context, let me try.

pitdicker: Because these macro's are evaluated at compile time the message is already quite nice in my opinion:
```
error[E0080]: evaluation of constant value failed
  --> src\macros.rs:52:20
   |
52 |         assert_eq!(date!(2023-04-31), NaiveDate::from_ymd_opt(2023, 4, 30).unwrap());
   |                    ^^^^^^^^^^^^^^^^^ the evaluated program panicked at 'invalid calendar date', src\macros.rs:52:20
   |
```

jtmoon79: > Because these macro's are evaluated at compile time the message is already quite nice in my opinion:
> ```
> error[E0080]: evaluation of constant value failed
>   --> src\macros.rs:52:20
>    |
> 52 |         assert_eq!(date!(2023-04-31), NaiveDate::from_ymd_opt(2023, 4, 30).unwrap());
>    |                    ^^^^^^^^^^^^^^^^^ the evaluated program panicked at 'invalid calendar date', src\macros.rs:52:20
>    |
> ```

I still think a nice touch here is to be specific about precisely which argument caused the panic.  Up to you.

jtmoon79 (approved):
Approved. However I like my idea of getting specific about precisely which argument made the date invalid. Up to you.

sorairolake:
I think it would be useful to have an initialization macro for `DateTime<Utc>`.

```rs
// DateTime<Utc>
datetime!(2023-09-08 7:03:25Z);

// DateTime<FixedOffset>
datetime!(2023-09-08 7:03:25+00:00);
```

pitdicker:
> I think it would be useful to have an initialization macro for `DateTime<Utc>`.

Agreed. But as I understand such a syntax, using a letter after a number, is not supported by simple macro's. And my goal was to keep it simple, without resorting to proc macro's.

Note that you can do
```rust
// DateTime<Utc>
datetime!(2023-09-08 7:03:25).and_utc();
```

pitdicker:
Found a way to accept a time with fractions of a second using the `stringify` macro!

So we can now also do `time!(7:03:15.01)`, and the same for `datetime!()`.

@djc What would be needed to push this PR forwards? I think it will be a really good addition to improve the ergonomics of chrono.

b.t.w. I opened https://github.com/rust-lang/rust/issues/123156 for the CI errors.

PR #1271, opened 2023-09-08
duplicate derives on ArchivedT types (for rkyv feature)

Awpteamoose:
WIthout the derives ArchivedDuration loses a lot of its usefulness.

### DIFF
--- a/src/naive/date.rs
+++ b/src/naive/date.rs
@@ -192,6 +192,10 @@ impl Days {
 /// [proleptic Gregorian date]: crate::NaiveDate#calendar-date
 #[derive(PartialEq, Eq, Hash, PartialOrd, Ord, Copy, Clone)]
 #[cfg_attr(feature = "rkyv", derive(Archive, Deserialize, Serialize))]
+#[cfg_attr(
+    feature = "rkyv",
+    archive_attr(derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug, Hash))
+)]
 pub struct NaiveDate {
     ymdf: DateImpl, // (year << 13) | of
 }
--- a/src/duration.rs
+++ b/src/duration.rs
@@ -52,6 +52,7 @@ macro_rules! try_opt {
 /// This also allows for the negative duration; see individual methods for details.
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug, Hash)]
 #[cfg_attr(feature = "rkyv", derive(Archive, Deserialize, Serialize))]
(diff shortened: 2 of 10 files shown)


pitdicker:
Thank you for working on this. I still haven't worked with rkyv and can't say anything meaningful yet.
Wat is the reason only `Duration` needs this and not the other types?

pitdicker:
We can ignore coverage for this PR in my opinion.
But do we have any tests at all for the rkyv feature?

Awpteamoose:
Other types could probably use it too, but personally I only need it for `Duration`. I could probably go around and see where it makes sense to add.

src/naive/date.rs
```diff
@@ -192,6 +192,10 @@ impl Days {
 /// [proleptic Gregorian date]: crate::NaiveDate#calendar-date
 #[derive(PartialEq, Eq, Hash, PartialOrd, Ord, Copy, Clone)]
 #[cfg_attr(feature = "rkyv", derive(Archive, Deserialize, Serialize))]
+#[cfg_attr(
+    feature = "rkyv",
+    archive_attr(derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug, Hash))
```
pitdicker, line 197: Deriving those traits for the archived type pretty much makes us promise that `NaiveDate` will not be modified in a way that makes comparing the type any different from a plain integer. But I think that is a useful property anyway.

pitdicker (approved):

pitdicker:
Unrelated to this PR: the `rkyv` derive creates types such as `ArchivedDuration` in private modules. Maybe we should have a top-level `rkyv` module that re-exports them?

cc @mkatychev.

pitdicker:
Opened a PR to add an `rkyv` module https://github.com/chronotope/chrono/pull/1304.

src/naive/date.rs
Awpteamoose: If `DateImpl` changes you'll have to implement `Eq`/`Ord` in a way consistent with the current implementation, in which case it should be possible to implement rkyv traits and `Eq`/`Ord` on the Archived type in the same way. What I mean is, unless I'm misunderstanding, deriving traits for the Archived type doesn't make you promise anything you don't promise already.

pitdicker: True. But that is not all that easy, and would amount to deserializing the `Archived` type. That seems to defeat the purpose.

Anyway I think we are good.

mkatychev:
> Opened a PR to add an `rkyv` module #1304.

Sounds good, I'll resolve the merge conflicts and add the tests there

djc (approved):
Seems okay, please squash during merge.

PR #1274, opened 2023-09-08
Documentation tweaks

pitdicker:
This is one of those things that have been sitting as a half-finished branch for a couple of months...

I would like to tweak the `inline`/`no-inline` attributes of re-exports in `lib.rs` so that the the useful types are better visible than the ones that are rarely needed return types.

## Before
afbeelding

## After
afbeelding


pitdicker:
This took a little more effort.

If you don't document a type right at the crate root, it must be in another public module or otherwise it will not be documented:
- I made the `round` module public, so we can hide the `DurationRound` trait and co. in there (I don't like that one all :smile:).
- `SecondsFormat` is moved from the private `datetime` module to public module `formatting`. Seems like a better place anyway.

And because types such as `NaiveDate` and `NaiveTime` now have a copy of their documentation in the main module, old manual links in their methods were broken. I updated them to rust paths (but did not go over all documentation to fix all old-style links, just the ones that were broken).

src/lib.rs
```diff
@@ -533,7 +533,7 @@ pub use offset::LocalResult;
 #[doc(inline)]
 pub use offset::{FixedOffset, Offset, TimeZone, Utc};
 
-mod round;
+pub mod round;
```
djc, line 536: I don't understand the argument put forth in the commit message. What's the rationale for this change?

pitdicker: The idea is to highlight the importants traits in the "Traits" section: `TimeZone, `Offset`, `Datelike` and `Timelike`.
I don't think `DurationRound` and `SubsecRound` are good enough to be part of the list in the main documentation.
We can only hide it in the list if the `round` module is public.

To say it stronger, in my opinion the `DurationRound` trait is something that begs for a replacement.

djc: Okay, sure, but if that's the goal IMO the other changes to that goal should be in the same commit as that. In general, the first commit here seems too large/unfocused (it probably doesn't help that I'm completely unfamiliar with the semantics of `doc(no_inline)`)...

pitdicker: O, sorry. That was only the goal of that commit. No: the goal of the PR was tweaking `doc(incline)` and `doc(no_inline)`.

If the original module is public:
- a re-export comes under the list "Re-exports".
- `doc(incline)` on a re-export copies the documentation, so it now lives at two places.

If the original module is private:
- a re-export moves the documentation.
- `doc(no_incline)` makes the documentation unreachable.

src/lib.rs
```diff
@@ -477,70 +477,75 @@ extern crate alloc;
...
 pub use offset::Local;
-#[doc(no_inline)]
-pub use offset::{FixedOffset, LocalResult, Offset, TimeZone, Utc};
+pub use offset::LocalResult;
+#[doc(inline)]
+pub use offset::{FixedOffset, Offset, TimeZone, Utc};
 
-mod round;
+pub mod round;
+#[doc(no_inline)]
```
pitdicker, line 537: Oops, that means this line does nothing :smile:.

djc (approved):

PR #1275, opened 2023-09-08
Add `timestamp_nanos_opt`, deprecate `timestamp_nanos`

pitdicker:
`NaiveDateTime::timestamp_nanos` and `DateTime::timestamp_nanos` are two of the few methods that can panic on somewhat reasonable values and have no alternative that returns `Option`.

I added `timestamp_nanos_opt` and went ahead and deprecated the the old methods, like all other fallible methods.

The new method lead to a nice little simplification in `duration_round` and `duration_trunc`.

### DIFF
--- a/src/datetime/serde.rs
+++ b/src/datetime/serde.rs
@@ -157,6 +157,14 @@ pub mod ts_nanoseconds {
     ///
     /// Intended for use with `serde`s `serialize_with` attribute.
     ///
+    /// # Errors
+    ///
+    /// An `i64` with nanosecond precision can span a range of ~584 years. This function returns an
+    /// error on an out of range `DateTime`.
+    ///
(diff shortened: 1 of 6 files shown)


src/datetime/serde.rs
```diff
@@ -180,7 +188,10 @@ pub mod ts_nanoseconds {
     where
         S: ser::Serializer,
     {
-        serializer.serialize_i64(dt.timestamp_nanos())
+        serializer.serialize_i64(
```
djc, line 191: For all of these, why not yield errors here?

(Let's not have a `.` in `expect()` messages.)

djc (approved):

src/datetime/serde.rs
pitdicker: Good point, they should just return an error.

Issue #944, opened 2023-01-23
Json round-tripping changes leap seconds


Ekleog:
The following code triggers the second `assert_eq!` with a confusing error message:
```rust
fn main() {
    let time = chrono::NaiveTime::from_num_seconds_from_midnight_opt(0, 1_500_000_000).unwrap();
    assert_eq!(time, time, "time was equal to itself");
    let time2 = serde_json::from_slice(&serde_json::to_vec(&time).unwrap()).unwrap();
    assert_eq!(time, time2, "time was not equal to its json-roundtripped self"); // triggers with confusing error message
}
```
[(playground)](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=dc64d062203b7e0642a98d5005d78980)
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `00:00:01.500`,
 right: `00:00:01.500`: time was not equal to its json-roundtripped self', src/main.rs:5:5
```

This issue was again found by my fuzzer… and TBH I would have rather not had to care :sweat_smile: 

My thoughts are:
- Probably we should change the Debug repr of leap seconds, so it's obvious there's a leap second at play in this error message? I'm willing to do a PR but not sure what the output should be. Maybe `00:00:00+1.500`? Or `00:00:00(leap).500`?
- Defaulting to TAI (#21) would have avoided this issue in the first place
- I have literally no idea how to deal with json roundtripping of the value. Either we introduce a specific notation for leap seconds time in the JSON, but then that'd probably break literally all the other json parsers that wouldn't be ready for this kind of input. Or we just acknowledge that json serialization of non-TAI is lossy, but then it'd really be great if we had a way of serializing to TAI.

What do you think?

djc:
Changing the `Debug` representation makes sense to me, although I'm not sure how we should represent them -- something like `+1.500` seems easily confused with time zones so I feel like it should be more explicit though maybe less verbose than `(leap)` -- just enough to make it clear that something rare is going on.

We can discuss TAI in the linked issue. I think we'll have to accept that JSON serialization of non-TAI time is lossy.

Ekleog:
Hmm… maybe `00:00:00.leap.500` then, for the time that is currently being serialized as `00:00:01.500`?

ijackson:
[...]

PR #1283, opened 2023-09-12
Don't allow strange nanoseconds in initialization methods

pitdicker:
As discussed in https://github.com/chronotope/chrono/issues/944.

A leap second that is not on a minute boundary never happens in reality. It still makes sense to be able to represent them. But the less trivial it is to create them, the less confusion they can cause (?).

We have to add an extra check in `NaiveTime::from_hms_nano_opt` and `NaiveTime::from_num_seconds_from_midnight_opt`.

This also effects the following methods:
- `DateTime<Tz>::from_timestamp` (brand-new :smile:)
- `NaiveTime::from_hms_milli`
- `NaiveTime::from_hms_milli_opt`
- `NaiveTime::from_hms_micro`
- `NaiveTime::from_hms_micro_opt`
- `NaiveTime::from_hms_nano`
- `NaiveTime::from_hms_nano_opt`
- `NaiveTime::from_num_seconds_from_midnight`
- `NaiveDate::and_hms_milli`
- `NaiveDate::and_hms_milli_opt`
- `NaiveDate::and_hms_micro`
- `NaiveDate::and_hms_micro_opt`
- `NaiveDate::and_hms_nano`
- `NaiveDate::and_hms_nano_opt`
- `NaiveDateTime::from_timestamp`
- `NaiveDateTime::from_timestamp_opt`
- `TimeZone::timestamp`
- `TimeZone::timestamp_opt`

A strange leap-second can still be created from an existing value by using a `DateTime` with a fractional minute offset, and by changing a time with `Timelike::with_nanosecond`.


src/naive/date.rs
```diff
@@ -861,8 +861,8 @@ impl NaiveDate {
 
     /// Makes a new `NaiveDateTime` from the current date, hour, minute, second and millisecond.
     ///
-    /// The millisecond part can exceed 1,000
-    /// in order to represent the [leap second](./struct.NaiveTime.html#leap-second-handling).
+    /// The millisecond part is allowed exceed 1,000,000,000 in order to represent a [leap second](
```
djc, line 864: "allowed exceed" -> "allowed to exceed" for all of these?

Nit: please try to keep the first line of your commit messages within ~72 characters.

src/naive/time/mod.rs
```diff
@@ -458,7 +458,7 @@ impl NaiveTime {
     #[inline]
     #[must_use]
     pub const fn from_num_seconds_from_midnight_opt(secs: u32, nano: u32) -> Option<NaiveTime> {
-        if secs >= 86_400 || nano >= 2_000_000_000 {
+        if secs >= 86_400 || (nano >= 1_000_000_000 && secs % 60 != 59) || nano >= 2_000_000_000 {
```
djc, line 461: I would order the leap second conditional last here, since it is the most complex.

src/naive/date.rs
pitdicker: Oops, fixed.

pitdicker:
I forgot the `arbitrary` feature, let me take another look

djc (approved):

PR #1285, opened 2023-09-12
Add time zone timestamp micros

### LINKED ISSUE #1284: Add timestamp_micros to Utc
I wanted to know if there is a reason for having the functions `timestamp_millis` and `timestamp_nanos` in the trait `TimeZone` but not `timestamp_micros`.
If there is no reason not to have it, will it be possible to implement it?

emikitas:
I added `TimeZone::timestamp_micros_opt` and made `TimeZone::timestamp_millis_opt` use `NaiveDateTime::from_timestamp_millis` to reduce logic duplication.
resolves #1284

### DIFF
--- a/src/offset/mod.rs
+++ b/src/offset/mod.rs
@@ -431,6 +431,22 @@ pub trait TimeZone: Sized + Clone {
         self.timestamp_opt(secs, nanos as u32).unwrap()
     }
 
+    /// Makes a new `DateTime` from the number of non-leap microseconds
+    /// since January 1, 1970 0:00:00 UTC (aka "UNIX timestamp").
+    ///
+    /// #Example
+    /// ```
+    /// use chrono::{Utc, TimeZone};
+    ///
+    /// assert_eq!(Utc.timestamp_micros(1431648000000).unwrap().timestamp(), 1431648);
+    /// ```
+    fn timestamp_micros(&self, micros: i64) -> LocalResult<DateTime<Self>> {
+        match NaiveDateTime::from_timestamp_micros(micros) {
+            Some(dt) => LocalResult::Single(self.from_utc_datetime(&dt)),
+            None => LocalResult::None,
+        }
+    }
+
     /// Parses a string with the specified format string and returns a
     /// `DateTime` with the current offset.
     ///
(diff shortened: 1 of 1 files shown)


src/offset/mod.rs
```diff
@@ -556,4 +574,23 @@ mod tests {
...
+        let dt = Utc.timestamp_micros_opt(-1).unwrap();
+        assert_eq!(dt.to_string(), "1969-12-31 23:59:59.999999 UTC");
+        let dt = Utc.timestamp_micros_opt(-60_000_000).unwrap();
+        assert_eq!(dt.to_string(), "1969-12-31 23:59:00 UTC");
+        let dt = Utc.timestamp_micros_opt(-3_600_000_000).unwrap();
+        assert_eq!(dt.to_string(), "1969-12-31 23:00:00 UTC");
+    }
+
+    #[test]
+    fn test_micros_example() {
```
pitdicker, line 593: I don't think you need to copy the doctest here.

src/offset/mod.rs
```diff
@@ -431,6 +431,24 @@ pub trait TimeZone: Sized + Clone {
...
+    /// Makes a new `DateTime` from the number of non-leap microseconds
+    /// since January 1, 1970 0:00:00 UTC (aka "UNIX timestamp").
+    ///
+    /// #Example
+    /// ```
+    /// use chrono::{Utc, TimeZone};
+    ///
+    /// assert_eq!(Utc.timestamp_micros_opt(1431648000000).unwrap().timestamp(), 1431648);
+    /// ```
+    fn timestamp_micros_opt(&self, micros: i64) -> LocalResult<DateTime<Self>> {
```
pitdicker, line 443: Can this also use `NaiveTime::from_timestamp_micros`?

@djc Do we want the `_opt` suffix?

src/offset/mod.rs
emikitas: Thanks, I wasn't aware this gets tested automatically :)

src/offset/mod.rs
emikitas: I tested using `NaiveTime::from_timestamp_micros` locally, it seems to work as expected.
Once we decide if you'll rather not have the `_opt` suffix, I'll push my changes.

djc: Let's skip the `opt` suffix!

pitdicker (approved):
Maybe squash the last two commits?

djc (approved):

PR #1286, opened 2023-09-12
Make methods on `NaiveDateTime` const where possible

pitdicker:
I had to copy a few methods from the `Datelike` and `Timelike` traits to const methods on `NaiveDate` and `NaiveTime`, to make use of them in const methods on `NaiveDateTime`.

Otherwise the changes are fairly straightforward.

Only `DateTime<Tz>` to go after this (and a handful of methods that depend on `Duration`).

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -622,6 +622,14 @@ impl DateTime<Utc> {
         NaiveDateTime::from_timestamp_opt(secs, nsecs).as_ref().map(NaiveDateTime::and_utc)
     }
 
+    // FIXME: remove when our MSRV is 1.61+
+    // This method is used by `NaiveDateTime::and_utc` because `DateTime::from_naive_utc_and_offset`
+    // can't be made const yet.
+    // Trait bounds in const function / implementation blocks were not supported until 1.61.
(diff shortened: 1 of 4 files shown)


pitdicker:
The CI pointed out an issue because of our MSRV:
```
error[E0658]: trait bounds other than `Sized` on const fn parameters are unstable
   --> src\datetime\mod.rs:98:6
    |
98  | impl<Tz: TimeZone> DateTime<Tz> {
    |      ^^
...
123 |     pub const fn from_naive_utc_and_offset(datetime: NaiveDateTime, offset: Tz::Offset) -> DateTime<Tz> {
    |     --------------------------------------------------------------------------------------------------- function declared as const here
    |
    = note: see issue #93706 <https://github.com/rust-lang/rust/issues/93706> for more information

For more information about this error, try `rustc --explain E0658`.
error: could not compile `chrono` due to previous error
```

I have added a workaround so that we can have at least one constructor for a `DateTime`: `NaiveDateTime::and_utc`.

jtmoon79 (approved):
Looks great.

djc (approved):
Please squash the first three commits into the fourth one: they add code that goes unused.

Issue #1282, opened 2023-09-11
Add `CalendarDuration` type


pitdicker:
The goal is to add a type that can describe durations with the flexibility of an ISO 8601 duration. Because we already have enough confusion with types named `Duration`, I propose `CalendarDuration` as name.

It will have components that can describe a 'nominal duration' (to use the terms of ISO 8601), and an `accurate duration`.

```rust
#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
pub struct CalendarDuration {
    // Components with a nominal duration
    months: u32,
    days: u32,
    // Components with an accurate duration
    seconds: u32,
    nanos: u32,
}
```

ISO 8601 describes more possible components: years, weeks, hours and minutes. But these can all be expressed as a multiple of the components proposed above: years as 12 months, weeks as 7 days, hours as 3600 seconds, and minutes as 60 seconds.

ISO 8601 does not have a nanosecond component, but allows the last smallest component to have fractions. So we can map a fraction to `nanos`.

In theory a minute may not be always equal to 60 seconds, in the presence of a leap second. We currently can't do calculations with leap seconds. But if/when we do, this type may need a little revising.

Checklist for implementation:
- [ ] Add `CalendarDuration` type
- [ ] Initilization methods
- [ ] Implement `From` `Day`, `Month`, `std::time::Duration`
- [ ] Methods to get the components of a `CalendarDuration`
- [ ] `Display` implementation
- [ ] Parsing from ISO 8601 formatted string
- [ ] Implement `Add<CalendarDuration> for CalendarDuration`
- [ ] Various `diff_*` methods on `NaiveDate`, `NaiveDateTime` and `DateTime` to create a `CalendarDuration`
- [ ] `add_calendar_duration` methods on `NaiveDate`, `NaiveDateTime` and `DateTime`
- [ ] `sub_calendar_duration` methods on `NaiveDate`, `NaiveDateTime` and `DateTime`
- [ ] Other methods?

pitdicker:
The natural order to add a `CalendarDuration` to a some value is to first add the `months` component, and then the `days` component, and then the accurate component.

What to do if a result does not exist? That can be the case of the date is for example 2023-08-31 and you try to add one month. The result, 2023-09-31, does not exist. Returning 2023-10-01 does not seem right; adding one month to a date in August should not return a date in October.  Also returning 2023-09-30 does not seem right, the period in between is less than a month. In those cases we should return `None`.

[...]

PR #1290, opened 2023-09-14
Add `CalendarDuration` type

pitdicker:
See https://github.com/chronotope/chrono/issues/1282.

As a first PR this adds nothing but the type, methods to set and get the individual components, and a `Display` implementation.

We have 5 components: months, days, minutes, seconds and nanoseconds.
Example of ways to initialize a `CalendarDuration`:
```rust
use chrono::CalendarDuration;
// Duration in calendar months and days.
CalendarDuration::new().months(15).days(5);
// Duration in calendar days, and 3 hours.
CalendarDuration::new().days(5).hms(3, 0, 0).unwrap();
// Duration that will precisely count the number of seconds, including leap seconds.
CalendarDuration::new().seconds(23_456).millis(789).unwrap();
```

As described in https://github.com/chronotope/chrono/issues/1282#issuecomment-1719937171 we squeeze both the minutes and seconds components into one `u32`.
The idea is that it is too strange and niche to have two large components that only differ in how they count leap seconds.
So we either have:
- a large number of minutes, and sub-minute seconds (which works almost like ignoring leap-seconds exist)
- a large number of minutes (which accurately count leap seconds)

The `Debug` implementation will format the type as if it is a struct with fields for 5 components.

### DIFF
--- a/src/calendar_duration.rs
+++ b/src/calendar_duration.rs
@@ -1,0 +1,257 @@
+use core::num::NonZeroU64;
+
+use crate::{expect, try_opt};
+
+/// Duration type capable of expressing a duration as a combination of multiple units.
+///
+/// A duration can be expressed as a combination of four components: *months*, *days*, *minutes*,
+/// and *seconds and nanoseconds*. This type is compatible with the definition of an ISO 8601
+/// duration.
+///
+/// # Components
+///
+/// The *months* and *days* components are called components with a "nominal" durations, and can
+/// also be used to express years and weeks respectively.
+///
+/// The *minutes*, and *seconds and nanoseconds* components are called components with an "accurate"
+/// duration. Hours can be expressed with the *minutes* component. The definition of a minute is not
+/// exactly 60 seconds; it can also be 61 or 59 in the presence of leap seconds.
+///
+/// To keep operations with a `CalendarDate` sane a large value for the *seconds* component is
+/// mutually exclusive with the presence of a *minutes* component. If the *minutes* component is set
+/// the *seconds* component is limited to 60 seconds.
+///
+/// **Note**: The distinction between *minutes and seconds* components and only a *seconds*
+/// component prepares the `CalendarDate` type for working with leap seconds accurately.
+/// However they currently behave the same because full support for leap seconds is not yet
+/// implemented in chrono.
+///
+/// | component           | limit                    |
+/// |---------------------|--------------------------|
+/// | months              | `u32::MAX`               |
+/// | days                | `u32::MAX`               |
+/// | minutes and seconds | `u64::MAX >> 8` and `60` |
+/// | seconds             | `u64::MAX >> 2`          |
+/// | nanoseconds         | `999_999_999`            |
+///
+/// # Examples
+///
+/// ```
+/// # use chrono::CalendarDuration;
+/// // Encoding 1½ year as a duration of 18 months:
+/// let _duration = CalendarDuration::new().with_months(18);
+///
+/// // Encoding 1½ year as a duration of 12 months and 183 days (366 / 2):
+/// let _duration = CalendarDuration::new().with_months(12).with_days(183);
+///
+/// // Encoding 1½ year as a duration of 548 days (365 + 366 / 2):
+/// let _duration = CalendarDuration::new().with_days(548);
+///
+/// // Encoding 1½ year as a duration in seconds:
+/// let _duration = CalendarDuration::new().with_seconds(548 * 24 * 60 * 60).unwrap();
+/// ```
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+pub struct CalendarDuration {
+    // Components with a nominal duration
+    months: u32,
+    days: u32,
+    // Components with an accurate duration
+    mins_and_secs: MinutesAndSeconds,
+    nanos: u32,
+}
+
+impl Default for CalendarDuration {
+    fn default() -> Self {
+        CalendarDuration::new()
+    }
+}
+
+impl CalendarDuration {
+    /// Create a new duration initialized to `0`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use chrono::CalendarDuration;
+    /// // Duration in calendar months and days.
+    /// let _ = CalendarDuration::new().with_months(15).with_days(5);
+    /// // Duration in calendar days, and 3 hours.
+    /// let _ = CalendarDuration::new().with_days(5).with_hms(3, 0, 0).unwrap();
+    /// // Duration that will precisely count the number of seconds, including leap seconds.
+    /// let _ = CalendarDuration::new().with_seconds(23_456).unwrap().with_nanos(789).unwrap();
+    /// ```
+    #[must_use]
+    pub const fn new() -> Self {
+        Self {
+            months: 0,
+            days: 0,
+            mins_and_secs: expect(MinutesAndSeconds::from_seconds(0), "in range"),
+            nanos: 0,
+        }
+    }
+
+    /// Set the months component of this duration to `months`.
+    #[must_use]
+    pub const fn with_months(mut self, months: u32) -> Self {
+        self.months = months;
+        self
+    }
+
+    /// Set the days component of this duration to `days`.
+    #[must_use]
+    pub const fn with_days(mut self, days: u32) -> Self {
+        self.days = days;
+        self
+    }
+
+    /// Sets the accurate component of this duration to a value in minutes and seconds.
+    ///
+    /// The definition of a minute is not exactly 60 seconds; it can also be 61 or 59 in the
+    /// presence of leap seconds.
+    ///
+    /// Durations made with this method are the closest thing to working like leap seconds don't
+    /// exists.
+    ///
+    /// # Errors
+    ///
+    /// Returns `None`:
+    /// - if `seconds` exceeds 60
+    /// - if the value of the minutes component (`hours * 60 + minutes`) would be out of range.
+    pub const fn with_hms(mut self, hours: u64, minutes: u64, seconds: u8) -> Option<Self> {
+        let minutes = try_opt!(try_opt!(hours.checked_mul(60)).checked_add(minutes));
+        self.mins_and_secs =
+            try_opt!(MinutesAndSeconds::from_minutes_and_seconds(minutes, seconds as u64));
+        Some(self)
+    }
+
+    /// Sets the accurate component of this duration to a value in seconds.
+    ///
+    /// This duration will count an exact number of seconds, which includes the counting of leap
+    /// seconds.
+    ///
+    /// The minutes component will be set to zero.
+    ///
+    /// # Errors
+    ///
+    /// Returns `None` if the value of the seconds component would be out of range.
+    #[must_use]
+    pub const fn with_seconds(mut self, seconds: u64) -> Option<Self> {
+        self.mins_and_secs = try_opt!(MinutesAndSeconds::from_seconds(seconds));
+        Some(self)
+    }
+
+    /// Sets the nanoseconds component of this duration to `nanos`.
+    ///
+    /// # Errors
+    ///
+    /// Returns `None` if `nanos` exceeds 999_999_999.
+    pub const fn with_nanos(mut self, nanos: u32) -> Option<Self> {
+        if nanos >= 1_000_000_000 {
+            return None;
+        }
+        self.nanos = nanos;
+        Some(self)
+    }
+
+    /// Returns the `months` component of this calendar duration.
+    pub const fn months(&self) -> u32 {
+        self.months
+    }
+
+    /// Returns the `days` component of this calendar duration.
+    pub const fn days(&self) -> u32 {
+        self.days
+    }
+
+    /// Returns the `minutes` and `seconds` components of this calendar duration.
+    pub const fn mins_and_secs(&self) -> (u64, u64) {
+        self.mins_and_secs.mins_and_secs()
+    }
+
+    /// Returns the `nanos` component of this calendar duration.
+    pub const fn nanos(&self) -> u32 {
+        self.nanos
+    }
+
+    /// Returns `true` if this `CalendarDuration` spans no time.
+    pub const fn is_zero(&self) -> bool {
+        let (mins, secs) = self.mins_and_secs();
+        self.months == 0 && self.days == 0 && mins == 0 && secs == 0 && self.nanos == 0
+    }
+}
+
+/// Type to encode either seconds, or minutes and up to 60 seconds.
+///
+/// We don't allow having both an `u64` with minutes and an `u64` with seconds.
+/// Having two components in a `CalendarDuration` whose only subtle difference is how they behave
+/// around leap seconds will make the type unintuitive.
+///
+/// # Encoding
+///
+/// - Seconds:                                     `seconds << 2 | 0b10`
+/// - Minutes and up to 60 seconds: `minutes << 8 | seconds << 2 | 0b11`
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+struct MinutesAndSeconds(NonZeroU64);
+
+impl MinutesAndSeconds {
+    const fn from_seconds(secs: u64) -> Option<Self> {
+        if secs >= (1 << 62) {
+            return None;
+        }
+        Some(Self(expect(NonZeroU64::new(secs << 2 | 0b10), "can't fail, non-zero")))
+    }
+
+    const fn from_minutes_and_seconds(mins: u64, secs: u64) -> Option<Self> {
+        if mins >= (1 << 56) || secs > 60 {
+            return None;
+        }
+        // The `(mins > 0) as u64` causes a value without minutes to have the same encoding as
+        // one created with `from_seconds`.
+        let val = mins << 8 | secs << 2 | (mins > 0) as u64 | 0b10;
+        Some(Self(expect(NonZeroU64::new(val), "can't fail, non-zero")))
+    }
+
+    /// Returns the `minutes` and `seconds` components.
+    const fn mins_and_secs(&self) -> (u64, u64) {
+        let value = self.0.get();
+        match value & 0b01 == 0 {
+            true => (0, value >> 2),
+            false => (value >> 8, (value >> 2) & 0b11_1111),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::CalendarDuration;
+
+    #[test]
+    fn test_basic_functionality() {
+        #[track_caller]
+        fn compare(d: CalendarDuration, months: u32, days: u32, mins: u64, secs: u64, nanos: u32) {
+            assert_eq!(d.months(), months);
+            assert_eq!(d.days(), days);
+            assert_eq!((d.mins_and_secs()), (mins, secs));
+            assert_eq!(d.nanos(), nanos);
+        }
+
+        let duration = CalendarDuration::new()
+            .with_months(1)
+            .with_days(2)
+            .with_hms(3, 4, 5)
+            .unwrap()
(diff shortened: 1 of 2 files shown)


pitdicker:
For initialization the methods work much like the builder pattern, except that we don't need a seperate builder type.
We have methods to initialize the individual components as `months()`, `days()`, `hms()`, `seconds()` and `nanos()`.
For convenience I added methods methods that initialize with other units, but with a name that hopefully helps to make clear on which component they act: `years_and_months()`, `weeks_and_days()`, `micros()` and `millis()`.

I have intentionally added only methods that set a field, not methods that add some value to a field. And also not `Add` implementations. It is quite easy to create a `CalendarDuration` with a combination of components that will be near-impossible to explain. In my opinion the type should not be treated as black boxes that you can just add together. You should be aware of the existing components when changing anything.

To get the value of the individual components, we have not five but four methods (because minutes and seconds are closely related): `months_component()`, `days_component()`, `mins_and_secs_component()`, and `nanos_component()`. No convenience methods here.

src/calendar_duration.rs
```diff
@@ -0,0 +1,379 @@
...
+            "can't fail because the `| 0b10` ensures the value is non-zero"
+        );
+        Some(self)
+    }
+
+    /// Sets the accurate component of this duration to a value in seconds.
+    ///
+    /// This duration will count an exact number of seconds, which includes the counting of leap
+    /// seconds.
+    pub const fn seconds(mut self, seconds: u32) -> Self {
```
ARBaart, line 194: If I understand correctly, the object is the builder so CalendarDuration::new().months(5).months(7) == CalendarDuration::new().months(7).

That is reasonably intuitive for months, but the struct suggests fields months, days, minutes, seconds, nanos. Because of that suggestion, I don't think a user would fully suspect CalendarDuration::new().hms(1, 2, 3).seconds(0) == CalendarDuration::new().

src/calendar_duration.rs
```diff
@@ -0,0 +1,244 @@
...
+/// // Encoding 1½ year as a duration of 12 months and 183 days (366 / 2):
+/// let _duration = CalendarDuration::new().months(12).days(183);
+///
+/// // Encoding 1½ year as a duration of 548 days (365 + 366 / 2):
+/// let _duration = CalendarDuration::new().days(548);
+///
+/// // Encoding 1½ year as a duration in seconds:
+/// let _duration = CalendarDuration::new().seconds(548 * 24 * 60 * 60);
+/// ```
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
```
djc: Would be nice to also have a `Default` impl here.

src/calendar_duration.rs
```diff
@@ -0,0 +1,244 @@
...
+pub struct CalendarDuration {
+    // Components with a nominal duration
+    months: u32,
+    days: u32,
+    // Components with an accurate duration
+    // `seconds` can either encode `minutes << 6 | seconds`, or just seconds.
+    seconds: u32,
+    // `nanos` encodes `nanoseconds << 2 | has_minutes << 1 | 1`.
+    nanos: NonZeroU32,
```
djc, line 38: I'd prefer to implement the boring version here and optimize memory usage later if we decide that is necessary. Since these are public fields anyway, there should be no need to prematurely optimize.

src/calendar_duration.rs
```diff
@@ -0,0 +1,244 @@
...
+        Self {
+            months: 0,
+            days: 0,
+            seconds: 0,
+            nanos: expect!(NonZeroU32::new(1), "can't fail, non-zero"),
+        }
+    }
+
+    /// Set the months component of this duration to `months`.
+    pub const fn months(mut self, months: u32) -> Self {
```
djc, line 130: I think we should call this `with_months()` and allow `months()` to be used for the getter (same for the other ones).

src/calendar_duration.rs
```diff
@@ -23,7 +24,7 @@ use crate::{expect, try_opt};
 /// // Encoding 1½ year as a duration in seconds:
 /// let _duration = CalendarDuration::new().seconds(548 * 24 * 60 * 60);
 /// ```
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
```
djc: Let's keep this out of the initial commit if you're going to replace it in the second commit, anyway.

src/calendar_duration.rs
```diff
@@ -35,6 +36,19 @@ pub struct CalendarDuration {
     nanos: NonZeroU32,
 }
 
+impl fmt::Debug for CalendarDuration {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let (mins, secs) = self.mins_and_secs_component();
+        f.debug_struct("CalendarDuration")
```
djc, line 44: How about we avoid enconding fields that are 0?

src/calendar_duration.rs
```diff
@@ -268,9 +270,31 @@ impl CalendarDuration {
     }
 }
 
+impl From<Days> for CalendarDuration {
```
djc, line 273: Let's skip the `From<Days>` and `From<Months>` impl pending decisions about whether we'll deprecate those.

src/calendar_duration.rs
```diff
@@ -49,6 +49,58 @@ impl fmt::Debug for CalendarDuration {
     }
 }
 
+impl fmt::Display for CalendarDuration {
```
djc, line 54: I think this should come with a `FromStr` impl and a bunch of round-tripping tests.

src/calendar_duration.rs
```diff
@@ -49,6 +49,58 @@ impl fmt::Debug for CalendarDuration {
...
+            f.write_fmt(format_args!("{}Y", years))?;
+        }
+        if months > 0 {
+            f.write_fmt(format_args!("{}M", months))?;
+        }
+        if self.days > 0 {
+            f.write_fmt(format_args!("{}D", self.days))?;
+        }
+        let nanos = self.nanos_component();
+        if self.seconds > 0 || nanos > 0 {
```
djc, line 72: Invert the condition to return early here, please.

src/calendar_duration.rs
```diff
@@ -49,6 +49,58 @@ impl fmt::Debug for CalendarDuration {
...
+            let (mins, secs) = self.mins_and_secs_component();
+            let hours = mins / 60;
+            let mins = mins % 60;
+            if hours > 0 {
+                f.write_fmt(format_args!("{}H", hours))?;
+            }
+            if mins > 0 {
+                f.write_fmt(format_args!("{}M", mins))?;
+            }
+            if secs > 0 || nanos > 0 {
```
djc, line 83: Invert + early return.

src/calendar_duration.rs
```diff
@@ -49,6 +49,58 @@ impl fmt::Debug for CalendarDuration {
...
+                f.write_fmt(format_args!("{}", secs))?;
+                if nanos > 0 {
```
djc, line 85: And here.

djc (commented):
Good start!

djc:
(Please avoid adding more stuff to this PR -- other than per review feedback.)

src/calendar_duration.rs
pitdicker: A parser will be a couple hundred lines. You you want to make it part of this initial PR?

djc: Yes, I think so.

pitdicker: A complete parser is quite non-trivial :laughing:. It would re-use ca. 75% of #1143.
That is partially because besides the `Pnn̲Ynn̲Mnn̲DTnn̲Hnn̲Mnn̲S` and `Pnn̲W` formats ISO 8601 allows `P<date>T<time>`, with all formats except *week date* supported for those.

Of course I'll start with `Pnn̲Ynn̲Mnn̲DTnn̲Hnn̲Mnn̲S`. That will only share the handling of fractions with #1143.

@djc Can you please review the date and time parsers from that PR, if I split them out into smaller PRs? I know you are busy :disappointed:. Meanwhile I'll continue this parser.

pitdicker:
Setting to draft for now. I have most of the parser ready, but need to clean up things a bit more before it is ready a second look.

src/calendar_duration.rs
pitdicker: Renamed. I was going for having short names for initialization, but there is something to say for making the getters have the short names.

src/calendar_duration.rs
pitdicker: We have this cfg in lib.rs: `#![deny(missing_debug_implementations)]`, so the initial version can't compile without.

src/calendar_duration.rs
pitdicker: Isn't it strange for a struct to leave out fields that are 0 from its `Debug` format?

src/calendar_duration.rs
pitdicker: Yes, that is better 👍.

src/calendar_duration.rs
djc: I'm confused, how can `P<date>T<time>` be a duration?

src/calendar_duration.rs
pitdicker: There is a little premature optimization going on. But my goal is more about having two clear modes of operation. My thought process went like this:

1. Include a flag for 'I care about leap seconds'/'leap seconds don't exist'.
   However that is not really something we can serialize, and also not enough to make `CalendarDuration` at least in theory capable of working with leap seconds correctly.
2. Have explicit `minutes` and `seconds` fields.
3. Suppose we pre-multiply the `minutes` field with `60`. This turns the fields into `seconds_in_utc` and `seconds_in_tai`.
   This is becoming a mess. Having two pretty much the same fields that only differ in how they count leap seconds within one duration seems unreasonable.
4. Introduce two sane modes of operation:
   - arbitrary large minutes and only sub-minute seconds
   - arbitrary large seconds

To encode that with rust types:
```rust
pub struct CalendarDuration {
    // Components with a nominal duration
    months: u32,
    days: u32,
    // Components with an accurate duration
    accurate: AccurateDuration,
}

enum AccurateDuration {
    MinutesAndSubMinuteSecondsAndNanos(u32, u8, u32),
    SecondsAndNanos(u32, u32);
}
```

We can simplify that a bit if we manually uphold the relation between `minutes` and `seconds` fields to:
```rust
pub struct CalendarDuration {
    // Components with a nominal duration
    months: u32,
    days: u32,
    // Components with an accurate duration
    minutes: u32,
    seconds: u32,
    nanos: u32,
}
```

The idea is that if you have an existing `CalendarDuration` and change it with `with_seconds`, it should switch to the `SecondsAndNanos` mode. Previous minutes and seconds should be forgotten. And if `minutes` is non-zero, `seconds` must be less than `61`.

If we already have to uphold a relation between `minutes` and `seconds` anyway, why not encode them in a single `u32` and get the smaller type size for free? And if we don't do it now but want to keep the option for the optimization open for later, we should already include the range checks.

I have to admit stuffing the discriminant inside the `nanos` field, and using the other bit for niche optimization, is a premature optimization. Can we please keep it?

pitdicker:
> (Please avoid adding more stuff to this PR -- other than per review feedback.)

Sorry. I added two more things, but minor I think: `#[must_use]` to a few methods, and a test to check the setters correctly work with our-of-range values.

src/calendar_duration.rs
djc: Okay, feel free to kill that!

src/calendar_duration.rs
djc: Seems fine to me.

src/calendar_duration.rs
djc: IMO that's way too much magic. I'd rather have the actual enum if we're going to do something like this.

pitdicker: I added an `MinutesAndSeconds` type to encode this in.

And I switched to accurate part of a duration to a 64-bit type.
With a 32-bit value we could only store an accurate duration of ca. 136 years. With a 64-bit type `CalendarDuration` can cover all uses over `std::time::Duration` and `TimeDelta` (forgetting the negative part). With this type we also never have to make methods fallible that construct a duration from the difference between two dates; the result always fits.

pitdicker:
Started working on this again.
@djc are you sure you don't want to have the parsing code split out from this PR so we can merge a minimal part first?

pitdicker:
Put a couple of hours into self-review and improving the initial documentation.

src/calendar_duration.rs
```diff
@@ -0,0 +1,294 @@
...
+    /// # use chrono::CalendarDuration;
+    /// // Duration in calendar months and days.
+    /// let _ = CalendarDuration::new().with_months(15).with_days(5);
+    /// // Duration in calendar days, and 3 hours.
+    /// let _ = CalendarDuration::new().with_days(5).with_hms(3, 0, 0).unwrap();
+    /// // Duration that will precisely count the number of seconds, including leap seconds.
+    /// let _ = CalendarDuration::new().with_seconds(23_456).unwrap().with_millis(789).unwrap();
+    /// ```
+    #[must_use]
+    pub const fn new() -> Self {
```
djc, line 255: Should add a `Default` implementation that calls this.

src/calendar_duration.rs
```diff
@@ -0,0 +1,294 @@
...
+        self.months = months;
+        self
+    }
+
+    /// Set the months component of this duration to `years * 12 + months`.
+    ///
+    ///  # Errors
+    ///
+    /// Returns `None` if the value of the months component would be out of range.
+    pub const fn with_years_and_months(mut self, years: u32, months: u32) -> Option<Self> {
```
djc, line 276: I'd prefer to skip this one for now, easy enough to implement from the outside.

src/calendar_duration.rs
```diff
@@ -0,0 +1,294 @@
...
+        self.days = days;
+        self
+    }
+
+    /// Set the days component of this duration to `weeks * 7 + days`.
+    ///
+    ///  # Errors
+    ///
+    /// Returns `None` if the value of the days component would be out of range.
+    pub const fn with_weeks_and_days(mut self, weeks: u32, days: u32) -> Option<Self> {
```
djc, line 293: Same here.

src/calendar_duration.rs
```diff
@@ -0,0 +1,294 @@
...
+    /// Returns `None` if `nanos` exceeds 999_999_999.
+    pub const fn with_nanos(mut self, nanos: u32) -> Option<Self> {
+        if nanos >= 1_000_000_000 {
+            return None;
+        }
+        self.nanos = nanos;
+        Some(self)
+    }
+
+    /// Sets the nanoseconds component of this duration to `micros * 1000`.
```
djc, line 347: Let's skip `micros` and `millis` for now.

src/calendar_duration.rs
```diff
@@ -0,0 +1,294 @@
...
+
+/// Type to encode either seconds, or minutes and up to 60 seconds.
+///
+/// We don't allow having both an `u64` with minutes and an `u64` with seconds.
+/// Having two components in a `CalendarDuration` whose only subtle difference is how they behave
+/// around leap seconds will make the type unintuitive.
+///
+/// # Encoding
+///
+/// - Seconds:                                     `seconds << 2 | 0b10`
```
djc, line 416: How much memory is this stuff saving? It's not obvious to me that this is worth it.

src/calendar_duration.rs
djc: Especially since the fields are private.

src/calendar_duration.rs
```diff
@@ -292,6 +293,16 @@ impl CalendarDuration {
     }
 }
 
+impl From<Duration> for CalendarDuration {
```
djc, line 398: Let's make this a `TryFrom` instead?

src/calendar_duration.rs
```diff
@@ -62,6 +62,12 @@ pub struct CalendarDuration {
     nanos: u32,
 }
 
+impl Default for CalendarDuration {
```
djc, line 149: (Let's fold this into the initial commit.)

djc (approved):
Let's split this into smaller chunks, first the `CalendarDuration` type itself (up to and including the `Default` commit, except squashing that) but removing some of the extra setters.

src/calendar_duration.rs
pitdicker: Yes, better :+1:.

src/calendar_duration.rs
pitdicker: Done. It feels strange to me tbh.

src/calendar_duration.rs
pitdicker: My main motivation is to encode the either *seconds* or *minutes and seconds* cases.

Still I prefer to make the type not larger than necessary, especially if one of the goals is to replace the `Days` and `Months` types. An enum will add 8 bytes to a type of currently 20 bytes; 40%.

src/calendar_duration.rs
pitdicker: I hope to have these eventually. Not only for convenience, but to match the components ISO 8601 defines.

pitdicker:
@djc You gave an approval, but want to have a final look?

src/calendar_duration.rs
djc: I spent some time thinking this over and would still prefer that we start with a "real" enum -- that doesn't rely on bit-shifting tricks. If we get an issue that complains about the size we can decide whether it's worth optimizing this. But it seems unlikely to me at this point that people will want to hold a ton of these in an `Vec` or similar.

pitdicker: With an enum we would still have most of the complexity.
- In my opinion we want to keep the range checks in `MinutesAndSeconds::from_seconds` and `from_minutes_and_seconds`.
  The restricted range for minutes gives as the guarantee `minutes * 60 + seconds` does not overflow.
  The checks would be needed to allow for this 'optimization' in the future.
- We still need the special case where we encode a value created with `from_minutes_and_seconds(0, s)` the same as `from_seconds(s)` so that they compare equal.
- We still need a match to get both values in `mins_and_secs()`.

Switching to an enum is not going to save many lines or maybe even none.

I just don't consider a couple of bitwise operations much of a problem and well worth it for the much nicer type size. Really don't want to drop them :disappointed:.

pitdicker: How are we going to proceed here?

djc: I still think this is a textbook case of premature optimization and think we should drop the bitshifting for now.

pitdicker: We have a difference of opinion here. If chrono gets a `CalendarDuration` as new functionality I want it to be as good as I can make it. Do you want to block the functionality on this detail?

punkstarman:
What's missing to get this merged?

djc:
> What's missing to get this merged?

The disagreement in the open comment thread.

src/calendar_duration.rs
djc: @pitdicker hasn't been active for a while now. If you want to take over this PR, probably fine to go ahead (but keep them as a co-author on the commits).

pitdicker:
@djc Sorry again be becoming inactive. Chrono was I nice project to work on, and I learned from you as a maintainer. I still hope to return someday but can't promise.

For this and the other PR's: don't have any concerns about authorship or my preferences. Feel free to take it in any way that might help chrono.
And for this PR: my repository probably has a couple of branches that build on top of this one with more features for the `CalenderDuration` type.

I still think the implementation here was conceptually simpler than an enum, but do as you please.

djc:
> @djc Sorry again be becoming inactive. Chrono was I nice project to work on, and I learned from you as a maintainer. I still hope to return someday but can't promise.

No need to apologize, thanks for all the effort you spent on chrono and good luck with your current and future endeavors!

PR #1294, opened 2023-09-15
fix: underflow during datetime->nanos conversion

### LINKED ISSUE #1289: Utc nanoseconds do not always roundtrip
# Reproducer

## Code
```rust
#[test]
fn nano_roundrip() {
    let nanos = i64::MIN + 2;
    let dt = Utc.timestamp_nanos(nanos);
    let nanos2 = dt.timestamp_nanos();
    assert_eq!(nanos, nanos2);
}
```

## Expected Result
Test passes.

## Actual Result
Test panics when calling `dt.timestamp_nanos()` (so we don't even get to the actual comparison:

`value can not be represented in a timestamp with nanosecond precision.`

# Technical Background
`dt.timestamp()` is `-9223372037`, however `nanos / 1_000_000_000` is `-9223372036`, hence I conclude that this IF branch was taken:

https://github.com/chronotope/chrono/blob/b64cedc6a8758d15265702895d917045ce1eebf8/src/offset/mod.rs#L427-L430

Hence this `checked_mul` overflows:

https://github.com/chronotope/chrono/blob/b64cedc6a8758d15265702895d917045ce1eebf8/src/naive/datetime/mod.rs#L482

~My guess is that the conversion routine (first code block) isn't entirely accurate.~ The conversion is alright, I rather think when converting back we have to be more careful.

crepererum:
Fixes #1289.

### DIFF
--- a/src/datetime/tests.rs
+++ b/src/datetime/tests.rs
@@ -1486,3 +1486,36 @@ fn locale_decimal_point() {
     assert_eq!(dt.format_localized("%T%.6f", ar_SY).to_string(), "18:58:00.123456");
     assert_eq!(dt.format_localized("%T%.9f", ar_SY).to_string(), "18:58:00.123456780");
 }
+
+/// This is an extended test for <https://github.com/chronotope/chrono/issues/1289>.
+#[test]
+fn nano_roundrip() {
+    const BILLION: i64 = 1_000_000_000;
+
+    for nanos in [
+        i64::MIN,
+        i64::MIN + 1,
+        i64::MIN + 2,
+        i64::MIN + BILLION - 1,
+        i64::MIN + BILLION,
+        i64::MIN + BILLION + 1,
+        -BILLION - 1,
+        -BILLION,
+        -BILLION + 1,
+        0,
+        BILLION - 1,
+        BILLION,
+        BILLION + 1,
+        i64::MAX - BILLION - 1,
+        i64::MAX - BILLION,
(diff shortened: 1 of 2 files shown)


djc (approved):
Thanks for the thorough test case!

src/datetime/tests.rs
```diff
@@ -1486,3 +1486,36 @@ fn locale_decimal_point() {
...
+        BILLION,
+        BILLION + 1,
+        i64::MAX - BILLION - 1,
+        i64::MAX - BILLION,
+        i64::MAX - BILLION + 1,
+        i64::MAX - 2,
+        i64::MAX - 1,
+        i64::MAX,
+    ] {
+        println!("nanos: {nanos}");
```
pitdicker, line 1516: Maybe remove this `println!`? Or can you change it so that the test compiles on Rust 1.57 (our MSRV)?

djc: Just needs `println!("nanos: {}", nanos)`.

crepererum: Adjusted the println to be compatible w/ Rust 1.57. I think keeping it helps debugging a failing tests due to the many cases[^1].

[^1]: I wish there would be a slightly nicer story for parametrized tests in "standard Rust" though.

djc: (Hopefully [there will be](https://epage.github.io/blog/2023/06/iterating-on-test/) in the not-too-distant future.)

PR #1295, opened 2023-09-15
Bump version to 0.4.31

pitdicker:


djc (approved):
Thanks!

pitdicker:
Published.

cmackenzie1:
> Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backward compatible functionality is introduced to the public API. It MUST be incremented if any public API functionality is marked as deprecated. It MAY be incremented if substantial new functionality or improvements are introduced within the private code. It MAY include patch level changes. Patch version MUST be reset to 0 when minor version is incremented.

https://semver.org/#spec-item-7 🫣

djc:
@cmackenzie1 that's not how pre-1.0 Cargo dependencies work. Cargo counts the left-most non-zero version as the version that govers incompatibility, so in a version like 0.4.31 the 4 really acts as the major version.

pitdicker:
@cmackenzie1 Sorry for the deprecations. You may be updating from a couple of versions before?
Hopefully the suggested alternatives help you to prevent panics in you library/application.

cmackenzie1:
> @cmackenzie1 that's not how pre-1.0 Cargo dependencies work. Cargo counts the left-most non-zero version as the version that govers incompatibility, so in a version like 0.4.31 the 4 really acts as the major version.

@djc today I learned! Thanks for the explanation.

PR #1296, opened 2023-09-15
Add ISO 8601 date parser

pitdicker:
@djc https://github.com/chronotope/chrono/pull/1143 seems a bit too large to review. But the first few commits make sense on their own.
This is the commit that adds an ISO 8601 date parser.

The parser would be used by `NaiveDate::parse_from_iso8601()`.
It is a component for the ISO 8601 parsers for `NaiveDateTime` and `DateTime<FixedOffset>`.

But ISO 8601 also specifies that the date components of a duration can use the date formats supported by this parser, with the exception of the formats for a week date.

### DIFF
--- a/src/format/parse.rs
+++ b/src/format/parse.rs
@@ -27,7 +27,7 @@ fn set_weekday_with_num_days_from_sunday(p: &mut Parsed, v: i64) -> ParseResult<
     })
 }
 
-fn set_weekday_with_number_from_monday(p: &mut Parsed, v: i64) -> ParseResult<()> {
+pub(super) fn set_weekday_with_number_from_monday(p: &mut Parsed, v: i64) -> ParseResult<()> {
     p.set_weekday(match v {
         1 => Weekday::Mon,
         2 => Weekday::Tue,
--- a/src/format/mod.rs
+++ b/src/format/mod.rs
@@ -44,6 +44,7 @@ mod parsed;
 
 // due to the size of parsing routines, they are in separate modules.
 mod parse;
+pub(crate) mod parse_iso8601;
 pub(crate) mod scan;
 
 pub mod strftime;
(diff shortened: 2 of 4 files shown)


pitdicker:
Maybe we should move all the ISO 8601 parsing code into a module of its own?

src/format/parse.rs
```diff
@@ -224,6 +224,103 @@ pub(crate) fn parse_rfc3339<'a>(parsed: &mut Parsed, mut s: &'a str) -> ParseRes
     Ok((s, ()))
 }
 
+#[derive(Copy, Clone, PartialEq, Eq)]
+pub(crate) enum Iso8601Format {
```
pitdicker, line 228: This enum serves no purpose yet. We need it to make sure that the date, time and offset components use the same format when parsing to a `NaiveDateTime`, `DateTime` or duration.

src/format/parse.rs
```diff
@@ -224,6 +224,103 @@ pub(crate) fn parse_rfc3339<'a>(parsed: &mut Parsed, mut s: &'a str) -> ParseRes
...
+/// | basic format, expanded representations    | ±Y̲YYYYMMDD    | ±Y̲YYYYDDD    | ±Y̲YYYYWwwD   |
+/// | extended format, expanded representations | ±Y̲YYYY-MM-DD  | ±Y̲YYYY-DDD   | ±Y̲YYYY-Www-D |
+///
+/// Returns `(remainder, Iso8601Format)`.
+//// - The ISO 8601 format of the date is returned so the calling function can check it matches the
+///   format of a time component (basic or extended format).
+pub(crate) fn parse_iso8601_date<'a>(
+    parsed: &mut Parsed,
+    mut s: &'a str,
+) -> ParseResult<(&'a str, Iso8601Format)> {
```
pitdicker, line 250: To reuse this method for parsing a duration it should take a flag to indicate it should not parse *week date* formats.

djc:
> My main motivation for adding an ISO 8601 parser is that chrono is currently unable to parse is own Debug output if the year is outside the 0..=9999 range supported by RFC 3339. Edit: I forgot the FromString implementation of DateTime is very forgiving and accepts this.
> 
> A second motivation is that this crate frequently mentions ISO 8601. Then it is strange it does not even have a parser for that format 😆.

As far as motivations go, I find these to be fairly weak, so I don't think I'm going to prioritize reviewing these for the near future.

PR #1299, opened 2023-09-18
improve doc comment and tests for timestamp_nanos_opt

mlegner:
- The doc comment for `DateTime::timestamp_nanos_opt` (added in #1275) incorrectly described a panic, which is actually an error.
- The lower and upper limits of valid timestamps were not precise (both for `DateTime` and `NaiveDateTime` and both `timestamp_nanos` and `timestamp_nanos_opt`). This commit adds the precise limits to all methods and doctests for those limits to `DateTime::timestamp_nanos_opt`.


djc (approved):
Thanks!

PR #1303, opened 2023-09-21
move MIN_UTC, MAX_UTC to a specialized impl

pitdicker:
Rebased version of https://github.com/chronotope/chrono/pull/935.

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -545,11 +545,6 @@ impl<Tz: TimeZone> DateTime<Tz> {
             .expect("writing rfc3339 datetime to string should never fail");
         result
     }
-
-    /// The minimum possible `DateTime<Utc>`.
-    pub const MIN_UTC: DateTime<Utc> = DateTime { datetime: NaiveDateTime::MIN, offset: Utc };
-    /// The maximum possible `DateTime<Utc>`.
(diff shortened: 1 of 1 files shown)


src/datetime/mod.rs
```diff
@@ -831,6 +826,11 @@ impl DateTime<Utc> {
     pub fn parse_from_str(s: &str, fmt: &str) -> ParseResult<DateTime<Utc>> {
         DateTime::<FixedOffset>::parse_from_str(s, fmt).map(|result| result.into())
     }
+
+    /// The minimum possible `DateTime<Utc>`.
+    pub const MIN_UTC: Self = DateTime { datetime: NaiveDateTime::MIN, offset: Utc };
```
djc, line 831: Should we lose the `_UTC` suffix now that it's sort of implied by the `<Utc>` type param?

djc (approved):

src/datetime/mod.rs
pitdicker: Previously you would have to write `let dt = DateTime<Utc>::MIN_UTC`, and with this change it becomes `let dt = DateTime::MIN_UTC`.
With the change I think keeping `_UTC` it makes code more readable.

PR #1305, opened 2023-09-21
Switch to `doc_auto_cfg`

pitdicker:
Since end 2021 docs.rs can make use of the [`doc_auto_cfg`](https://github.com/rust-lang/rust/pull/90502).
With it we can remove all the `#[cfg_attr(docsrs, doc(cfg(feature = "...")))]` lines.

Like the `doc_cfg` feature we used until now it is only available on nightly.

I also switched the `arbitrary` and `unstable-locales` features to visible on docs.rs.

Can be tested locally with:
```
RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --features=serde,rkyv,unstable-locales,arbitrary
```


pitdicker:
I have removed `RUSTFLAGS="--cfg docsrs"` from the cargo-deadlinks lint, and made the lint check all features.
The cfg does not influence the links anyway, and can run on stable.

djc (approved):

PR #1306, opened 2023-09-21
Deprecate standalone `format` functions

pitdicker:
Split out from https://github.com/chronotope/chrono/pull/1163.

In the `format` module we have 4 standalone [`format*` functions](https://docs.rs/chrono/latest/chrono/format/index.html#functions). I don't think they were ever intended to be public.
They need an `&mut Formatter<'_>` as argument, which can't be constructed outside of a `Display` implementation.

To deprecated these methods I inlined them in `DelayedFormat::Display`, and made the methods wrappers around that.

### DIFF
--- a/src/format/formatting.rs
+++ b/src/format/formatting.rs
@@ -7,8 +7,9 @@
 use alloc::string::{String, ToString};
 #[cfg(any(feature = "alloc", feature = "std"))]
 use core::borrow::Borrow;
-use core::fmt;
-use core::fmt::Write;
+#[cfg(any(feature = "alloc", feature = "std"))]
+use core::fmt::Display;
+use core::fmt::{self, Write};
 
 #[cfg(any(
     feature = "alloc",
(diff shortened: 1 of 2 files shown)


pitdicker:
Oops, 0.4.28 is no longer the next version to mentions in the deprecation :rofl:.

src/format/formatting.rs
```diff
@@ -161,6 +161,7 @@ impl<'a, I: Iterator<Item = B> + Clone, B: Borrow<Item<'a>>> Display for Delayed
 /// Internally used by `DelayedFormat`.
 #[cfg(any(feature = "alloc", feature = "std"))]
 #[cfg_attr(docsrs, doc(cfg(any(feature = "alloc", feature = "std"))))]
+#[deprecated(since = "0.4.32")]
```
djc, line 164: Let's add a `note` about what to use instead?

djc (approved):

PR #1307, opened 2023-09-22
Don't let `unstable-locales` imply the `alloc` feature

pitdicker:
I hope that at some point we add support for parsing different locales (should not be all that hard). And that we can make the new formatting API work without `alloc` (proven to be possible in https://github.com/chronotope/chrono/pull/1163).

In both cases it does not make much sense for the `unstable-locales` feature to imply the `alloc` feature.

This will break uses of chrono with `no-default-features` with the `unstable-locales` feature, that happen to also depend on `alloc`.
I feel we are allowed to make this change because of the name *unstable*-locales.

Split out from https://github.com/chronotope/chrono/pull/1163.


pitdicker:
All localized formatting methods now also depend on `alloc` or `std`.
`StrftimeItems::new_with_locale` becomes available on `no_std`.

djc:
I'm okay with this, but `any(feature = "alloc", feature = "std")` is silly because `std` should imply `alloc`.

pitdicker:
> I'm okay with this, but `any(feature = "alloc", feature = "std")` is silly because `std` should imply `alloc`.

It did not bother me enough yet :smile:. I have added a commit to clean that up.

djc (approved):

pitdicker:
Nice, this also reduced the number of feature combinations the CI tests with 35%.

PR #1310, opened 2023-09-23
Add `NaiveTime::overflowing_(add|sub)_offset`

pitdicker:
On its own these methods are not that useful. We need them to nicely implement `NaiveDateTime::checked_(add|sub)_offset`.
It does make the existing `Add` and `Sub` implementations simpler.

Split out from https://github.com/chronotope/chrono/pull/1069.


djc (approved):

pitdicker:
# :tada:

PR #1311, opened 2023-09-23
Remove `format::{format_localized, format_item_localized}`

pitdicker:
I forgot that in https://github.com/chronotope/chrono/pull/1306 `format::{format_localized, format_item_localized}` are behind the `unstable-locales` feature. Instead of deprecating I think we can just removed them.
That is what it did in the PR the commit was based on, but I messed up.

### DIFF
--- a/src/format/formatting.rs
+++ b/src/format/formatting.rs
@@ -188,53 +188,6 @@ pub fn format_item(
     .fmt(w)
 }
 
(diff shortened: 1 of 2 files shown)


pitdicker:
Also I can't imagine anyone has ever used these.

djc (approved):
Nice!

PR #1313, opened 2023-09-23
Add `NaiveDateTime::checked_(add|sub)_offset`

pitdicker:
These methods are const and public so they can be used with https://github.com/chronotope/chrono/pull/1270.

Because the offset is never more than 24 hours, we can use `NaiveDate::pred_opt()` and `NaiveDate::succ_opt()` Instead of `NaiveDate::add_days`. This gives a nice little performance improvement for many methods on `DateTime` (benchmarks in https://github.com/chronotope/chrono/pull/1069#issuecomment-1732226078).

Most of all we can use these methods to avoid out of range panics when converting datetimes from UTC to local or back.

Split out from https://github.com/chronotope/chrono/pull/1069.

### DIFF
--- a/src/naive/datetime/mod.rs
+++ b/src/naive/datetime/mod.rs
@@ -1619,6 +1619,15 @@ impl AddAssign<Duration> for NaiveDateTime {
     }
 }
 
+impl Add<FixedOffset> for NaiveDateTime {
+    type Output = NaiveDateTime;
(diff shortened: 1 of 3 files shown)


djc (approved):

pitdicker:
Thank you, this made my day :smile:.

PR #1317, opened 2023-09-26
Fix out of range panics in `DateTime` getters and setters

pitdicker:
Split out from https://github.com/chronotope/chrono/pull/1048.

> If the timestamp of a `DateTime` is near the end of the range of `NaiveDateTime` and the offset pushes the timestamp beyond that range, all kinds of methods currently panic. About 40 methods that are part of the public API. With the `Debug` implementation panicking being the most fun. See #1047.
> 
> Most of these methods fall in two categories:
> 
> * They panic on an intermediate value, but should be able to return a meaningful result. Examples are formatting, `year()`, `hour()`.
>
> * They can return `None`. Examples are `checked_add_days`, `with_month`.
> 
> 
> A somewhat easy fix is to slightly restrict the range of `NaiveDate`, so that we have some buffer space to represent the out-of-range datetimes. Everything that needs just the intermediate value will be fixed by this. But care should be taken to never let an invalid intermediate value escape to the library user.

This PR only fixes the first group of methods: those that panic on an intermediate value, but should be able to return a meaningful result. This includes formatting, and getters and setters of the `Datelike` and `Timelike` traits.

I made `MIN_YEAR` and `MAX_YEAR` 1 year smaller to have some buffer space. `DateTime::overflowing_naive_local` and `NaiveDateTime::overflowing_add_offset` are added to create a value in that buffer space.

The rest of the PR consists of just using these methods and adding tests.

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -22,7 +22,7 @@ use crate::format::{
     StrftimeItems, TOO_LONG,
 };
 #[cfg(feature = "alloc")]
-use crate::format::{write_rfc3339, DelayedFormat};
+use crate::format::{write_rfc2822, write_rfc3339, DelayedFormat};
 use crate::naive::{Days, IsoWeek, NaiveDate, NaiveDateTime, NaiveTime};
 #[cfg(feature = "clock")]
 use crate::offset::Local;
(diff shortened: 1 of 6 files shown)


pitdicker:
This does not yet fix the `with_year`, `checked_add_months`, `checked_sub_months`, `checked_add_days`, and `checked_sub_days` methods on `DateTime`, or parsing. If we want to support round-tripping these should all be able to create dates in the buffer space, and I am not yet sure how to do so in a clean way. Parts of that work remain in https://github.com/chronotope/chrono/pull/1048.

djc (approved):

PR #1318, opened 2023-09-26
Revert #806

pitdicker:
https://github.com/chronotope/chrono/pull/806 conveniently adds parsing methods to `DateTime<Utc>`. I don't believe it was fully realized at the time how much of a change this is going to be for users.

All `parse_from_*` methods are going to need type annotations on 0.5. These are widely used methods. I think this makes the common use more annoying.

Parsing to `DateTime<FixedOffset>` is what I consider the regular case. It is the type that best matches the format strings of RFC 3339, RFC 2822, ISO 8601, and any manual string that can be parsed as a `DateTime`.
```rust
// 0.4.x
let dt = DateTime::parse_from_rfc3339("2015-06-30T23:56:05+02:00")?;
let dt: DateTime<Utc> = DateTime::parse_from_rfc3339("2015-06-30T23:56:05+02:00")?.into();

// with #806:
let dt = DateTime::<FixedOffset>::parse_from_rfc3339("2015-06-30T23:56:05+02:00")?;
let dt = DateTime::<Utc>::parse_from_rfc3339("2015-06-30T23:56:05+02:00")?;
```

Parsing to `DateTime<Utc>` folds the offset in the datetime value, essentially losing the offset information.
In https://github.com/chronotope/chrono/issues/1075 the discussion was to go in the opposite direction for 0.5: if we parse a value with fields that the target type can't hold, we should return an error.

So I would like to revert for two reasons: all the extra type annotations, and to make it an explicit choice for users to drop the offset information.

cc @mqudsi.

### DIFF
--- a/bench/benches/chrono.rs
+++ b/bench/benches/chrono.rs
@@ -12,7 +12,7 @@ fn bench_datetime_parse_from_rfc2822(c: &mut Criterion) {
     c.bench_function("bench_datetime_parse_from_rfc2822", |b| {
         b.iter(|| {
             let str = black_box("Wed, 18 Feb 2015 23:16:09 +0000");
-            DateTime::<FixedOffset>::parse_from_rfc2822(str).unwrap()
+            DateTime::parse_from_rfc2822(str).unwrap()
         })
     });
 }
@@ -21,7 +21,7 @@ fn bench_datetime_parse_from_rfc3339(c: &mut Criterion) {
     c.bench_function("bench_datetime_parse_from_rfc3339", |b| {
         b.iter(|| {
             let str = black_box("2015-02-18T23:59:60.234567+05:00");
-            DateTime::<FixedOffset>::parse_from_rfc3339(str).unwrap()
+            DateTime::parse_from_rfc3339(str).unwrap()
         })
     });
 }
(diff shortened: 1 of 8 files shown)


djc (approved):
What is our most convenient way to convert from `DateTime<FixedOffset>` to `DateTime<Utc>`? Perhaps if we make that easy enough it compensates for this?

pitdicker:
> What is our most convenient way to convert from `DateTime<FixedOffset>` to `DateTime<Utc>`?

I think `let dt: DateTime<Utc> = DateTime::parse_from_rfc3339("2015-06-30T23:56:05+02:00")?.into();` is the best we have now, `into()` doesn't take type annotations.

> Perhaps if we make that easy enough it compensates for this?

We can add a method similar to `DateTime::fixed_offset()`. I am not sure about the best name. Maybe `DateTime::utc_only()`, to show it forgets about the offset?

> Should we deprecate these on 0.4.x?

There is no equivalent on 0.4.x.

djc:
But it doesn't forget, right? It just folds it into the time? I was thinking it might be called `into_utc()`.

pitdicker:
I don't know, but kind of want to make it clear this is a non-reversible conversion. But well, so is `DateTime::fixed_offset()` if you are converting from a `chrono-tz` timezone.

From the [API guidelines](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv):

> Conversions prefixed `as_` and `into_` typically *decrease abstraction*, either exposing a view into the underlying representation (`as`) or deconstructing data into its underlying representation (`into`). Conversions prefixed `to_`, on the other hand, typically stay at the same level of abstraction but do some work to change from one representation to another.

Sound like we should go with `to_*`. But `into_utc()` matches nicely with the current `into()` :+1:.

djc:
Ah yeah, I suppose `to_utc()` makes more sense.

PR #1325, opened 2023-09-26
Add `DateTime::to_utc`

pitdicker:
To have an easy way to convert any `DateTime<Tz>` to `DateTime<Utc>`.

As discussed in https://github.com/chronotope/chrono/pull/1318.


djc:
In my mind the proposed behavior was to adjust the `DateTime` for the offset, not to just discard the offset.

pitdicker:
The value is already stored in UTC, so it does nothing special compared to `into()`.

djc (approved):

PR #1328, opened 2023-09-27
Make `Display` format of `Duration` conform better to ISO 8601

pitdicker:
The `Display` format of `Duration` currently converts all values greater than 24 hours to a value with days. But in ISO 8601 `P1D` has a different meaning from `PT86400S`.

If we get a duration type that properly supports the ISO 8601 format with different components (https://github.com/chronotope/chrono/issues/1282), I think we should fix the simple `Duration` type to always write only seconds. We later may be able to share the parser and a `FromStr` implementation.

I ported the code to count the significant fractional digits in `nanos` from https://github.com/chronotope/chrono/pull/1290.
It should find a shared place once one of these PRs has landed.

### DIFF
--- a/src/time_delta.rs
+++ b/src/time_delta.rs
@@ -538,27 +538,30 @@ impl fmt::Display for TimeDelta {
         // but we need to print it anyway.
         let (abs, sign) = if self.secs < 0 { (-*self, "-") } else { (*self, "") };
 
-        let days = abs.secs / SECS_PER_DAY;
-        let secs = abs.secs - days * SECS_PER_DAY;
-        let hasdate = days != 0;
-        let hastime = (secs != 0 || abs.nanos != 0) || !hasdate;
-
         write!(f, "{}P", sign)?;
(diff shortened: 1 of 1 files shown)


src/duration.rs
```diff
@@ -413,27 +413,30 @@ impl fmt::Display for Duration {
...
-            } else {
-                write!(f, "T{}.{:09}S", secs, abs.nanos)?;
+
+        f.write_fmt(format_args!("T{}", abs.secs))?;
+
+        if abs.nanos > 0 {
+            // Count the number of significant digits, while removing all trailing zero's.
+            let mut figures = 9usize;
+            let mut fraction_digits = abs.nanos;
+            loop {
```
djc, line 428: What's the point of this loop?

pitdicker: To count the number of significant digits, and removing all trailing zero's. I.e. repeatedly divide `nanos` by 10.
I do not know a cleaner way to write it.

djc (approved):

PR #1330, opened 2023-09-28
Add `DateTime::try_to_rfc2822`, deprecate `DateTime::to_rfc2822`

pitdicker:
Split out from https://github.com/chronotope/chrono/pull/1144.

This adds a new method `try_to_rfc2822` that returns `None` instead of panicking on years outside of 0..9999.

### DIFF
--- a/bench/benches/chrono.rs
+++ b/bench/benches/chrono.rs
@@ -46,7 +46,9 @@ fn bench_datetime_to_rfc2822(c: &mut Criterion) {
                 .unwrap(),
         )
         .unwrap();
-    c.bench_function("bench_datetime_to_rfc2822", |b| b.iter(|| black_box(dt).to_rfc2822()));
+    c.bench_function("bench_datetime_to_rfc2822", |b| {
+        b.iter(|| black_box(dt).try_to_rfc2822().unwrap())
+    });
 }
 
 fn bench_datetime_to_rfc3339(c: &mut Criterion) {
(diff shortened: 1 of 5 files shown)


pitdicker:
@djc Sorry for all the pushes.
This PR and #1331 expose two more methods that can fail, that would be nice to convert to return `Result`s on 0.5.

I don't depend on them for any other work. But maybe nice to keep in mind for reviewing? They should not do anything surprising, it is more a matter of: do we want to deprecate them on 0.4?

djc:
So my main hesitation about this and #1331 has been that I find the `try_to_rfc2822()` method name ugly/unidiomatic. I wonder if there's something else we could do here that would clarify how RFC 2822-compatible datetimes are really a (strict?) subset of the values supported by our `DateTime` type, and if that could also help reducing fallibility in a new format API down the line.

For example, maybe we have a `Rfc2822` wrapper for `DateTime` and this becomes a `TryFrom` impl?

pitdicker:
> So my main hesitation about this and #1331 has been that I find the `try_to_rfc2822()` method name ugly/unidiomatic. I wonder if there's something else we could do here that would clarify how RFC 2822-compatible datetimes are really a (strict?) subset of the values supported by our `DateTime` type, and if that could also help reducing fallibility in a new format API down the line.
> 
> For example, maybe we have a `Rfc2822` wrapper for `DateTime` and this becomes a `TryFrom` impl?

I'm not too fond of the name either. Maybe we should just make the `to_rfc2822()` method infallible on 0.5 and leave it as it is for 0.4.x? 0.5 feels less far away than half a year ago.

A wrapper type such as `Rfc2822` to encode the invariant that a `DateTime` is in the range `0..10_000` does not seem all that useful to me. There is no other use for it except formatting. Then I'd rather have a fallible formatting method than a fallible conversion to another type as an extra step.

PR #1333, opened 2023-09-28
Use `overflowing_naive_local` in `DateTime::checked_add*`

pitdicker:
The final part of https://github.com/chronotope/chrono/pull/1048.

Using `overflowing_naive_local` in `DateTime::checked_(add|sub)_days` and `DateTime::checked_(add|sub)_months` prevents panics.

I thought it would be good if users can rely on another property: that a clear no-op works as a no-op. I.e.`DateTime::checked_add_days(Days::new(0))` should always return the original value.
To do that, `checked_(add|sub)_days` and `checked_(add|sub)_months` have a fast path that returns `Some(self)` if it is a no-op.
They already had a fast path anyway, we just start relying on it.

`with_year` gains a fast path to have the same property.

What we can't promise is that you can add or subtract and create a value that is just out of range for the local time but still in range for the UTC value. That would require alternative methods on `NaiveDate` for `add_days`, `checked_*_months` and `with_year` that have a different bound checks. It just doesn't seem worth it to me.

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -436,13 +436,17 @@ impl<Tz: TimeZone> DateTime<Tz> {
     /// # Errors
     ///
     /// Returns `None` if:
-    /// - The resulting date would be out of range.
     /// - The local time at the resulting date does not exist or is ambiguous, for example during a
     ///   daylight saving time transition.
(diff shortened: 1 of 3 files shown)


djc (approved):

pitdicker:
Thank you!

PR #1334, opened 2023-09-29
Fix `Duration::MIN.abs()` (adjust `Duration::MIN` by 1 millisecond)

pitdicker:
Currently `Duration::MIN.abs()` will create a value that is 1 millisecond beyond `Duration::MAX.abs()`.
We can choose to add a fallible variant of `Duration::abs`.
But given that `Duration::MIN` is pretty much chosen arbitrarily as `i64::MIN` *milli*seconds, I propose to redefine it as `-i64::MAX` milliseconds and make the range symmetrical.

### DIFF
--- a/src/duration.rs
+++ b/src/duration.rs
@@ -61,10 +61,10 @@ pub struct Duration {
     nanos: i32, // Always 0 <= nanos < NANOS_PER_SEC
 }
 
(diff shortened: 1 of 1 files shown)


djc (approved):

pitdicker:
Wow, approved before the CI is even ready!

PR #1335, opened 2023-10-01
Refactor formatting

pitdicker:
We currently have a 234-line `format_inner` function.
In https://github.com/chronotope/chrono/pull/1163 I found a way to make it a bit nicer and faster.
This is only the initial refactoring, that splits it up into three methods on `DelayedFormat`.

Every commit is as close to a copy-paste as compiles and passes rustfmt.

### DIFF
--- a/src/format/formatting.rs
+++ b/src/format/formatting.rs
@@ -112,20 +112,16 @@ impl<'a, I: Iterator<Item = B> + Clone, B: Borrow<Item<'a>>> DelayedFormat<I> {
         let name_and_diff = (offset.to_string(), offset.fix());
         DelayedFormat { date, time, off: Some(name_and_diff), items, locale: Some(locale) }
     }
-}
 
-#[cfg(feature = "alloc")]
-impl<'a, I: Iterator<Item = B> + Clone, B: Borrow<Item<'a>>> Display for DelayedFormat<I> {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+    fn format(&self, w: &mut impl Write) -> fmt::Result {
         #[cfg(feature = "unstable-locales")]
         let locale = self.locale;
         #[cfg(not(feature = "unstable-locales"))]
         let locale = None;
 
-        let mut result = String::new();
         for item in self.items.clone() {
             format_inner(
-                &mut result,
+                w,
                 self.date,
                 self.time,
                 self.off.as_ref(),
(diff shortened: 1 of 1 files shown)


Zomtir (approved):
Looks sane. Reviewing this is quite something, mostly because the huge indentation and lack of encapsulation of the old code. So this patch is surely a welcome improvement.

I know that the original `format_inner` function used a simple `w` as the main parameter name. By the time I arrived at the third patch I already forgot and had to go back. Maybe `write` would be more verbose for people with light memory like me.

Same goes for for `Some(v)`. Still trying to figure out what it stands for... a padding **v**alue?

There were a few memory allocation changes from pass-by-reference to pass-by-copy (and back). I didn't track those through-full, but in the compiler I trust.

pitdicker:
Thank you for reviewing!

> There were a few memory allocation changes from pass-by-reference to pass-by-copy (and back). I didn't track those through-full, but in the compiler I trust.

Yes, this is just a copy-paste and minimal changes from by-reference to by-value as needed to keep the code compiling.

`v` is short for `value`. It is used within `format_numeric` for whatever value needs to be formatted as an integer.

For context:
- #1196 was a large rewrite of our formatting code. Parts of it are performance improvements and making the code `no_std` compatible. The goal is to make it easier to re-use parsed format strings, and to avoid panics while formatting.
- #1163 was split out as a first step, including only the performance improvements and `no_std` compatibility.
- This PR is split out from #1163 as just the most basic refactor.
- I have another branch to build on this one and port over the performance improvements.

I hope this PR can land before it rots and needs to be redone from 0 to keep it a copy-paste. Hmmm, already has a conflict, I'll have to see about that.

pitdicker:
> Looks sane. Reviewing this is quite something, mostly because the huge indentation and lack of encapsulation of the old code. So this patch is surely a welcome improvement.

Rustfmt kept changing the formatting of this huge method. So part of the reason to split it over multiple methods is to avoid that.

pitdicker:
> I hope this PR can land before it rots and needs to be redone from 0 to keep it a copy-paste. Hmmm, already has a conflict, I'll have to see about that.

My own fault with #1473 :smile:. That's doable.

src/format/formatting.rs
```diff
@@ -422,6 +362,75 @@ fn format_inner(
     }
 }
 
+#[cfg(feature = "alloc")]
+fn format_numeric(
```
djc: How about attaching this to `Numeric` as a method?

src/format/formatting.rs
```diff
@@ -431,6 +274,161 @@ fn format_numeric(
     }
 }
 
+#[cfg(feature = "alloc")]
+fn format_fixed(
```
djc: Same here.

src/format/formatting.rs
pitdicker: Later commits make it a method on `DelayedFormat`. That also seems like a good place?

pitdicker:
@djc Do you want to give this a second look?

src/format/formatting.rs
djc: Right, that seems to be actually better.

djc (approved):
Looks great!

PR #1336, opened 2023-10-01
Don't use deprecated method in `impl Arbitrary for DateTime`

pitdicker:
Missed in https://github.com/chronotope/chrono/pull/1175. I also modified the CI to check the `arbitrary` feature.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -79,7 +79,7 @@ jobs:
       - uses: taiki-e/install-action@cargo-hack
       - uses: Swatinem/rust-cache@v2
       - run: |
-          cargo hack check --feature-powerset --optional-deps serde \
+          cargo hack check --feature-powerset --optional-deps arbitrary,serde \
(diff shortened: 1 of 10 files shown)


pitdicker:
It turns out arbitrary doesn't support `no_std` (see https://github.com/rust-fuzz/arbitrary/issues/38, https://github.com/rust-fuzz/arbitrary/pull/74).

Because there is a chance it may in the future, I thought it best to not make the `arbitrary` feature imply `std`, but to put our implementations behind an extra `std` feature gate.

pitdicker:
Converted to draft for now. Maybe my comment in https://github.com/rust-fuzz/arbitrary/pull/74#issuecomment-1742096990 gets a reply.

pitdicker:
This is a small PR that replaces the use of a deprecated method and sets up a CI test for the `arbitrary` feature. I don't think we need to wait any longer on seeing changes to arbitrary to merge this.

djc (approved):

PR #1337, opened 2023-10-02
Make remaining methods const

pitdicker:
Depends on https://github.com/chronotope/chrono/pull/1137.

This finishes the work to make all methods const where possible.
Where it is not possible is parsing and formatting, and anything that involves traits including `DateTime<tz>`.

### DIFF
--- a/src/duration.rs
+++ b/src/duration.rs
@@ -1144,6 +1144,40 @@ mod tests {
     }
 
     #[test]
+    fn test_duration_const() {
+        const ONE_WEEK: Duration = Duration::weeks(1);
+        const ONE_DAY: Duration = Duration::days(1);
+        const ONE_HOUR: Duration = Duration::hours(1);
+        const ONE_MINUTE: Duration = Duration::minutes(1);
+        const ONE_SECOND: Duration = Duration::seconds(1);
+        const ONE_MILLI: Duration = Duration::milliseconds(1);
+        const ONE_MICRO: Duration = Duration::microseconds(1);
+        const ONE_NANO: Duration = Duration::nanoseconds(1);
+        let combo: Duration = ONE_WEEK
+            + ONE_DAY
+            + ONE_HOUR
+            + ONE_MINUTE
+            + ONE_SECOND
+            + ONE_MILLI
+            + ONE_MICRO
+            + ONE_NANO;
+
+        assert!(ONE_WEEK != Duration::zero());
+        assert!(ONE_DAY != Duration::zero());
+        assert!(ONE_HOUR != Duration::zero());
+        assert!(ONE_MINUTE != Duration::zero());
+        assert!(ONE_SECOND != Duration::zero());
+        assert!(ONE_MILLI != Duration::zero());
+        assert!(ONE_MICRO != Duration::zero());
+        assert!(ONE_NANO != Duration::zero());
+        assert_eq!(
+            combo,
+            Duration::seconds(86400 * 7 + 86400 + 3600 + 60 + 1)
+                + Duration::nanoseconds(1 + 1_000 + 1_000_000)
+        );
+    }
+
+    #[test]
     #[cfg(feature = "rkyv-validation")]
     fn test_rkyv_validation() {
         let duration = Duration::seconds(1);
--- a/src/duration.rs
+++ b/src/duration.rs
@@ -75,6 +75,23 @@ pub(crate) const MAX: Duration = Duration {
 };
 
 impl Duration {
+    /// Makes a new `Duration` with given number of seconds and nanoseconds.
+    ///
+    /// # Errors
+    ///
+    /// Returns `None` when the duration is out of bounds, or if `nanos` ≥ 1,000,000,000.
+    pub(crate) const fn new(secs: i64, nanos: u32) -> Option<Duration> {
+        if secs < MIN.secs
+            || secs > MAX.secs
+            || nanos > 1_000_000_000
+            || (secs == MAX.secs && nanos > MAX.nanos as u32)
+            || (secs == MIN.secs && nanos < MIN.nanos as u32)
+        {
+            return None;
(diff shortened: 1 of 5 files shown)


src/naive/time/mod.rs
```diff
@@ -580,46 +580,30 @@ impl NaiveTime {
...
     /// ```
     #[must_use]
-    pub fn overflowing_add_signed(&self, mut rhs: OldDuration) -> (NaiveTime, i64) {
-        let mut secs = self.secs;
-        let mut frac = self.frac;
-
-        // check if `self` is a leap second and adding `rhs` would escape that leap second.
-        // if it's the case, update `self` and `rhs` to involve no leap second;
-        // otherwise the addition immediately finishes.
+    pub const fn overflowing_add_signed(&self, rhs: OldDuration) -> (NaiveTime, i64) {
```
jtmoon79, line 583: There appear to be significant logic changes to `overflowing_add_signed`. Currently, `test_time_overflowing_add` does not have any asserts involving fractional seconds.
https://github.com/chronotope/chrono/blob/v0.4.31/src/naive/time/tests.rs#L117

I strongly recommend adding more asserts to `test_time_overflowing_add` specifically targeting this change in logic.

jtmoon79 (changes requested):

src/naive/time/mod.rs
pitdicker: > There appear to be significant logic changes to `overflowing_add_signed`.

True, this method was pretty convoluted because it comes from a time before `rem_euclid`, like we had in more places in chrono.

This code is *extremely* well-tested, better than any other part of chrono. The tests just work on the higher-level methods that wrap `test_time_overflowing_add`.

src/duration.rs
```diff
@@ -257,40 +263,30 @@ impl Duration {
 
     /// Add two durations, returning `None` if overflow occurred.
     #[must_use]
-    pub fn checked_add(&self, rhs: &Duration) -> Option<Duration> {
-        let mut secs = try_opt!(self.secs.checked_add(rhs.secs));
+    pub const fn checked_add(&self, rhs: &Duration) -> Option<Duration> {
+        // No overflow checks here because we stay comfortably wihtin the range of an `i64`.
```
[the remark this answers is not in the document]
pitdicker: Thank you, fixed.

pitdicker:
I hoped to make this the last PR before suggesting a 0.4.32 release. And then 4 months evaporated :roll_eyes:.
Took care of the merge conflicts.

pitdicker:
Updated. Now that our MSRV is 1.61 it is possible to also make `Duration::to_std` const.

src/duration.rs
```diff
@@ -350,6 +350,15 @@ impl Duration {
         }
         Ok(StdDuration::new(self.secs as u64, self.nanos as u32))
     }
+
+    /// This duplicates `Neg::neg` because trait methods can't be const yet.
+    pub(crate) const fn neg(self) -> Duration {
+        if self.nanos == 0 {
+            Duration { secs: -self.secs, nanos: 0 }
+        } else {
+            Duration { secs: -self.secs - 1, nanos: NANOS_PER_SEC - self.nanos }
```
djc, line 359: I'd prefer to spell this more like this:

```rust
let (secs_diff, nanos) = match self.nanos {
    0 => (0, 0),
    nanos => (1, NANOS_PER_SEC - nanos),
}

Duration { secs: -self.secs - secs_diff, nanos }
```

djc (approved):

src/duration.rs
pitdicker: It was a direct copy of the code in the trait method just below it. But this is nice, updated.

pitdicker:
I'll wait with merging until after https://github.com/chronotope/chrono/pull/1385, to make this the one that has to rebase.

danwilliams:
@pitdicker I'm concerned about some of the changes brought in here. I like the new `Duration::new()` function - that's a great addition! - but I'm seeing what I perceive as an increased brittleness in the code through direct checks on hard-coded values such as `-i64::MAX` instead of their representatives.

I can understand why this is, but I feel it is a wrong path.

Over in https://github.com/chronotope/chrono/issues/1392 we have discussed reworking some of the constants - it seems to me that this would be a good opportunity to correct some of these patterns at the same time. What do you think? 🙂 Would you be open to me submitting a PR that not only streamlines the constants but also ensures they are used in the various checks that are currently diverse?

pitdicker:
You are right. It would especially be nice to replace the `-i64::MAX` with a proper constant. Unfortunately I also did it in one place of your new code to make the function const (CI was instantly red after my naive rebase). I'll look forward to what you come up with 👍.

For now I'm glad to have the changes to make the `Duration` methods const land. It was based on https://github.com/chronotope/chrono/pull/638 and https://github.com/chronotope/chrono/pull/1137.

src/duration.rs
```diff
@@ -82,6 +75,23 @@ pub(crate) const MAX: Duration = Duration {
...
 impl Duration {
+    /// Makes a new `Duration` with given number of seconds and nanoseconds.
+    ///
+    /// # Errors
+    ///
+    /// Returns `None` when the duration is out of bounds, or if `nanos` ≥ 1,000,000,000.
+    pub(crate) const fn new(secs: i64, nanos: u32) -> Option<Duration> {
+        if secs < MIN.secs
+            || secs > MAX.secs
+            || nanos > 1_000_000_000
```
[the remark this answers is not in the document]
pitdicker: Thank you for catching that!

pitdicker:
> Would you be open to me submitting a PR that not only streamlines the constants but also ensures they are used in the various checks that are currently diverse?

@danwilliams Keep an eye on https://github.com/chronotope/chrono/pull/1351 that is also touching constants in the `duration` module.

PR #1342, opened 2023-10-28
docs: Improve deprecation of `TimeZone::datetime_from_str`

tmccombs:
Add recommending using `NaiveDate::parse_from_str` as well


djc (approved):

djc:
Thanks, and sorry for the long delay!

PR #1343, opened 2023-10-28
Split `clock` feature into `clock` and `now`

### LINKED ISSUE #1341: Split `clock` feature into `clock-now` and `clock-tz`
The `clock` feature currently pulls in timezone features which pulls in a bunch of crates to deal with timezones (`iana-time-zone`, `android-tz`). Some of these crates pull in system libraries that negatively affect startup time for Deno, which means we need to create an inconvenient shim for getting the current UTC time: https://github.com/denoland/deno/blob/main/cli/main.rs.

It would be great if the feature could be split into smaller sub-features: one for reading the current time, and one for reading tz data, which would allow us to use `Utc::now()` without requiring timezone data.

mmastrac:
Fixes #1341

Split a `now` feature off from `clock` that enables `Utc::now()` only. This function is useful on its own, and can be included without the small-but-measurable overhead of adding timezone support.

For future compatibility, the feature `clock` implies the feature `now`. 

## Build matrix

A `now`-only build run was added to the matrix, though this might be overkill.

## Tests

This PR updates a few tests to ensure that `--features now` and `--feature clock` both continue work. I added a hidden `cfg` to one test to ensure that it can pass in both `clock` and `now` feature sets.

Tested the following builds successfully:

```
cargo test
cargo test --no-default-features --features=now
cargo test --no-default-features --features=clock
```


mmastrac:
One workflow run appears to be failing on a hashbrown dependency which doesn't make sense. I'll dig into it.

djc (approved):

djc:
Thanks!

PR #1351, opened 2023-11-04
Create the subsec_millis and subsec_micros methods for Duration

### LINKED ISSUE #1348: Add `subsec` methods to `Duration`
The standard library's `Duration` type has two useful methods:
- `subsec_micros`
- `subsec_millis`

This has the benefit of returning a `u32` instead of a `Option<i64>`. I suggest the same three methods be added to chrono 0.4. I see we already have `subsec_nanos` in the `0.4.x` branch, although it wasn't released yet. So this proposal just adds the two remaining methods.

Also, for consistency, should this method be called `subsec_nanoseconds` instead of `subsec_nanos`? That's what's used in `DateTime`.

botahamec:
Fixes #1348

### DIFF
--- a/src/duration.rs
+++ b/src/duration.rs
@@ -19,24 +19,10 @@ use std::error::Error;
 #[cfg(feature = "rkyv")]
 use rkyv::{Archive, Deserialize, Serialize};
 
-/// The number of nanoseconds in a microsecond.
-const NANOS_PER_MICRO: i32 = 1000;
-/// The number of nanoseconds in a millisecond.
-const NANOS_PER_MILLI: i32 = 1_000_000;
-/// The number of nanoseconds in seconds.
-const NANOS_PER_SEC: i32 = 1_000_000_000;
-/// The number of microseconds per second.
-const MICROS_PER_SEC: i64 = 1_000_000;
-/// The number of milliseconds per second.
-const MILLIS_PER_SEC: i64 = 1000;
-/// The number of seconds in a minute.
-const SECS_PER_MINUTE: i64 = 60;
-/// The number of seconds in an hour.
-const SECS_PER_HOUR: i64 = 3600;
-/// The number of (non-leap) seconds in days.
-const SECS_PER_DAY: i64 = 86_400;
-/// The number of (non-leap) seconds in a week.
-const SECS_PER_WEEK: i64 = 604_800;
+use crate::{
+    MICROS_PER_SEC, MILLIS_PER_SEC, NANOS_PER_MICRO, NANOS_PER_MILLI, NANOS_PER_SEC, SECS_PER_DAY,
+    SECS_PER_HOUR, SECS_PER_MINUTE, SECS_PER_WEEK,
+};
 
 macro_rules! try_opt {
     ($e:expr) => {
--- a/src/duration.rs
+++ b/src/duration.rs
@@ -232,6 +232,20 @@ impl Duration {
         }
     }
 
+    /// Returns the number of microseconds such that
+    /// `subsec_micros() + num_seconds() * MICROS_PER_SEC` is the total number of
+    /// microseconds in the duration.
+    pub const fn subsec_micros(&self) -> i32 {
+        self.subsec_nanos() / NANOS_PER_MICRO
+    }
+
+    /// Returns the number of milliseconds such that
+    /// `subsec_millis() + num_seconds() * MILLIS_PER_SEC` is the total number of
+    /// milliseconds in the duration.
+    pub const fn subsec_millis(&self) -> i32 {
+        self.subsec_nanos() / NANOS_PER_MILLI
+    }
+
     /// Returns the total number of whole milliseconds in the duration,
     pub const fn num_milliseconds(&self) -> i64 {
         // A proper Duration will not overflow, because MIN and MAX are defined
@@ -645,6 +659,33 @@ mod tests {
         );
         assert_eq!(Duration::days(i64::MAX / NANOS_PER_DAY + 1).num_nanoseconds(), None);
         assert_eq!(Duration::days(-i64::MAX / NANOS_PER_DAY - 1).num_nanoseconds(), None);
+    }
+
+    #[test]
+    fn test_duration_subsec_nanos() {
+        assert_eq!(Duration::zero().subsec_nanos(), 0);
+        assert_eq!(Duration::nanoseconds(1).subsec_nanos(), 1);
+        assert_eq!(Duration::nanoseconds(-1).subsec_nanos(), -1);
+        assert_eq!(Duration::seconds(1).subsec_nanos(), 0);
+        assert_eq!(Duration::nanoseconds(1_000_000_001).subsec_nanos(), 1);
+    }
+
+    #[test]
+    fn test_duration_subsec_micros() {
+        assert_eq!(Duration::zero().subsec_micros(), 0);
+        assert_eq!(Duration::microseconds(1).subsec_micros(), 1);
+        assert_eq!(Duration::microseconds(-1).subsec_micros(), -1);
+        assert_eq!(Duration::seconds(1).subsec_micros(), 0);
+        assert_eq!(Duration::microseconds(1_000_001).subsec_micros(), 1);
+    }
+
+    #[test]
+    fn test_duration_subsec_millis() {
+        assert_eq!(Duration::zero().subsec_millis(), 0);
+        assert_eq!(Duration::milliseconds(1).subsec_millis(), 1);
+        assert_eq!(Duration::milliseconds(-1).subsec_millis(), -1);
+        assert_eq!(Duration::seconds(1).subsec_millis(), 0);
+        assert_eq!(Duration::milliseconds(1_000_001).subsec_millis(), 1);
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


src/duration.rs
```diff
@@ -232,6 +232,20 @@ impl Duration {
         }
     }
 
+    /// Returns the number of nanoseconds such that
+    /// `subsec_micros() + num_seconds() * MICROS_PER_SEC` is the total number of
```
jtmoon79, line 236: > Returns the number of *nano*seconds ...

I would have guess the function name `subsec_micros` alluded to *micro*seconds.  This seems confusing to me.

src/duration.rs
```diff
@@ -232,6 +232,20 @@ impl Duration {
...
+    /// microseconds in the duration.
+    pub const fn subsec_micros(&self) -> i32 {
+        self.subsec_nanos() / NANOS_PER_MICRO
+    }
+
+    /// Returns the number of nanoseconds such that
```
jtmoon79, line 242: > Returns the number of *nano*seconds such that ...

I suspect this docstring and the docstring above are copy+pasta from `subsec_nanos`. 🤨

jtmoon79 (changes requested):

src/duration.rs
botahamec: Oof, yeah. I missed that part when I was reading through. Thanks for catching that

src/duration.rs
```diff
@@ -232,6 +232,20 @@ impl Duration {
...
 
+    /// Returns the number of microseconds such that
+    /// `subsec_micros() + num_seconds() * MICROS_PER_SEC` is the total number of
+    /// microseconds in the duration.
+    pub const fn subsec_micros(&self) -> i32 {
+        self.subsec_nanos() / NANOS_PER_MICRO
+    }
+
+    /// Returns the number of milliseconds such that
+    /// `subsec_millis() + num_seconds() * MILLIS_PER_SEC` is the total number of
```
jtmoon79, line 243: I like this specificity.

@pitdicker should `MILLIS_PER_SEC` be made `pub`?  If a user wants to follow this docstring implied suggestion (compute `subsec_millis() + num_seconds() * MILLIS_PER_SEC`) then they'll want to access `MILLIS_PER_SEC`.  This docstring is taken from `subsec_nanos` which has the same implied suggestion using `NANOS_PER_SEC`, so `NANOS_PER_SEC` should probably also be `pub`.

Stretch goal would be to make all the constants in that clump of constants above into `pub`.

src/duration.rs
```diff
@@ -232,6 +232,20 @@ impl Duration {
         }
     }
 
+    /// Returns the number of microseconds such that
+    /// `subsec_micros() + num_seconds() * MICROS_PER_SEC` is the total number of
```
jtmoon79, line 236: @pitdicker same question as below regarding `pub MICROS_PER_SEC`.

src/duration.rs
djc: I think it makes sense to make these constants part of the public API.

botahamec: A made a change like that to the docstrings. I'm wondering if those constants should stay in the `duration` module. They might make more sense if they were put in the `lib.rs`. I can't imagine anyone looking for these constants intuiting that they'd be in `duration`.

djc: Makes sense to me!

botahamec: Just made that change, but I do have one more question. Now that these are public, I find it a bit strange that some of them are `i32` and some are `i64`. I think this makes sense for what we're using them for, and if it's useful for us to have them that way, then it's probably useful for someone else too, but I wanted to bring it in case someone thinks we should make it more consistent.

djc (commented):
I'd like this split into clean commits.

* One for moving the constants (and making them `pub`)
* One for adding new methods (with complete docs and tests)
* One for improving docs on current methods

src/duration.rs
jtmoon79: I believe the goal for using `i32` is to lessen resource "footprint" chrono objects.

src/lib.rs
```diff
@@ -585,6 +585,25 @@ pub mod rkyv {
     pub use crate::weekday::ArchivedWeekday;
 }
 
+/// The number of nanoseconds in a microsecond.
+pub const NANOS_PER_MICRO: i32 = 1000;
+/// The number of nanoseconds in a millisecond.
+pub const NANOS_PER_MILLI: i32 = 1_000_000;
+/// The number of nanoseconds in seconds.
```
jtmoon79, line 592: nitpick:

    /// The number of nanoseconds in a second.

src/lib.rs
```diff
@@ -585,6 +585,25 @@ pub mod rkyv {
...
+pub const NANOS_PER_SEC: i32 = 1_000_000_000;
+/// The number of microseconds per second.
```
jtmoon79, line 594: nitpick:

    /// The number of microseconds in a second.

src/lib.rs
```diff
@@ -585,6 +585,25 @@ pub mod rkyv {
...
+pub const MICROS_PER_SEC: i64 = 1_000_000;
+/// The number of milliseconds per second.
```
jtmoon79, line 596: nitpick:

    /// The number of milliseconds in a second.

jtmoon79 (approved):
LGTM with nitpicks about inconsistent docstring

botahamec:
Sorry I took a while
@jtmoon79 I addressed your comments and rebased to make my commits match @djc's suggestion

src/lib.rs
```diff
@@ -585,6 +585,25 @@ pub mod rkyv {
...
 
+/// The number of nanoseconds in a microsecond.
+pub const NANOS_PER_MICRO: i32 = 1000;
+/// The number of nanoseconds in a millisecond.
+pub const NANOS_PER_MILLI: i32 = 1_000_000;
+/// The number of nanoseconds in a second.
+pub const NANOS_PER_SEC: i32 = 1_000_000_000;
+/// The number of microseconds in a second.
+pub const MICROS_PER_SEC: i64 = 1_000_000;
+/// The number of millisecondsin a second.
```
jtmoon79, line 596: typo

src/lib.rs
```diff
@@ -585,6 +585,25 @@ pub mod rkyv {
...
+pub const MILLIS_PER_SEC: i64 = 1000;
+/// The number of seconds in a minute.
+pub const SECS_PER_MINUTE: i64 = 60;
+/// The number of seconds in an hour.
+pub const SECS_PER_HOUR: i64 = 3600;
+/// The number of (non-leap) seconds in days.
```
jtmoon79, line 602: nitpick: _... seconds in a day._

jtmoon79 (approved):

jtmoon79 (approved):

src/lib.rs
botahamec: That's not a nitpick lol. Fixed

src/duration.rs
```diff
@@ -221,7 +221,9 @@ impl Duration {
         }
     }
 
-    /// Returns the number of nanoseconds such that
+    /// Returns the number of nanoseconds in the fractional part of the duration.
```
djc, line 210: These improvements for the new methods should be introduced in the same commit as the methods themselves, as I explained in my previous comment.

src/lib.rs
```diff
@@ -585,6 +585,25 @@ pub mod rkyv {
...
 }
 
+/// The number of nanoseconds in a microsecond.
+pub const NANOS_PER_MICRO: i32 = 1000;
+/// The number of nanoseconds in a millisecond.
+pub const NANOS_PER_MILLI: i32 = 1_000_000;
+/// The number of nanoseconds in a second.
+pub const NANOS_PER_SEC: i32 = 1_000_000_000;
+/// The number of microseconds in a second.
+pub const MICROS_PER_SEC: i64 = 1_000_000;
```
pitdicker, line 595: The types of these constants are 'whatever needs the fewest casts in our current code'.

That is: everything that has to do with seconds or more is using the type of the `Duration.secs` field.
Everything that has to do with subsecond units uses the type of the `Duration.nanos` field.
For most constants that seems like a good default.

`NANOS_PER_SEC` and `MICROS_PER_SEC` are at the boundary. `NANOS_PER_SEC` is currently an `i32`, and we sometimes cast it to an `i64`.

Maybe we should change `MICROS_PER_SEC` to also be an `i32`. Argument is consistency with `NANOS_PER_SEC`, and that widening casts are infallible and available with a `From` impl (even though this is just a constant and would not fail anyway.).

pitdicker:
@botahamec Do you want to push this PR over the finish line? Please adjust the commits a bit more matching @djc's request, and adjust the type of `MICROS_PER_SEC`.

Note that we will hopefully merge two PRs the next few days that touch surrounding code: https://github.com/chronotope/chrono/pull/1337 and https://github.com/chronotope/chrono/pull/1385. You may want to wait with rebasing until that is done.

pitdicker:
> Note that we will hopefully merge two PRs the next few days that touch surrounding code: #1337 and #1385. You may want to wait with rebasing until that is done.

They are merged now.

pitdicker:
@botahamec Can you let us now if you are still interested in working on this PR?

pitdicker:
Closing due to inactivity.

PR #1357, opened 2023-11-14
Add from_timestamp_nanos

Ali-Mirghasemi:
Add and implement `from_timestamp_nanos` and add unit test for it

### Known issue

if you input `i64::MAX` function give you a valid datetime, but for `i64::MIN` it gives you `None`
i think it's because time unit is nanoseconds

### DIFF
--- a/src/naive/datetime/mod.rs
+++ b/src/naive/datetime/mod.rs
@@ -13,7 +13,7 @@ use core::{fmt, str};
 #[cfg(feature = "rkyv")]
 use rkyv::{Archive, Deserialize, Serialize};
 
-use crate::duration::Duration as OldDuration;
+use crate::duration::{Duration as OldDuration, NANOS_PER_SEC};
 #[cfg(feature = "alloc")]
 use crate::format::DelayedFormat;
 use crate::format::{parse, parse_and_remainder, ParseError, ParseResult, Parsed, StrftimeItems};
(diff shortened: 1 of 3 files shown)


src/naive/datetime/mod.rs
```diff
@@ -197,6 +197,39 @@ impl NaiveDateTime {
...
+    /// // Negative timestamps (before the UNIX epoch) are supported as well.
+    /// let timestamp_nanos: i64 = -2208936075_000_000_000; //Mon Jan 01 1900 14:38:45 GMT+0000
+    /// let naive_datetime = NaiveDateTime::from_timestamp_nanos(timestamp_nanos);
+    /// assert!(naive_datetime.is_some());
+    /// assert_eq!(timestamp_nanos, naive_datetime.unwrap().timestamp_nanos_opt().unwrap());
+    /// ```
+    #[inline]
+    #[must_use]
+    pub const fn from_timestamp_nanos(nanos: i64) -> Option<NaiveDateTime> {
+        let secs = nanos.div_euclid(1_000_000_000);
```
djc, line 227: This should probably reuse the `NANOS_PER_SEC` `const` that we have defined somewhere.

Ali-Mirghasemi: `NANOS_PER_SEC` is in duration.rs mod and is inaccessible for me to use it, the solution is make them public for use

jtmoon79: `NANOS_PER_SEC` will be `pub` after #1351

djc: You can make it `pub(crate)` for this change.

Ali-Mirghasemi: So which one should i use for more compatibility with new commits?
- add `pub(crate)` in `duration.rs`
or
- make them `pub`

djc: `pub(crate)`, as said in my previous comment.

djc (approved):

djc:
Thanks!

PR #1361, opened 2023-11-20
Update windows-bindgen to 0.52

djc:


djc:
Merged #1360 instead.

PR #1368, opened 2023-12-12
Expose rkyv features as features for chrono users.

gz:
rkyv by default serializes usize as u32. This isn't ideal on most modern platforms and unfortunately is configured through a feature flag.

If we just set `default-features = false` in the rkyv Cargo dependency, the crate fails to compile because all the size features are mutually exclusive. On the other hand if we want to e.g., change the serialization of usize to 64-bit and we also want to use chrono this currently fails to compile because chrono always enables rkyv/size_32.

This re-exports the relevant rkyv features so users can choose which serialization to enable. The approach is similar to what the ordered-float crate does:

https://github.com/reem/rust-ordered-float/blob/8111b345372632893af0b8aa12152f3dc7278aba/Cargo.toml#L37

Note aside: rkyv with `default-features = false is what was recommended on the rkyv discord by the maintainer (see screenshot below). But it seems that the approach ordered-float took is better as it doesn't break backward compatibility (and compilation of chrono library without specifying any features).

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ unstable-locales = ["pure-rust-locales"]
 rkyv-16 = ["rkyv", "rkyv?/size_16"]
 rkyv-32 = ["rkyv", "rkyv?/size_32"]
 rkyv-64 = ["rkyv", "rkyv?/size_64"]
-rkyv-validation = ["rkyv?/validation"]
+rkyv-validation = ["rkyv", "rkyv?/validation"]
 # Features for internal use only:
 __internal_bench = []
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ rust-version = "1.61.0"
 name = "chrono"
 
 [features]
+# Don't forget to adjust `ALL_NON_EXCLUSIVE_FEATURES` in CI scripts when adding a feature or an optional dependency.
 default = ["clock", "std", "oldtime", "wasmbind"]
 alloc = []
 libc = []
@@ -27,7 +28,12 @@ now = ["std"]
 oldtime = []
 wasmbind = ["wasm-bindgen", "js-sys"]
 unstable-locales = ["pure-rust-locales"]
-rkyv-validation = ["rkyv/validation"]
+# Note that rkyv-16, rkyv-32, and rkyv-64 are mutually exclusive.
+rkyv-16 = ["rkyv", "rkyv?/size_16"]
+rkyv-32 = ["rkyv", "rkyv?/size_32"]
+rkyv-64 = ["rkyv", "rkyv?/size_64"]
+rkyv-validation = ["rkyv?/validation"]
+# Features for internal use only:
 __internal_bench = []
 
 [dependencies]
@@ -35,7 +41,7 @@ num-traits = { version = "0.2", default-features = false }
 rustc-serialize = { version = "0.3.20", optional = true }
 serde = { version = "1.0.99", default-features = false, optional = true }
 pure-rust-locales = { version = "0.7", optional = true }
-rkyv = { version = "0.7.41", optional = true }
+rkyv = { version = "0.7.43", optional = true, default-features = false }
 arbitrary = { version = "1.0.0", features = ["derive"], optional = true }
 
 [target.'cfg(all(target_arch = "wasm32", not(any(target_os = "emscripten", target_os = "wasi"))))'.dependencies]
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -1,4 +1,9 @@
 name: codecov
+
+env:
+  # It's really `--all-features`, but not adding the mutually exclusive features from rkyv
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -26,11 +26,16 @@ clock = ["std", "winapi", "iana-time-zone", "android-tzdata"]
 wasmbind = ["wasm-bindgen", "js-sys"]
 unstable-locales = ["pure-rust-locales"]
 __internal_bench = []
+rkyv     = ["rkyv_32"]
```
djc, line 29: Nits: would prefer to elide the fancy formatting with extra spaces and use `-` instead of `_` for chrono's features. Please rename `_ck` to `-validation`.

src/datetime/mod.rs
```diff
@@ -28,7 +28,7 @@ use crate::offset::Local;
 use crate::offset::{FixedOffset, Offset, TimeZone, Utc};
 use crate::{Datelike, Months, TimeDelta, Timelike, Weekday};
 
-#[cfg(feature = "rkyv")]
+#[cfg(any(feature = "rkyv_16", feature = "rkyv_32", feature = "rkyv_64"))]
```
djc, line 31: This seems fairly ugly to spread all over the place. I propose that we instead use an `rkyv-any` (open to better naming suggestions) feature that is enabled by all the size features (add a comment that it is for internal use only, maybe?).

djc (approved):
This makes sense to me, thanks!

If you want to use this any time soon, I recommend retargeting the PR/rebasing the branch on 0.4.x.

Cargo.toml
gz: Fixed

src/datetime/mod.rs
gz: Fixed with an internal feature named `__internal-rkyv` since there was already a feature `__internal-bench`.

gz:
Thanks for the review, I updated this and also made some changes to CI scripts to account for the fact that --all-features no longer works out of the box in a new commit.

gz:
> If you want to use this any time soon, I recommend retargeting the PR/rebasing the branch on 0.4.x.

Is there a timeline for the 0.5 release? I'm happy to send a PR for 0.4 if you think the 0.5 release is still a few months out.

gz:
Looks like the last test-failure I get is

```
Run cargo test --lib \
error: failed to parse manifest at `/home/runner/work/chrono/chrono/Cargo.toml`

Caused by:
  namespaced features with the `dep:` prefix are only allowed on the nightly channel and requires the `-Z namespaced-features` flag on the command-line
Error: Process completed with exit code 101.
```

and it's due to https://github.com/chronotope/chrono/pull/1368/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R37

I added it due to the suggestion here:

> I propose that we instead use an rkyv-any (open to better naming suggestions) feature that is enabled by all the size features (add a comment that it is for internal use only, maybe?).

I'm not quite sure if it's possible to specify the feature with the same intent without using `dep:`? The alternatives are to go back to the uglier `any(f1, f2, f3)` approach or raise msrv to 1.60 (which is when this was stabilized). Any preferences on how to proceed?

djc:
> Is there a timeline for the 0.5 release? I'm happy to send a PR for 0.4 if you think the 0.5 release is still a few months out.

There is not, you definitely want to be on 0.4.x for now (and we do regular merges from 0.4.x -> main).

djc:
> I'm not quite sure if it's possible to specify the feature with the same intent without using `dep:`? The alternatives are to go back to the uglier `any(f1, f2, f3)` approach or raise msrv to 1.60 (which is when this was stabilized). Any preferences on how to proceed?

I think it is possible, just replace `dep:rkyv` with `rkyv`.

gz:
>I think it is possible, just replace dep:rkyv with rkyv.

I made this change here: https://github.com/chronotope/chrono/pull/1368/commits/8a8b9e001674c9dd3f7e3fb076fa6c12a943dfb3

Not that I don't think it's fully backwards compatible due to `rkyv` now having `default-features = false`

e.g., I believe that before if someone specified just `rkyv` as a feature flag it would compile with rkyv-32 enabled.

Now `cargo build --features rkyv` says:

```
error: Package `chrono v0.5.0-alpha.1 (/home/gz/workspace/chrono)` does not have feature `rkyv`. It has an optional dependency with that name, but that dependency uses the "dep:" syntax in the features table, so it does not have an implicit feature with that name.
```

but `cargo build --features rkyv-64`, `cargo build --features rkyv-32` etc. works as expected. Is this what you had in mind?

gz:
@djc I rebased it on the 0.4 branch

djc:
Thanks!

pitdicker:
This PR was a breaking change, chrono no longer builds with just `features = ["rkyv"]` as reported in https://github.com/chronotope/chrono/issues/1380.

This breaks our build on docs.rs for 0.4.32, although that is easy to fix.

PR #1373, opened 2024-01-08
Add Months::num_months() and num_years()

danwilliams:
# Summary of changes

The changes in this PR are extremely simple and trivial: to present two new functions that provide general ease and compatibility in using `Months`, in similar fashion to `Duration`.

The problem being solved is that the `Months` type is a one-way black box. Once created, it is impossible to verify what it contains. Although its primary purpose is to provide a vector for effecting changes, without being able to inspect its value it is impossible to verify in tests, or support in a wider context. Additionally, extending the type from an external perspective will not help with this.

Therefore, this PR simply adds two new methods to `Months`:

  - `Months::num_months()` - This returns the number of months represented by the type, and is the primary objective of the PR.
  - `Months::num_years()` - This is a convenience function that is added to fit with the functions provided elsewhere.

The naming of the functions follows those provided by `Duration`, as this is the most similar type.

# Questions and points of note

## Constants

There is no readily-available `MONTHS_PER_YEAR` or similar - there is one in `src/offset/local/tz_info/rule.rs`, but nothing usable. Please advise if one should be added instead of using the magic number `12` - in which case, should it be local to the module, and private, or would Chrono benefit from a more public one?

## Annotations

Observing the code elsewhere, for instance in `Duration`, some functions are annotated as `inline`, and some as `must_use`, but not entirely consistently. In an attempt to fit with the codebase as well as possible, the two new functions have been annotated as `inline`, but not as `must_use`. Please advise as to what the general Chrono preference is in this regard.

## Tests

Tests have been added to try to satisfy what seems sensible and appropriate to check, in a manner that fits in with the other Chrono tests. Hopefully they have been placed in an acceptable location.

## Branch point

The branch point chosen has been the latest tagged release, with the PR targeting the 0.4.x branch. Hopefully that is correct.

### DIFF
--- a/src/month.rs
+++ b/src/month.rs
@@ -230,14 +230,8 @@ impl Months {
 
     /// Returns the total number of months in the `Months` instance.
     #[inline]
-    pub const fn num_months(&self) -> u32 {
+    pub const fn as_u32(&self) -> u32 {
         self.0
-    }
-
-    /// Returns the total number of whole years in the `Months` instance.
-    #[inline]
-    pub const fn num_years(&self) -> u32 {
-        self.0 / 12
     }
 }
 
@@ -366,21 +360,10 @@ mod tests {
     }
 
     #[test]
-    fn test_months_num_months() {
-        assert_eq!(Months::new(0).num_months(), 0);
-        assert_eq!(Months::new(1).num_months(), 1);
-        assert_eq!(Months::new(u32::MAX).num_months(), u32::MAX);
-    }
-
-    #[test]
-    fn test_months_num_years() {
-        assert_eq!(Months::new(0).num_years(), 0);
-        assert_eq!(Months::new(1).num_years(), 0);
-        assert_eq!(Months::new(11).num_years(), 0);
-        assert_eq!(Months::new(12).num_years(), 1);
-        assert_eq!(Months::new(23).num_years(), 1);
-        assert_eq!(Months::new(24).num_years(), 2);
-        assert_eq!(Months::new(u32::MAX).num_years(), u32::MAX / 12);
+    fn test_months_as_u32() {
+        assert_eq!(Months::new(0).as_u32(), 0);
+        assert_eq!(Months::new(1).as_u32(), 1);
+        assert_eq!(Months::new(u32::MAX).as_u32(), u32::MAX);
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


djc:
How are you using `Months` that you need this? I somewhat regret adding `Months` to the API and it's not obvious to me that it should be used in downstream code to the effect that these sorts of methods are needed.

danwilliams:
@djc I can understand that, but I'm not against `Months` existing, as it serves a purpose 🙂

The trouble is, as soon as you start using it (and it is the natural way to carry out calendar-based changes), you then need to be able to test and verify the code that's using it, and that starts to rely upon it. So anywhere that `Months` is used, needs to be able to verify the quantity that the instance represents. Otherwise, the type is not really usable.

I hope that helps...?

djc:
I don't think you've really answered the question of how/why you're using it?

danwilliams:
@djc apologies, I thought I had.

In code using Chrono at present, `Months` is being used where appropriate.

In some places we have code that emits `Months` as part of the general interface dealing with dates, times, durations, etc. - all based on Chrono.

In order to perform appropriate testing, it is necessary to confirm and validate what value of `Months` is being given to the instances in use.

This applies more broadly, as any type used for any purpose needs to be able to be inspected to the degree where its correctness is verifiable. Hence I had thought my previous explanation would help there. However, I trust this clarification has helped resolve my earlier oversight 🙂

djc:
Okay, I'll take an `as_u32()` method (and would prefer to avoid the added convenience method for now). The way forward here will be #1290.

danwilliams:
@djc no problem! I will amend and resubmit the PR 👍 

I took a quick look at #1291, and that looks interesting. Are you open to discussion on it? I have a complete and working implementation of Intervals, which include calendar-based durations, which I was considering adapting for potential submission to Chrono at some future point. I did not know work had already been done on something similar. From doing that I can see some drawbacks in the current implementation in #1291, which it might be worth discussing? Perhaps that's why development on the feature has stalled? If it's all in hand and no help is wanted, I will back politely away 🙂

djc:
Happy to take your feedback in #1290 (not #1291).

djc (approved):

djc:
Thanks!

Issue #264, opened 2018-07-14
Creating a DateTime from epoch milliseconds


nevi-me:
I've been able to convert a `DateTime` to milliseconds with `timestamp_millis()`, but I am struggling to find a trait member that allows me to get a DateTime with milliseconds.

Is it currently possible to do this?

evanxg852000:
Possible way
```rust
use chrono::{DateTime, TimeZone, NaiveDateTime, Utc};

fn main() {
    let time  = chrono::Utc::now();
    let ts_millis = time.timestamp_millis();
    println!("timestamp milli {} -> {}", time, ts_millis);

    let ts_secs = ts_millis / 1000;
    let ts_ns = (ts_millis % 1000) * 1_000_000 ;
    let dt = DateTime::<Utc>::from_utc(NaiveDateTime::from_timestamp(ts_secs, ts_ns as u32), Utc);

    println!("timestamp milli {} -> {}", dt, dt.timestamp_millis());

}
```

calebpitan:
Wouldn't it be better if it were `u64` as in `DateTime::from_timestamp_millis(millis: u64)` than `i64`?

PR #1374, opened 2024-01-09
Add from_timestamp_millis to DateTime<Utc>

xmakro:
As discussed in https://github.com/chronotope/chrono/issues/264#issuecomment-1882616604


djc:
Do you have an actual use case for the micros variant?

djc:
Okay, I would prefer to defer the addition of micros in that case until someone comes up with a use case for it. IME millis timestamps are much more common than micros timestamps since they're used a bunch in the JS ecosystem.

djc (approved):

djc:
Thanks!

Issue #1364, opened 2023-11-22
panic with_timezone Jan 1, 1601 00:00 UTC


davehorner:
I'm trying to do an ls \\.\pipe in nushell which causes a panic.
The ls \\.\pipe lists the pipes on windows and according to powershell that date is 12/31/1600 7:00PM.

https://github.com/nushell/nushell/issues/10464

I have some details in my PR which sets and unsets the panic handler around chrono to handle the panic thrown from Chrono.
https://github.com/nushell/nushell/pull/10558

Is there a way to make another interface/function which does not unwrap panic when called? 

wrapping chrono in panic hooks to handle panic'ing unwraps on Jan 1, 1601 00:00 UTC and other reasons unknown is not ideal and I would love to have a better solution.

```
fn unix_time_to_local_date_time(secs: i64) -> Option<DateTime<Local>> {
    let old_hook = std::panic::take_hook();
    std::panic::set_hook(Box::new(|_info| {}));
    let result = std::panic::catch_unwind(|| match Utc.timestamp_opt(secs, 0) {
        LocalResult::Single(t) => Some(t.with_timezone(&Local)),
        _ => None,
    });
    std::panic::set_hook(old_hook);
    match result {
        Ok(r) => r,
        _ => None,
    }
}
```

I'm not sure what chrono should tell me the timezone is in 1600, but I would prefer it not be via a panic.

djc:
Does this work better with #1017?

davehorner:
Are you suggesting to attempt to patch my system with that PR and attempt again?

djc:
Yup, I suggest you use a `[patch]` section (see https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section) to use a local clone of that branch to see if that helps.

davehorner:
I forked chrono, pulled in win_timezone, `git merge -X theirs win_timezone` on main and pushed it up to https://github.com/davehorner/chrono.git.

I tried to make it as hard on myself as possible and modified `.cargo\config.toml` to include:

```
[patch.crates-io]
chrono = { git = 'https://github.com/davehorner/chrono.git', branch="win_timezone" }
```

I did a ripgrep and ran through all the Cargo.toml that had a chrono dependency, placing the patch section in each.  I also added it to the Cargo.toml workspace root file.

```
warning: patch for the non root package will be ignored, specify patch at the workspace root:
package:   c:\w\rust\nushell\crates\nu-cli\Cargo.toml
[...]

PR #1377, opened 2024-01-18
Workaround for Windows local time prior to 1601

ChrisDenton:
This PR makes the minimum necessary changes to workaround #1364. It's not a full fix but it will allow people to use chrono to get the local time of files with a creation time of zero. I believe this to be worthwhile in the interim until a better patch lands.

If getting the offset for 1601 fails with `ERROR_INVALID_PARAMETER` then this patch instead calculates the offset for 1602. This works only because the 1600's are long enough ago that dynamic timezones aren't a concern and negative timezone offsets will always be much less than a year (typically a few hours).

### DIFF
--- a/src/datetime/tests.rs
+++ b/src/datetime/tests.rs
@@ -1267,6 +1267,17 @@ fn test_datetime_from_timestamp_millis() {
 
 #[test]
 #[cfg(feature = "clock")]
+fn test_datetime_before_windows_api_limits() {
+    // dt corresponds to `FILETIME = 147221225472` from issue 651.
+    // This used to fail on Windows for timezones with an offset of -5:00 or greater.
+    // The API limits years to 1601..=30827.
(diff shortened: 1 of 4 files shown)


ChrisDenton:
On second thought, I've scoped it to only hit the fallback case on 1601-01-01 (UTC). This better expresses the intent and timezones offsets are always going to be less than 24hrs, so from 1601-01-02 onwards `ERROR_INVALID_PARAMETER` is a real error.

djc:
@pitdicker do you think you'll have time to rework #1017 sometime soonish?

pitdicker:
Yes, I'll get back to it this weekend.

pitdicker:
The complete fix in https://github.com/chronotope/chrono/pull/1017 is merged. @ChrisDenton Thank you for your efforts here.

PR #1383, opened 2024-01-24
Make `rkyv` feature imply `size_32`

### LINKED ISSUE #1380: Uhm build fails?
Something is wrong, the lint doesn't compile, can somebody check that please?
It's not fun to see red failing badges.

pitdicker:
Should fix the [build failure on docs.rs](https://docs.rs/crate/chrono/0.4.32), see https://github.com/chronotope/chrono/pull/1368#issuecomment-1908337824 and https://github.com/chronotope/chrono/issues/1380.

Fixes #1380.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,10 +29,11 @@ oldtime = []
 wasmbind = ["wasm-bindgen", "js-sys"]
 unstable-locales = ["pure-rust-locales"]
 # Note that rkyv-16, rkyv-32, and rkyv-64 are mutually exclusive.
-rkyv-16 = ["rkyv", "rkyv?/size_16"]
-rkyv-32 = ["rkyv", "rkyv?/size_32"]
-rkyv-64 = ["rkyv", "rkyv?/size_64"]
-rkyv-validation = ["rkyv", "rkyv?/validation"]
+rkyv = ["dep:rkyv", "rkyv/size_32"]
+rkyv-16 = ["dep:rkyv", "rkyv?/size_16"]
+rkyv-32 = ["dep:rkyv", "rkyv?/size_32"]
+rkyv-64 = ["dep:rkyv", "rkyv?/size_64"]
+rkyv-validation = ["rkyv?/validation"]
 # Features for internal use only:
 __internal_bench = []
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "chrono"
-version = "0.4.32"
+version = "0.4.33"
 description = "Date and time library for Rust"
 homepage = "https://github.com/chronotope/chrono"
 documentation = "https://docs.rs/chrono/"
--- a/src/month.rs
+++ b/src/month.rs
@@ -1,6 +1,6 @@
 use core::fmt;
 
-#[cfg(feature = "rkyv")]
+#[cfg(any(feature = "rkyv", feature = "rkyv-16", feature = "rkyv-32", feature = "rkyv-64"))]
 use rkyv::{Archive, Deserialize, Serialize};
 
 use crate::OutOfRange;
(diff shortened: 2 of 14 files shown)


djc:
We can write `rkyv = ['dep:rkyv', 'rkyv/size_32']` if our MSRV is new enough (which I think it is?).

pitdicker:
I'll give it a try.

djc (approved):
Nice!

pitdicker:
Now the feature `rkyv-validation` no longer implies the `rkyv` feature and fails to build. Let me give it one more round :smile:.

pitdicker:
Do we want to bump to 0.4.33 in this PR?

Cargo.toml
```diff
@@ -29,10 +29,11 @@ oldtime = []
 wasmbind = ["wasm-bindgen", "js-sys"]
 unstable-locales = ["pure-rust-locales"]
 # Note that rkyv-16, rkyv-32, and rkyv-64 are mutually exclusive.
-rkyv-16 = ["rkyv", "rkyv?/size_16"]
-rkyv-32 = ["rkyv", "rkyv?/size_32"]
-rkyv-64 = ["rkyv", "rkyv?/size_64"]
-rkyv-validation = ["rkyv", "rkyv?/validation"]
+rkyv = ["dep:rkyv", "rkyv/size_32"]
+rkyv-16 = ["dep:rkyv", "rkyv?/size_16"]
```
djc, line 33: BTW, I think this might be a pre-existing issue but I don't think we need the `?` here? AIUI it's meant to denote that the `size_16` feature should only be enabled if the `rkyv` dependency is enabled, which is always the case for these. Want to fix that up here?

Cargo.toml
```diff
@@ -29,10 +29,11 @@ oldtime = []
...
+rkyv-32 = ["dep:rkyv", "rkyv?/size_32"]
+rkyv-64 = ["dep:rkyv", "rkyv?/size_64"]
+rkyv-validation = ["dep:rkyv", "rkyv?/validation"]
```
djc, line 36: So I think this one is now odd because `rkyv-validation` doesn't pick a size? I'd propose that we leave `dep:rkyv` out of `rkyv-validation` and only leave `rkyv?/validation` in there so that it will only activate the downstream feature if any of the other features have been used to enable `rkyv` (and pick a size).

djc (commented):
> Do we want to bump to 0.4.33 in this PR?

Yes, please!

Cargo.toml
pitdicker: I tried both of your suggestions, but then it doesn't work out for all combinations:
```
cargo check --features=rkyv-64,rkyv-validation
    Checking rkyv v0.7.43
error: "size_32" and "size_64" are mutually-exclusive features. You may need to set `default-features = false` or compile with `--no-default-features`.
   --> C:\Users\paul\.cargo\registry\src\index.crates.io-6f17d22bba15001f\rkyv-0.7.43\src\macros.rs:127:1
    |
127 | / core::compile_error!(
128 | |     "\"size_32\" and \"size_64\" are mutually-exclusive features. You may need to set \
129 | |     `default-features = false` or compile with `--no-default-features`."
130 | | );
    | |_^

error: could not compile `rkyv` (lib) due to previous error
```

src/month.rs
```diff
@@ -30,9 +30,9 @@ use crate::OutOfRange;
 // Actual implementation is zero-indexed, API intended as 1-indexed for more intuitive behavior.
 #[derive(PartialEq, Eq, Copy, Clone, Debug, Hash, PartialOrd, Ord)]
 #[cfg_attr(feature = "rustc-serialize", derive(RustcEncodable, RustcDecodable))]
-#[cfg_attr(feature = "rkyv", derive(Archive, Deserialize, Serialize))]
 #[cfg_attr(
-    feature = "rkyv",
+    any(feature = "rkyv", feature = "rkyv-16", feature = "rkyv-32", feature = "rkyv-64"),
+    derive(Archive, Deserialize, Serialize),
```
pitdicker, line 35: Sorry, you reviewed right when I made this change. Learned today that multiple attributes can be combined in the same `cfg_attr`.

pitdicker:
Set the date to today. Do we want to get https://github.com/chronotope/chrono/pull/1384 in first?

Cargo.toml
djc: I think we can make do without the `validation` -> `dep:rkyv` edge? It seems to work for me.

src/month.rs
djc: Nice!

Cargo.toml
pitdicker: 👍 Done

djc:
Published to crates.io, thanks!

pitdicker:
That is quick. You too!

pitdicker:
And we have docs again https://docs.rs/chrono/0.4.33/chrono/

PR #1388, opened 2024-01-26
Specify branch for GitHub Actions badge and fix link

pitdicker:
I think it makes more sense for the GitHub Actions badge to match the latest run on the branch that that particular readme is on.
For the main branch I'll make it point to the `main` with the next merge.

This also fixes the link, that seems to have been broken for quite a while.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,8 @@
 [![Chat][discord-image]][discord]
 [![codecov.io][codecov-img]][codecov-link]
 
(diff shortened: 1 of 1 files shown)


pitdicker:
@djc Just saw your comment on switching the main branch. If we go that route we may only want to fix the link here.

djc (approved):

djc:
See #1390, so this might need some changes?

djc (approved):

PR #1394, opened 2024-01-28
0.5.x: Remove everything deprecated

pitdicker:
Remove everything that is deprecated on the 0.5 branch.

I was wondering if it was better to rename the replacement methods at the same time as removing the deprecated methods, or to leave that for another PR. I.e. renaming things like `FixedOffset::east_opt` to `FixedOffset::east`.

### DIFF
--- a/bench/benches/chrono.rs
+++ b/bench/benches/chrono.rs
@@ -37,7 +37,7 @@ fn bench_datetime_from_str(c: &mut Criterion) {
 }
 
 fn bench_datetime_to_rfc2822(c: &mut Criterion) {
-    let pst = FixedOffset::east_opt(8 * 60 * 60).unwrap();
+    let pst = FixedOffset::east(8 * 60 * 60).unwrap();
     let dt = pst
         .from_local_datetime(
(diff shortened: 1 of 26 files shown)


pitdicker:
We may want to backport the first two commit (documentation and test fixes) to the 0.4 branch.

djc:
> We may want to backport the first two commits (documentation and test fixes) to the 0.4 branch.

That sounds good to me.

djc:
> I was wondering if it was better to rename the replacement methods at the same time as removing the deprecated methods, or to leave that for another PR. I.e. renaming things like `FixedOffset::east_opt` to `FixedOffset::east`.

Yes, I think we should do that in the same commit.

pitdicker:
Because the renames are touching so many lines I split them out in separate commits. Easy to squash if you do prefer them combined though.

I don't really want to make a lot of changes to the file for `NaiveDate` because I think it would create a messy merge conflict the open PR to refactor that file. So I skipped the renames there for now.

djc (approved):

PR #1400, opened 2024-02-02
Make methods on `DateTime` const where possible

pitdicker:
Since rust 1.61 it is possible to make these methods const.

In theory we should be careful with making methods on `DateTime` const. If we discover a method need to change and use something from the `TimeZone` trait it can no longer be const, which would be a breaking change.

But all the methods that can be made const in this PR clearly operate on UTC values, or return just a field of the `DateTime` type.

This completes the work to make the API const were possible :tada:.


djc (approved):

pitdicker:
@djc Thank you for all the reviews the last days!

PR #1403, opened 2024-02-02
Remove check for DurationExceedsTimestamp

### LINKED ISSUE #1375: Truncating a timestamp close to EPOCH fails
Due to this check in `duration_truc`:
https://github.com/chronotope/chrono/blob/main/src/round.rs#L221-L223

Expected behaviour: I should be able to truncate the timestamp to 0

joroKr21:
This check is not necessary and prevented rounding and truncation from working correctly on timestamps close to the Unix epoch.

fixes #1375 

### DIFF
--- a/src/round.rs
+++ b/src/round.rs
@@ -758,23 +758,40 @@ mod tests {
 
     #[test]
     fn test_duration_trunc_close_to_epoch() {
-        let dt = NaiveDate::from_ymd_opt(1970, 1, 1).unwrap().and_hms_opt(0, 0, 15).unwrap();
         let span = TimeDelta::minutes(15);
+
+        let dt = NaiveDate::from_ymd_opt(1970, 1, 1).unwrap().and_hms_opt(0, 0, 15).unwrap();
         assert_eq!(dt.duration_trunc(span).unwrap().to_string(), "1970-01-01 00:00:00");
 
         let dt = NaiveDate::from_ymd_opt(1969, 12, 31).unwrap().and_hms_opt(23, 59, 45).unwrap();
-        let span = TimeDelta::minutes(15);
         assert_eq!(dt.duration_trunc(span).unwrap().to_string(), "1969-12-31 23:45:00");
     }
 
     #[test]
     fn test_duration_round_close_to_epoch() {
-        let dt = NaiveDate::from_ymd_opt(1970, 1, 1).unwrap().and_hms_opt(0, 0, 15).unwrap();
         let span = TimeDelta::minutes(15);
+
+        let dt = NaiveDate::from_ymd_opt(1970, 1, 1).unwrap().and_hms_opt(0, 0, 15).unwrap();
         assert_eq!(dt.duration_round(span).unwrap().to_string(), "1970-01-01 00:00:00");
 
         let dt = NaiveDate::from_ymd_opt(1969, 12, 31).unwrap().and_hms_opt(23, 59, 45).unwrap();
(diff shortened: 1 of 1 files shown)


pitdicker:
Thank you @joroKr21. I'll have to think this through.

src/round.rs
```diff
@@ -239,19 +233,6 @@ where
...
-    /// ``` rust
-    /// # use chrono::{DurationRound, Duration, RoundingError, TimeZone, Utc};
-    /// let dt = Utc.with_ymd_and_hms(1970, 12, 12, 0, 0, 0).unwrap();
-    ///
-    /// assert_eq!(
-    ///     dt.duration_round(Duration::days(365)),
-    ///     Err(RoundingError::DurationExceedsTimestamp),
-    /// );
-    /// ```
-    DurationExceedsTimestamp,
```
pitdicker, line 253: This enum is part of our public API, so `DurationExceedsTimestamp` can't be removed. That would be a breaking change.

src/round.rs
```diff
@@ -181,9 +181,6 @@ where
             return Err(RoundingError::DurationExceedsLimit);
         }
         let stamp = naive.timestamp_nanos_opt().ok_or(RoundingError::TimestampExceedsLimit)?;
-        if span > stamp.abs() {
-            return Err(RoundingError::DurationExceedsTimestamp);
-        }
```
pitdicker, line 186: If you remove this check, I expect it will introduce a panic if the date is large and the duration is even larger. Say a date in the year 200.000 and rounding to a duration of 300.000 years.

Could you add a test for this case, and do something to avoid the panic?

src/round.rs
pitdicker: I would only update the doc of `DurationExceedsTimestamp`.

pitdicker (approved):

src/round.rs
joroKr21: I couldn't find any. Added a test for timestamps that are close to the min / max of nanosecond precision

pitdicker: Thank you 👍. @djc will also want to have a look, but it seems good to merge for me.

djc (approved):
Just needs a rebase.

joroKr21:
Looks like CI is failing on something unrelated

pitdicker:
All CI failures are with the reason:
```
 error[E0635]: unknown feature `stdsimd`
  --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ahash-0.7.7/src/lib.rs:33:42
   |
33 | #![cfg_attr(feature = "stdsimd", feature(stdsimd))]
   |                                          ^^^^^^^
```

Curious. In any case unrelated to your work.

pitdicker:
We don't seem to be the only one https://github.com/thin-edge/thin-edge.io/pull/2687.

djkoloski:
rkyv 0.7 will not be updating to a newer version of hashbrown because it would be a server breaking change. 0.8 won't be out for a while either. I did create a branch of 0.7 that updates hashbrown yesterday (`0.7-hashbrown-0.14`) so you could change your dependency to a git dependency for nightly CI testing if that would help. Unfortunately I don't know of a better solution to this problem besides hashbrown releasing a fixed 0.12 version without `stdsimd`.

djc:
There's also a PR to request an update to ahash 0.7: https://github.com/tkaitchuck/aHash/pull/201.

PR #1405, opened 2024-02-02
Fixes for `Duration::new`

pitdicker:
Thanks to @Skgland in https://github.com/chronotope/chrono/pull/1337#discussion_r1475695141 for noticing.

### DIFF
--- a/src/duration.rs
+++ b/src/duration.rs
@@ -83,7 +83,7 @@ impl Duration {
     pub const fn new(secs: i64, nanos: u32) -> Option<Duration> {
         if secs < MIN.secs
             || secs > MAX.secs
-            || nanos > 1_000_000_000
+            || nanos >= 1_000_000_000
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
If we're going to make this public, should we instead do a div_mod for nanos and add the modulo to the seconds?

pitdicker:
> If we're going to make this public, should we instead do a div_mod for nanos and add the modulo to the seconds?

I like that that it now is just the most basic way to set the fields of a `Duration`.

[`std::time::Duration::new`](https://doc.rust-lang.org/std/time/struct.Duration.html#method.new) does carry over nanoseconds beyond 1.000.000.000 to seconds. That makes it fallible:

> #### Panics
>
> This constructor will panic if the carry from the nanoseconds overflows the seconds counter.

PR #1406, opened 2024-02-02
Rename `Duration` to `TimeDelta`, add type alias

pitdicker:
We have discussed something like this many times. Reasons to rename `Duration` to `TimeDelta`:

* A `Duration` is usually understood to be strictly positive. Our type can be both positive and negative. That makes it more flexible, which is a good thing for a date and time library. The name `TimeDelta` can better represent a signed value.
* The standard library made its `Duration` type unsigned, as that is a better match for system APIs. We want to avoid confusion with that type.
* The name `OldDuration` we currently use as an alias in our documentation to ease the confusion a little is just sad (and not really helping).
* On our 0.5 branch we have already renamed it to `TimeDelta`, and we expect users to eventually migrate. By making the name available now we can ease the migration a little when the time comes.
* The name difference between our 0.4 and 0.5 branches is my biggest source of merge conflicts. It would make me happy if we can make them the same.

The old name `Duration` will remain available as a type alias for `TimeDelta`, so this is not a breaking change.


pitdicker (commented):
Did a review of all lines. It was not a trivial find-and-replace, but it seems like I didn't mangle things.

djc:
You refer to `TimeSpan` a number of times in your OP. Which is it -- `TimeDelta` or `TimeSpan`?

pitdicker:
> You refer to `TimeSpan` a number of times in your OP. Which is it -- `TimeDelta` or `TimeSpan`?

`TimeDelta`. Working on chrono makes me feel dumb surprisingly often.

pitdicker:
There was a little discussion on possible names in https://github.com/chronotope/chrono/pull/1137#issuecomment-1627612073. I don't care deeply about the names, but `TimeDelta` and `TimeSpan` have my preference. It seems you and @esheppa had a preference for `TimeDelta` as that is what is already on the 0.5 branch.

src/lib.rs
```diff
@@ -467,6 +467,8 @@ extern crate alloc;
 
 mod duration;
 pub use duration::Duration;
+/// Alias of [`Duration`]
```
djc: This should be formatted as a type, that is, with empty lines around it and not between a bunch of imports.

src/lib.rs
```diff
@@ -52,23 +52,18 @@
...
-//! months.
-//!
-//! Chrono does not yet natively support
-//! the standard [`Duration`](https://doc.rust-lang.org/std/time/struct.Duration.html) type,
-//! but it will be supported in the future.
-//! Meanwhile you can convert between two types with
-//! [`Duration::from_std`](https://docs.rs/time/0.1.40/time/struct.Duration.html#method.from_std)
-//! and
-//! [`Duration::to_std`](https://docs.rs/time/0.1.40/time/struct.Duration.html#method.to_std)
+//! ### Time span / Duration
```
djc, line 55: Let's use "Time delta" in the title here?

djc (approved):
Okay -- do we want to deprecate the `Duration` alias, and if so, when?

pitdicker:
Thank you! Looking forward to less merge conflicts (going to introduce plenty of new ones though now that I'm starting to work on the 0.5 branch).

> Okay -- do we want to deprecate the `Duration` alias, and if so, when?

I feel that deprecating the type would cause a lot of churn while bringing little benefit. New code can start using `TimeDelta` and be compatible with 0.5. But personally I'd leave it available until 0.5.

PR #1411, opened 2024-02-04
Remove partial support for handling `-0000` offset

pitdicker:
RFC 2822 and others have a special encoding to indicate the offset of a datetime is unknown: `-00:00`.

We currently have code in `format::parse` that ensures we do not provide an offset to the `Parsed` struct if the offset is `-0000`.
`timezone_offset_2822` returns an `Option<i32>` that should be `None` to differentiate this case from `+0000`.

But then `timezone_offset_2822` does return `Some(0)` on the `-0000` input. That makes the `Option` and the code to handle it useless.

I would like to see proper support for encoding this value in a `FixedOffset`, as in https://github.com/chronotope/chrono/pull/1042.

But for now the inconsistency just messes up the error variants we return.
Because I am currently working on converting the parsing code from `ParsedError` to a new `Error` type for 0.5, I just need something consistent to work on.

### DIFF
--- a/src/format/parse.rs
+++ b/src/format/parse.rs
@@ -216,7 +216,8 @@ pub(crate) fn parse_rfc3339<'a>(parsed: &mut Parsed, mut s: &'a str) -> ParseRes
     }
 
     let offset = try_consume!(scan::timezone_offset(s, |s| scan::char(s, b':'), true, false, true));
-    if offset <= -86_400 || offset >= 86_400 {
+    const MAX_RFC3339_OFFSET: i32 = 86_400;
+    if !(-MAX_RFC3339_OFFSET..=MAX_RFC3339_OFFSET).contains(&offset) {
         return Err(OUT_OF_RANGE);
     }
     parsed.set_offset(i64::from(offset))?;
--- a/src/format/parse.rs
+++ b/src/format/parse.rs
@@ -142,10 +142,7 @@ fn parse_rfc2822<'a>(parsed: &mut Parsed, mut s: &'a str) -> ParseResult<(&'a st
     }
 
     s = scan::space(s)?; // mandatory
-    if let Some(offset) = try_consume!(scan::timezone_offset_2822(s)) {
-        // only set the offset when it is definitely known (i.e. not `-0000`)
-        parsed.set_offset(i64::from(offset))?;
-    }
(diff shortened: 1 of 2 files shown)


pitdicker:
Tacked on a commit to return the correct error from `parse_rfc3339_relaxed`.

It was returning `NOT_ENOUGH` ('Given set of fields is not enough to make a requested date and time value'), while it should be `TOO_SHORT` ('The input string has been prematurely ended).

src/format/parse.rs
```diff
@@ -216,7 +216,11 @@ pub(crate) fn parse_rfc3339<'a>(parsed: &mut Parsed, mut s: &'a str) -> ParseRes
     }
 
     let offset = try_consume!(scan::timezone_offset(s, |s| scan::char(s, b':'), true, false, true));
-    if offset <= -86_400 || offset >= 86_400 {
+    // This range check is similar to the one in `FixedOffset::east_opt`, so it would be redundant.
+    // But it is possible to read the offset directly from `Parsed`. We want to only successfully
+    // populate `Parsed` if the input is fully valid RFC 3339.
+    const MAX_OFFSET: i32 = 23 * 3600 + 59 * 60;
```
djc, line 219: I'd like for this to be two commits:

* One for using a constant and the reformulated `Range::contains()`
* One for changing the value from 86400 to 863400

I also think `23 * 3600 + 59 * 60` is a surprising way to express this? I think (24 * 60 - 1) * 60 might be clearer.

djc (approved):

src/format/parse.rs
pitdicker: > I also think `23 * 3600 + 59 * 60` is a surprising way to express this? I think (24 * 60 - 1) * 60 might be clearer.

It comes from the definition in RFC 3339 as copied in comment of this function:
```
// time-hour = 2DIGIT ; 00-23
// time-numoffset = ("+" / "-") time-hour ":" time-minute
// time-minute = 2DIGIT ; 00-59
// time-offset = "Z" / time-numoffset
```

djc: Right. (23 * 60 + 59) * 60 could also work, I guess, but I find the previous suggestion more intuitive. I'll leave it up to you.

pitdicker: Done. Added 'Max for the hours field is `23`, and for the minutes field `59`.' as a comment.

PR #1413, opened 2024-02-05
support to use dot as timezone offset separator

biuiko:
# Summary of changes
This PR adds support for using a dot `.` as a separator for timezone offset, like `+9.5` means `+09:30`. Currently, the library only supports using a colon `:` as the separator for offset, but some users prefer using a dot for consistency with other time-related formats.

# Changes
To achieve this, I have made the following changes:
- update parsing logic: add function `scan::timezone_offset_with_dot_separator` to calculate offset
- add specifier `%.z` to formatting and parsing

### DIFF
--- a/src/format/formatting.rs
+++ b/src/format/formatting.rs
@@ -22,10 +22,10 @@ use crate::{NaiveDate, NaiveTime, Weekday};
 use super::locales;
 #[cfg(all(feature = "unstable-locales", feature = "alloc"))]
 use super::Locale;
-#[cfg(any(feature = "alloc", feature = "serde", feature = "rustc-serialize"))]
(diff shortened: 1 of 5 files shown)


pitdicker:
Fractional hours as timezone offset. That is a terrible format. Where did you encounter it?

pitdicker:
Thank you @biuiko for contributing to chrono.

ISO 8601 is the prime example of a format that supports fractional values for components of a time or duration. Both are not (yet) supported by chrono. But even that format does not suggest using fractional values for an offset.

In https://github.com/chronotope/chrono/pull/1083 I already had to make our parsing code for offsets quite a bit more complex to handle real-world cases. (Should yet that ready for review again...) Adding this as another piece of complexity needs a very good use case, such as a common library that generates such strings or some standard.

PR #1420, opened 2024-02-09
Update to pure-rust-locales 0.8.0

### LINKED ISSUE #1349: Time formatting on en_US locale with '%X' not expected
Hi all, 
I am trying to format some dates, based on the user's `locale`, therefore I am using the `localized` base functions. However, when doing a `Display` of date, I get unexpected result, or maybe I am missing something? 

`platform: linux` (ubuntu 20.04)
`chrono version: 0.4.31`

```rust
use chrono::prelude::*;
fn main() {
    let locale_nl = Locale::nl_NL;
    let locale_us = Locale::en_US;
    let local: DateTime<Local> = Local::now();

    let fmt_time_nl = local.format_localized("%X", locale_nl);
    let fmt_time_us = local.format_localized("%X", locale_us);
    let fmt_date_nl = local.format_localized("%x", locale_nl);
    let fmt_date_us = local.format_localized("%x", locale_us);
    println!("Time NL: {}", fmt_time_nl);
    println!("Time US: {}", fmt_time_us);
    println!("Date NL: {}", fmt_date_nl);
    println!("Date US: {}", fmt_date_us);

    ....
    
}


```
Yields the following output:
```
Time NL: 20:34:51
Time US: 08
Date NL: 02-11-23
Date US: 11/02/2023

```
Thus, date seems fine, but not `time`.

I saw that I could do '%r' for 12 hour clock time(us locale), which would yield the correct result, and for `nl_NL` it would fallback to `%X`, but `Display`ing just `08` as time seems unexpected. I would expect to also get `mins` and `secs`, not only `h`.

Thanks for your time!

pitdicker:
Included a test for the fixed formatting of `%X` in some of the affected locales.

Fixes #1349.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ __internal_bench = []
 num-traits = { version = "0.2", default-features = false }
 rustc-serialize = { version = "0.3.20", optional = true }
 serde = { version = "1.0.99", default-features = false, optional = true }
-pure-rust-locales = { version = "0.7", optional = true }
+pure-rust-locales = { version = "0.8", optional = true }
 rkyv = { version = "0.7.43", optional = true, default-features = false }
 arbitrary = { version = "1.0.0", features = ["derive"], optional = true }
 
--- a/src/format/strftime.rs
+++ b/src/format/strftime.rs
@@ -905,6 +905,22 @@ mod tests {
     }
 
     #[test]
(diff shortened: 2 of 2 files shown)


djc (approved):

pitdicker:
Thank you for all your reviews today!

Issue #1049, opened 2023-05-06
Converting the API to return Results


pitdicker:
There are now multiple issues or comments to make chrono 0.5 return `Result`s instead of `Option`s, and panic less. There is some design work that needs to be done first. I don't see any concrete, high-level proposals. Maybe this issue can serve for exploration.

See https://github.com/chronotope/chrono/issues/815

cc @Zomtir, @udoprog

pitdicker:
# Error types

As a start I'll propose an `Error` enum.
The current causes for errors and panics in the parts of the API I have looked add fall in the following categories:

```rust
pub enum ChronoError {
    InvalidParameter, // example: hour == 25
    Overflow, // example: year = 500_000
    DateDoesNotExist, // example: 31 February
    InconsistentDate(NaiveDate), // example: parsing Sunday 2023-04-21 (should be Friday)
    ParsingError(ParsingError), // collect all other reasons for failing to parse something in a nested enum.
    // Maybe:
    TzLoadingError(/* todo */), // Catchall for all errors caused by the OS in `offset::local`
}
```

And a specialized type for errors caused by timezone conversion:

```rust
pub enum LocalResult<T> {
    Single(T),
    Ambiguous(T, T),
    InGap(T, T),
    Error(ChronoError)
}
```

pitdicker:
# Goals
1. Chrono should allow writing software that handles dates and times correctly and doesn't panic.
2. Chrono should have an API that is pleasant to use.
3. Errors should be informative and/or actionable.

## Pleasant to use
The deprecations of many potentially panicking methods in 0.4.23 have made chrono very unpleasant to use, with `unwrap()` all over the place.

Many edge conditions may simply never arise in real-world use. Making every method that might error in some obscure situation return `Result` makes the code that uses chrono hard to read. Why is this `unwrap`? Is it because of some condition that will never arise, or is it taking sketchy short-cuts?

Example: `NaiveDate::succ` may fail if the following date falls outside of the range of dates representable by a `NaiveDate`. But how much real-world code has to deal with dates around the year 262143? In this case making the default method panic on overflow is best.

See https://github.com/chronotope/chrono/issues/815#issuecomment-1312841773

## Allow use without panics
For every API that may panic, there should be an equivalent that is guaranteed to never panic (and otherwise that is a bug).

Example: `NaiveDate::try_succ`.

### Errors should be actionable.

[...]

PR #1421, opened 2024-02-09
Make conversions to and from `std::time::SystemTime` non-panicking

ahcodedthat:
As discussed in in #1049 (specifically [here](https://github.com/chronotope/chrono/issues/1049#issuecomment-1928641000) and [here](https://github.com/chronotope/chrono/issues/1049#issuecomment-1929431760)), this PR replaces the panicking `From` implementations between `std::time::SystemTime` and `chrono::DateTime` with non-panicking `TryFrom` implementations.

I see that #1410 introduces a new error type, but it has not yet landed, so this PR uses the existing `OutOfRange` error type. Should I open a PR against #1410's branch instead, and use the new error type?


pitdicker (commented):

> I see that #1410 introduces a new error type, but it has not yet landed, so this PR uses the existing `OutOfRange` error type. Should I open a PR against #1410's branch instead, and use the new error type?

I prefer what you did here until our error work is a bit further along. It would be even better to target the 0.4 branch and get this improvements available earlier.

I did not do a good review yet, just an initial comment.

src/datetime/mod.rs
```diff
@@ -1536,42 +1538,67 @@ impl str::FromStr for DateTime<Local> {
 }
 
 #[cfg(feature = "std")]
-impl From<SystemTime> for DateTime<Utc> {
-    fn from(t: SystemTime) -> DateTime<Utc> {
+impl TryFrom<SystemTime> for DateTime<Utc> {
```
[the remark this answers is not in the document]
ahcodedthat: I can't. `impl From<T>` and `impl TryFrom<T>` are mutually exclusive. The standard library has a blanket `impl TryFrom<T>` [here](https://doc.rust-lang.org/stable/src/core/convert/mod.rs.html#803-805) that covers every `impl From<T>`.

src/datetime/mod.rs
```diff
@@ -1536,42 +1538,67 @@ impl str::FromStr for DateTime<Local> {
...
+    type Error = OutOfRange;
+
+    fn try_from(t: SystemTime) -> Result<DateTime<Utc>, OutOfRange> {
         let (sec, nsec) = match t.duration_since(UNIX_EPOCH) {
```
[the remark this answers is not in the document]
ahcodedthat: It already did, even before this PR; that's what [this `Err` branch](https://github.com/chronotope/chrono/blob/dc196062650c05528cbe259e340210f0340a05d1/src/datetime/mod.rs#L1622) handles. The test [does try the conversion with a date before the epoch](https://github.com/chronotope/chrono/blob/dc196062650c05528cbe259e340210f0340a05d1/src/datetime/tests.rs#L1178), so it definitely works.

src/datetime/mod.rs
```diff
@@ -1536,42 +1538,67 @@ impl str::FromStr for DateTime<Local> {
...
+
+    fn try_from(dt: DateTime<Tz>) -> Result<SystemTime, OutOfRange> {
+        let sec: i64 = dt.timestamp();
+        let sec_abs: u64 = sec.unsigned_abs();
         let nsec = dt.timestamp_subsec_nanos();
         if sec < 0 {
             // unlikely but should be handled
-            UNIX_EPOCH - Duration::new(-sec as u64, 0) + Duration::new(0, nsec)
+            UNIX_EPOCH
+                .checked_sub(Duration::new(sec_abs, 0))
```
pitdicker, line 1597: Why not just `Duration::new(-sec as u64, 0)`?

src/datetime/mod.rs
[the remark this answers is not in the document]
pitdicker: That's unfortunate :disappointed:.

src/offset/utc.rs
```diff
@@ -74,18 +76,33 @@ impl Utc {
...
     /// let now_with_offset = Utc::now().with_timezone(&offset);
     /// ```
+    ///
+    /// # Panics
+    ///
+    /// Panics if the system clock is set to a time in the extremely distant past or future, such
+    /// that it is out of the range representable by `DateTime<Utc>`. It is assumed that this
+    /// crate will no longer be in use by that time.
+    ///
+    /// If panic in this situation is not acceptable, use [`SystemTime::now`] instead, like this:
```
pitdicker, line 86: I am not sure this is an advice we want to give in general, or want to give this prominent.

pitdicker (commented):
Good work, very nice to get these improvements!

I only have some minor requests.

src/offset/utc.rs
```diff
@@ -74,18 +76,33 @@ impl Utc {
...
+    ///
+    /// ```
+    /// # use chrono::{DateTime, Utc};
+    /// # use std::time::SystemTime;
+    /// # let _ =
+    /// DateTime::<Utc>::try_from(SystemTime::now())
+    /// # ;
+    /// ```
+    ///
+    /// This will yield a <code>Result&lt;DateTime&lt;Utc&gt;, [OutOfRange]&gt;</code>,
```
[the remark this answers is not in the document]
ahcodedthat: Per your later comment, I've gone ahead and removed this entirely, but in case you're wondering, I did this ugly HTML thing so that `OutOfRange` would be a link. As far as I know, the only way to make an intra-doc link inside an inline code element is by using HTML like that. It is admittedly ugly in source code, though.

src/offset/utc.rs
```diff
@@ -74,18 +76,33 @@ impl Utc {
...
         not(any(target_os = "emscripten", target_os = "wasi"))
     )))]
     #[must_use]
     pub fn now() -> DateTime<Utc> {
-        let now =
-            SystemTime::now().duration_since(UNIX_EPOCH).expect("system time before Unix epoch");
-        let naive =
-            NaiveDateTime::from_timestamp(now.as_secs() as i64, now.subsec_nanos()).unwrap();
-        Utc.from_utc_datetime(&naive)
+        SystemTime::now().try_into().unwrap()
```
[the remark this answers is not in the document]
ahcodedthat: Ok, done.

src/datetime/mod.rs
```diff
@@ -1536,42 +1538,67 @@ impl str::FromStr for DateTime<Local> {
...
+    fn try_from(t: SystemTime) -> Result<DateTime<Utc>, OutOfRange> {
         let (sec, nsec) = match t.duration_since(UNIX_EPOCH) {
-            Ok(dur) => (dur.as_secs() as i64, dur.subsec_nanos()),
+            Ok(dur) => {
+                let sec = i64::try_from(dur.as_secs()).map_err(|_| OutOfRange::new())?;
+                let nsec = dur.subsec_nanos();
+                (sec, nsec)
+            }
             Err(e) => {
                 // unlikely but should be handled
```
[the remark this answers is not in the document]
ahcodedthat: Ok, done.

src/datetime/mod.rs
```diff
@@ -1536,42 +1538,67 @@ impl str::FromStr for DateTime<Local> {
...
+    fn try_from(dt: DateTime<Tz>) -> Result<SystemTime, OutOfRange> {
+        let sec: i64 = dt.timestamp();
+        let sec_abs: u64 = sec.unsigned_abs();
         let nsec = dt.timestamp_subsec_nanos();
         if sec < 0 {
             // unlikely but should be handled
-            UNIX_EPOCH - Duration::new(-sec as u64, 0) + Duration::new(0, nsec)
+            UNIX_EPOCH
+                .checked_sub(Duration::new(sec_abs, 0))
+                .and_then(|t| t.checked_add(Duration::new(0, nsec)))
```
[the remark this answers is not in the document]
ahcodedthat: Ah, you're right. I've gone ahead and changed it.

src/datetime/mod.rs
ahcodedthat: If `sec == i64::MIN` and the compiler is configured to panic on overflow, then `-sec` will panic. I used `sec.unsigned_abs()` to avoid that.

I could replace `let sec_abs` with two separate invocations of `sec.unsigned_abs()`, if you'd like.

src/offset/utc.rs
ahcodedthat: Ok, I've removed it.

src/datetime/mod.rs
pitdicker: I missed the `unsigned_` part. Very nice, better than something like `checked_abs`. Compliments for also catching this case (I used to know this, but would have made the mistake).

pitdicker (approved):
@djc Will probably also want to have a look.

src/datetime/mod.rs
```diff
@@ -1536,42 +1538,74 @@ impl str::FromStr for DateTime<Local> {
...
+                let nsec = dur.subsec_nanos();
+                (sec, nsec)
+            }
             Err(e) => {
-                // unlikely but should be handled
+                // `t` is before the Unix epoch. `e.duration()` is how long before the epoch it
+                // is.
                 let dur = e.duration();
-                let (sec, nsec) = (dur.as_secs() as i64, dur.subsec_nanos());
+                let sec: u64 = dur.as_secs();
```
djc, line 1556: Please remove the type annotations here, which should be unnecessary for `as_secs()` and `subsec_nanos()`. You can use `i64::try_from()` for the conversion.

src/datetime/mod.rs
```diff
@@ -1536,42 +1538,74 @@ impl str::FromStr for DateTime<Local> {
...
+                let nsec: u32 = dur.subsec_nanos();
                 if nsec == 0 {
+                    // Overflow safety: `sec` was returned by `dur.as_secs()`, and is guaranteed to
+                    // be non-negative. Negating a non-negative signed integer cannot overflow.
                     (-sec, 0)
                 } else {
-                    (-sec - 1, 1_000_000_000 - nsec)
+                    // Overflow safety: In addition to the above, `-x - 1`, where `x` is a
+                    // non-negative signed integer, also cannot overflow.
+                    let sec: i64 = -sec - 1;
```
djc, line 1566: Don't think the type annotations are needed here, either.

src/datetime/mod.rs
```diff
@@ -1536,42 +1538,74 @@ impl str::FromStr for DateTime<Local> {
...
 
 #[cfg(feature = "std")]
-impl<Tz: TimeZone> From<DateTime<Tz>> for SystemTime {
-    fn from(dt: DateTime<Tz>) -> SystemTime {
-        let sec = dt.timestamp();
+impl<Tz: TimeZone> TryFrom<DateTime<Tz>> for SystemTime {
+    type Error = OutOfRange;
+
+    fn try_from(dt: DateTime<Tz>) -> Result<SystemTime, OutOfRange> {
+        let sec: i64 = dt.timestamp();
```
djc, line 1593: Same here.

src/datetime/mod.rs
```diff
@@ -1536,42 +1538,74 @@ impl str::FromStr for DateTime<Local> {
...
+        let sec_abs: u64 = sec.unsigned_abs();
         let nsec = dt.timestamp_subsec_nanos();
         if sec < 0 {
-            // unlikely but should be handled
-            UNIX_EPOCH - Duration::new(-sec as u64, 0) + Duration::new(0, nsec)
+            // `dt` is before the Unix epoch.
+            let mut t =
+                UNIX_EPOCH.checked_sub(Duration::new(sec_abs, 0)).ok_or(OutOfRange::new())?;
```
djc, line 1599: If we're going to call `OutOfRange::new()`, should probably use `ok_or_else()` instead?

src/offset/utc.rs
```diff
@@ -74,18 +76,22 @@ impl Utc {
...
         not(any(target_os = "emscripten", target_os = "wasi"))
     )))]
     #[must_use]
     pub fn now() -> DateTime<Utc> {
-        let now =
-            SystemTime::now().duration_since(UNIX_EPOCH).expect("system time before Unix epoch");
-        let naive =
-            NaiveDateTime::from_timestamp(now.as_secs() as i64, now.subsec_nanos()).unwrap();
-        Utc.from_utc_datetime(&naive)
+        SystemTime::now().try_into().expect(
```
djc, line 92: I think we'll want to make this fallible, but can leave this for a future PR.

src/datetime/mod.rs
ahcodedthat: Ok, done.

src/datetime/mod.rs
ahcodedthat: Ok, done.

src/datetime/mod.rs
ahcodedthat: Ok, done.

src/datetime/mod.rs
ahcodedthat: Ok, done.

I didn't think it would matter either way, since `OutOfRange::new` doesn't do anything at run time. :man_shrugging:

djc (approved):

src/offset/utc.rs
pitdicker: Opened an issue https://github.com/chronotope/chrono/issues/1440.

PR #1433, opened 2024-02-12
Refactor `internals` module part 5

pitdicker:
These are the simple commits that followed after #1212 but where split out.

- Split `naive::date` into separate files to get the same folder structure as `naive::time` and `naive::datetime`.
- Move the `NaiveWeek` type to its own module. It needs an extra `new` method because its fields are now private.

### DIFF
--- a/src/naive/date/mod.rs
+++ b/src/naive/date/mod.rs
@@ -33,7 +33,7 @@ use crate::format::{
     Parsed, StrftimeItems,
 };
 use crate::month::Months;
-use crate::naive::{IsoWeek, NaiveDateTime, NaiveTime, NaiveWeek};
+use crate::naive::{Days, IsoWeek, NaiveDateTime, NaiveTime, NaiveWeek};
 use crate::{expect, try_opt};
 use crate::{Datelike, TimeDelta, Weekday};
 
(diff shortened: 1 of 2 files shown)


src/naive/date/mod.rs
```diff
@@ -1427,7 +1432,7 @@ impl NaiveDate {
     /// specified.
     #[inline]
     pub const fn week(&self, start: Weekday) -> NaiveWeek {
-        NaiveWeek { date: *self, start }
+        NaiveWeek::new(*self, start)
```
djc, line 1340: This should go before `new()`.

src/naive/date/serde.rs
```diff
@@ -0,0 +1,78 @@
...
+    where
+        D: de::Deserializer<'de>,
+    {
+        deserializer.deserialize_str(NaiveDateVisitor)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::naive::date::{test_decodable_json, test_encodable_json};
```
djc, line 54: I think `test_decodable_json` and `test_encodable_json` should move into `tests`.

djc (commented):
So I think we discussed this before, but still not a big fan of the module split as proposed in this PR. All of `serde`, `rustc_serialize` and `week` end up being too small to be worthy of a separate file, IMO. I would suggest moving everything that's not an `impl NaiveDate` or `impl .. for NaiveDate` into a `types` module, maybe?

pitdicker:
Yes, the `rustc-serialize` and `serde` modules are very small. Shall I keep them in the `date/mod.rs` file?

And move `NaiveWeek` and `Day` into `naive/mod.rs`?

src/naive/date/mod.rs
pitdicker: Just to make sure: we are looking at `NaiveWeek::new` and `NaiveDate::week`. I don't think there is much I can reorder? Or I don't understand your commen.?

djc: If `week()` uses `new()`, it should be defined before it.

src/naive/date/serde.rs
pitdicker: Is it okay if I do that as a separate PR for these functions in all four modules (including `naive::time`, `naive::datetime` and `datetime`?

djc: Yup!

src/naive/date/mod.rs
pitdicker: Okay, I have been using that principle lately.

djc: It's still not looking the way I'd expect it to?

pitdicker: It is not part of the same impl block or the same type.

djc: Ahh, sorry!

djc (approved):

PR #1436, opened 2024-02-13
0.5.x: Rename `NaiveDate::*_opt` methods

pitdicker:
This continues the renaming op all `_opt` methods after the removal of their deprecated panicking variants in #1394. Only the `NaiveDate` type was left out of that PR. Turns out for the best, as the find-and-replace touches ca. 1000 lines.

The first three commits are a cosmetic fix for some tests that I would like to backport to the main branch in a later PR (I'm collecting a couple).

The last two commits remove 4 lines that mentioned removed `*_opt` methods in comments that slipped through in #1394 and the last merge of main.


Zomtir (approved):
Does what it describes. Only removes the _opt suffix and renames two macros.

djc (approved):

PR #1437, opened 2024-02-13
CI: test cross-compiling to `x86_64-unknown-illumos` instead of Solaris

pitdicker:
Apparently cross dropped support for the `x86_64-sun-solaris` target (https://github.com/cross-rs/cross/pull/1432), making our CI fail.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -169,7 +169,7 @@ jobs:
     strategy:
       matrix:
         target:
-          - x86_64-sun-solaris
+          - x86_64-unknown-illumos
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4


djc:
@steveklabnik as someone who might know some Illumos users, do you happen to have any feedback on this? Is this the right thing to be doing? (Not sure if Oxide systems end up using chrono somehow.)

steveklabnik:
Thanks for the cc, I have pinged work chat, let's see.

steveklabnik:
Yeah, this seems good: `x86_64-unknown-illumos` is the target we actually use. Thanks again for checking in!

djc (approved):

pitdicker:
@steveklabnik Thank you!

PR #1438, opened 2024-02-13
`Mdf` performance and docs

pitdicker:
The `Mdf` types reason for existence is to have a way to convert between ordinal dates and month-day dates efficiently. It is a core part of the performance of chrono, we use it internally all over the place. Before converting it to return `Result`s in the 0.5 branch I need a benchmark to not materially regress its performance.

Turns out I already regressed it by ca. 25% (at least according to the new benchmark) since chrono 0.4.24 :scream:.
After the benchmark in the first commit, the next three commits recover this lost performance by reverting needless checks:
- #1043 introduced checks in `Mdf::new` that are also covered by the table lookup `Mdf` exists for.
- #1432 tries to make `NaiveDate::from_mdf` share code with `NaiveDate::from_ordinal_and_flags`. It looks clean, but that introduces a bunch of unneeded checks. I added a dedicated method `Mdf::ordinal_and_flags` and made `NaiveDate::from_mdf` use that.
- #1207 introduced a useless check against zero with `NonZeroI32::new`.

Then there are two commits that improve the performance a little bit more.
And while working on it I wrote some doc comments for the methods on `Mdf`.

### Benchmark results

Chrono 0.4.24
```
bench_date_from_ymd     time:   [4.3541 ns 4.3737 ns 4.3958 ns]
```

Before
```
bench_date_from_ymd     time:   [5.5149 ns 5.5508 ns 5.5919 ns]
```

After
```
bench_date_from_ymd     time:   [4.2512 ns 4.2572 ns 4.2637 ns]
```

----

I have the commits that extend this PR to convert an `Mdf` to return `Result`s ready, but would like to base them on this PR to the main branch.

### DIFF
--- a/bench/benches/chrono.rs
+++ b/bench/benches/chrono.rs
@@ -8,6 +8,16 @@ use chrono::prelude::*;
 use chrono::Locale;
 use chrono::{DateTime, FixedOffset, Local, TimeDelta, Utc, __BenchYearFlags};
 
+fn bench_date_from_ymd(c: &mut Criterion) {
+    c.bench_function("bench_date_from_ymd", |b| {
+        let expected = NaiveDate::from_ymd_opt(2024, 2, 12);
+        b.iter(|| {
+            let (y, m, d) = black_box((2024, 2, 12));
+            assert_eq!(NaiveDate::from_ymd_opt(y, m, d), expected)
+        })
+    });
+}
+
 fn bench_datetime_parse_from_rfc2822(c: &mut Criterion) {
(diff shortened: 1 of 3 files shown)


pitdicker:
I rebased the documentation commit to be earlier in the PR, that should make reviewing a bit easier.

And there was one commit I had left after the `internals` refactor that makes the constants used in `YearFlags` less magical and updates the documentation. It kind of seems to fit in with this PR.

djc (approved):

pitdicker:
Nice, thank you!

PR #1439, opened 2024-02-14
`Parsed` fixes and documentation

### LINKED ISSUE #55: Example of Parsing to Parsed in README.MD
From my closed PR #54 it might be helpful to have an example of parsing formats using the [parse](https://github.com/lifthrasiir/rust-chrono/blob/master/src/format/parse.rs#L206) function with a `Parsed` struct in the readme for partial date parsing.

pitdicker:
This is the first part of #1418 and applies to the branch for 0.4.x.

We have two small bugs in our implementation of `Parsed`:

* If there is a timestamp and an offset field, the offset is added to the timestamp. But the definition of a Unix timestamp is that the value is in UTC, so this is not correct.
* We were returning `OUT_OF_RANGE` if the year value didn't match with the values of `year_div_100` or `year_mod_100`. It should return `IMPOSSIBLE` instead.

I extended the documentation of `Parsed` to describe why it exists (the resolution algorithm) and to give an example (fixes #55). Partly taken from the blog post before chrono 0.2: https://lifthrasiir.github.io/rustlog/worklog-2015-02-19.html

All the methods now have more accurate documentation that describes their error causes.

The documentation got a couple of rounds of self-review.

### DIFF
--- a/src/format/parsed.rs
+++ b/src/format/parsed.rs
@@ -892,7 +892,7 @@ impl Parsed {
 
     /// Returns a parsed fixed time zone offset out of given fields.
     pub fn to_fixed_offset(&self) -> ParseResult<FixedOffset> {
-        self.offset.and_then(FixedOffset::east_opt).ok_or(OUT_OF_RANGE)
+        FixedOffset::east_opt(self.offset.ok_or(NOT_ENOUGH)?).ok_or(OUT_OF_RANGE)
     }
 
     /// Returns a parsed timezone-aware date and time out of given fields.
--- a/src/format/parsed.rs
+++ b/src/format/parsed.rs
@@ -125,7 +125,7 @@ use crate::{DateTime, Datelike, TimeDelta, Timelike, Weekday};
 pub struct Parsed {
     /// Year.
     ///
-    /// This can be negative unlike [`year_div_100`](#structfield.year_div_100)
+    /// This can be negative, unlike [`year_div_100`](#structfield.year_div_100)
     /// and [`year_mod_100`](#structfield.year_mod_100) fields.
     pub year: Option<i32>,
 
@@ -141,7 +141,7 @@ pub struct Parsed {
 
     /// Year in the [ISO week date](../naive/struct.NaiveDate.html#week-date).
     ///
-    /// This can be negative unlike [`isoyear_div_100`](#structfield.isoyear_div_100) and
+    /// This can be negative, unlike [`isoyear_div_100`](#structfield.isoyear_div_100) and
     /// [`isoyear_mod_100`](#structfield.isoyear_mod_100) fields.
     pub isoyear: Option<i32>,
 
@@ -232,13 +232,27 @@ impl Parsed {
         Parsed::default()
     }
 
-    /// Tries to set the [`year`](#structfield.year) field from given value.
+    /// Set the 'year' field to the given value.
+    ///
+    /// The value can be negative, unlike the 'year divided by 100' and 'year modulo 100' fields.
+    ///
+    /// # Errors
+    ///
+    /// Returns `OUT_OF_RANGE` if `value` is outside the range of an `i32`.
+    ///
+    /// Returns `IMPOSSIBLE` if this field was already set to a different value.
     #[inline]
     pub fn set_year(&mut self, value: i64) -> ParseResult<()> {
         set_if_consistent(&mut self.year, i32::try_from(value).map_err(|_| OUT_OF_RANGE)?)
     }
 
-    /// Tries to set the [`year_div_100`](#structfield.year_div_100) field from given value.
+    /// Set the 'year divided by 100' field to the given value.
+    ///
+    /// # Errors
+    ///
+    /// Returns `OUT_OF_RANGE` if `value` is negative or if it is greater than `i32::MAX`.
+    ///
+    /// Returns `IMPOSSIBLE` if this field was already set to a different value.
     #[inline]
     pub fn set_year_div_100(&mut self, value: i64) -> ParseResult<()> {
         if value < 0 {
@@ -247,7 +261,20 @@ impl Parsed {
         set_if_consistent(&mut self.year_div_100, i32::try_from(value).map_err(|_| OUT_OF_RANGE)?)
     }
 
-    /// Tries to set the [`year_mod_100`](#structfield.year_mod_100) field from given value.
+    /// Set the 'year modulo 100' field to the given value.
+    ///
+    /// When set it implies that the year is not negative.
+    ///
+    /// If this field is set while the 'year divided by 100' field is missing (and the full 'year'
+    /// field is also not set), it assumes a default value for the 'year divided by 100' field.
+    /// The default is 19 when `year_mod_100 >= 70` and 20 otherwise.
+    ///
+    /// # Errors
+    ///
+    /// May return `OUT_OF_RANGE` if `value` is negative or if it is greater than 99.
+    /// Currently only checks the value is positive and not out of range for an `i32`.
+    ///
+    /// Returns `IMPOSSIBLE` if this field was already set to a different value.
     #[inline]
     pub fn set_year_mod_100(&mut self, value: i64) -> ParseResult<()> {
         if value < 0 {
@@ -256,13 +283,32 @@ impl Parsed {
         set_if_consistent(&mut self.year_mod_100, i32::try_from(value).map_err(|_| OUT_OF_RANGE)?)
     }
 
-    /// Tries to set the [`isoyear`](#structfield.isoyear) field from given value.
+    /// Set the 'year' field that is part of an [ISO 8601 week date] to the given value.
+    ///
+    /// The value can be negative, unlike the 'year divided by 100' and 'year modulo 100' fields.
+    ///
+    /// [ISO 8601 week date]: crate::NaiveDate#week-date
+    ///
+    /// # Errors
+    ///
+    /// Returns `OUT_OF_RANGE` if `value` is outside the range of an `i32`.
+    ///
+    /// Returns `IMPOSSIBLE` if this field was already set to a different value.
     #[inline]
     pub fn set_isoyear(&mut self, value: i64) -> ParseResult<()> {
         set_if_consistent(&mut self.isoyear, i32::try_from(value).map_err(|_| OUT_OF_RANGE)?)
     }
 
-    /// Tries to set the [`isoyear_div_100`](#structfield.isoyear_div_100) field from given value.
+    /// Set the 'year divided by 100' field that is part of an [ISO 8601 week date] to the given
+    /// value.
+    ///
+    /// [ISO 8601 week date]: crate::NaiveDate#week-date
+    ///
+    /// # Errors
+    ///
+    /// Returns `OUT_OF_RANGE` if `value` is negative or if it is greater than `i32::MAX`.
+    ///
+    /// Returns `IMPOSSIBLE` if this field was already set to a different value.
     #[inline]
     pub fn set_isoyear_div_100(&mut self, value: i64) -> ParseResult<()> {
         if value < 0 {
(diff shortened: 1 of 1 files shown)


src/format/parsed.rs
```diff
@@ -309,7 +309,7 @@ impl Parsed {
                 // we should filter a negative full year first.
                 (Some(y), q, r @ Some(0..=99)) | (Some(y), q, r @ None) => {
                     if y < 0 {
-                        return Err(OUT_OF_RANGE);
+                        return Err(IMPOSSIBLE);
```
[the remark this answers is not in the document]
pitdicker: The comment right above it says:
```
                // if there is a full year *and* also quotient and/or modulo,
                // check if present quotient and/or modulo is consistent to the full year.
                // since the presence of those fields means a positive full year,
                // we should filter a negative full year first.
```

That seems to cover it?

src/format/parsed.rs
```diff
@@ -9,13 +9,116 @@ use crate::naive::{NaiveDate, NaiveDateTime, NaiveTime};
...
+/// `Parsed` is used internally by all parsing functions in chrono. It is a public type so that it
+/// can be used to write custom parsers that re-use the resolving algorithm, or to inspect the
+/// results of a string parsed with chrono without converting it to concrete types.
+///
+/// # Resolving algorithm
+///
+/// Resolving date/time parts is littered with lots of corner cases, which is why common date/time
+/// parsers do not correctly implement it.
+///
+/// Chrono provides a complete and correct resolution algorithm via the `Parsed` type.
```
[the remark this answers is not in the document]
pitdicker: I even watered it down from "Therefore I'm glad to announce that Chrono 0.2 has a complete and known-to-be-correct resolution algorithm" in bold from the blog post 🤣.

It is complete though. Once we have resolved for example a date, all other fields are compared to what chrono would calculate those fields to be for that date. I'll tone it down to just 'complete'.

src/format/parsed.rs
```diff
@@ -570,6 +570,17 @@ impl Parsed {
     ///
     /// Gregorian year and ISO week date year can have their century number (`*_div_100`) omitted,
     /// the two-digit year is used to guess the century number then.
+    ///
+    /// It checks all given date fields are consistent with each other.
+    ///
+    /// # Errors
+    ///
+    /// - Returns `IMPOSSIBLE`  if any of the date fields conflict.
```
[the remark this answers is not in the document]
pitdicker: > But also, not sure to what level of detail we want to document this stuff? Like, once we're at the point where we're documenting all the possible scenarios in which a given error value can be returned we're just restating the code in prose which is painful to maintain and inefficient to consume. After all, this is an open source project -- it's pretty easy to click "view source" on docs.rs.

I don't expect this to cause any trouble for maintaining the code base. The causes for errors are not going to change.

My problem with using chrono in the past was that it was hard to understand all the causes why a function might fail. So I very much like documentation such as this. It makes me think deeper and better about the corner cases around the issue I'm solving at that moment.

Clicking "view source" is not really enough for these methods.
The errors in `Parsed::to_datetime` can come from the method itself, from `FixedOffset::east_opt`, `TimeZone::from_local_datetime`, `Parsed::to_naive_datetime_with_offset`, `NaiveDateTime::from_timestamp_opt`, `Parsed::to_naive_time`, `Parsed::to_naive_date` and a couple of methods of `NaiveDate`.
That is a lot of levels deep to examine!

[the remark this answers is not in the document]
pitdicker: I'll do my best to clean it up though.

pitdicker:
Found more oddity to document for the `to_naive_datetime_with_offset` method:
> The offset is assumed to have a given value. It is not compared against the offset field set in the `Parsed` type, so it is allowed to be inconsistent.

Actually that whole method is just weird to work with. The logic should be inverted to return a `NaiveDateTime`, and an optional `FixedOffset` if there is enough information available. In September I planned to deprecate that method and replace it with something more sensible but never made the PR.

pitdicker:
> * If there is a timestamp and an offset field, the offset is added to the timestamp. But the definition of a Unix timestamp is that the value is in UTC, so this is not correct.

I have to admit I read the current code wrong. It is working exactly like it should. The timestamp is assumed to be in UTC, it adds an offset, and then checks the rest of the values as local date and time. Dropped the first commit.

src/format/parsed.rs
```diff
@@ -9,13 +9,117 @@ use crate::naive::{NaiveDate, NaiveDateTime, NaiveTime};
...
+///
+/// - `set_*` methods to set fields you have available. They do a basic range check, and if the
+///   same field is set more than once it is checked for consistency.
 ///
 /// - `to_*` methods try to make a concrete date and time value out of set fields.
-///   It fully checks any remaining out-of-range conditions and inconsistent/impossible fields.
+///   They fully check that all fields are consistent and whether the date/datetime exists.
+///
+/// `Parsed` is used internally by all parsing functions in chrono. It is a public type so that it
+/// can be used to write custom parsers that re-use the resolving algorithm, or to inspect the
```
djc, line 23: "re-use" -> "reuse"

src/format/parsed.rs
```diff
@@ -570,6 +570,18 @@ impl Parsed {
     ///
     /// Gregorian year and ISO week date year can have their century number (`*_div_100`) omitted,
     /// the two-digit year is used to guess the century number then.
+    ///
+    /// It checks all given date fields are consistent with each other.
+    ///
+    /// # Errors
+    ///
+    /// This method returns:
+    /// - `IMPOSSIBLE` if any of the date fields conflict.
```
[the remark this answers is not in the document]
pitdicker: > (I also feel `IMPOSSIBLE` is not a great name for conflicting field values, and we should maybe rename `NOT_ENOUGH` to `INCOMPLETE`?)

Agreed. I planned `Error::Inconsistent` for the first, and didn't think much about the other. `Error::Incomplete` sounds good. Something for the 0.5.x branch?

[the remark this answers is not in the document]
pitdicker: > For these, I do think we should at least have nested bullet lists if there are multiple potential causes for the same error value.

I just separated it by primary error cause (as I see it), matching what I have for the `Error` type for the 0.5 branch. 🤷, I'll change it back when this code switches to the new type.

pitdicker:
Found one more missing error case in `Parsed::to_fixed_offset`. The use of `Option`s covered it up :smile:.
If there is no offset field set, it should return `NOT_ENOUGH` instead of `OUT_OF_RANGE`.

src/format/parsed.rs
```diff
@@ -9,13 +9,117 @@ use crate::naive::{NaiveDate, NaiveDateTime, NaiveTime};
...
-/// - `set_*` methods try to set given field(s) while checking for the consistency.
-///   It may or may not check for the range constraint immediately (for efficiency reasons).
+/// There are two classes of methods:
+///
+/// - `set_*` methods to set fields you have available. They do a basic range check, and if the
+///   same field is set more than once it is checked for consistency.
 ///
 /// - `to_*` methods try to make a concrete date and time value out of set fields.
-///   It fully checks any remaining out-of-range conditions and inconsistent/impossible fields.
+///   They fully check that all fields are consistent and whether the date/datetime exists.
```
djc, line 20: Side thought: arguably the `to_()` methods should become `impl TryFrom<Parsed> for _` impls?

src/format/parsed.rs
```diff
@@ -9,13 +9,117 @@ use crate::naive::{NaiveDate, NaiveDateTime, NaiveTime};
...
+///
+/// `Parsed` is used internally by all parsing functions in chrono. It is a public type so that it
+/// can be used to write custom parsers that reuse the resolving algorithm, or to inspect the
+/// results of a string parsed with chrono without converting it to concrete types.
+///
+/// # Resolving algorithm
+///
+/// Resolving date/time parts is littered with lots of corner cases, which is why common date/time
+/// parsers do not correctly implement it.
```
djc, line 29: "do not correctly implement it" -> "do not implement it correctly" is a better word order here.

src/format/parsed.rs
```diff
@@ -232,13 +232,27 @@ impl Parsed {
         Parsed::default()
     }
 
-    /// Tries to set the [`year`](#structfield.year) field from given value.
+    /// Set the 'year' field to the given value.
+    ///
+    /// The value can be negative unlike the 'year divided by 100' and 'year modulo 100' fields.
```
djc, line 237: I think we need commas before "unlike". When referencing values stored in other fields, would suggest referencing the field directly, like `year_div_100` or even ```[`Parsed::year_div_100`]```?

src/format/parsed.rs
```diff
@@ -247,7 +261,20 @@ impl Parsed {
...
+    /// When set it implies that the year is not negative.
+    ///
+    /// If this field is set while the 'year divided by 100' field is missing (and the full 'year'
+    /// field is also not set), it assumes a default value for the 'year divided by 100' field.
+    /// The default is 19 when `year_mod_100 >= 70` and 20 otherwise.
+    ///
+    /// # Errors
+    ///
+    /// May return `OUT_OF_RANGE` if `value` is negative or if it is greater than 99.
+    /// Currently only checks the value is positive and not out of range for an `i32`.
```
djc, line 275: I think this sounds pretty weird. "May return blah" but actually "Currently" doesn't? I think we should describe what it does, and not what it doesn't do. So maybe replace "Currently only checks [..]" with "For performance reasons, precise range checks may be deferred in favor of checking on conversion into a target type."?

djc (approved):

djc:
This is shaping up nicely!

src/format/parsed.rs
pitdicker: In a few cases they can. Good to keep in mind! I'll come back to it later (before 0.5).

src/format/parsed.rs
pitdicker: The part about performance has never really been true or benchmarked. See https://github.com/chronotope/chrono/pull/1418#issuecomment-1954102235.

With incomplete range checks (current):
```
bench_datetime_parse_from_rfc2822
                        time:   [156.24 ns 157.01 ns 157.85 ns]
  2 (2.00%) high mild
  5 (5.00%) high severe

bench_datetime_parse_from_rfc3339
                        time:   [104.25 ns 104.54 ns 104.86 ns]
  2 (2.00%) high mild
  6 (6.00%) high severe
```

With complete range checks:
```
bench_datetime_parse_from_rfc2822
                        time:   [156.47 ns 156.81 ns 157.16 ns]
                        change: [-0.3688% +0.2339% +0.7850%] (p = 0.44 > 0.05)
                        No change in performance detected.
  2 (2.00%) high mild
  5 (5.00%) high severe

bench_datetime_parse_from_rfc3339
                        time:   [102.69 ns 103.07 ns 103.49 ns]
                        change: [-2.4267% -1.7659% -1.1091%] (p = 0.00 < 0.05)
                        Performance has improved.
  4 (4.00%) high mild
  4 (4.00%) high severe
```

So basically identical.

pitdicker: > I think this sounds pretty weird. "May return blah" but actually "Currently" doesn't? I think we should describe what it does, and not what it doesn't do.

My hope was to open a PR to add the complete range checks (instead of the incomplete current ones) as soon as this one is merged and remove all those lines again. Does that sound okay or should I polish this a bit more?

src/format/parsed.rs
pitdicker: You are right.

src/format/parsed.rs
pitdicker: > When referencing values stored in other fields, would suggest referencing the field directly, like `year_div_100` or even `` [`Parsed::year_div_100`] ``?

I choose to avoid referencing the fields directly so it is easier to make them private (my proposal for the 0.5.x branch).

djc: Okay, suggest making them `#[doc(hidden)]` on this branch.

pitdicker: :+1: I was think the same, I'll do it with a PR that adds the range checks and setter methods.

pitdicker:
> Found one more missing error case in `Parsed::to_fixed_offset`.

Wow, Apparently I found and fixed it last year also in #1042 😆.

PR #1442, opened 2024-02-14
CI: only run `cargo hack check` on Linux

pitdicker:
The `cargo hack-check` runs take ca. 10 minutes longer to run on Windows and MacOS then the other CI jobs.
None of our features interact with our platform integration code, so these CI jobs always succeed if the other jobs do.
Seems removing them is an easy way to cut our CI times to a third.

(Sorry, tiny PR.)


djc (approved):
Nice!

Issue #1050, opened 2023-05-06
Reduce the `Datelike` and `Timelike` traits


pitdicker:
Chrono would become more correct and convenient if the `Datelike` and `Timelike` traits were removed in 0.5.

I propose to implement their methods directly on the types that currently implement the traits.

- Is anyone even depending on these traits? Writing code that is generic over types implementing `Datelike` or `Timelike` is asking for incorrect results and errors. It kind of defeats the thoughtful API of chrono to write generic code that mixes types like that.
- When implemented on `DateTime` the `with_` method should return `LocalResult`, so library users can choose how to deal with ambiguous or non-existing local datetimes. The current workaround to use them is cumbersome. See https://github.com/chronotope/chrono/issues/1049#issuecomment-1537157454 example 4.
- Trait methods can't be made const for the foreeseable future, which would be a gap in the const API of chrono (once it has one that is).
- The methods are currently less discoverable, but this can be considered a documentation issue.
- Library users would need to do one import less because the methods would just be available on the type.

jtmoon79:
FWIW I also get consistently hung up on the `Datelike` and `Timelike` traits; e.g. I spend tens of minutes searching for something that I know I've seen work elsewhere to then discover I need to `use ...Datelike` or `Timelike`. It's happened a few times and it's a little bit exasperating. At the same time, this could be fixed with better documentation.

> The methods are currently less discoverable, but this can be considered a documentation issue.

djc:
I'm not very convinced by your arguments so far to remove them. Abstracting over a `Date` and the date components of a `Time` seems conceptually reasonable to me, if maybe something that doesn't come up a lot.

However, if you can show that over some large body of Rust code there is negligible code that abstracts over `Datelike` or `Timelike` (with `T: Datelike` or `impl Datelike` bounds), we could consider it.

olliemath:
My 2c: as the author of a library depending on chrono it's useful for me to offer a single function that takes a `T: Datelike`, rather than 3 boilerplate functions that take the various concrete date types

pitdicker:
My thoughts have changed a bit on this issue. I now think it is best to *shrink* the `Datelike` and `Timelike` traits.
They currently offer getter and setter methods.

The getter methods are fine as they are:

```rust
pub trait Datelike: Sized {
    // Required methods
    fn year(&self) -> i32;
    fn month(&self) -> u32;
    fn month0(&self) -> u32;
    fn day(&self) -> u32;
[...]

PR #1444, opened 2024-02-14
0.5: Converting the API to return Results (part 2): `Mdf` type

pitdicker:
# :tada:

This and #1436 are the last ingredients before we can convert a good portion of the `naive` module to return `Result`s.

In the first commit I adjusted four methods on `Mdf` to return a `Result`.

To ensure we correctly return either `InvalidParameter` or `DoesNotExist` on error, the `MDL_TO_OL` lookup table can now encode a second error case. We had 128 unused negative values available to do so, and it doesn't cost us any performance (compared to always returning the same variant).

The second commit changes `NaiveDate::from_ymd_opt` to return a `Result` to have something to benchmark against.

### Benchmark
```
bench_date_from_ymd     time:   [4.5764 ns 4.5945 ns 4.6141 ns]
```

Compared to returning an `Option` this is a slight regression of 8% with `main`, and compared to 0.4.24 a regression of ca. 5%. In my opinion it stays in the right ballpark, and seems to be unavoidable.

### Alternative
As an alternative I tried what happens if `Mdf` returns an `Option`, and we map it to an error in the callers. That brings ca. 20% overhead compared to `main`, and is a lot less nice to work with. Better to return `Result`s everywhere.

cc @Zomtir

### DIFF
--- a/src/format/parsed.rs
+++ b/src/format/parsed.rs
@@ -390,7 +390,7 @@ impl Parsed {
         let (verified, parsed_date) = match (given_year, given_isoyear, self) {
             (Some(year), _, &Parsed { month: Some(month), day: Some(day), .. }) => {
                 // year, month, day
-                let date = NaiveDate::from_ymd(year, month, day).ok_or(OUT_OF_RANGE)?;
+                let date = NaiveDate::from_ymd(year, month, day).map_err(|_| OUT_OF_RANGE)?;
                 (verify_isoweekdate(date) && verify_ordinal(date), date)
             }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -129,20 +129,20 @@
 //! # fn doctest() -> Option<()> {
 //!
 //! let dt = Utc.with_ymd_and_hms(2014, 7, 8, 9, 10, 11).unwrap(); // `2014-07-08T09:10:11Z`
-//! assert_eq!(dt, NaiveDate::from_ymd(2014, 7, 8)?.and_hms(9, 10, 11).unwrap().and_local_timezone(Utc).unwrap());
+//! assert_eq!(dt, NaiveDate::from_ymd(2014, 7, 8).unwrap().and_hms(9, 10, 11).unwrap().and_local_timezone(Utc).unwrap());
 //!
 //! // July 8 is 188th day of the year 2014 (`o` for "ordinal")
 //! assert_eq!(dt, NaiveDate::from_yo(2014, 189)?.and_hms(9, 10, 11).unwrap().and_utc());
 //! // July 8 is Tuesday in ISO week 28 of the year 2014.
 //! assert_eq!(dt, NaiveDate::from_isoywd(2014, 28, Weekday::Tue)?.and_hms(9, 10, 11).unwrap().and_utc());
 //!
-//! let dt = NaiveDate::from_ymd(2014, 7, 8)?.and_hms_milli(9, 10, 11, 12).unwrap().and_local_timezone(Utc).unwrap(); // `2014-07-08T09:10:11.012Z`
-//! assert_eq!(dt, NaiveDate::from_ymd(2014, 7, 8)?.and_hms_micro(9, 10, 11, 12_000).unwrap().and_local_timezone(Utc).unwrap());
-//! assert_eq!(dt, NaiveDate::from_ymd(2014, 7, 8)?.and_hms_nano(9, 10, 11, 12_000_000).unwrap().and_local_timezone(Utc).unwrap());
+//! let dt = NaiveDate::from_ymd(2014, 7, 8).unwrap().and_hms_milli(9, 10, 11, 12).unwrap().and_local_timezone(Utc).unwrap(); // `2014-07-08T09:10:11.012Z`
+//! assert_eq!(dt, NaiveDate::from_ymd(2014, 7, 8).unwrap().and_hms_micro(9, 10, 11, 12_000).unwrap().and_local_timezone(Utc).unwrap());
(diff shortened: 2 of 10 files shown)


src/format/parsed.rs
```diff
@@ -390,7 +390,7 @@ impl Parsed {
         let (verified, parsed_date) = match (given_year, given_isoyear, self) {
             (Some(year), _, &Parsed { month: Some(month), day: Some(day), .. }) => {
                 // year, month, day
-                let date = NaiveDate::from_ymd_opt(year, month, day).ok_or(OUT_OF_RANGE)?;
+                let date = NaiveDate::from_ymd_opt(year, month, day).map_err(|_| OUT_OF_RANGE)?;
```
pitdicker, line 393: This could be a proper mapping between the variants of `Error` and `ParsedResult`.
Because this will be one of the first places to touch when converting the parsing code (where it will become just `?`), and because that code is still getting the preparations before we can convert it to `Result`s, I'd like to wait with it until that time.

pitdicker:
I may have forgotten the doc tests (among things) in the enthusiasm :blush:. Sorry CI.

src/naive/internals.rs
```diff
@@ -322,14 +329,19 @@ impl Mdf {
...
+    ///
+    /// Returns [`Error::DoesNotExist`] if the given day does not exist in the given month.
     #[inline]
-    pub(super) const fn ordinal(&self) -> Option<u32> {
+    pub(super) const fn ordinal(&self) -> Result<u32, Error> {
         let mdl = self.0 >> 3;
-        match MDL_TO_OL[mdl as usize] {
-            XX => None,
-            v => Some((mdl - v as u8 as u32) >> 1),
+        let adjustment = MDL_TO_OL[mdl as usize];
```
Zomtir, line 338: Would it be possible to keep the matching? That way the adjustment variable has a tighter scope.

Secondly `x < 0` is not always the same as `x == -1`. Currently there is no other value than `-1`, but it is slightly more future proof to immediately detect `-2` as a different error. A more suitable error would probably be `Error::Undefined`, but it is out of range in this case.
```
        match MDL_TO_OL[mdl as usize] {
            adjustment if adjustment > 0 => Ok((mdl - adjustment as u8 as u32) >> 1),
            adjustment if adjustment == 0 => Err(Error::DoesNotExist),
            adjustment if adjustment == 1 => Err(Error::InvalidArgument),
            _ => Err(Error::OutOfRange),
        }
```

Zomtir (commented):
Looks very fine from what I see. I suggested a minor change to the -1/0/XX mapping.

src/lib.rs
```diff
@@ -129,20 +129,20 @@
 //! # fn doctest() -> Option<()> {
```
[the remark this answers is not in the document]
pitdicker: What do you mean with 'rat tail'?

I don't look how the doctests are starting to look, but trust we can clean them up nicely after converting some more parts of the API.

src/naive/internals.rs
```diff
@@ -134,46 +136,47 @@ const MAX_MDL: u32 = (12 << 6) | (31 << 1) | 1;
 
 // The next table are adjustment values to convert a date encoded as month-day-leapyear to
 // ordinal-leapyear. OL = MDL - adjustment.
-// Dates that do not exist are encoded as `XX`.
+// Error cases are encoded as `-1` and `XX` (`0`). For a quick check all positive values are
+// adjustments, `XX` indicates a date does not exist, and negative values are for invalid input.
 const XX: i8 = 0;
```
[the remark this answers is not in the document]
pitdicker: Actually I did just that but switched to this. It felt a bit confusing, looking like Roman numerals. I don't mind how we encode it. Maybe @djc has a preference?

src/naive/internals.rs
pitdicker: My first attempt used `cmp(&0)`, but that didn't work in const context. I expect making the match more precise instead of positive/0/negative would effect the performance a bit. (Can't run a reliable benchmark on the device I am working on right now.)

src/lib.rs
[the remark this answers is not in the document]
Zomtir: Returning a result in this doctest requires too many changes at this point in time because most *_opt() functions still return an Option. "Rat tail" is a local saying for having to deal with too many consequences for a certain action.

src/naive/internals.rs
pitdicker: Very nice!

djc (approved):

pitdicker:
:tada: Nice.

@Zomtir If you want to help out the time has come. Drop a note if you start something so I'll steer clear :smile:.

It seems good to do the conversion with only one method or a very small number of methods per commit. And starting with methods in the `naive` module that don't depend on pieces that have not yet been converted.

Converting `LocalResult`, `ParseError` and `TimeDelta` require some preparations first that I'll look into. Let's ignore anything around those for now.

Zomtir:
> Drop a note if you start something so I'll steer clear

I started with NaiveDate and got it to compile: 5d476d3b

Docs are not updated yet.

pitdicker:
Nice! That requires surprising little changes. I hope you can split it up in smaller commits.
Take your time to make the docs, tests and doctests work on more specific, now that they can be.

Could you skip the `DateTime` type? The `map_local` type is going to pass on multiple error causes, such as `OutOfRange`, `DoesNotExist`, OS errors, and trouble with a TZ string or TZif file. I'd like to be a bit more careful with that type and leave that to its own PR.

We should also have a discussion around the `Datelike` and `Timelike` traits before converting them. For the naive types their signature is fine. For `Datetime` it would be better if they returned `LocalResult` to give users a way to deal with timezone issues. See #1050.

PR #1450, opened 2024-02-21
Deprecate panicking constructors of `TimeDelta`

### LINKED ISSUE #1408: Deprecate panicking `TimeDelta` initializers
The following methods can panic and have a `try_` variant:
```rust
impl Duration {
    pub const fn weeks(weeks: i64) -> Duration;
    pub const fn days(days: i64) -> Duration;
    pub const fn hours(hours: i64) -> Duration;
    pub const fn minutes(minutes: i64) -> Duration;
    pub const fn seconds(seconds: i64) -> Duration;
    pub const fn milliseconds(milliseconds: i64) -> Duration;
}
```

The replacement methods have been introduced only recently in 0.4.32, and `Duration::try_milliseconds` has not seen a release yet.

In all other places in chrono we have deprecated such panicking methods, and I suppose we should do the same here.
I wonder when would be the right time?

cc @danwilliams

pitdicker:
This PR touches a lot of lines, but I would only call the first 5 commits and the last one interesting.

- In a few places that only work with dates I replaced `TimeDelta::days` and `TimeDelta::weeks` with the `Days` type, which should be a little faster and conceptually a better match.
- In `NaiveDateTime::{checked_add_signed, checked_sub_signed}` we could remove a workaround now that we have `try_seconds`.
- In `Parsed::to_naive_date` I factored out the logic to calculate a date from a week starting on Monday or Sunday to `resolve_week_date`. And instead of calculating a `TimeDelta` with the number of days to add to January 1 we just calculate the ordinal directly and set the ordinal.
- All other commits replace each method that is to be deprecated with its `try_` variant.

Fixes #1408.

### DIFF
--- a/src/datetime/tests.rs
+++ b/src/datetime/tests.rs
@@ -1511,20 +1511,20 @@ fn test_datetime_sub_assign() {
     let datetime = naivedatetime.and_utc();
     let mut datetime_sub = datetime;
 
-    datetime_sub -= TimeDelta::minutes(90);
-    assert_eq!(datetime_sub, datetime - TimeDelta::minutes(90));
+    datetime_sub -= TimeDelta::try_minutes(90).unwrap();
+    assert_eq!(datetime_sub, datetime - TimeDelta::try_minutes(90).unwrap());
 
     let timezone = FixedOffset::east_opt(60 * 60).unwrap();
     let datetime = datetime.with_timezone(&timezone);
     let datetime_sub = datetime_sub.with_timezone(&timezone);
 
-    assert_eq!(datetime_sub, datetime - TimeDelta::minutes(90));
+    assert_eq!(datetime_sub, datetime - TimeDelta::try_minutes(90).unwrap());
 
     let timezone = FixedOffset::west_opt(2 * 60 * 60).unwrap();
     let datetime = datetime.with_timezone(&timezone);
     let datetime_sub = datetime_sub.with_timezone(&timezone);
 
-    assert_eq!(datetime_sub, datetime - TimeDelta::minutes(90));
+    assert_eq!(datetime_sub, datetime - TimeDelta::try_minutes(90).unwrap());
 }
 
 #[test]
--- a/src/datetime/tests.rs
+++ b/src/datetime/tests.rs
@@ -1500,8 +1500,8 @@ fn test_datetime_add_assign_local() {
 
     // ensure we cross a DST transition
     for i in 1..=365 {
-        datetime_add += TimeDelta::days(1);
-        assert_eq!(datetime_add, datetime + TimeDelta::days(i))
+        datetime_add += TimeDelta::try_days(1).unwrap();
+        assert_eq!(datetime_add, datetime + TimeDelta::try_days(i).unwrap())
     }
 }
 
@@ -1709,8 +1709,8 @@ fn test_datetime_sub_assign_local() {
 
     // ensure we cross a DST transition
     for i in 1..=365 {
-        datetime_sub -= TimeDelta::days(1);
-        assert_eq!(datetime_sub, datetime - TimeDelta::days(i))
+        datetime_sub -= TimeDelta::try_days(1).unwrap();
+        assert_eq!(datetime_sub, datetime - TimeDelta::try_days(i).unwrap())
     }
 }
 
--- a/src/datetime/tests.rs
+++ b/src/datetime/tests.rs
@@ -307,7 +307,7 @@ fn test_nanosecond_range() {
     // Far beyond range
     let maximum = "2262-04-11T23:47:16.854775804UTC";
     let parsed: DateTime<Utc> = maximum.parse().unwrap();
-    let beyond_max = parsed + TimeDelta::days(365);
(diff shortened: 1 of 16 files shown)


src/naive/date/mod.rs
```diff
@@ -1140,7 +1140,8 @@ impl NaiveDate {
         let (year2_div_400, year2_mod_400) = div_mod_floor(year2, 400);
         let cycle1 = yo_to_cycle(year1_mod_400 as u32, self.ordinal()) as i64;
         let cycle2 = yo_to_cycle(year2_mod_400 as u32, rhs.ordinal()) as i64;
-        TimeDelta::days((year1_div_400 as i64 - year2_div_400 as i64) * 146_097 + (cycle1 - cycle2))
+        let days = (year1_div_400 as i64 - year2_div_400 as i64) * 146_097 + (cycle1 - cycle2);
+        expect!(TimeDelta::try_days(days), "range of NaiveDate is MUCH less than TimeDelta")
```
djc, line 1159: This error message is a bit too opinionated for my taste, can we just say it is out of range?

src/naive/datetime/mod.rs
```diff
@@ -527,14 +519,9 @@ impl NaiveDateTime {
...
-        let (time, rhs) = self.time.overflowing_add_signed(rhs);
-
-        // early checking to avoid overflow in TimeDelta::seconds
-        if rhs <= (-1 << MAX_SECS_BITS) || rhs >= (1 << MAX_SECS_BITS) {
-            return None;
-        }
-
-        let date = try_opt!(self.date.checked_add_signed(TimeDelta::seconds(rhs)));
+        let (time, rhs_days) = self.time.overflowing_add_signed(rhs);
+        let rhs_days = try_opt!(TimeDelta::try_seconds(rhs_days));
```
djc, line 532: It seems pretty confusing to pass something called `rhs_days` to `try_seconds()`.

src/datetime/tests.rs
```diff
@@ -307,7 +307,7 @@ fn test_nanosecond_range() {
     // Far beyond range
     let maximum = "2262-04-11T23:47:16.854775804UTC";
     let parsed: DateTime<Utc> = maximum.parse().unwrap();
-    let beyond_max = parsed + TimeDelta::days(365);
+    let beyond_max = parsed + Days::new(365);
```
djc, line 310: Hmm, maybe we should resurrect your work on `CalendarDuration`?

src/time_delta.rs
```diff
@@ -102,6 +102,10 @@ impl TimeDelta {
     /// Panics when the duration is out of bounds.
     #[inline]
     #[must_use]
+    #[deprecated(
+        since = "0.4.35",
+        note = "Use `TimeDelta::try_weeks` instead or consider the `Days` type"
```
djc, line 107: So I don't really want to recommend the `Days` type here.

src/naive/date/mod.rs
pitdicker: It is const but otherwise functionally the same as
```rust
TimeDelta::try_days(days).unwrap() // range of NaiveDate is MUCH less than TimeDelta
```

The doc comment a little above says:
```
    /// This does not overflow or underflow at all,
    /// as all possible output fits in the range of `TimeDelta`.
```

The range of `TimeDelta` is ca. 585 million years, the range of `NaiveDate` ca. 525.000 years.

I can change it to `"always in rage"`.

src/datetime/tests.rs
pitdicker: I'd like to. Part 1 is ready for review :smile:. Maybe wait a few more weeks until we are good underway with the `Result` stuff?

src/naive/datetime/mod.rs
pitdicker: I suppose `remainder` is less confusing.

src/naive/date/mod.rs
djc: That works for me.

src/naive/datetime/mod.rs
djc: Sounds good.

src/naive/datetime/mod.rs
```diff
@@ -527,14 +519,9 @@ impl NaiveDateTime {
...
     pub const fn checked_add_signed(self, rhs: TimeDelta) -> Option<NaiveDateTime> {
-        let (time, rhs) = self.time.overflowing_add_signed(rhs);
-
-        // early checking to avoid overflow in TimeDelta::seconds
-        if rhs <= (-1 << MAX_SECS_BITS) || rhs >= (1 << MAX_SECS_BITS) {
-            return None;
-        }
-
-        let date = try_opt!(self.date.checked_add_signed(TimeDelta::seconds(rhs)));
+        let (time, rhs_days) = self.time.overflowing_add_signed(rhs);
```
djc, line 531: `rhs_days` -> `remainder` here as well?

djc (approved):

src/naive/datetime/mod.rs
pitdicker: Oops, stupid!

PR #1451, opened 2024-02-21
Rustfmt doctests

pitdicker:
The formatting of the examples in our documentation doesn't match the rustfmt formatted style of modern code. Also there are some very long lines in there.

I wished there was a way to run rustfmt on the doctests... Turns out there is on nightly!
```
cargo fmt -- --config format_code_in_doc_comments=true
```

It is still unstable, and with good reason apparently. It and messes up our doc comments and regular comments in a couple of places. And it makes changes that conflict with regular rustfmt.

In this PR I took the formatted doctests and filtered out the weird parts.

### DIFF
--- a/src/time_delta.rs
+++ b/src/time_delta.rs
@@ -670,6 +670,7 @@ mod tests {
         assert_eq!(TimeDelta::milliseconds(-999).num_seconds(), 0);
         assert_eq!(TimeDelta::milliseconds(-1001).num_seconds(), -1);
     }
+
     #[test]
     fn test_duration_seconds_max_allowed() {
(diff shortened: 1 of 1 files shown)


djc:
We should probably come up with a better solution for these multi-line chains...

pitdicker:
Yes. At least the unwraps should become a `?` when the methods return a `Result`.

pitdicker:
It missed the doctests that were somewhat hidden with `#![cfg_attr(feature = "clock", doc = "```rust")]`, added them now.

djc (approved):

PR #1455, opened 2024-02-23
Improve docs for crate features

### LINKED ISSUE #1434: Inconsistent crate features documentation
PR #1343 added a new `now` feature, however, this feature is only documented in `README.md` and not in the features list on `docs.rs`.

Compare README.md:
https://github.com/chronotope/chrono/blob/3016c30014064bd1c67d285bd23c9e232b9efb80/README.md?plain=1#L49-L67

With the crate rustdocs:
https://github.com/chronotope/chrono/blob/3016c30014064bd1c67d285bd23c9e232b9efb80/src/lib.rs#L20-L44

I see there are also some differences in the other features listed under "optional features" too.

edmorley:
The `chrono` crate makes use of several Cargo features:
https://github.com/chronotope/chrono/blob/02c68d69a1ff8e6461384a770b87737b5096eae3/Cargo.toml#L19-L46

These features are documented in both `README.md` and the rustdocs in `lib.rs, however, the lists in each place were missing some features and also inconsistent in their descriptions.

I've also removed the unused `[wasm-bindgen]` link definition.

Fixes #1434.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,21 +50,31 @@ See [docs.rs](https://docs.rs/chrono/latest/chrono/) for the API reference.
 
 Default features:
 
-* `alloc`: Enable features that depend on allocation (primarily string formatting)
+* `alloc`: Enable features that depend on allocation (primarily string formatting).
 * `std`: Enables functionality that depends on the standard library. This is a superset of `alloc`
   and adds interoperation with standard library types and traits.
 * `clock`: Enables reading the local timezone (`Local`). This is a superset of `now`.
-* `now`: Enables reading the system time (`now`)
+* `now`: Enables reading the system time (`now`).
 * `wasmbind`: Interface with the JS Date API for the `wasm32` target.
 
 Optional features:
 
-* `serde`: Enable serialization/deserialization via serde.
-* `rkyv`: Enable serialization/deserialization via rkyv.
+* `serde`: Enable serialization/deserialization via [serde].
+* `rkyv`: Deprecated, use the `rkyv-*` features.
+* `rkyv-16`: Enable serialization/deserialization via [rkyv], using 16-bit integers for integral `*size` types.
+* `rkyv-32`: Enable serialization/deserialization via [rkyv], using 32-bit integers for integral `*size` types.
+* `rkyv-64`: Enable serialization/deserialization via [rkyv], using 64-bit integers for integral `*size` types.
+* `rkyv-validation`: Enable rkyv validation support using `bytecheck`.
 * `rustc-serialize`: Enable serialization/deserialization via rustc-serialize (deprecated).
-* `arbitrary`: construct arbitrary instances of a type with the Arbitrary crate.
+* `arbitrary`: Construct arbitrary instances of a type with the Arbitrary crate.
 * `unstable-locales`: Enable localization. This adds various methods with a `_localized` suffix.
   The implementation and API may change or even be removed in a patch release. Feedback welcome.
+* `oldtime`: This feature no longer has any effect; it used to offer compatibility with the `time` 0.1 crate.
+
+Note: The `rkyv{,-16,-32,-64}` features are mutually exclusive.
+
+[serde]: https://github.com/serde-rs/serde
+[rkyv]: https://github.com/rkyv/rkyv
 
 ## Rust version requirements
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24,30 +24,35 @@
 //!
 //! Default features:
 //!
-//! - `alloc`: Enable features that depend on allocation (primarily string formatting)
+//! - `alloc`: Enable features that depend on allocation (primarily string formatting).
 //! - `std`: Enables functionality that depends on the standard library. This
 //!   is a superset of `alloc` and adds interoperation with standard library types
 //!   and traits.
-//! - `clock`: Enables reading the system time (`now`) that depends on the standard library for
-//! UNIX-like operating systems and the Windows API (`winapi`) for Windows.
+//! - `clock`: Enables reading the local timezone (`Local`). This is a superset of `now`.
+//! - `now`: Enables reading the system time (`now`).
 //! - `wasmbind`: Interface with the JS Date API for the `wasm32` target.
 //!
 //! Optional features:
 //!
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -28,26 +28,32 @@
...
-//! - [`serde`][]: Enable serialization/deserialization via serde.
-//! - `rkyv`: Enable serialization/deserialization via rkyv.
+//! - `serde`: Enable serialization/deserialization via [serde].
+//! - `rkyv`: Enable serialization/deserialization via [rkyv]. This is equivalent to `rkyv-32`.
+//! - `rkyv-16`: Enable serialization/deserialization via [rkyv], using 16-bit integers for integral `*size` types.
+//! - `rkyv-32`: Enable serialization/deserialization via [rkyv], using 32-bit integers for integral `*size` types.
+//! - `rkyv-64`: Enable serialization/deserialization via [rkyv], using 64-bit integers for integral `*size` types.
+//! - `rkyv-validation`: Enable rkyv validation support using `bytecheck`.
+//! - `rustc-serialize`: Enable serialization/deserialization via rustc-serialize (deprecated).
 //! - `arbitrary`: construct arbitrary instances of a type with the Arbitrary crate.
```
pitdicker, line 44: Now that everything starts with a capital, update this line also?

src/lib.rs
```diff
@@ -28,26 +28,32 @@
...
+//! - `clock`: Enables reading the local timezone (`Local`). This is a superset of `now`.
+//! - `now`: Enables reading the system time (`now`).
 //! - `wasmbind`: Interface with the JS Date API for the `wasm32` target.
 //!
 //! Optional features:
 //!
-//! - [`serde`][]: Enable serialization/deserialization via serde.
-//! - `rkyv`: Enable serialization/deserialization via rkyv.
+//! - `serde`: Enable serialization/deserialization via [serde].
+//! - `rkyv`: Enable serialization/deserialization via [rkyv]. This is equivalent to `rkyv-32`.
```
pitdicker, line 38: We should probably discourage the use of this feature in the documentation, we only have it for compatibility. The replacement features where introduced in #1368.
No preference from my side on how to word it. Maybe just "`rkyv`: Deprecated, use the `kkyv-*` features"?

pitdicker (commented):
Thank you @edmorley for straightening this out. One nit and one suggestion.

src/lib.rs
edmorley: Sounds good to me!

edmorley:
I've force pushed rather than adding a "Fix review comments" commit since it looks like this repo doesn't normally use squash merges. To make the re-review easier, the diff between this version and the last can be seen here:
https://github.com/chronotope/chrono/compare/88eaa06ee4b8b5551efaceceaf62d7a5227eb0fd..1ca1df1654755f01402c5986802169802a1c7f3c

pitdicker (approved):
Thank you!
And thanks for looking at how we like the commits :smile:.

@djc will probably also want to have a look.

README.md
```diff
@@ -54,17 +54,27 @@ Default features:
...
-* `now`: Enables reading the system time (`now`)
+* `now`: Enables reading the system time (`now`).
 * `wasmbind`: Interface with the JS Date API for the `wasm32` target.
 
 Optional features:
 
-* `serde`: Enable serialization/deserialization via serde.
-* `rkyv`: Enable serialization/deserialization via rkyv.
+* `serde`: Enable serialization/deserialization via [serde].
+* `rkyv`: Deprecated, use the `kkyv-*` features.
```
djc, line 63: Nit: "kkyv" -> "rkyv".

README.md
```diff
@@ -54,17 +54,27 @@ Default features:
...
+* `rkyv-16`: Enable serialization/deserialization via [rkyv], using 16-bit integers for integral `*size` types.
+* `rkyv-32`: Enable serialization/deserialization via [rkyv], using 32-bit integers for integral `*size` types.
+* `rkyv-64`: Enable serialization/deserialization via [rkyv], using 64-bit integers for integral `*size` types.
+* `rkyv-validation`: Enable rkyv validation support using `bytecheck`.
 * `rustc-serialize`: Enable serialization/deserialization via rustc-serialize (deprecated).
-* `arbitrary`: construct arbitrary instances of a type with the Arbitrary crate.
+* `arbitrary`: Construct arbitrary instances of a type with the Arbitrary crate.
 * `unstable-locales`: Enable localization. This adds various methods with a `_localized` suffix.
   The implementation and API may change or even be removed in a patch release. Feedback welcome.
+* `oldtime`: This feature no langer has a function, but once offered compatibility with the `time` 0.1 crate.
```
djc, line 72: "langer" -> "longer". The formulation "This feature no longer has a function" reads odd to me, maybe replace with something like "This feature no longer has any effect; it used to offer..."?

djc (approved):
Consistency is good, but on balance I would probably prefer to avoid capitalizing after `:`.

README.md
edmorley: Agree the wording is strange - I copied it verbatim from `lib.rs`. I'll adjust both places now.

README.md
edmorley: Ah I copied this verbatim from the review suggestion (but didn't notice there was a typo in the suggestion :-))

edmorley:
> Consistency is good, but on balance I would probably prefer to avoid capitalizing after :.

So I tried switching to using lowercase after the colons, however, that didn't work so well for the items in the list that have multiple sentences after the colon, since then the first sentence is not capitalised, but later ones are.

Searching around, it seems that it's acceptable to capitalise after a colon when using a complete sentence:
https://www.grammarly.com/blog/capitalization-after-colons/

As such, I've added a couple of missing trailing periods, making all entries be complete sentences, so that they are suitable for capitalisation.

edmorley:
I've force pushed with the latest changes; diff here:
https://github.com/chronotope/chrono/compare/1ca1df1654755f01402c5986802169802a1c7f3c..ed6fb2592f66f497169ed9bb322d2c5e6e51643d

pitdicker:
Thank you @edmorley.

PR #1459, opened 2024-02-25
Simplify error value of `parse_internal`

pitdicker:
If there was an error `parse_internal` returned a tuple with the remainder of the input and the error.
None of the callers use the remainder portion, so we can simplify a bit.


djc (approved):
Nice!

PR #1460, opened 2024-02-25
Convert `NaiveDate::{from_yo|from_isoywd}` to return `Result`

pitdicker:
cc @Zomtir. I would like to depend on these methods when working on the `Parsed` type, and you don't seem to have started on them yet.

### DIFF
--- a/src/naive/date/mod.rs
+++ b/src/naive/date/mod.rs
@@ -216,55 +216,58 @@ impl NaiveDate {
     ///
     /// # Errors
     ///
-    /// Returns `None` if:
-    /// - The specified week does not exist in that year (for example 2023 week 53).
-    /// - The value for `week` is invalid (for example: `0`, `60`).
-    /// - If the resulting date is out of range for `NaiveDate`.
+    /// This method returns:
+    /// - [`Error::DoesNotExist`] if specified week does not exist in that year (for example 2023
+    ///    week 53).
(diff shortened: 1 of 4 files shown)


pitdicker:
Simplified `from_isoywd` a bit and made that into a commit that can apply to the main branch.

src/naive/date/mod.rs
```diff
@@ -216,55 +216,59 @@ impl NaiveDate {
...
+    pub const fn from_isoywd(year: i32, week: u32, weekday: Weekday) -> Result<NaiveDate, Error> {
         let flags = YearFlags::from_year(year);
         let nweeks = flags.nisoweeks();
         if week == 0 || week > nweeks {
-            return None;
+            if week == 0 || week > 53 {
+                return Err(Error::InvalidArgument);
+            } else {
+                return Err(Error::DoesNotExist);
+            }
```
djc, line 271: Nit, would suggest the following:

```rust
return Err(match week == 0 || week > 53 {
    true => Error::InvalidArgument,
    false => Error::DoesNotExist,
});
```

djc (approved):

pitdicker:
I'll keep this PR open a little longer until we have a new merge of main with the 0.5.x branch.

src/naive/date/mod.rs
pitdicker: For the next time just say 'Use match'. I should know by now :laughing:.

PR #1462, opened 2024-02-26
Update windows-bindgen to 0.53

pitdicker:
See #1461. @dependabot can't update the version line in the generated file.

Version 0.53 now build with rust 1.70 instead of 1.73, and defines it as the MSRV https://github.com/microsoft/windows-rs/pull/2766.


Cargo.toml
```diff
@@ -53,7 +53,7 @@ js-sys = { version = "0.3", optional = true }       # contains FFI bindings for
 windows-targets = { version = "0.52", optional = true }
 
 [target.'cfg(windows)'.dev-dependencies]
-windows-bindgen = { version = "0.52" }
+windows-bindgen = { version = "0.53" } # The MSRV of its windows-metatada 0.53 dependency is 1.70
```
djc, line 56: Since this is a dev-dependency, why is this comment useful?

djc (approved):

Cargo.toml
pitdicker: Because the version requirement bit me once :smile:.
Just serving as documentation of the minimum rust version we need to do development on chrono on Windows.

PR #1463, opened 2024-02-27
Improve error checking of `SYSTEMTIME` values

pitdicker:
Add more checks that values are in range in the Windows `SYSTEMTIME` type.
We should be able to trust the Windows API; still it is good to have complete checks.

### DIFF
--- a/src/offset/local/windows.rs
+++ b/src/offset/local/windows.rs
@@ -9,7 +9,6 @@
 // except according to those terms.
 
 use std::cmp::Ordering;
-use std::convert::TryFrom;
 use std::mem::MaybeUninit;
 use std::ptr;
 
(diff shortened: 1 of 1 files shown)


src/offset/local/windows.rs
```diff
@@ -144,39 +143,54 @@ impl TzInfo {
...
 }
 
-fn system_time_from_naive_date_time(st: SYSTEMTIME, year: i32) -> Option<NaiveDateTime> {
+fn naive_date_time_from_system_time(
+    st: SYSTEMTIME,
+    year: i32,
+) -> Result<Option<NaiveDateTime>, ()> {
     if st.wYear == 0 && st.wMonth == 0 {
-        return None; // No DST transitions for this year in this timezone.
+        return Ok(None); // No DST transitions for this year in this timezone.
```
djc, line 165: Why is this? It doesn't intuitively make sense to me that this needs to return `Result<Option<_>, _>` instead of just `Result`.

pitdicker: In a way `SYSTEMTIME` is an `Option`-like type. If it is `0` it means there is no transition date that year. If it is non-zero, it can be a concrete date or a rule.

We can resolve the rule to a concrete date within this function, so besides returning `NaiveDateTime` we only need to encode two other cases: `SYSTEMTIME` was `None`, or there was an error.

djc: The function is called `naive_date_time_from_system_time()`, which makes it non-obvious to me why we would need to return `None` in case there are no DST transitions.

pitdicker: I'll add a doc comment and think up a better name :smile:.

djc: Maybe the name is fine -- the context of `SYSTEMTIME` being `Option`-like wasn't clear to me; if we can resolve that with documentation it would okay.

pitdicker:
@djc Can I ask for a second look at this PR when you have the time? It seems like you were mostly done with the review.

djc (approved):

pitdicker:
Thank you for all the reviews past days!

PR #1465, opened 2024-02-27
Add range checks to `Parsed::set` and hide fields in docs

pitdicker:
This does not remove the range checks from the `Parsed::to_*` methods, because as long as the fields are public they can be set to an out of range value. Something to do on the 0.5.x branch.

### DIFF
--- a/src/format/parsed.rs
+++ b/src/format/parsed.rs
@@ -125,89 +125,47 @@ use crate::{DateTime, Datelike, TimeDelta, Timelike, Weekday};
 #[allow(clippy::manual_non_exhaustive)]
 #[derive(Clone, PartialEq, Eq, Debug, Default, Hash)]
 pub struct Parsed {
-    /// Year.
-    ///
-    /// This can be negative, unlike [`year_div_100`](#structfield.year_div_100)
-    /// and [`year_mod_100`](#structfield.year_mod_100) fields.
+    #[doc(hidden)]
     pub year: Option<i32>,
-
-    /// Year divided by 100. Implies that the year is >= 1 BCE when set.
-    ///
-    /// Due to the common usage, if this field is missing but
-    /// [`year_mod_100`](#structfield.year_mod_100) is present,
-    /// it is inferred to 19 when `year_mod_100 >= 70` and 20 otherwise.
+    #[doc(hidden)]
     pub year_div_100: Option<i32>,
-
-    /// Year modulo 100. Implies that the year is >= 1 BCE when set.
+    #[doc(hidden)]
     pub year_mod_100: Option<i32>,
-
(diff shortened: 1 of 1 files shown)


src/format/parsed.rs
```diff
@@ -575,6 +577,175 @@ impl Parsed {
         set_if_consistent(&mut self.offset, i32::try_from(value).map_err(|_| OUT_OF_RANGE)?)
     }
 
+    /// Get the 'year' field if set.
+    ///
+    /// See also [`set_year`](Parsed::set_year).
+    #[inline]
+    pub fn year(&self) -> Option<i32> {
```
djc, line 542: If we're going to have these getters we can also refer to these in those earlier docs where you didn't want to refer to the fields directly?

djc (approved):

src/format/parsed.rs
pitdicker: I'll add similar "See also ..." lines to the `set` methods.

djc: That's not really what I'm after. I don't like `'year divided by 100'" "descriptions" because they are kind of imprecise references to something that actually has a precise reference in the code, with more information. So IMO we should get rid of the "descriptions" in favor of precise references.

pitdicker: I am not sure I like linking to a method as if it is a field any better.
Added as a commit. Is this more what you had in mind?

pitdicker:
@djc When you have time, can you look at the direction of the documentation in the last commit? Then I'll merge squash it into the right commit.

I would like to start building on top of this PR :smile:.

src/format/parsed.rs
pitdicker: Can you write one line as an example of a sentence? I only didn't do it until now because I can't come up with something that makes sense to me.

pitdicker: Example of our current options:

Set the 'year' field to the given value.
Set the `year` field to the given value.
Set the `year()` field to the given value.

Set the 'year divided by 100' field that is part of an ISO 8601 week date to the given value.
Set the `isoyear_div_100` field, that is part of an ISO 8601 week date, to the given value.
Set the `isoyear_div_100()` field, that is part of an ISO 8601 week date, to the given value.

Set the 'week number starting with Sunday' field to the given value.
Set the `week_from_sun` week number field to the given value.
Set the `week_from_sun()` week number field to the given value.

djc: I think I'd prefer the second option in the end. We'll eventually get [read-only fields](https://rust-lang.github.io/rfcs/3323-restrictions.html) at which point we could conceivably make these `pub mut(crate)` (or however it ends up being spelled).

djc (approved):

Issue #1448, opened 2024-02-19
0.5.x: Timezone conversion result (`LocalResult`)


pitdicker:
*Part of this has been discussed before in #716 and #1049, and was my motivation to start contributing to chrono in #1017.*

The calculation to go from a local time (wall clock time) to an instant in UTC can end up in three cases:
- a single, simple result
- an ambiguous result when the clock is turned backwards during transition due to for example DST.
- no result when the clock is turned forwards during transition due to for example DST.

Our current `LocalResult` type matches those three cases:
```rust
pub enum LocalResult<T> {
    Single(T),
    Ambiguous(T, T), /* first, last */
    None,
}
```

The type of `T` is usually an offset or `DateTime`.

As we discovered this simple definition has some issues.

### No way to distinguish errors from the `None` case
The type that implements `TimeZone` and returns a `LocalResult` may not be completely under our/your control. Prime example is our `Local` type. It wraps platform APIs, which can fail for various cases. Our current code returns `LocalResult::None` on such a failure.

Now if a method returns `LocalResult::None` there is no way for the user to know if it is because the datetime falls within a gap, or because there is an error.

### Not enough info to handle the `None` case
Suppose we want to write a function that returns exactly how many seconds have passed since midnight, taking into account DST transitions.
That should be simple. Take the supplied `DateTime`, create a `NaiveDateTime` with the time set to 00:00:00, use `TimeZone::from_local_datetime` to create a `DateTime` at midnight and subtract.

What if some government decided to do a DST transition right at midnight, say from 23:30 to 00:30? `TimeZone::from_local_datetime` would return `LocalResult::None`, and our function can't return a result. However both 23:30 (from before moving the clock forwards) and 00:30 map to the same time in UTC, and that is the time you could consider midnight for this calculation.

`LocalResult::None` would be more useful if it returned more info than `None`, such as the UTC time during the gap and the value of `T` before and after the transition.

### Related problem: wrapping other APIs that don't give enough info
Other APIs may return a single value, even when it would be ambiguous or doesn't exist.
And I suppose there could be API's, like chrono 0.4.x, that can return two results for ambiguous cases but can't return more info for the 'gap' case.

I'm not sure if there is anything special we want to do here. But it may be useful for callers to know that some platform or library may return bogus results around transitions, with no way to detect that.

### When to use a `LocalResult`

[...]

PR #1466, opened 2024-02-28
0.5: Convert API to return `Result`: `Datelike::with_year`

Zomtir:
Follow up commit for #1444. Split from #1445.

- Converts `Datelike::with_year` to return `Result<T,E>`
- Converts the implementations `NaiveDate::with_year` and `NaiveDateTime::with_year` to return `Result<T,E>`
- Converts the implementation `DateTime<Tz>::with_year` to return `Result<T,E>` as well
  - Adds `DateTime<Tz>::map_local_result` as transitionary helper function to prevent dependency on `DateTime<Tz>::map_local`
  - Adds `LocalResult<T>::map_result_unique` to have a transitionary, centralized mapping function until #1448 is settled

The function `map_result_unique` is NOT meant to outline future behaviour in any form. The idea is that there is some centralized mapping to a `Result` that can easily tracked for the purpose of #1448. Local (non-central) mappings will be prone to spread inconsistent behaviour.

However the future of `LocalResult` looks like, it will be easy to track the current usage and adapt the implementation and documentation by searching for the `map_result_unique` usage. For that reason I did not document the behaviour of `map_result_unique` anywhere else than in `src/offset/mod.rs`, but referenced it instead.

cc @pitdicker

### DIFF
--- a/src/naive/date/mod.rs
+++ b/src/naive/date/mod.rs
@@ -1201,44 +1201,39 @@ impl NaiveDate {
     ///
     /// # Errors
     ///
-    /// Returns `None` if:
-    /// - The resulting date does not exist (February 29 in a non-leap year).
-    /// - The year is out of range for a `NaiveDate`.
+    /// Returns:
+    /// - [`Error::DoesNotExist`] if the resulting date does not exist.
+    /// - [`Error::OutOfRange`] if `year` is out of range for a `NaiveDate`.
     ///
     /// # Examples
     ///
     /// ```
-    /// use chrono::NaiveDate;
-    ///
-    /// assert_eq!(
-    ///     NaiveDate::from_ymd(2015, 9, 8).unwrap().with_year(2016),
-    ///     Some(NaiveDate::from_ymd(2016, 9, 8).unwrap())
-    /// );
-    /// assert_eq!(
-    ///     NaiveDate::from_ymd(2015, 9, 8).unwrap().with_year(-308),
-    ///     Some(NaiveDate::from_ymd(-308, 9, 8).unwrap())
-    /// );
+    /// # use chrono::{NaiveDate, Error};
+    /// assert_eq!(NaiveDate::from_ymd(2015, 9, 8)?.with_year(2016), NaiveDate::from_ymd(2016, 9, 8));
+    /// assert_eq!(NaiveDate::from_ymd(2015, 9, 8)?.with_year(-308), NaiveDate::from_ymd(-308, 9, 8));
+    /// # Ok::<(), Error>(())
     /// ```
     ///
-    /// A leap day (February 29) is a case where this method can return `None`.
+    /// A leap day (February 29) in a non-leap year will return [`Err(Error::DoesNotExist)`].
     ///
     /// ```
-    /// # use chrono::NaiveDate;
-    /// assert!(NaiveDate::from_ymd(2016, 2, 29).unwrap().with_year(2015).is_none());
-    /// assert!(NaiveDate::from_ymd(2016, 2, 29).unwrap().with_year(2020).is_some());
+    /// # use chrono::{NaiveDate, Error};
+    /// assert_eq!(NaiveDate::from_ymd(2016, 2, 29)?.with_year(2015), Err(Error::DoesNotExist));
+    /// assert!(NaiveDate::from_ymd(2016, 2, 29)?.with_year(2020).is_ok());
+    /// # Ok::<(), Error>(())
     /// ```
     ///
-    /// Don't use `with_year` if you want the ordinal date to stay the same:
+    /// Don't use `with_year` if you want the ordinal date to stay the same.
     ///
     /// ```
-    /// # use chrono::NaiveDate;
-    /// assert_ne!(
-    ///     NaiveDate::from_yo(2020, 100).unwrap().with_year(2023).unwrap(),
-    ///     NaiveDate::from_yo(2023, 100).unwrap() // result is 2023-101
-    /// );
+    /// # use chrono::{NaiveDate, Error};
+    /// let date = NaiveDate::from_yo(2020, 100)?.with_year(2023);
+    /// assert_ne!(date, NaiveDate::from_yo(2023, 100));
+    /// assert_eq!(date, NaiveDate::from_yo(2023, 99));
+    /// # Ok::<(), Error>(())
     /// ```
     #[inline]
-    pub const fn with_year(&self, year: i32) -> Option<NaiveDate> {
+    pub const fn with_year(&self, year: i32) -> Result<NaiveDate, Error> {
         // we need to operate with `mdf` since we should keep the month and day number as is
         let mdf = self.mdf();
 
@@ -1246,7 +1241,7 @@ impl NaiveDate {
         let flags = YearFlags::from_year(year);
         let mdf = mdf.with_flags(flags);
 
-        ok!(NaiveDate::from_mdf(year, mdf))
+        NaiveDate::from_mdf(year, mdf)
     }
 
     /// Makes a new `NaiveDate` with the month number (starting from 1) changed.
@@ -1276,7 +1271,7 @@ impl NaiveDate {
     /// use chrono::{Datelike, Error, NaiveDate};
     ///
     /// fn with_year_month(date: NaiveDate, year: i32, month: u32) -> Option<NaiveDate> {
-    ///     date.with_year(year)?.with_month(month)
+    ///     date.with_year(year).ok()?.with_month(month)
     /// }
     /// let d = NaiveDate::from_ymd(2020, 2, 29).unwrap();
     /// assert!(with_year_month(d, 2019, 1).is_none()); // fails because of invalid intermediate value
--- a/src/naive/datetime/mod.rs
+++ b/src/naive/datetime/mod.rs
@@ -728,28 +728,22 @@ impl NaiveDateTime {
     ///
     /// # Errors
     ///
-    /// Returns `None` if:
-    /// - The resulting date does not exist (February 29 in a non-leap year).
-    /// - The year is out of range for a `NaiveDate`.
+    /// Returns:
+    /// - [`Error::DoesNotExist`] if the resulting date does not exist.
+    /// - [`Error::OutOfRange`] if `year` is out of range for a `NaiveDate`.
     ///
-    /// # Example
+    /// # Examples
     ///
     /// ```
-    /// use chrono::{NaiveDate, NaiveDateTime};
-    ///
-    /// let dt: NaiveDateTime = NaiveDate::from_ymd(2015, 9, 25).unwrap().and_hms(12, 34, 56).unwrap();
-    /// assert_eq!(
-    ///     dt.with_year(2016),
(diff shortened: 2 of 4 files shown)


pitdicker:
@Zomtir Thank you!
Why do you keep trying to work on something that I consider one of the last pieces to convert, and don't pick something that is easy and ready without needing workarounds? I keep suggesting it friendly:

From https://github.com/chronotope/chrono/pull/1444#issuecomment-1949143027
> It seems good to do the conversion with only one method or a very small number of methods per commit. And starting with **methods in the `naive` module that don't depend on pieces that have not yet been converted**.
> 
> Converting `LocalResult`, `ParseError` and `TimeDelta` require some preparations first that I'll look into. **Let's ignore anything around those for now.**

From https://github.com/chronotope/chrono/pull/1444#issuecomment-1950285985
> **Could you skip the `DateTime` type?** The `map_local` type is going to pass on multiple error causes, such as `OutOfRange`, `DoesNotExist`, OS errors, and trouble with a TZ string or TZif file. I'd like to be a bit more careful with that type and leave that to its own PR.
> 
> **We should also have a discussion around the `Datelike` and `Timelike` traits before converting them.** For the naive types their signature is fine. For `DateTime` it would be better if they returned `LocalResult` to give users a way to deal with timezone issues (in my opinion). See #1050.

Repeated in https://github.com/chronotope/chrono/pull/1445#issuecomment-1951912557.

From https://github.com/chronotope/chrono/pull/1445#issuecomment-1952075921
> > That is not addressing the problem. I cannot change the signature of an implementation without changing the signature of the trait.
> 
> True. **That is why I suggested not touching the trait in this PR yet**, and wait a bit.

And in https://github.com/chronotope/chrono/pull/1445#issuecomment-1963398311
> May I suggest the following methods as a first step ...

Zomtir:
@pitdicker I have neither issue with the pull request idling until it gets useful and/or rebasing it once changes are made. Don't worry.

I am looking into `from_num_days_from_ce` as we speak and will leave a comment on #1445.

pitdicker:
With #1490 these methods are no longer part of a trait and can be adjusted one type at a time.
Now is a good time to work on the methods on `NaiveDate`, `NaiveTime` and `NaiveDateTime`.
I'd still like to avoid `DateTime`.

Zomtir:
I rebased the changes onto the the current 0.5.x. Now only `NaiveDate` and `NaiveDateTime` are affected.

- API change of `with_year` to return `Result`
- Linking is now more prominent in the documentation to follow the 'DRY' principle. There is little use in repeating the same information when the function has no own logic (e.g. `NaiveDateTime::with_year` directly calls `NaiveDate::with_year`).
- For the same reason the examples in `NaiveDateTime::with_year` now only provides an example where the *NaiveDateTime* is used.

CC @pitdicker

src/naive/date/mod.rs
```diff
@@ -1195,58 +1195,51 @@ impl NaiveDate {
 
     /// Makes a new `NaiveDate` with the year number changed, while keeping the same month and day.
     ///
-    /// This method assumes you want to work on the date as a year-month-day value. Don't use it if
-    /// you want the ordinal to stay the same after changing the year, of if you want the week and
-    /// weekday values to stay the same.
```
pitdicker, line 1200: I moved this documentation from the `Datelike` trait to here in #1476. Did you remove it on purpose or is it a rebase thing?

src/naive/date/mod.rs
```diff
@@ -1195,58 +1195,51 @@ impl NaiveDate {
...
-    ///     NaiveDate::from_ymd(2015, 9, 8).unwrap().with_year(2016),
-    ///     Some(NaiveDate::from_ymd(2016, 9, 8).unwrap())
-    /// );
-    /// assert_eq!(
-    ///     NaiveDate::from_ymd(2015, 9, 8).unwrap().with_year(-308),
-    ///     Some(NaiveDate::from_ymd(-308, 9, 8).unwrap())
-    /// );
+    /// # use chrono::{NaiveDate, Error};
+    /// assert_eq!(NaiveDate::from_ymd(2015, 9, 8)?.with_year(2016),
+    ///            NaiveDate::from_ymd(2016, 9, 8));
```
pitdicker, line 1208: This doesn't follow the rustfmt style anymore #1451.
I run it with `cargo fmt -- --config format_code_in_doc_comments=true` and discard any stray changes.

src/naive/date/mod.rs
```diff
@@ -1195,58 +1195,51 @@ impl NaiveDate {
...
+    /// assert_eq!(NaiveDate::from_ymd(2015, 9, 8)?.with_year(-308),
+    ///            NaiveDate::from_ymd(-308, 9, 8));
+    /// # Ok::<(), Error>(())
     /// ```
     ///
-    /// A leap day (February 29) is a case where this method can return `None`.
+    /// A leap day (February 29) in a non-leap year will return [`Err(Error::DoesNotExist)`].
```
pitdicker, line 1214: :+1: more precise.

src/naive/date/mod.rs
```diff
@@ -1195,58 +1195,51 @@ impl NaiveDate {
...
     ///
     /// ```
-    /// # use chrono::NaiveDate;
-    /// assert!(NaiveDate::from_ymd(2016, 2, 29).unwrap().with_year(2015).is_none());
-    /// assert!(NaiveDate::from_ymd(2016, 2, 29).unwrap().with_year(2020).is_some());
+    /// # use chrono::{NaiveDate, Error};
+    /// assert!(NaiveDate::from_ymd(2016, 2, 29)?.with_year(2015).is_err());
```
pitdicker, line 1218:
```suggestion
    /// assert_eq!(NaiveDate::from_ymd(2016, 2, 29)?.with_year(2015), Err(Error::DoesNotExist));
```

src/naive/date/mod.rs
```diff
@@ -1195,58 +1195,51 @@ impl NaiveDate {
...
     ///
-    /// Don't use `with_year` if you want the ordinal date to stay the same:
+    /// Don't use `with_year` if you want the ordinal date to stay the same.
     ///
     /// ```
-    /// # use chrono::NaiveDate;
+    /// # use chrono::{NaiveDate, Error};
     /// assert_ne!(
-    ///     NaiveDate::from_yo(2020, 100).unwrap().with_year(2023).unwrap(),
+    ///     NaiveDate::from_yo(2020, 100).unwrap().with_year(2023)?,
```
pitdicker, line 1228:
```suggestion
    ///     NaiveDate::from_yo(2020, 100)?.with_year(2023)?,
```

src/naive/datetime/mod.rs
```diff
@@ -728,28 +728,20 @@ impl NaiveDateTime {
     ///
     /// # Errors
     ///
-    /// Returns `None` if:
-    /// - The resulting date does not exist (February 29 in a non-leap year).
-    /// - The year is out of range for a `NaiveDate`.
+    /// - See [`NaiveDate::with_year`] if the intermediate conversion to [`NaiveDate`] fails.
+    /// - Cannot fail on the [`NaiveTime`] component.
```
pitdicker, line 732: I prefer to just repeat the error cases instead of letting users chaise links in the documentation. It is not going to increase the number of lines :smile:.

src/naive/datetime/mod.rs
```diff
@@ -728,28 +728,20 @@ impl NaiveDateTime {
...
-    /// use chrono::{NaiveDate, NaiveDateTime};
-    ///
-    /// let dt: NaiveDateTime = NaiveDate::from_ymd(2015, 9, 25).unwrap().and_hms(12, 34, 56).unwrap();
-    /// assert_eq!(
-    ///     dt.with_year(2016),
-    ///     Some(NaiveDate::from_ymd(2016, 9, 25).unwrap().and_hms(12, 34, 56).unwrap())
-    /// );
-    /// assert_eq!(
-    ///     dt.with_year(-308),
-    ///     Some(NaiveDate::from_ymd(-308, 9, 25).unwrap().and_hms(12, 34, 56).unwrap())
```
pitdicker, line 747: Why remove this example?

pitdicker (commented):
A couple of comments. I think it is mostly caused by the rebase?

src/naive/date/mod.rs
Zomtir: Rebase artifact... will fix.

src/naive/date/mod.rs
Zomtir: Thanks, I was getting crazy with the comments not formatting automatically.

pitdicker: :laughing: With all the changes the renames and `?`s we're  introducing I am glad we have this. Needs nightly though and has its quicks. (And I have a couple of compiles ICEs that I should report...)

src/naive/datetime/mod.rs
pitdicker: The next level is `DateTime`, and that only introduces one more case through the `LocalResult` type (by its new name by then). And that will be encoded in the method signature. I'm not afraid yet.

When it comes to the parsing methods there is an argument for linking to the documentation of `Parsed`. Let's see if it gets unwieldy.

src/naive/date/mod.rs
Zomtir: I will insert an assert_eq as well.

src/naive/datetime/mod.rs
Zomtir: No clue what happened there.

src/naive/date/mod.rs
```diff
@@ -1201,52 +1201,46 @@ impl NaiveDate {
...
     /// ```
     ///
-    /// Don't use `with_year` if you want the ordinal date to stay the same:
+    /// Don't use `with_year` if you want the ordinal date to stay the same.
     ///
     /// ```
-    /// # use chrono::NaiveDate;
-    /// assert_ne!(
-    ///     NaiveDate::from_yo(2020, 100).unwrap().with_year(2023).unwrap(),
-    ///     NaiveDate::from_yo(2023, 100).unwrap() // result is 2023-101
```
Zomtir, line 1237: So actually this comment was wrong and the result is 99...

src/naive/date/mod.rs
```diff
@@ -1201,52 +1201,46 @@ impl NaiveDate {
     ///
     /// # Errors
     ///
-    /// Returns `None` if:
-    /// - The resulting date does not exist (February 29 in a non-leap year).
-    /// - The year is out of range for a `NaiveDate`.
+    /// - Returns [`Error::DoesNotExist`] when the resulting date does not exist.
```
pitdicker, line 1204: @djc suggested at some point putting 'This method returns:' above the list, and removing 'Returns' from every line. I'm trying to stick to that style but am not consistent yet.

src/naive/date/mod.rs
pitdicker: Oops. Now I wonder who wrote that originally...

src/naive/date/mod.rs
djc: Would prefer that over repeating it.

djc (approved):

src/naive/date/mod.rs
Zomtir: Done

PR #1468, opened 2024-02-28
Convert `FixedOffset::{east, west}` to return `Result`

pitdicker:
The easiest methods to convert yet!

Both `Error::InvalidArgument` and `Error::OutOfRange` could be reasonable return types.

The real-world limit of offsets is currently 14 hours, which puts area's using it 2 hours to the other side of the international date line.
The type limit is generous and somewhat arbitrarily limited to be less than 24 hours. At some point it turned out limiting the offset to less than a day keeps the surrounding code saner.

Because the limit is artificial and specific to chrono I went with `Error::OutOfRange`.

cc @Zomtir.

### DIFF
--- a/src/offset/fixed.rs
+++ b/src/offset/fixed.rs
@@ -11,7 +11,7 @@ use rkyv::{Archive, Deserialize, Serialize};
 
 use super::{LocalResult, Offset, TimeZone};
 use crate::format::{scan, OUT_OF_RANGE};
-use crate::{NaiveDateTime, ParseError};
+use crate::{Error, NaiveDateTime, ParseError};
 
 /// The time zone with fixed offset, from UTC-23:59:59 to UTC+23:59:59.
 ///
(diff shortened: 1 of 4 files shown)


pitdicker:
The spelling of the existing docs doesn't seems entirely correct English. I'll adjust that.

src/offset/fixed.rs
```diff
@@ -154,9 +154,7 @@ impl fmt::Display for FixedOffset {
 impl arbitrary::Arbitrary<'_> for FixedOffset {
     fn arbitrary(u: &mut arbitrary::Unstructured) -> arbitrary::Result<FixedOffset> {
         let secs = u.int_in_range(-86_399..=86_399)?;
-        let fixed_offset = FixedOffset::east(secs)
-            .expect("Could not generate a valid chrono::FixedOffset. It looks like implementation of Arbitrary for FixedOffset is erroneous.");
-        Ok(fixed_offset)
+        Ok(FixedOffset::east(secs).unwrap())
```
Zomtir, line 157: Could the unwrapping be prevented by mapping the error? Untested: `FixedOffset::east(secs).or(Err(arbitrary::Error))`

Zomtir (approved):

src/offset/fixed.rs
pitdicker: The range we supply to `arbitrary::Unstructured::int_in_range` is exactly equal to the range `FixedOffset` supports, and I don't see that changing. This should simply never fail, so I even considered the `expect` to be overkill.

PR #1472, opened 2024-02-29
Disable wasmbind feature by default

### LINKED ISSUE #1164: Disable wasmbind feature by default
When compiling a dependency that transitively includes chrono for a wasm32 target, chrono implicitly pulls in wasm_bindgen. However, this is not desirable if you are not using wasm bindgen, but compiling for a plain wasm target. wasm_bindgen imposes requirements on the output wasm file, e.g. imports it needs from the JS environment, which might not always be available. Assuming that all wasm32-unknown-unknown targets are using wasm_bindgen is not a safe assumption.

If chrono was a direct dependency, I could turn off the default features, but since it is a transitive dependency this is not possible. Therefore I think it should be turned off by default. Turning _on_ features in transitive dependencies is possible via [feature unification](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification). You'd make chrono a direct dependency and add the wasmbind feature, and it will also be enabled for other dependencies. However the opposite is not possible.

erickt:
This removes `wasmbind` from the default feature set, which stops chrono from implicitly depending upon wasm-bindgen and js-sys. This is helpful for a few reasons:

* It reduces the default dependency set by default for non-wasm projects, which shrinks the download size.

* Projects like Fuchsia have a policy where 3rd party crates need to be audited. While we don't use wasm-bindgen, we can't opt out of it by setting `default-features = false` because of [feature unification] ends up enabling chrono's default feature. See this [cargo issue] for more details. `wasm-bindgen` is large and complicated, so it's pretty expensive for us to update.

While I'd love to land this in the main branch, it's technically a breaking issue since downstream dependencies are probably not enabling the `wasmbind` feature if they need it.

Fixes #1164

[feature unification]: https://doc.rust-lang.org/cargo/reference/features.html#feature-unification
[cargo issue]: https://github.com/rust-lang/cargo/issues/4463

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ name = "chrono"
 
 [features]
 # Don't forget to adjust `ALL_NON_EXCLUSIVE_FEATURES` in CI scripts when adding a feature or an optional dependency.
-default = ["clock", "std", "wasmbind"]
+default = ["clock", "std"]
 alloc = []
 libc = []
 winapi = ["windows-targets"]


pitdicker (approved):

djc:
Thanks!

amitu:
I would like to know the plan for releasing 0.5? I need this feature, and if its expected to land shortly I can wait, but if its going to take long I would consider removing `chrono` and try to compile my project somehow else.

pitdicker:
If you are in a position to remove chrono, you probably are also in the position to use it without default features?

Issue #1471, opened 2024-02-28
Deprecate all timestamp-related methods on `NaiveDateTime`


pitdicker:
The definition of a timestamp is that it is a value in UTC.
We don't know how a `NaiveDateTime` relates to UTC. So having methods on this type to convert to and from timestamps is in theory not correct.

The time crate chooses to not have any timestamp related methods on its `PrimitiveDateTime` for this reason.

Workarounds are not that problematic.
If the `NaiveDate` is in UTC `dt.timestamp()` would become `dt.and_utc().timestamp()`.
`NaiveDate::from_timestamp(ts, 0)` would become `Utc.timestamp(ts, 0).naive_utc()`.

Deprecating these methods can help users notice invalid assumptions, and removes 13 methods from our API surface.

djc:
This makes sense to me.

PR #1473, opened 2024-02-29
Deprecate timestamp methods on `NaiveDateTime`

pitdicker:
As discussed in #1471 a timestamp is defined to be in UTC. The `NaiveDateTime` type doesn't know the offset from UTC, so is is technically wrong to have these methods. Users should use the similar methods on the `DateTime<Utc>` type (or the `TimeZone` trait) instead.

Converting from `NaiveDateTime` to `DateTime<Utc>` is simple with `.and_utc()`, and in the other direction with `.naive_utc()`.

------

The first commit introduces two extra functions on `DateTime<Utc>` to get feature parity with `NaiveDateTime`: `from_timestamp_micros` and `from_timestamp_nanos`.

Later commits move over our implementation and tests from `NaiveDateTime` to `DateTime`.

Then we avoid using the timestamp methods on `NaiveDateTime` and deprecate the them.

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -1881,6 +1881,14 @@ where
     }
 }
 
+/// Number of days between Januari 1, 1970 and December 31, 1 BCE which we define to be day 0.
+/// 4 full leap year cycles until December 31, 1600     4 * 146097 = 584388
+/// 1 day until January 1, 1601                                           1
+/// 369 years until Januari 1, 1970                      369 * 365 = 134685
+/// of which floor(369 / 4) are leap years          floor(369 / 4) =     92
+/// except for 1700, 1800 and 1900                                       -3 +
+///                                                                  --------
+///                                                                  719163
 const UNIX_EPOCH_DAY: i64 = 719_163;
 
 #[cfg(all(test, any(feature = "rustc-serialize", feature = "serde")))]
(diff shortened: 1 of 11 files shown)


src/datetime/mod.rs
```diff
@@ -1878,6 +1881,8 @@ where
     }
 }
 
+const UNIX_EPOCH_DAY: i64 = 719_163;
```
djc, line 1884: Pre-existing issue, but can you add a comment on what this number is/how you can check it?

djc (approved):
Good stuff! Let's do a release soon after merging this so we can get some feedback?

src/datetime/mod.rs
pitdicker: Like I know 😆. Added a comment.

pitdicker:
> Let's do a release soon after merging this so we can get some feedback?

Good idea, I'll start preparing release notes.

Do we want to include #1450, deprecating the panicking constructors of `TimeDelta`, as well?
In my opinion if we are going to introduce some churn it is better to do it all at once than do it twice a few weeks apart.
It is now five weeks since introducing their fallible alternatives.

djc:
> Do we want to include #1450, deprecating the panicking constructors of `TimeDelta`, as well? In my opinion if we are going to introduce some churn it is better to do it all at once than do it twice a few weeks apart. It is now five weeks since introducing their fallible alternatives.

Yes, that seems okay.

PR #1474, opened 2024-02-29
Return error when rounding with a zero duration

Dav1dde:
The division by zero (modulo) panics, `duration_round` already checks for 0 and returns the original value.

### DIFF
--- a/src/round.rs
+++ b/src/round.rs
@@ -230,6 +230,9 @@ where
         }
         let stamp =
             naive.and_utc().timestamp_nanos_opt().ok_or(RoundingError::TimestampExceedsLimit)?;
+        if span == 0 {
+            return Ok(original);
+        }
         let delta_down = stamp % span;
         match delta_down.cmp(&0) {
             Ordering::Equal => Ok(original),
@@ -619,6 +622,11 @@ mod tests {
                     .unwrap(),
             )
             .unwrap();
+
+        assert_eq!(
+            dt.duration_trunc(TimeDelta::zero()).unwrap().to_string(),
+            "2016-12-31 23:59:59.175500 UTC"
+        );
 
         assert_eq!(
             dt.duration_trunc(TimeDelta::try_milliseconds(10).unwrap()).unwrap().to_string(),
--- a/src/round.rs
+++ b/src/round.rs
@@ -102,7 +102,7 @@ const fn span_for_digits(digits: u16) -> u32 {
 /// Both rounding and truncating are done via [`TimeDelta::num_nanoseconds`] and
 /// [`DateTime::timestamp_nanos_opt`]. This means that they will fail if either the
 /// `TimeDelta` or the `DateTime` are too big to represented as nanoseconds. They
-/// will also fail if the `TimeDelta` is bigger than the timestamp.
+/// will also fail if the `TimeDelta` is bigger than the timestamp, negative or zero.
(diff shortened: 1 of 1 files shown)


Dav1dde:
codecov error seems unrelated

pitdicker:
@Dav1dde Thank you for fixing this!

My opinion of this module remains low 😞.

In theory rounding or truncating to multiples of 0 (nano)seconds doesn't make sense. Wouldn't it be better to return an error instead?

Dav1dde:
> In theory rounding or truncating to multiples of 0 (nano)seconds doesn't make sense. Wouldn't it be better to return an error instead?

In my usecase returning the original value is what I would need. Thinking about it a bit more, to get that same behaviour it should also be possible to truncate by the smallest represented unit in the timestamp (1 ns), which does make more sense.

But I think being consistent with `duration_round` is better than returning an error, even if it does probably make more sense.

pitdicker:
If @djc agrees I would like to see both rounding and truncating return an error on zero, and leave it up to the caller to use sane arguments. I.e. make not only negative values return `RoundingError::DurationExceedsLimit` but also zero.

djc:
@Dav1dde what are you using these methods for? What are you trying to achieve?

Dav1dde:
> @Dav1dde what are you using these methods for? What are you trying to achieve?

Aggregating data by timestamp with a given granularity. 

Zero was just the first thing I reached to, though 1ns would make technically more sense, it also requires you to know the minimum precision and in an interface that only accepts seconds `1ns` is not possible to supply -> zero being the easy alternative.

Dav1dde:
> @Dav1dde Would you be willing to update this PR to return consistently return an error when rounding to multiples of zero?

Sure, if that is the desired behaviour.

While overall the error makes more sense, I don't think that is a good idea. Because it is different behaviour to what `duration_round` does. A user replacing `duration_round` with `duration_trunc` will now get an error or the other way around instead of an error they get a value. I think the behaviour on chrono should be consistent here.

pitdicker:
Yes, both should error.

Dav1dde:
> Yes, both should error.

Should I update both.

pitdicker:
Please :smile:

src/round.rs
```diff
@@ -102,7 +102,7 @@ const fn span_for_digits(digits: u16) -> u32 {
 /// Both rounding and truncating are done via [`TimeDelta::num_nanoseconds`] and
 /// [`DateTime::timestamp_nanos_opt`]. This means that they will fail if either the
 /// `TimeDelta` or the `DateTime` are too big to represented as nanoseconds. They
-/// will also fail if the `TimeDelta` is bigger than the timestamp.
+/// will also fail if the `TimeDelta` is bigger than the timestamp, negative or zero.
```
pitdicker, line 105: Interesting. Another weirdness in this module. 'will fail if `TimeDelta` is bigger than the timestamp.' So rounding to hours will fail for values within the first hour of 1970-01-01 😞. Don't worry about it for this PR though.

pitdicker (approved):
Thank you, looks good!

We add release notes to GitHub releases, nothing you need to update.

pitdicker:
@djc Do you have time to look at this PR?

djc (approved):
LGTM (but let's squash during merge?).

PR #1412, opened 2024-02-04
Serde rfc 2822

chrisranderson:
> ### Thanks for contributing to chrono!
>If your feature is semver-compatible, please target the main branch;
> for semver-incompatible changes, please target the `0.5.x` branch.

I'm not sure what this means - every change is compatible with semver, right? Are you asking if it's a backwards-incompatible change?

> Please consider adding a test to ensure your bug fix/feature will not break in the future.

Done.

PR #1477, opened 2024-03-01
Add RFC 2822 serde module

### LINKED ISSUE #1292: Offer an option to serialize/deserialize to/from RFC 2822
For work, I'm currently implementing RSS feeds. These use RFC 2822 date/times for the `pubDate` field. It would be nice if there was a `serde::rfc2822` module that I could use similar to the other options we currently offer there.

pitdicker:
Taking over from #1412.

Fixes #1292.

### DIFF
--- a/src/datetime/serde.rs
+++ b/src/datetime/serde.rs
@@ -1213,6 +1213,92 @@ pub mod ts_seconds_option {
     }
 }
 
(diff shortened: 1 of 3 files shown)


pitdicker:
To be honest: while I opened this PR to take over, it is no priority for me. We can add an optional variant or review any time.

PR #1479, opened 2024-03-02
Parsing preparations

pitdicker:
These issues came up when trying to convert our parsing code to the new error types (only 750 errors to go... :laughing:).

I would like to restrict the visibility of `ParseError` to the `format` module on 0.5, and make all parsing go through `format::parsed`.
That means the `FromStr` implementations of `DateTime<FixedOffset>` and `FixedOffset` should use the standard parsing methods instead of reaching directly for the functions in `format::{parse, scan}`.

In `timezone_offset` I simplified digit parsing a bit because it did not have quite the places to hook up new error values as I wanted. It remains a pretty broken parser though.

Further I noticed a `0i64.checked_sub(v).ok_or(OUT_OF_RANGE)` line. Subtracting a positive `i64` from `0` can never fail. Then I noticed we don't support parsing negative timestamps at all! That is fixed in the last commit. Parsing `i64::MIN` was a little tricky.

### DIFF
--- a/src/format/parse.rs
+++ b/src/format/parse.rs
@@ -1292,7 +1292,7 @@ mod tests {
         check("12345678", &[internal_fixed(TimezoneOffsetPermissive)], Err(INVALID));
         check("+1", &[internal_fixed(TimezoneOffsetPermissive)], Err(TOO_SHORT));
         check("+12", &[internal_fixed(TimezoneOffsetPermissive)], parsed!(offset: 43_200));
-        check("+123", &[internal_fixed(TimezoneOffsetPermissive)], Err(TOO_SHORT));
+        check("+123", &[internal_fixed(TimezoneOffsetPermissive)], Err(TOO_LONG));
         check("+1234", &[internal_fixed(TimezoneOffsetPermissive)], parsed!(offset: 45_240));
         check("-1234", &[internal_fixed(TimezoneOffsetPermissive)], parsed!(offset: -45_240));
         check("−1234", &[internal_fixed(TimezoneOffsetPermissive)], parsed!(offset: -45_240)); // MINUS SIGN (U+2212)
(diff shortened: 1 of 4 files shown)


src/format/scan.rs
```diff
@@ -14,7 +14,7 @@ use crate::Weekday;
 /// More than `max` digits are consumed up to the first `max` digits.
 /// Any number that does not fit in `i64` is an error.
 #[inline]
-pub(super) fn number(s: &str, min: usize, max: usize) -> ParseResult<(&str, i64)> {
+pub(super) fn number(s: &str, min: usize, max: usize, positive: bool) -> ParseResult<(&str, i64)> {
```
djc, line 17: What are we trying to do here? Why do we need to support negative timestamps? Having to pass in `true` for `positive` all over the place doesn't look great.

pitdicker: > Why do we need to support negative timestamps?

We support negative timestamps everywhere, it seems like an oversight to not support them with the `%s` formatting specifier when parsing.

> Having to pass in `true` for `positive` all over the place doesn't look great.

My problem was that I wanted to do a complete fix instead of partial. I.e. to also support parsing `i64::MIN` (which can be stored in `Parsed` even though it is out of range for `NaiveDateTime`, and is something I expect users may test against). Because `-i64::MIN` can't be represented in an `i64` this function should either return an `u64`, or it must return a value with the correct sign. Returning with the correct sign was the minimal change.

pitdicker: :+1: I'll do that.

pitdicker:
Added another commit to remove some duplicated length checks. They where just before `scan::nanosecond_fixed`, which uses `scan::number` that does the same checks.

src/format/scan.rs
```diff
@@ -15,6 +15,16 @@ use crate::Weekday;
 /// Any number that does not fit in `i64` is an error.
 #[inline]
 pub(super) fn number(s: &str, min: usize, max: usize) -> ParseResult<(&str, i64)> {
+    let (s, n) = negative_number(s, min, max)?;
```
djc, line 18: I'm very confused by the logic of parsing a non-negative number by first parsing a negative number and then negating it...

pitdicker: I thought it was quite nice. Positive numbers are not 'better' than negative numbers, and in this case negative numbers had the one integer larger range we needed.

Updated to parse to an `u64` and convert with a tricky cast instead.

src/format/parse.rs
```diff
@@ -424,25 +424,16 @@ where
                     }
 
                     &Internal(InternalFixed { val: InternalInternal::Nanosecond3NoDot }) => {
-                        if s.len() < 3 {
```
djc, line 428: Can you explain in the commit message why they're duplicate (that is, how these invariants are upheld)? It's not clear in the diff.

src/format/scan.rs
```diff
@@ -15,6 +15,28 @@ use crate::Weekday;
...
+    Ok((s, n.try_into().map_err(|_| OUT_OF_RANGE)?))
+}
+
+/// Tries to parse the negative number from `min` to `max` digits.
+///
+/// The absence of digits at all is an unconditional error.
+/// More than `max` digits are consumed up to the first `max` digits.
+/// Any number that does not fit in `i64` is an error.
+#[inline]
+pub(super) fn negative_number(s: &str, min: usize, max: usize) -> ParseResult<(&str, i64)> {
```
djc, line 28: I'd prefer to inline this in the caller (more like before), since there aren't any other users, anyway.

Is there a way to more directly express the precondition for negating the number? This feels kind of roundabout -- even the previous `0i64.checked_sub(v).ok_or()` feels clearer.

PR #1481, opened 2024-03-02
Limit TZ string to 23:59:59 offsets

### LINKED ISSUE #1480: Panic with TZ="pBB24"
Gave my first try fuzzing chrono, starting with setting the TZ env variable.
It came up with "pBB24"

We accept an offset of +24 hours for a time zone named "pBB".
But in chrono we define that an offset can be at most 23:59:59, and return `LocalResult::None` (and then panic) when trying to do anything with `Local`.

pitdicker:
Fixes #1480.


pitdicker:
I may want to look at this a bit more comprehensively.

PR #1492, opened 2024-03-07
Remove unnessary Copy and Send impls

### LINKED ISSUE #25: UTC Offset is not Send (cannot send DateTime across mps::channel)
I have a struct with a `DateTime<UTC>` field that I would like to send across a channel, like this:

    pub struct Signal {
      pub time: DateTime<UTC>,
      pub raw: String
    }

But when I want to create the channel, `rustc` complains that

    error: the trait `core::marker::Send` is not implemented for the type `<chrono::offset::utc::UTC as chrono::offset::TimeZone>::Offset` [    E0277]    
    
    let (new_signal_sender, new_signal_rcvr) = channel::<Signal>();
                                               ^~~~~~~~~~~~~~~~~
    note: `<chrono::offset::utc::UTC as chrono::offset::TimeZone>::Offset` cannot be sent between threads safely
    let (new_signal_sender, new_signal_rcvr) = channel::<Signal>();

Now `Send` is easily the part of Rust I understand the least, but I would have expected a DateTime object to be "sendable".

erickt:
Fuchsia is doing an [audit of chrono 0.4.34](http://fxrev.dev/999580), and we noticed an unncessary unsafe of `Send` for `DateTime`. While it's valid since `Tz::Offset` is `Send`, and `NaiveDateTime` only has `u32` fields, there's a potential hazard if `NaiveDateTime` ever grows unsendable fields (unlikely as that is).

This (and the `Copy` impl) were added 9 years ago to fix #25, which stemmed from early versions of associate traits not working properly with auto-traits. This has since been fixed, and is no longer necessary with the MSRV 1.61.0.

To make sure this stays working, it includes a test that makes sure that `DateTime<Utc>` is `Send` and `Copy` just to be safe.

### DIFF
--- a/src/datetime/tests.rs
+++ b/src/datetime/tests.rs
@@ -1866,3 +1866,12 @@ fn nano_roundrip() {
         assert_eq!(nanos, nanos2);
     }
 }
+
+#[test]
+fn test_datetime_utc_is_copy_and_send() {
+    fn is_copy<T: Copy>(_: T) {}
+    fn is_send<T: Send>(_: T) {}
+
+    is_copy(DateTime::<Utc>::MIN_UTC);
+    is_send(DateTime::<Utc>::MIN_UTC);
+}
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -49,7 +49,7 @@ mod tests;
 /// There are some constructors implemented here (the `from_*` methods), but
 /// the general-purpose constructors are all via the methods on the
 /// [`TimeZone`](./offset/trait.TimeZone.html) implementations.
-#[derive(Clone)]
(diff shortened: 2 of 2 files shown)


djc (approved):
Great, thanks!

src/datetime/tests.rs
```diff
@@ -1866,3 +1866,12 @@ fn nano_roundrip() {
         assert_eq!(nanos, nanos2);
     }
 }
+
+#[test]
+fn test_datetime_utc_is_copy_and_send() {
```
pitdicker, line 1871: We have the same test ca. 600 lines above.

djc:
Removing the duplicate test in #1494.

PR #1493, opened 2024-03-07
0.5: Make fields of `Parsed` private

pitdicker:
The first commit changes the signature of the `Parsed::set_*` methods to be usable in a builder pattern, returning `ParseResult<&mut Parsed>`.

Now that they can be used in a builder pattern, the second commit replaces the `parsed!` macro in `format::parse`. That accounts for 250 of the changed lines in this PR.

Then we can change the fields of `Parsed` to private, and update the tests to no longer set them to invalid values.

The final commit removes some range checks from `Parsed::_to` that have now become unnecessary.

### DIFF
--- a/src/format/parsed.rs
+++ b/src/format/parsed.rs
@@ -562,7 +562,6 @@ impl Parsed {
     /// - `IMPOSSIBLE` if any of the date fields conflict.
     /// - `NOT_ENOUGH` if there are not enough fields set in `Parsed` for a complete date.
     /// - `OUT_OF_RANGE`
-    ///   - if any of the date fields of `Parsed` are set to a value beyond their acceptable range.
     ///   - if the value would be outside the range of a [`NaiveDate`].
     ///   - if the date does not exist.
     pub fn to_naive_date(&self) -> ParseResult<NaiveDate> {
@@ -580,7 +579,7 @@ impl Parsed {
                 // check if present quotient and/or modulo is consistent to the full year.
                 // since the presence of those fields means a positive full year,
                 // we should filter a negative full year first.
-                (Some(y), q, r @ Some(0..=99)) | (Some(y), q, r @ None) => {
+                (Some(y), q, r) => {
                     if y < 0 {
                         return Err(IMPOSSIBLE);
                     }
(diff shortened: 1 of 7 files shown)


src/format/parsed.rs
```diff
@@ -201,8 +203,9 @@ impl Parsed {
     ///
     /// Returns `IMPOSSIBLE` if this field was already set to a different value.
     #[inline]
-    pub fn set_year(&mut self, value: i64) -> ParseResult<()> {
-        set_if_consistent(&mut self.year, i32::try_from(value).map_err(|_| OUT_OF_RANGE)?)
+    pub fn set_year(&mut self, value: i64) -> ParseResult<&mut Parsed> {
```
djc, line 185: Not a 100% sure it's a good idea here, but have you considered doing an owned builder pattern here? That is, taking `mut self` and yielding `Self`. Given that all the fields are `Copy` anyway, seems like it might be nice? In particular, you can chain straight off the constructor (`new().set_day(31)`).

(Nit: can we use `Self` instead of `Parsed`?)

src/format/parsed.rs
```diff
@@ -124,49 +124,28 @@ use crate::{DateTime, Datelike, TimeDelta, Timelike, Weekday};
...
-    pub minute: Option<u32>,
-    #[doc(hidden)]
-    pub second: Option<u32>,
-    #[doc(hidden)]
-    pub nanosecond: Option<u32>,
-    #[doc(hidden)]
-    pub timestamp: Option<i64>,
-    #[doc(hidden)]
-    pub offset: Option<i32>,
+    year: Option<i32>,
```
djc, line 129: Perhaps we should add documentation on each field to document invariants we should enforce (in particular, in setters)?

djc:
I would also like to just get rid of `Parsed::new()` in favor of relying on `Parsed::default()`.

src/format/parsed.rs
pitdicker: This one seemed the most flexible. The next commit shows you can chain as `Parsed::new().set_year(2015)?.set_month(2)?.set_day(4)?.set_hour(14)?.set_minute(37)?.set_second(5)?.set_offset(32_400)`.
But you can also still write `parsed.set_month(3)` without needing to assign the returned value.

An owned builder would allow the methods to be const, but I don't see much advantage in that.

> (Nit: can we use `Self` instead of `Parsed`?)
Sure, I'll update. Out of curiosity: I find repeating the type a bit more readable, but maybe you have a reason `Self` is nicer to use?

pitdicker:
> I would also like to just get rid of `Parsed::new()` in favor of relying on `Parsed::default()`.

Just realized I only dislike `Default::default()` because of the name 😆.

src/format/parsed.rs
djc: It's true that this allows using `parsed.set_month()` on its own more easily. How common is that pattern?



> Sure, I'll update. Out of curiosity: I find repeating the type a bit more readable, but maybe you have a reason `Self` is nicer to use?

Why do you find it more readable? Generally my reasoning is that code should try not to repeat itself, and for example it would be nice if renaming `Parsed` doesn't lead to a lot of low-value diffs. `Self` is also generally more concise. In some of the projects I help maintain we've turned on `warn(clippy::use_self)` to enforce this.

pitdicker:
Sorry for the failing CI: I'll fix it at a better PC later today.

src/format/parsed.rs
pitdicker: > The next commit shows you can chain as

I got confused. Like you say we normally need to create a binding and can't continue from a temporary with borrowing setters.
Because it worked I though something must have changed around temporaries in recent rust versions.
But we can only chain directly in some cases, such as in `assert_eq!`.

> It's true that this allows using `parsed.set_month()` on its own more easily. How common is that pattern?

At least in chrono it is the common pattern. We parse a field, set it in the `Parsed` struct, and do a new iteration of the parsing loop.
It is only in tests that we know all the values at the same time and chaining them is helpful.

Outside chrono I expect this type sees very little use (it may deserve a bit more).

-----

Do you want to choose the pattern here? I don't mind.

And thanks for the explanation on `Self`. The rename argument doesn't seem to apply much in this case, but it is good to have general principles.

djc (approved):

PR #1495, opened 2024-03-08
Convert timestamp methods on `DateTime` to return `Result`s

pitdicker:
All the preparations are in!
Also adds an `ok_or!` macro.

cc @Zomtir.

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -28,8 +28,8 @@ use crate::offset::Local;
 use crate::offset::{FixedOffset, Offset, TimeZone, Utc};
 #[cfg(any(feature = "clock", feature = "std"))]
 use crate::OutOfRange;
-use crate::{expect, ok, try_opt};
-use crate::{Datelike, Months, TimeDelta, Timelike, Weekday};
+use crate::{expect, ok, try_ok_or, try_opt};
+use crate::{Datelike, Error, Months, TimeDelta, Timelike, Weekday};
 
 #[cfg(any(feature = "rkyv", feature = "rkyv-16", feature = "rkyv-32", feature = "rkyv-64"))]
 use rkyv::{Archive, Deserialize, Serialize};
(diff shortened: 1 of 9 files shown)


src/datetime/tests.rs
```diff
@@ -217,15 +219,15 @@ fn test_datetime_from_timestamp_nanos() {
     let nanos = parsed.timestamp_nanos().unwrap();
     assert_eq!(
         Some(DateTime::from_timestamp_nanos(nanos)),
```
Zomtir, line 221: I'd prefer to do the comparison in the `Result` domain:

```
    assert_eq!(
        DateTime::from_timestamp_nanos(nanos),
        DateTime::from_timestamp(nanos / A_BILLION, (nanos % A_BILLION) as u32)
    );
```

src/datetime/mod.rs
```diff
@@ -282,7 +282,8 @@ impl<Tz: TimeZone> DateTime<Tz> {
             subsec_nanos -= 1_000_000_000;
             timestamp += 1;
         }
-        try_opt!(timestamp.checked_mul(1_000_000_000)).checked_add(subsec_nanos)
+        let ts = try_err!(ok_or!(timestamp.checked_mul(1_000_000_000), Error::OutOfRange));
```
Zomtir, line 285: If I remember correctly, we want the value to be unwrapped anyway in all instances. I sugges `try_ok_or` instead that directly unwraps T or returns the error early.

`try_err!(ok_or!(..))` vs `try_ok_or!`

I don't really mind chaining macros, but wanted to have it mentioned.

Zomtir (approved):

djc (approved):

pitdicker:
Updated to have a `to_or` macro and used `?` in the doctests (starting to get payoffs :smile:).

pitdicker:
Sorry, just after your review.

It is a combination of `try_err!(ok_or!(..))`.

djc:
I don't like the `to_or!()` name -- `try_ok_or!()` looks better to me.

src/datetime/tests.rs
```diff
@@ -234,7 +236,7 @@ fn test_datetime_from_timestamp_nanos() {
     for secs in secs_test.iter().copied() {
         assert_eq!(
             Some(DateTime::from_timestamp_nanos(secs * 1_000_000_000)),
-            DateTime::from_timestamp(secs, 0)
+            DateTime::from_timestamp(secs, 0).ok()
```
Zomtir, line 239: This one slipped through.

pitdicker: Oops, thank you.

pitdicker:
> I don't like the `to_or!()` name -- `try_ok_or!()` looks better to me.

Yea, that is better.

djc (approved):

PR #1497, opened 2024-03-08
Deprecate `Parsed::to_naive_datetime_with_offset`, simplify logic

pitdicker:
I would like to deprecate `Parsed::to_naive_datetime_with_offset` and replace it with `Parsed::to_datetime_or_utc` and `Parsed::to_naive_datetime`.

The reason is that in my opinion any use of `Parsed::to_naive_datetime_with_offset` had its logic inside out. Because why would you need to supply an offset in order to parse something as `NaiveDateTime`? It needed it just in case `Parsed` contained a UTC timestamp and also local date and time fields, that have to be checked against each other.

What makes the code much more natural is to parse into a `DateTime<FixedOffset>` first, and discard the offset if all you need is a `NaiveDateTime`.

With this changes we can add two pieces of functionality:
- Allow creating a `NaiveDateTime` from only a timestamp with offset.
- Fix roundtripping values near `DateTime::{MIN, MAX}` where the offset pushes the local value out of range, where we were limited by the API of `Parsed` before (not implemented here yet).

We do now also accept something like:
```rust
let fmt = "%Y-%m-%d %H:%M:%S = timestamp %s";
assert!(NaiveDateTime::parse_from_str("1970-01-01 01:00:00 = timestamp 0", fmt).is_err());
```

It will assume the date and time are not in UTC but have an offset of one hour.

--------

We unfortunately can't reuse the new logic in `Parsed::to_datetime_or_utc` in the old `Parsed::to_naive_datetime_with_offset`. The old method does not check the offset field for consistency, which it left up to the callers.

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -1027,6 +1027,16 @@ impl DateTime<FixedOffset> {
     ///         .unwrap())
     /// );
     /// ```
+    ///
+    /// All parsed fields should be consistent to each other, otherwise it's an error.
+    ///
+    /// ```
+    /// # use chrono::DateTime;
+    /// # let parse_from_str = DateTime::parse_from_str;
+    /// let fmt = "%Y-%m-%d %H:%M:%S %z = timestamp %s";
+    /// assert!(parse_from_str("2001-09-09 01:46:39 +00:00 = timestamp 999999999", fmt).is_ok());
+    /// assert!(parse_from_str("1970-01-01 00:00:00 +00:00 = timestamp 1", fmt).is_err());
+    /// ```
     pub fn parse_from_str(s: &str, fmt: &str) -> ParseResult<DateTime<FixedOffset>> {
         let mut parsed = Parsed::new();
         parse(&mut parsed, s, StrftimeItems::new(fmt))?;
--- a/src/format/parsed.rs
+++ b/src/format/parsed.rs
@@ -970,8 +970,10 @@ impl Parsed {
 
     /// Returns a parsed naive date and time out of given fields.
     ///
-    /// The offset field will be ignored. If there is a timestamp field set, all other date and time
-    /// fields must be in UTC.
+    /// If there are not enough date and time fields set for a complete `NaiveDateTime`, the date
+    /// and time are derived from the timestamp if present:
+    /// - If the offset field is set the local date and time will be calculated.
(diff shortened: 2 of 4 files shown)


djc:
> I would like to deprecate `Parsed::to_naive_datetime_with_offset` and replace it with `Parsed::to_datetime_or_utc` and `Parsed::to_naive_datetime`.

I'm going to stop reading here because `to_datetime_or_utc()` seems suspect and unclear. What are the intended semantics here and why are they better than what they're supposed to replace?

djc:
That seems like a weird design to me. The caller should use `to_naive_datetime()` instead and add the offset separately.

pitdicker:
That was my first thought. But then it is no longer possible to create a `NaiveDateTime` from a timestamp while checking other date/time fields for consistency, because it doesn't know the offset between those. If you want to be able to parse a timestamp into a `NaiveDateTime`, the consequence is you have to be aware of an offset.

djc:
Can you give an example?

pitdicker:
(From a doc comment) The existing code is able to parse and check for consistency a string like:
```rust
assert!(NaiveDateTime::parse_from_str(
    "1970-01-01 00:00:00 +00:00 = timestamp 1",
    "%Y-%m-%d %H:%M:%S %z = timestamp %s"
)
.is_err())
assert!(NaiveDateTime::parse_from_str(
    "1970-01-01 00:00:00 = timestamp 1000000",
    "%Y-%m-%d %H:%M:%S = timestamp %s"
)
.is_err())
```

pitdicker:
Agreed, that example is!
It does fit in the general idea of the `Parsed` type, that it makes sure it checks all relevant fields for consistency before returning a concrete type.

djc:
If we're saying that timestamps only make sense in a UTC-based context, then parsing a `NaiveDateTime` from a format string containing a timestamp just doens't make sense.

pitdicker:
You would suggest removing support for that on 0.5?
We should then maybe also remove the last support for timestamps from the serde modules for `NaiveDateTime`.

djc:
> You would suggest removing support for that on 0.5? We should then maybe also remove the last support for timestamps from the serde modules for `NaiveDateTime`.

Yes, I think that would make sense.

pitdicker:
Okay. I'll close this PR and report back for the 0.5 branch later. If it doesn't have to be shared my may convert the methods to return `Result` first.

PR #1498, opened 2024-03-08
Convert time-related `with_*` methods to return `Result`

pitdicker:
Convert `NaiveTime::{with_hour, with_minute, with_second, with_nanosecond}` and `NaiveDateTime::{with_hour, with_minute, with_second, with_nanosecond}` to return `Result`.
This turned out to be a refreshingly mechanical change.

I took the opportunity to use `?` in the doctests of the relevant methods.
At some point when we are a bit further along we should convert more doctests and regular tests to use `?`.

cc @Zomtir

### DIFF
--- a/src/naive/datetime/mod.rs
+++ b/src/naive/datetime/mod.rs
@@ -1016,7 +1016,7 @@ impl NaiveDateTime {
     /// ```
     #[inline]
     pub const fn with_second(&self, sec: u32) -> Option<NaiveDateTime> {
-        Some(NaiveDateTime { time: try_opt!(self.time.with_second(sec)), ..*self })
+        Some(NaiveDateTime { time: try_opt!(ok!(self.time.with_second(sec))), ..*self })
     }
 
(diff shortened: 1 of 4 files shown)


src/naive/datetime/mod.rs
```diff
@@ -987,7 +987,7 @@ impl NaiveDateTime {
     /// ```
     #[inline]
     pub const fn with_minute(&self, min: u32) -> Option<NaiveDateTime> {
-        Some(NaiveDateTime { time: try_opt!(self.time.with_minute(min)), ..*self })
+        Some(NaiveDateTime { time: try_opt!(ok!(self.time.with_minute(min))), ..*self })
```
djc: Why don't we have a `try_ok_opt!()` or whatever macro?

djc (approved):

src/naive/datetime/mod.rs
pitdicker: They could have use that macro also. It was only there to split up the commits, and replaced in a later commit.

PR #1499, opened 2024-03-09
Fallible variants for `TimeZone::from_utc_*`

pitdicker:
As discussed in https://github.com/chronotope/chrono/issues/1448#issuecomment-1972580791:
> What would we need to change to propagate the error instead of unwrapping on [this line](https://github.com/chronotope/chrono/blob/v0.4.34/src/offset/local/mod.rs#L185) in `Local::offset_from_utc_datetime`?
> 
> Three methods:
> * `TimeZone::offset_from_utc_datetime`
> * `Timezone::from_utc_datetime`
> * `DateTime::with_timezone`
> 
> Those methods currently come with the assumption that finding the offset of a UTC datetime in some time zone will never fail. This is reasonable in a mathematical sense, but ignores that a `TimeZone` implementation can run into errors, such as our `Local` type or a more flexible type that uses the OS time zone information.
> 
> Within chrono it turns out these methods are used almost always in methods that are already fallible. Or, for `DateTime::with_timezone`, there are the infallible `DateTime::to_utc` and `DateTime::fixed_offset` methods which cover many uses.

The first commit adds `TimeZone::offset_from_utc_datetime_opt`, `Timezone::from_utc_datetime_opt` and `DateTime::with_timezone_opt`.
The second makes our platform implementations for `Local` return an `Option` (instead of going through `LocalResult`).

The third commit goes through all uses of the existing three methods and changes them to propagate the error, or switch to the infallible `DateTime::to_utc` and `DateTime::fixed_offset`. In cases where we still need to panic I used `expect` instead of hiding the panic case behind another layer.

For the rest I adjusted the tests to make it easy to switch to fallible methods on the 0.5 branch.

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -416,6 +416,21 @@ impl<Tz: TimeZone> DateTime<Tz> {
         tz.from_utc_datetime(&self.datetime)
     }
 
+    /// Changes the associated time zone.
+    ///
+    /// The returned `DateTime` references the same instant of time from the perspective of the
+    /// provided time zone.
+    ///
+    /// # Errors
+    ///
+    /// In theory every instant of time should exist in every time zone. Still errors may originate
+    /// from for example an OS API, in which case this method returns `None`.
+    #[inline]
+    #[must_use]
+    pub fn with_timezone_opt<Tz2: TimeZone>(&self, tz: &Tz2) -> Option<DateTime<Tz2>> {
+        tz.from_utc_datetime_opt(&self.datetime)
+    }
+
     /// Fix the offset from UTC to its current value, dropping the associated timezone information.
     /// This it useful for converting a generic `DateTime<Tz: Timezone>` to `DateTime<FixedOffset>`.
     #[inline]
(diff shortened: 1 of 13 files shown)


src/datetime/mod.rs
```diff
@@ -416,6 +416,21 @@ impl<Tz: TimeZone> DateTime<Tz> {
...
+    /// The returned `DateTime` references the same instant of time from the perspective of the
+    /// provided time zone.
+    ///
+    /// # Errors
+    ///
+    /// In theory every instant of time should exist in every time zone. Still errors may originate
+    /// from for example an OS API, in which case this method returns `None`.
+    #[inline]
+    #[must_use]
+    pub fn with_timezone_opt<Tz2: TimeZone>(&self, tz: &Tz2) -> Option<DateTime<Tz2>> {
```
djc, line 431: Something I've been thinking about: for newly fallible API, shouldn't we directly start to yield some kind of `Result` instead of sticking with `Option` everywhere?

pitdicker: I sometimes think the same. On the other hand it would make the API seem even less consistent if we end up with a handful of methods that return `Result` and the rest returns `Option`.

pitdicker:
I feel like these changes add friction to the 0.4 versions for little benefit.
Our implementation evolved to be infallible by either falling back to UTC (except for bugs), or returning `LocalResult::None`. That does not seem like something to just change in a minor version. So the new fallible methods should pretty much always return `Ok`.

I'd like to only do this on the 0.5 branch and directly return an error type.

PR #1501, opened 2024-03-09
Rename `LocalResult` to `TzResolution`, add alias

pitdicker:
The name `LocalResult` is unfortunate because in Rust types called `*Result` are used as return values for error cases.
However in our case it is not an error (at least not usually) but a choice of how to deal with time zone transitions due to for example DST.

We can do the rename on the 0.5.x branch or on main. If we think `TzResolution` is a better name that helps in understanding the use we may as well start using it on 0.4.x. With a type alias with the old name that is easy.

Doing the rename on main also has the advantage of keeping the branches a bit more similar. But if 0.5.x is a better target I'll rebase of course.

The first commit is a simple find-replace-rustfmt.
The second commit extends the documentation.

### DIFF
--- a/src/datetime/rustc_serialize.rs
+++ b/src/datetime/rustc_serialize.rs
@@ -2,7 +2,7 @@ use super::DateTime;
 use crate::format::SecondsFormat;
 #[cfg(feature = "clock")]
 use crate::offset::Local;
-use crate::offset::{FixedOffset, LocalResult, TimeZone, Utc};
+use crate::offset::{FixedOffset, MappedLocalTime, TimeZone, Utc};
 use core::fmt;
 use core::ops::Deref;
 use rustc_serialize::{Decodable, Decoder, Encodable, Encoder};
@@ -13,16 +13,16 @@ impl<Tz: TimeZone> Encodable for DateTime<Tz> {
     }
 }
 
-// lik? function to convert a LocalResult into a serde-ish Result
-fn from<T, D>(me: LocalResult<T>, d: &mut D) -> Result<T, D::Error>
+// Function to convert a MappedLocalTime into a serde-ish Result
+fn from<T, D>(me: MappedLocalTime<T>, d: &mut D) -> Result<T, D::Error>
 where
     D: Decoder,
     T: fmt::Display,
 {
     match me {
-        LocalResult::None => Err(d.error("value is not a legal timestamp")),
-        LocalResult::Ambiguous(..) => Err(d.error("value is an ambiguous timestamp")),
-        LocalResult::Single(val) => Ok(val),
+        MappedLocalTime::None => Err(d.error("value is not a legal timestamp")),
+        MappedLocalTime::Ambiguous(..) => Err(d.error("value is an ambiguous timestamp")),
+        MappedLocalTime::Single(val) => Ok(val),
     }
 }
 
--- a/src/offset/mod.rs
+++ b/src/offset/mod.rs
@@ -39,7 +39,7 @@ pub use self::utc::Utc;
 
 /// The conversion result from the local time to the timezone-aware datetime types.
 #[derive(Clone, PartialEq, Debug, Copy, Eq, Hash)]
-pub enum LocalResult<T> {
+pub enum MappedLocalTime<T> {
     /// Given local time representation is invalid.
     /// This can occur when, for example, the positive timezone transition.
     None,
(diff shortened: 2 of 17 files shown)


pitdicker:
@djc before spending effort on reviewing this, I'm interested in you opinion on https://github.com/chronotope/chrono/issues/1448#issuecomment-1986965331. Although the type becomes a bit more complex, it seems overall like a simplification to remove the offset methods from `TimeZone` and make the enum only generic over `Offset` types.

djc:
I don't love the name?

- Abbreviations on type/field names are kinda unidiomatic
- I don't think the term "resolution" is all that common for resolving a date/time to a timezone offset, at least in non-expert usage?

(I feel like it's probably slightly better than `LocalResult` but if we're going to do this I'd like to feel better about the name we end up picking.)

pitdicker:
Hmm. The result of finding the offset of a local datetime within a timezone, which needs a decision on how to deal with times that fall in a transition. What more names can we think up for it?

`ResolveOffset` maybe? We need the user to decide what offset he wants in case of a transition.

djc:
`CurrentOffset`, `ProjectedOffset`, `ExpectedOffset`?

pitdicker:
Those also don't seem to be it. The result is not current, a projection, and there may be no expectation? Resolve, resolution or result seem closer to me.

Maybe something that has 'map' or 'disambiguation' (https://peps.python.org/pep-0495/) in the name? I don't like something as long as disambiguation though.

pitdicker:
`LocalTimeMapping`?

djc:
`MappedTime`?

djc:
Seems like a decent option...

pitdicker:
The least bad :laughing:.

With this name I do think it is best to combine it with making the type only return `DateTime`s and not offsets anymore, the last idea in https://github.com/chronotope/chrono/issues/1448#issuecomment-1987220921. So targeting the 0.5 branch.

Have you already formed an opinion on that?
In short: We remove `TimeZone::offset_from_* methods. An offset can easily be derived from a `DateTime` with `.offset()`. That makes the `TimeZone` trait and this type simpler.

djc:
> With this name I do think it is best to combine it with making the type only return `DateTime`s and not offsets anymore, the last idea in [#1448 (comment)](https://github.com/chronotope/chrono/issues/1448#issuecomment-1987220921). So targeting the 0.5 branch.

Why/how is that related to the naming? In principle I like the idea of bringing the new name to 0.4.x as an alias to potentially get some feedback.

> Have you already formed an opinion on that? In short: We remove the `TimeZone::offset_from_*` methods. An offset can easily be derived from a `DateTime` with `.offset()`. That makes the `TimeZone` trait and this type simpler.

I have not, and this description doesn't help me form an opinion. Are there commits that flesh out this idea?

pitdicker:
> Why/how is that related to the naming?

It is a bit strange if [`TimeZone::offset_from_local_datetime`](https://docs.rs/chrono/latest/chrono/offset/trait.TimeZone.html#tymethod.offset_from_local_datetime) returns an offset in a type with the name `MappedLocalTime`.
But I'm fine with doing it on 0.4, that method should be little-used.

> I have not, and this description doesn't help me form an opinion. Are there commits that flesh out this idea?

I put effort in the description in the linked comment. But I'll try to make some commits after the rename.

src/datetime/rustc_serialize.rs
```diff
@@ -13,16 +13,16 @@ impl<Tz: TimeZone> Encodable for DateTime<Tz> {
     }
 }
 
-// lik? function to convert a LocalResult into a serde-ish Result
-fn from<T, D>(me: LocalResult<T>, d: &mut D) -> Result<T, D::Error>
+// lik? function to convert a MappedLocalTime into a serde-ish Result
```
djc, line 16: "lik?"?

src/offset/mod.rs
```diff
@@ -37,21 +37,47 @@ pub use self::local::Local;
...
+/// * An ambiguous result when the clock is turned backwards during a transition due to for example
+///   DST.
+/// * No result when the clock is turned forwards during a transition due to for example DST.
+///
+/// When the clock is turned backwards it creates a _fold_ in local time, during which the local
+/// time is _ambiguous_. When the clock is turned forwards it creates a _gap_ in local time, during
+/// which the local time is _missing_, or does not exist.
+///
+/// Chrono does not return a default choice or invalid data during time zone transitions, but has
+/// the `MappedLocalTime` to help deal with the result correctly.
```
djc, line 54: "has the `MappedLocalTime` to" -> "has the `MappedLocalTime` type to".

djc (approved):

src/datetime/rustc_serialize.rs
pitdicker: I have no clue. Remove it?

djc: Yup.

pitdicker:
Something weird happened with my branches, I'll figure it out

PR #1515, opened 2024-03-14
0.5: Let `TimeDelta::{weeks, days, hours, minutes, seconds}` take `i32`

pitdicker:
Making these methods fallible is not ergonomic. Especially because they are usually used with small values that comfortably with in an `i32`.

This PR converts  `TimeDelta::{weeks, days, hours, minutes, seconds}` to being infallible, and to take an `i32` as argument.
They can no longer be used to create values that cover the entire range of `TimeDelta`. If that is needed users can use `TimeDelta::new`.
Also they no longer support easy round-trip with their `num_*` variants that return an `i64`. Constructing a value is much more common that round-tripping, so it is best to make constructing ergonomic.

Because the range of `TimeDelta` is defined as `-i64::MAX..=i64::MAX` the `TimeDelta::milliseconds` constructor still takes an `i64` as argument and remains fallible. That way an easy mapping between `TimeDelta` and a single integer remains easy.
If users want a infallible alternative, they can multiply by `1000` and use `TimeDelta::microseconds`.


djc (approved):
This is a great idea.

How do you feel about just ditching the `try_()` equivalents? People that need the extreme range can just use `TimeDelta::new()`.

pitdicker:
> How do you feel about just ditching the `try_()` equivalents? People that need the extreme range can just use `TimeDelta::new()`.

I'm all for it, we did in #1488.

PR #1519, opened 2024-03-15
Slightly improve serde documentation

### LINKED ISSUE #1484: Deserializing DateTime from unix timestamp is not implemented, despite description
The Deserialize Trait can't deserialize DateTimes from timestamps despite the error message you get. I tried tracking it down in the source code, I hope the cited references are correct.

The Deserialize Trait only calls the `deserialize_str()` function

https://github.com/chronotope/chrono/blob/f4adc28dce9a139ef17c7044e2adafde029cb17d/src/datetime/serde.rs#L75-L82

where, in spite of the `expecting()` stating it accepts **"a formatted date and time string or a unix timestamp"** it only calls the `parse()` function

https://github.com/chronotope/chrono/blob/f4adc28dce9a139ef17c7044e2adafde029cb17d/src/datetime/serde.rs#L53-L66

which in turn is implemented to only try and deserialize rfc3339 strings

https://github.com/chronotope/chrono/blob/f4adc28dce9a139ef17c7044e2adafde029cb17d/src/format/parse.rs#L523-L534

pitdicker:
Fixes #1484.

### DIFF
--- a/src/datetime/serde.rs
+++ b/src/datetime/serde.rs
@@ -23,10 +23,12 @@ pub struct MicroSecondsTimestampVisitor;
 #[derive(Debug)]
 pub struct MilliSecondsTimestampVisitor;
 
-/// Serialize into an ISO 8601 formatted string.
+/// Serialize to an RFC 3339 formatted string
 ///
-/// See [the `serde` module](./serde/index.html) for alternate
-/// serializations.
+/// As an extension to RFC 3339 this can serialize `DateTime`s outside the range of 0-9999 years
+/// using an ISO 8601 syntax (which prepends an `-` or `+`).
+///
+/// See [the `serde` module](crate::serde) for alternate serializations.
 impl<Tz: TimeZone> ser::Serialize for DateTime<Tz> {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
@@ -54,7 +56,7 @@ impl<'de> de::Visitor<'de> for DateTimeVisitor {
     type Value = DateTime<FixedOffset>;
 
     fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
-        formatter.write_str("a formatted date and time string or a unix timestamp")
+        formatter.write_str("an RFC 3339 formatted date and time string")
     }
 
     fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>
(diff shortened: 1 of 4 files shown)


pitdicker:
No clue yet why a warning shows up for the `naive::time` module on this PR but not on main:
```
error: redundant explicit link target
   --> src/naive/time/mod.rs:193:18
    |
193 | /// [`DateTime`](crate::DateTime) with fractional minutes as offset, or using
    |      ----------  ^^^^^^^^^^^^^^^ explicit target is redundant
    |      |
    |      because label contains path that resolves to same destination
    |
    = note: when a link's destination is not specified,
            the label is used to resolve intra-doc links
    = note: `-D rustdoc::redundant-explicit-links` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(rustdoc::redundant_explicit_links)]`
help: remove explicit link target
    |
193 | /// [`DateTime`] with fractional minutes as offset, or using
    |     ~~~~~~~~~~~~
```

pitdicker:
> No clue yet why a warning shows up for the `naive::time` module on this PR but not on main:

Seems to be a bug in the error reporting of rustdoc. The redundant target was in `libs.rs` but for a similar `DateTime` link.

src/datetime/serde.rs
```diff
@@ -23,10 +23,9 @@ pub struct MicroSecondsTimestampVisitor;
 #[derive(Debug)]
 pub struct MilliSecondsTimestampVisitor;
 
-/// Serialize into an ISO 8601 formatted string.
+/// Serialize into an ISO 8601 formatted string, with a format similar to RFC 3339.
```
djc, line 26: This seems weirdly imprecise, and its precision doesn't match the string used for `expecting()`.

src/datetime/serde.rs
```diff
@@ -65,13 +64,9 @@ impl<'de> de::Visitor<'de> for DateTimeVisitor {
     }
 }
 
-/// Deserialize a value that optionally includes a timezone offset in its
-/// string representation
+/// Deserialize an RFC 3339 formatted string into a `DateTime<FixedOffset>`.
```
djc, line 67: FWIW, it's somewhat on purpose that the first line of docstrings doesn't have a `.`. These are sentence fragments, not full sentences.

src/datetime/serde.rs
pitdicker: It is the RFC 3339 format with the addition that it can (de)serialize datetimes outside the `0..=9999` years range using the ISO 8601 syntax. I'll try to word it better.

src/datetime/serde.rs
pitdicker: Really? Okay.

djc (approved):

PR #1523, opened 2024-03-16
Rename `TimeZone::timestamp*` to `TimeZone::and_timestamp*`

pitdicker:
We have the pattern to use the name `and_` for methods that add information to convert to a more complete type.
`NaiveDate` had `and_time` and `and_hms*`, `NaiveDateTime` has `and_local_timezone`.

This makes the methods on `TimeZone` consistent. `TimeZone::timestamp*` is renamed to `TimeZone::and_timestamp*` and `TimeZone::with_ymd_and_hms` to `TimeZone::and_ymd_and_hms*`.

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -64,7 +64,7 @@ impl<Tz: TimeZone> DateTime<Tz> {
     /// passing it through FFI.
     ///
     /// For regular use you will probably want to use a method such as
-    /// [`TimeZone::from_local_datetime`] or [`NaiveDateTime::and_local_timezone`] instead.
+    /// [`TimeZone::from_local_datetime`] or [`NaiveDateTime::in_timezone`] instead.
     ///
     /// # Example
     ///
@@ -337,7 +337,7 @@ impl<Tz: TimeZone> DateTime<Tz> {
         // datetime.
         self.overflowing_naive_local()
             .checked_add_months(months)?
-            .and_local_timezone(Tz::from_offset(&self.offset))
+            .in_timezone(self.timezone())
             .single()
     }
 
@@ -374,7 +374,7 @@ impl<Tz: TimeZone> DateTime<Tz> {
         // datetime.
         self.overflowing_naive_local()
             .checked_sub_months(months)?
-            .and_local_timezone(Tz::from_offset(&self.offset))
+            .in_timezone(self.timezone())
             .single()
     }
 
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -648,7 +648,7 @@ impl<Tz: TimeZone> DateTime<Tz> {
     /// ```
     #[must_use]
     pub fn with_time(&self, time: NaiveTime) -> MappedLocalTime<Self> {
-        self.timezone().from_local_datetime(self.overflowing_naive_local().date().and_time(time))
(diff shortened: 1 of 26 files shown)


djc:
So we may have this `and_()` pattern but I feel like `with_()` is much more common in the Rust ecosystem. Maybe we should be moving in the other direction instead? (I think that was the idea behind `with_ymd_and_hms()`, at least.)

Can we do some part of this on main?

pitdicker:
Fine by me. Then we should probably rename the current `with_` methods such as `with_year` to something like `replace_year`? That matches time-rs.

djc:
How are the semantics for current `with_()` methods different from those for current `and_()` methods, in your opinion?

pitdicker:
Currently the `with_()` methods return the same type and replace some part of the date, time or offset.
The `and_()` methods add more information to turn it into another type.

pitdicker:
Other methods that start with `and` are `NaiveDate::{and_time, and_hms*}`, `NaiveDateTime::{and_local_timezone, and_utc}`.

And one more that starts with `with` is `DateTime::with_timezone`.

pitdicker:
I'll collect some info in an issue.

pitdicker:
Added commits to rename all methods from #1527 except `TimeZone::{from_local_datetime, from_utc_datetime}`.

I am not 100% conviced with the new names for `NaiveDateTime::{and_local_timezone, and_utc}`. Naming them `in_` does match natural language. And `assume_` seems like saying the user may not be sure for some reason.

Example code:
```rust
let pst = FixedOffset::east(8 * 60 * 60)?;
// in_timezone
let dt = NaiveDate::from_ymd(2018, 1, 11)?.at_hms(10, 5, 13)?.in_timezone(pst);
// assume_timezone
let dt = NaiveDate::from_ymd(2018, 1, 11)?.at_hms(10, 5, 13)?.assume_timezone(pst);

// in_utc
let timestamp = NaiveDate::from_ymd(1970, 1, 1)?.at_hms(0, 0, 0).in_utc().timestamp();
// assume_utc
let timestamp = NaiveDate::from_ymd(1970, 1, 1)?.at_hms(0, 0, 0).assume_utc().timestamp();
```

src/datetime/mod.rs
```diff
@@ -527,13 +527,12 @@ impl<Tz: TimeZone> DateTime<Tz> {
...
     /// ```rust
-    /// # use chrono::{FixedOffset, SecondsFormat, TimeZone, Utc, NaiveDate};
+    /// # use chrono::{FixedOffset, SecondsFormat, TimeZone, NaiveDate};
     /// let dt = NaiveDate::from_ymd(2018, 1, 26)
     ///     .unwrap()
     ///     .at_hms_micro(18, 30, 9, 453_829)
     ///     .unwrap()
-    ///     .and_local_timezone(Utc)
-    ///     .unwrap();
+    ///     .and_utc();
```
djc: This seems weird if a later commit is just going to change it again to `in_utc()`.

djc (approved):

src/datetime/mod.rs
pitdicker: I agree. This commit is merged on the main branch in #1530. I'll wait until main is merged into the 0.5.x branch again and then rebase this PR.

PR #1524, opened 2024-03-16
Convert `TryFrom<SystemTime>` to our error type

pitdicker:
This does not yet convert `TryFrom<SystemTime> for DateTime<Local>`.
I'd like to work on the methods that can have a failure cause in the `TimeZone` trait later.

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -1712,13 +1712,13 @@ impl str::FromStr for DateTime<Local> {
 
 #[cfg(feature = "std")]
 impl TryFrom<SystemTime> for DateTime<Utc> {
-    type Error = OutOfRange;
+    type Error = Error;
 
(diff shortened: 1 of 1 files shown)


src/datetime/mod.rs
```diff
@@ -1753,7 +1753,7 @@ impl TryFrom<SystemTime> for DateTime<Local> {
     type Error = OutOfRange;
 
     fn try_from(t: SystemTime) -> Result<DateTime<Local>, OutOfRange> {
-        DateTime::<Utc>::try_from(t).map(|t| t.with_timezone(&Local))
+        DateTime::<Utc>::try_from(t).map(|t| t.with_timezone(&Local)).map_err(|_| OutOfRange::new())
```
djc, line 1756: Why not do this one at the same time?

djc (approved):

PR #1525, opened 2024-03-17
Be consistent about taking `self` by value or by reference

pitdicker:
We are currently inconsistent in when we take `self` by value and when by reference.

### DateTime

For `DateTime` I propose to let all methods take `self` by reference because the type is generic and not guaranteed to be `Copy`. Otherwise users of a `TimeZone` provider such as `tzfile` would have to clone the `DateTime` if they want to use they value after calling the method.

For this reason the methods that are part of the `Datelike` and `Timelike` trait should also keep taking `self` by reference.

### Copy types

For a `Copy` type it is recommended to pass `self` by value if its size is at most the size of one or two pointers. This is a bit more optimal than by reference for non-inlined functions. There is a Clippy lint `trivially_copy_pass_by_ref` that suggest passing such values by copy, it is recommended by the [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#f16-for-in-parameters-pass-cheaply-copied-types-by-value-and-others-by-reference-to-const), and similar advise can be found on the Rust Reddit, StackOverflow and [URLO](https://users.rust-lang.org/t/guidelines-for-self-ownership-on-copy-types/61262) (which even mentions chrono).

So what to do for `NaiveDate` and `IsoWeek` as 4-byte `Copy` types is clear: passing them by reference is less optimal than by value.

What to do for types of 8 and 12 bytes such as `NaiveTime` and `NaiveDateTime`, keeping in mind that there are 32-bit and 64-bit targets? When it is less clear I think the real question to pick what is optimal is: can all method arguments be passed in registers or do they have to spill to the stack? On x86 there are only four or five general-purpose registers. So we can pass up to four arguments by reference, or the total size of `self` and the arguments should at most be 16 bytes.

There are a handful of cases where passing `self` by reference would create room for passing another argument by value: `NaiveTime::{checked_add_signed, checked_sub_signed}` and `NaiveDateTime::checked_add_signed, checked_sub_signed, signed_duration_since}`. The latter would have two 12-byte values. Does it make any difference?

I made a benchmark for `NaiveDateTime::checked_add_signed` and the `i686-unknown-linux-gnu` target, and marked `checked_add_signed` as `#[inline(never)]`. Any changes between pass by reference and pass by value were in the noise. Based on the benchmark I think it is fine to simply always pass our `Copy` types by value, including the 12-byte types `NaiveDateTime` and `TimeDelta`.

This all of course only makes a difference when the methods are not inlined, despite the `[inline]` hint. And in the rare case when someone calls a method as a function it is a bit more ergonomic to use `self` by value.

The impact on user code is minimal and nice, as the diff of each commit shows.

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -311,7 +311,7 @@ impl<Tz: TimeZone> DateTime<Tz> {
     /// Returns `None` if the resulting date would be out of range.
     #[inline]
     #[must_use]
-    pub fn checked_add_signed(self, rhs: TimeDelta) -> Option<DateTime<Tz>> {
+    pub fn checked_add_signed(&self, rhs: TimeDelta) -> Option<DateTime<Tz>> {
         let datetime = self.datetime.checked_add_signed(rhs).ok()?;
         let tz = self.timezone();
         Some(tz.from_utc_datetime(datetime))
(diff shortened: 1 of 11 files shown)


djc (approved):
Great analysis, thanks!

PR #1529, opened 2024-03-19
Refactor `unix::Cache`

pitdicker:
I put in the effort to slice the first commit of #1457 into 17 commits that each are readable as a refactor.

The goal is to move all logic for determining and parsing the time zone with fallbacks in one layer.
This will make it possible to properly hook up errors, and to make the cache invalidation work beyond the most basic cases.
But those improvements are not included in this PR.

The last commit, adding support for the `TZ_DIR` variable, is the only noticeable change in functionality.

### DIFF
--- a/src/offset/local/unix.rs
+++ b/src/offset/local/unix.rs
@@ -53,15 +53,6 @@ const TZDB_LOCATION: &str = "/usr/share/lib/zoneinfo";
 #[cfg(not(any(target_os = "android", target_os = "aix")))]
 const TZDB_LOCATION: &str = "/usr/share/zoneinfo";
 
-fn fallback_timezone() -> Option<TimeZone> {
-    let tz_name = iana_time_zone::get_timezone().ok()?;
-    #[cfg(not(target_os = "android"))]
-    let bytes = fs::read(format!("{}/{}", TZDB_LOCATION, tz_name)).ok()?;
-    #[cfg(target_os = "android")]
-    let bytes = android_tzdata::find_tz_data(&tz_name).ok()?;
-    TimeZone::from_tz_data(&bytes).ok()
-}
-
 impl Cache {
     fn tz_info(&mut self) -> &TimeZone {
         self.refresh_cache();
@@ -125,9 +116,8 @@ impl Cache {
         self.source = Source::new(env_tz.as_deref());
         self.zone = Some(
             self.read_from_tz_env_or_localtime(env_tz.as_deref())
-                .ok()
-                .or_else(fallback_timezone)
-                .unwrap_or_else(TimeZone::utc),
+                .or_else(|_| self.read_with_tz_name())
+                .unwrap_or_else(|_| TimeZone::utc()),
         );
     }
 
@@ -135,6 +125,17 @@ impl Cache {
     /// Read from `/etc/localtime` if the variable is not set.
     fn read_from_tz_env_or_localtime(&self, env_tz: Option<&str>) -> Result<TimeZone, ()> {
         TimeZone::local(env_tz).map_err(|_| ())
+    }
+
+    /// Get the IANA time zone name of the system by whichever means the `iana_time_zone` crate gets
+    /// it, and try to read the corresponding TZif data.
+    fn read_with_tz_name(&self) -> Result<TimeZone, ()> {
+        let tz_name = iana_time_zone::get_timezone().map_err(|_| ())?;
+        #[cfg(not(target_os = "android"))]
+        let bytes = fs::read(format!("{}/{}", TZDB_LOCATION, tz_name)).map_err(|_| ())?;
+        #[cfg(target_os = "android")]
+        let bytes = android_tzdata::find_tz_data(&tz_name).ok()?;
+        TimeZone::from_tz_data(&bytes).map_err(|_| ())
     }
 }
 
--- a/src/offset/local/unix.rs
+++ b/src/offset/local/unix.rs
@@ -13,7 +13,7 @@ use std::collections::hash_map;
 use std::env;
 use std::fs;
 use std::hash::Hasher;
-use std::path::PathBuf;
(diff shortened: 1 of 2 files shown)


pitdicker:
Of course, and thank you.

Removed one commit that in the end didn't do much.

As a note: the reason to make everything methods on the `Cache` type is because as a follow-up I not only want to set `self.zone` but also for every source that is tried the inputs we should watch to invalidate the cache.

src/offset/local/unix.rs
```diff
@@ -105,14 +105,39 @@ fn current_zone(var: Option<&str>) -> TimeZone {
...
+
+        // we pass through the year as the year of a local point in time must either be valid in that locale, or
+        // the entire time was skipped in which case we will return MappedLocalTime::None anyway.
+        self.zone
+            .find_local_time_type_from_local(d.and_utc().timestamp(), d.year())
+            .expect("unable to select local time type")
+            .and_then(|o| FixedOffset::east_opt(o.offset()))
+    }
+
+    // Refresh our cached data if necessary.
```
djc: Let's use an actual docstring for this?

src/offset/local/unix.rs
```diff
@@ -162,10 +150,11 @@ impl Cache {
             (Source::Environment { .. }, Source::LocalTime { .. })
             | (Source::LocalTime { .. }, Source::Environment { .. }) => true,
             // stay as file, but mtime has changed
-            (Source::LocalTime { mtime: old_mtime }, Source::LocalTime { mtime })
-                if old_mtime != mtime =>
-            {
-                true
+            (Source::LocalTime, Source::LocalTime) => {
```
djc, line 94: Why is this better? It seems less precise to me.

src/offset/local/unix.rs
```diff
@@ -136,41 +136,45 @@ impl Cache {
...
-                    }
-                    // cache can be reused
-                    _ => false,
-                };
-
-                if out_of_date {
-                    self.zone = current_zone(env_ref);
-                }
+        if self.needs_update() {
+            let env_tz = env::var("TZ").ok();
```
djc: The duplication between here and `needs_update()` doesn't seem great...

src/offset/local/unix.rs
```diff
@@ -24,19 +24,25 @@ pub(super) fn offset_from_local_datetime(local: &NaiveDateTime) -> MappedLocalTi
...
+            last_checked: SystemTime::UNIX_EPOCH,
+        }
+    ) };
 }
 
+#[derive(PartialEq)]
 enum Source {
     Environment { hash: u64 },
     LocalTime,
+    Uninitialized,
```
djc: Can we just represent this as `Option<Source>` instead? Seems more idiomatic to me.

src/offset/local/unix.rs
```diff
@@ -24,19 +24,25 @@ pub(super) fn offset_from_local_datetime(local: &NaiveDateTime) -> MappedLocalTi
...
+        maybe_cache.borrow_mut().offset(*d, local)
     })
 }
 
-// we have to store the `Cache` in an option as it can't
-// be initialized in a static context.
 thread_local! {
-    static TZ_INFO: RefCell<Option<Cache>> = Default::default();
+    static TZ_INFO: RefCell<Cache> = const { RefCell::new(
+        Cache {
```
djc: Why? This doesn't seem like an improvement.

djc (commented):
I've tried to review this, but I've found it challenging.

Most of the changes don't look like improvements to me -- they mostly add more code, and seem to make it all more complex. Tell me why/how I'm wrong?

src/offset/local/unix.rs
pitdicker: Oops. Yes, that was the intention.

pitdicker:
> I've tried to review this, but I've found it challenging.
> 
> Most of the changes don't look like improvements to me -- they mostly add more code, and seem to make it all more complex. Tell me why/how I'm wrong?

When looking at it commit by commit it looks that way. Once you get to "Move parsing of `TZ` variable to `unix` module" we can start to share code that the previous commits where setting up for.

And this mostly prepares for the final commit in #1457. Could you give the end result a global look?

src/offset/local/unix.rs
pitdicker: I'll update it to use an `Option`.

src/offset/local/unix.rs
pitdicker: Hmm, maybe I should keep an mtime around. It is not guaranteed to be the real modification time. We should either use ctime or keep track of the mtime.

Comparing ctime to `last_checked` is simpler though.

src/offset/local/unix.rs
pitdicker: It is a preparation for the next commit: splitting up the `offset` function.
The old code did all in the initialization in `offset()` using `Cache::default` on first use within a thread.

The `offset_from_utc_datetime` and `offset_from_local_datetime` both had to go through the `offset` function even though they had little in common (and preferably not the same return type), so they could use the same `get_or_insert_with`.

The goal of this commit is to move the concern of initialization from the caller the `Cache` type.

src/offset/local/unix.rs
pitdicker: Thought the same! But it seemed premature optimization to remove this one duplicate read of an environment variable. We only do so if the variable has changed, which should be rare.

We can keep the variable around of course. It will have to be returned or passed to a couple of methods. More than what is visible in this PR as the final commit in https://github.com/chronotope/chrono/pull/1457 adds more methods.

djc:
I'm really sorry but even after looking at all of the changes in #1457 today it doesn't seem more clear -- and it adds 80 lines of code, is that all for the new functionality? That seems to remove some tests so, if anything, seems to be undercounting library code. It seems to add lots of small functions which IMO just make it harder to follow the flow.

PR #1533, opened 2024-03-21
Add `MappedLocalTime::try_map`

pitdicker:
Passing on one error in `unix::Cache::offset` requires a match that is very similar to the one in `TimeZone::from_local_datetime`.
I added `MappedLocalTime::try_map` that returns `MappedLocalTime::None` that can be used in both places.

For the 0.5.x branch it will be easy to convert this method to return a `Result` instead.

### DIFF
--- a/src/offset/mod.rs
+++ b/src/offset/mod.rs
@@ -121,6 +121,24 @@ impl<T> MappedLocalTime<T> {
             MappedLocalTime::Ambiguous(min, max) => MappedLocalTime::Ambiguous(f(min), f(max)),
         }
     }
+
+    /// Maps a `MappedLocalTime<T>` into `MappedLocalTime<U>` with given function.
+    ///
+    /// Returns `MappedLocalTime::None` if the function returns `None`.
+    #[must_use]
+    pub(crate) fn and_then<U, F: FnMut(T) -> Option<U>>(self, mut f: F) -> MappedLocalTime<U> {
+        match self {
+            MappedLocalTime::None => MappedLocalTime::None,
+            MappedLocalTime::Single(v) => match f(v) {
+                Some(new) => MappedLocalTime::Single(new),
+                None => MappedLocalTime::None,
+            },
+            MappedLocalTime::Ambiguous(min, max) => match (f(min), f(max)) {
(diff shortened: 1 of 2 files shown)


src/offset/mod.rs
```diff
@@ -121,6 +121,24 @@ impl<T> MappedLocalTime<T> {
             MappedLocalTime::Ambiguous(min, max) => MappedLocalTime::Ambiguous(f(min), f(max)),
         }
     }
+
+    /// Maps a `MappedLocalTime<T>` into `MappedLocalTime<U>` with given function.
+    ///
+    /// Returns `MappedLocalTime::None` if the function returns `None`.
+    #[must_use]
+    pub(crate) fn try_map<U, F: FnMut(T) -> Option<U>>(self, mut f: F) -> MappedLocalTime<U> {
```
djc, line 129: I think the `Option` variant of this is called `and_then()`, maybe reuse that name?

djc (approved):
This PR on its own doesn't look well-motivated... I'd probably keep this commit in the `Cache` PR but am okay if you want to get it merged first.

src/offset/mod.rs
pitdicker: :+1: `and_then()` is better.

pitdicker:
> This PR on its own doesn't look well-motivated...

Not sure what you mean. Maybe as another motivation: to implement addition and substraction for a `CalendarDuration` (#1282) months and days should be added in localtime with a `MappedLocalTime` as result. Then seconds should be added using a method such as this so the type remains a `MappedLocalTime`.

But I have to admit I'm not sure if the current two uses in this PR for the 0.4 versions will remain useful for 0.5 with one of the plans in [1448#issuecomment-1987220921](https://github.com/chronotope/chrono/issues/1448#issuecomment-1987220921). So the method is `pub(crate)` for now.

djc:
Ah, sorry, I missed that there are sort of two users for the new method. Looks a little surprising because one of the callers did not previously have the extensive logic implemented in the method.

pitdicker:
Yes, that was the issue I set out to fix :smile:. Sorry for not saying so better.

PR #1534, opened 2024-03-21
Use different method to ignore feature-dependent doctests

pitdicker:
This will enable an almost clean run with `cargo +nightly fmt -- --config format_code_in_doc_comments=true`.
It did not handle a combination of doc comments and `#[cfg_attr(feature = ..., doc = "...")]` well.

### DIFF
--- a/src/format/strftime.rs
+++ b/src/format/strftime.rs
@@ -258,8 +258,8 @@ impl<'a> StrftimeItems<'a> {
     ///
     /// # Example
     ///
-    #[cfg_attr(not(any(feature = "alloc", feature = "std")), doc = "```ignore")]
-    #[cfg_attr(any(feature = "alloc", feature = "std"), doc = "```rust")]
+    /// ```
+    /// # #[cfg(feature = "alloc")] {
     /// use chrono::format::{Locale, StrftimeItems};
     /// use chrono::{FixedOffset, TimeZone};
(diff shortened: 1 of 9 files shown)


src/format/strftime.rs
```diff
@@ -279,6 +278,7 @@ impl<'a> StrftimeItems<'a> {
     /// assert_eq!(fmtr.to_string(), "2023년 07월 11일");
     /// let fmtr = dt.format_with_items(StrftimeItems::new_with_locale("%x", Locale::ja_JP));
     /// assert_eq!(fmtr.to_string(), "2023年07月11日");
+    /// # }
```
djc, line 281: I don't see a matching opening brace...

djc (approved):
Nice!

src/format/strftime.rs
```diff
@@ -259,8 +258,8 @@ impl<'a> StrftimeItems<'a> {
     ///
     /// # Example
     ///
-    #[cfg_attr(not(any(feature = "alloc", feature = "std")), doc = "```ignore")]
-    #[cfg_attr(any(feature = "alloc", feature = "std"), doc = "```rust")]
+    /// ```
+    /// # #[cfg(feature = "alloc")] {
```
pitdicker, line 262: It is up here.

PR #1535, opened 2024-03-21
Main documentation: simplify links and reflow text

pitdicker:
It seems out main documentation needs a good rewrite, not everything matches what we would recommend today.
But for now I just simplified the links and reflowed the surrounding text.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,32 +1,33 @@
 //! # Chrono: Date and Time for Rust
 //!
-//! Chrono aims to provide all functionality needed to do correct operations on dates and times in the
-//! [proleptic Gregorian calendar](https://en.wikipedia.org/wiki/Proleptic_Gregorian_calendar):
+//! Chrono aims to provide all functionality needed to do correct operations on dates and times in
+//! the [proleptic Gregorian calendar]:
 //!
-//! * The [`DateTime`](https://docs.rs/chrono/latest/chrono/struct.DateTime.html) type is timezone-aware
-//!   by default, with separate timezone-naive types.
+//! * The [`DateTime`] type is timezone-aware by default, with separate timezone-naive types.
 //! * Operations that may produce an invalid or ambiguous date and time return `Option` or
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1,21 +1,23 @@
 //! # Chrono: Date and Time for Rust
 //!
 
-//! Chrono aims to provide all functionality needed to do correct operations on dates and times in the
-//! [proleptic Gregorian calendar](https://en.wikipedia.org/wiki/Proleptic_Gregorian_calendar):
+//! Chrono aims to provide all functionality needed to do correct operations on dates and times in
```
djc, line 4: How did you reflow? It doesn't look very consistent.

pitdicker: Manually against a line at the 100-characters column in my editor. And sorry, it should be a bit better now. I missed a few lines, and later changes to a few links messed up parts that I did before.

src/lib.rs
```diff
@@ -35,16 +36,19 @@
 //!
 //! - `serde`: Enable serialization/deserialization via [serde].
 //! - `rkyv`: Deprecated, use the `rkyv-*` features.
-//! - `rkyv-16`: Enable serialization/deserialization via [rkyv], using 16-bit integers for integral `*size` types.
-//! - `rkyv-32`: Enable serialization/deserialization via [rkyv], using 32-bit integers for integral `*size` types.
-//! - `rkyv-64`: Enable serialization/deserialization via [rkyv], using 64-bit integers for integral `*size` types.
+//! - `rkyv-16`: Enable serialization/deserialization via [rkyv],
```
djc, line 39: In particular, why use the shorter lines here?

pitdicker: The alternative would be
```rust
//! - `rkyv-16`: Enable serialization/deserialization via [rkyv], using 16-bit integers for integral
//!   `*size` types.
```

A manual wrap in this case seemed a little nicer.

djc: Fair enough.

djc (approved):

PR #1540, opened 2024-03-24
run cargo publish, package

jtmoon79:
Test that `cargo publish` and `cargo package` succeed.


pitdicker:
Thank you! What errors would this catch that our regular build and test CI runs don't catch?

pitdicker:
Just added a similar Ci job for chrono-tz in https://github.com/chronotope/chrono-tz/pull/158.
So I now see its use :smile:.

@jtmoon79 Want to finish up this PR?

PR #1541, opened 2024-03-25
Add: Support wasm32-unknown-emscripten

sevenc-nanashi:
I added support for wasm32-unknown-emscripten.
With this PR, chrono will get timezone support.


.github/workflows/test.yml
```diff
@@ -121,6 +121,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
+      - uses: mymindstorm/setup-emsdk@v14
+        if: matrix.target == 'wasm32-unknown-emscripten'
```
pitdicker, line 125: These lines are not needed I think?

Cargo.toml
```diff
@@ -26,7 +26,7 @@ std = ["alloc"]
 clock = ["winapi", "iana-time-zone", "android-tzdata", "now"]
 now = ["std"]
 oldtime = []
-wasmbind = ["wasm-bindgen", "js-sys"]
+wasmbind = ["wasm-bindgen", "js-sys", "emscripten-functions"]
```
pitdicker, line 29: Emscripten should get its own `dep:` dependency or feature.

ci/emscripten-test/js/test.js
```diff
@@ -0,0 +1,37 @@
...
+  assert(diffUtc < 1000);
+  assert(diffLocal < 1000);
+
+  console.log("")
+  console.log("Test #2: Timezone offset should be the same:")
+  const timezoneJs = new Date().getTimezoneOffset();
+  const timezoneWasm = chrono.timezoneOffset();
+  console.log(`Timezone (JS): ${timezoneJs} min`);
+  console.log(`Timezone (Wasm): ${timezoneWasm} sec (${timezoneWasm / 60} min)`);
+  assert(timezoneJs === timezoneWasm / 60);
```
pitdicker, line 33: This would be tricky if the time zone has DST. But with `Asia/Tokyo` it seems all right. Maybe measure the time zone in JS before and after the rust method, and assert that at least one must match? And a comment to explain why.

src/offset/local/mod.rs
```diff
@@ -16,7 +16,7 @@ use crate::naive::{NaiveDate, NaiveDateTime, NaiveTime};
 use crate::Date;
 use crate::{DateTime, Utc};
 
-#[cfg(unix)]
+#[cfg(all(unix, not(all(target_arch = "wasm32", feature = "wasmbind", target_os = "emscripten"))))]
```
pitdicker, line 19: Why do we need this? Does the emscripten target identify as `unix`?

src/offset/utc.rs
```diff
@@ -109,6 +101,19 @@ impl Utc {
         let now = js_sys::Date::new_0();
         DateTime::<Utc>::from(now)
     }
+
+    /// Returns a `DateTime` which corresponds to the current date and time.
+    #[cfg(all(target_arch = "wasm32", feature = "wasmbind", target_os = "emscripten"))]
+    #[must_use]
+    pub fn now() -> DateTime<Utc> {
+        use emscripten_functions::emscripten::run_script_string;
+        // In order to handle number than i32, string is used to store the number
```
pitdicker, line 110:
```suggestion
        // `run_script_string` instead of `run_script_int` so we can parse the return value as `i64` instead of `i32`.
```

src/offset/utc.rs
```diff
@@ -109,6 +101,19 @@ impl Utc {
...
+        let now = run_script_string("Date.now().toString()")
+            .expect("Date.now() failed")
+            .parse::<i64>()
+            .unwrap();
```
pitdicker, line 114: Better to have something different than `unwrap` here. Maybe use `map` or `and_then`?

pitdicker (commented):

@ALEX11BR As the author of emscripten-functions would you be willing to comment on if this is the right direction for chrono on emscripten?

.github/workflows/test.yml
sevenc-nanashi: They are: https://github.com/chronotope/chrono/actions/runs/8430535599/job/23086540259?pr=1541

src/offset/local/mod.rs
sevenc-nanashi: Yes.

<details>
<summary>Details</summary>

```rs
❯ cat test.rs
#[cfg(unix)]
fn i_am_unix() {}

#[cfg(target_arch = "wasm32")]
fn i_am_wasm32() {}

#[cfg(target_os = "emscripten")]
fn i_am_emscripten() {}

fn main() {}

❯ rustc test.rs --target wasm32-unknown-unknown
warning: function `i_am_wasm32` is never used
 --> test.rs:5:4
  |
5 | fn i_am_wasm32() {}
  |    ^^^^^^^^^^^
  |
  = note: `#[warn(dead_code)]` on by default

warning: 1 warning emitted

❯ rustc test.rs --target wasm32-unknown-emscripten
warning: function `i_am_unix` is never used
 --> test.rs:2:4
  |
2 | fn i_am_unix() {}
  |    ^^^^^^^^^
  |
  = note: `#[warn(dead_code)]` on by default

warning: function `i_am_wasm32` is never used
 --> test.rs:5:4
  |
5 | fn i_am_wasm32() {}
  |    ^^^^^^^^^^^

warning: function `i_am_emscripten` is never used
 --> test.rs:8:4
  |
8 | fn i_am_emscripten() {}
  |    ^^^^^^^^^^^^^^^

warning: 3 warnings emitted

```

</details>

djc:
I don't find this PR convincing. The crate you're advocating we start depending on:

* Was created by an otherwise unknown single engineer only 7 months ago
* Has had a total of 700 downloads in those 7 months

I don't think that's a stable base for us to build on.

Apart from that I'd like to hear a rationale for adding an emscripten target, which I think has been superseded in most use cases by wasm32-unknown-unknown and WASI.

ALEX11BR:
> Thank you for working on this!
> 
> @ALEX11BR As the author of emscripten-functions would you be willing to comment on if this is the right direction for chrono on emscripten?

Looks mostly OK to me.
I had a remark to make with regards to the use of `run_script` functions where direct bindings to the used JS functions are available.

pitdicker:
@sevenc-nanashi Sorry for the wait. Could you reply to the concerns of @djc? They match my concerns on discord.

I know little about the current status of emscripten. Stdweb seems to be no more, and `emscripten-functions` is a crate that doesn't see much use yet.

sevenc-nanashi:
Oops, sorry for the late reply.

In most cases, using wasm-pack (wasn32-unknown-unknown) or wasi is enough to build a new software.
But emscripten has a big advantage: it can compile C source (in other words, crates using cc can work with emscripten)
I used emscripten to port the software written in Rust.
(It's just my opinion: Isn't some software written in Rust becoming runnable on browser so fascinating?)

I don't care about closing this PR as lack of use cases.

pitdicker:
How hard would it be to link to `emscripten_run_script_string` directly?

pitdicker:
Hmm. Cargo requires that there is at most one crate that links to a C library. So a project using stdweb is incompatible with emscripten-functions, and would become incompatible with chrono if we link directly.

sevenc-nanashi:
> `emscripten_run_script_string` directly?

It shouldn't be so hard I guess.

pitdicker:
Unfortunately because of the state of the `wasm32-unknown-emscripten` target I think it is best not to merge this PR for now 😞. Thank you for all your efforts @sevenc-nanashi!

PR #1547, opened 2024-03-27
Swap `MappedLocalTime` and `LocalResult` type alias

pitdicker:
In #1501 I thought renaming `LocalResult` to `MappedLocalTime` would be entirely without consequences as long as there is a type alias with the old name. Turns out there is one case where a type alias behaves differently :disappointed:. You can't import enum variants from a type alias with `use chrono::LocalResult::*`.

We have two choices:
- revert the rename entirely for the 0.4.x versions.
- change the enum definition to be on `LocalResult`, and use the type alias everywhere else.

The second choice has my preference. The docs look mostly sane, but break down once you come to the trait implementations section:

afbeelding

...

afbeelding

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -3,8 +3,8 @@ cff-version: 1.2.0
 message: Please cite this crate using these information.
 
 # Version information.
-date-released: 2024-03-26
-version: 0.4.36
+date-released: 2024-03-27
+version: 0.4.37
 
 # Project information.
 abstract: Date and time library for Rust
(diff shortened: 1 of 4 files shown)


djc (approved):
Definitely annoying that importing the variants doesn't work correctly.

pitdicker:
Published.

I'll look through the rust issue tracker if there is anything about type aliases for enums.

pitdicker:
Opened https://github.com/rust-lang/rust/issues/123131.

PR #1550, opened 2024-03-30
Inline `test_encodable_json` and `test_decodable_json` functions

pitdicker:
The `test_encodable_json` and `test_decodable_json` functions in the modules of `NaiveDate`, `NaiveTime`, `NaiveDateTime` and `DateTime` used to help testing the serde and `rustc-serialize` implementations.

Now that we only have serde we can inline them in into the tests.


djc (approved):
Nice!

PR #1551, opened 2024-03-30
Index `Month` enum from 1

pitdicker:
For 0.5 I'd like to make the `Month` enum less useless.

By defining `Month::January` as 1 users can do simple casting, and use it for example in:
```rust
let date = NaiveDate::from_ymd(2015, Month::March as u32, 14);
```

We can also choose to make `Datelike::month` return `Month` (#727). I didn't do so here for two reasons:
- For my personal projects using integers for months was perfectly fine and matches that `year()` and `day()` also return integers. I find working with integers more convenient than wordy month names.
- Returning en enum will introduce extra checks in the `month` method on `NaiveDate` that can't be optimized out without unsafe code. We know `Mdf::from_ol(/* ... */).month()` will reaturn a value in `1..=12`, but the compiler doesn't.

### DIFF
--- a/src/month.rs
+++ b/src/month.rs
@@ -8,19 +8,19 @@ use crate::OutOfRange;
 /// The month of the year.
 ///
 /// This enum is just a convenience implementation.
-/// The month in dates created by DateLike objects does not return this enum.
+/// The month in dates created by `Datelike` objects does not return this enum.
 ///
-/// It is possible to convert from a date to a month independently
+/// It is possible to convert from a date to a month using `try_from`:
 /// ```
 /// use chrono::prelude::*;
 /// use chrono::Month;
 ///
 /// let date = Utc.at_ymd_and_hms(2019, 10, 28, 9, 10, 11).unwrap();
 /// // `2019-10-28T09:10:11Z`
-/// let month = Month::try_from(u8::try_from(date.month()).unwrap()).ok();
-/// assert_eq!(month, Some(Month::October))
+/// let month = Month::try_from(date.month() as u8).unwrap();
+/// assert_eq!(month, Month::October)
 /// ```
-/// Or from a Month to an integer usable by dates
+/// Or from a `Month` to an integer with simple casting:
 /// ```
 /// # use chrono::prelude::*;
 /// # use chrono::Month;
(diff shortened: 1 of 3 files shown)


src/month.rs
```diff
@@ -42,29 +41,29 @@ use crate::OutOfRange;
 #[cfg_attr(all(feature = "arbitrary", feature = "std"), derive(arbitrary::Arbitrary))]
 pub enum Month {
     /// January
-    January = 0,
+    January = 1,
```
djc, line 44: Hmm, should we also make this explicitly `repr(u8)` then?

djc (approved):

src/month.rs
```diff
@@ -8,29 +8,28 @@ use crate::OutOfRange;
...
+/// assert_eq!(month, Month::October)
 /// ```
-/// Or from a Month to an integer usable by dates
+/// Or from a `Month` to an integer with simple casting:
 /// ```
 /// # use chrono::prelude::*;
 /// # use chrono::Month;
 /// let month = Month::January;
-/// let dt = Utc.at_ymd_and_hms(2019, month.number_from_month(), 28, 9, 10, 11).unwrap();
+/// let dt = Utc.at_ymd_and_hms(2019, month as u32, 28, 9, 10, 11).unwrap();
```
djc, line 28: Should we change `at_ymd_and_hms()` (and other places) to take `u8` for `month`?

src/month.rs
pitdicker: As I understand it the only extra thing `repr(u8)` on an enum allows is unsafe pointer casting.

src/month.rs
pitdicker: I don't know. A number of methods could take smaller arguments, such as `NaiveTime::from_hms`. Is there an advantage for users in taking a smaller type?

src/month.rs
djc: Maybe, but to me it's also like a signal to the users about how this type works?

src/month.rs
djc: Smaller types would probably allow more optimization potential.

src/month.rs
pitdicker: With `repr(u8)` it no longer has niche optimization, but I suppose there is little use for that.

djc: Okay, I would actually say that's a good reason not to add it.

src/month.rs
pitdicker: I'll investigate and open an issue or PR. It does mean users have to do potentially fallible casts more often, so I'm not sure it is better for egonomics.

PR #1552, opened 2024-03-30
0.5: Revise exposed crate features

pitdicker:
Remove the old crate features `libc`, `old_time`, `rkyv` and `winapi`.
Switch to the `dep:` syntax for optional dependencies that where exposed as features: `pure-rust-locales`, `android-tzdata`, `iana-time-zone`, `js-sys`, `wasm-bindgen` and `windows-targets`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -77,9 +77,8 @@ jobs:
       - uses: taiki-e/install-action@cargo-hack
       - uses: Swatinem/rust-cache@v2
       - run: |
-          cargo hack check --feature-powerset --optional-deps arbitrary,serde \
-            --skip __internal_bench,iana-time-zone,oldtime,pure-rust-locales,libc,winapi,rkyv-validation,wasmbind \
-            --mutually-exclusive-features arbitrary,rkyv,rkyv-16,rkyv-32,rkyv-64,serde \
+          cargo hack check --feature-powerset --skip rkyv-validation,wasmbind \
+            --mutually-exclusive-features arbitrary,rkyv-16,rkyv-32,rkyv-64,serde \
             --all-targets
         # run using `bash` on all platforms for consistent
         # line-continuation marks
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,26 +16,37 @@ rust-version = "1.61.0"
 [lib]
 name = "chrono"
 
-[features]
-# Don't forget to adjust `ALL_NON_EXCLUSIVE_FEATURES` in CI scripts when adding a feature or an optional dependency.
+[features] # Don't forget to adjust `ALL_NON_EXCLUSIVE_FEATURES` in CI scripts when adding a feature or an optional dependency.
 default = ["clock", "std"]
+# Enable features that depend on allocation (primarily string formatting).
 alloc = []
+# Implements `TryFrom<SystemTime>` and `std::error::Error` for error types, implies of `alloc`.
 std = ["alloc"]
+# Enables reading the local timezone (`Local`), implies `now`.
 clock = [
   "dep:windows-targets",
   "dep:iana-time-zone",
   "dep:android-tzdata",
   "now",
 ]
+# Enables reading the system time (`now()`), on most platforms using the standard library.
 now = []
+# Enable features that depend on allocation (primarily string formatting).
 serde = ["dep:serde"]
 # Note that rkyv-16, rkyv-32, and rkyv-64 are mutually exclusive.
+# Enable serialization/deserialization via rkyv, using 16-bit integers for integral `*size` types.
 rkyv-16 = ["dep:rkyv", "rkyv?/size_16"]
+# Enable serialization/deserialization via rkyv, using 32-bit integers for integral `*size` types.
 rkyv-32 = ["dep:rkyv", "rkyv?/size_32"]
+# Enable serialization/deserialization via rkyv, using 64-bit integers for integral `*size` types.
 rkyv-64 = ["dep:rkyv", "rkyv?/size_64"]
+# Enable rkyv validation support using `bytecheck`.
 rkyv-validation = ["dep:rkyv", "rkyv?/validation"]
+# Construct arbitrary instances of a type with the arbitrary crate.
 arbitrary = ["dep:arbitrary"]
+# Interface with the JS Date API for the `wasm32` target.
 wasmbind = ["dep:wasm-bindgen", "dep:js-sys"]
+# Enable localization. This adds various methods with a `_localized` suffix.
 unstable-locales = ["dep:pure-rust-locales"]
 
 [dependencies]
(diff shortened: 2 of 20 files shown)


pitdicker:
Also updated the description in `lib.rs` and the readme, and removed the `__unstable_bench` feature.

The bench feature was only used for benchmarking `YearFlags::from_year`. All that method does is:
```rust
let year = year.rem_euclid(400);
YEAR_TO_FLAGS[year as usize]
```

A division and a table lookup doesn't seem all that interesting to benchmark, let alone have a feature for.

Cargo.toml
```diff
@@ -16,31 +16,38 @@ rust-version = "1.61.0"
...
 
-[features]
-# Don't forget to adjust `ALL_NON_EXCLUSIVE_FEATURES` in CI scripts when adding a feature or an optional dependency.
-default = ["clock", "std"]
-alloc = []
-libc = []
-winapi = ["windows-targets"]
-std = ["alloc"]
-clock = ["winapi", "iana-time-zone", "android-tzdata", "now"]
+[features] # Don't forget to adjust `ALL_NON_EXCLUSIVE_FEATURES` in CI scripts when adding a feature or an optional dependency.
```
djc, line 19: This commit has a lot going on, can you split it in smaller pieces? For example, one for introducing the use of `dep:`, one for cleaning up the `clock`-related stuff, other stuff...

Also I think I'd prefer to have single-line comments on the same line as the feature, and keep lists of features on one line.

pitdicker: Done. Second try on making the commits somewhat reasonable :smile:.

> Also I think I'd prefer to have single-line comments on the same line as the feature, and keep lists of features on one line.

Sorry, I didn't write down what I was trying to do. libs.rs can show documentation for crate features but I am not sure of the details. Having a comment above the feature works for sure.

We have a CI run that checks the format of cargo.toml. I think that is quite useless personally. It wants the list of features for `clock` to be on multiple lines.

Cargo.toml
```diff
@@ -18,16 +18,16 @@ name = "chrono"
 
 [features]
 # Don't forget to adjust `ALL_NON_EXCLUSIVE_FEATURES` in CI scripts when adding a feature or an optional dependency.
-default = ["clock", "std"]
-alloc = []
-std = ["alloc"]
+default = ["clock"]
```
djc, line 20: Not sure about this change? In our code, do we actually only need `std` for things that are related to `clock`? Or is this more of an incidental dependency rather than a conceptual one?

.github/workflows/test.yml
```diff
@@ -87,7 +87,6 @@ jobs:
       - run: cargo test --no-default-features --features=alloc
       - run: cargo test --no-default-features --features=unstable-locales
       - run: cargo test --no-default-features --features=alloc,unstable-locales
-      - run: cargo test --no-default-features --features=now
```
djc, line 92: (Depends on the outcome of my other comment.)

djc:
Much better, I think that split was definitely worth it!

Cargo.toml
pitdicker: We used to have custom code for getting the current time from the OS but today we are only a wrapper around `SystemTime::now()`.

djc: I guess what I mean is, is that the only use of `std` in the library? I guess all the other stuff is just `alloc`?

pitdicker: We only use `std` for the `Error` trait, `SystemTime::now()`, and the stuff in the unix module.

djc: Okay, but our usage for the `Error` trait doesn't seem to be scoped to `clock` only, right?

pitdicker: I've removed the commit.

README.md
```diff
@@ -50,12 +50,10 @@ See [docs.rs](https://docs.rs/chrono/latest/chrono/) for the API reference.
 
 Default features:
 
+* `clock`: Enables reading the local timezone (`Local`), superset of `now`.
+* `now`: Enables reading the system time (`now()`) using the standard library.
+* `std`: Implements `std::error::Error` for error types, superset of `alloc`.
```
djc, line 55: I'd like to avoid the "superset of" wording, instead using something like "implies" or "also enables". Should we also make explicit that `now` enables `std` here?

djc (approved):

README.md
djc: I think this is in firmly in [Hyrum's Law](https://www.hyrumslaw.com/) territory, and generally will be seen as semver-breaking anyway if we decide to go back on it so might as well document it.

pitdicker: Haha, yes that is something to worry about. But I don't think that will be a problem here.

I added a commit "Don't let `now` imply `std` feature". Do you want to have a look?

With it the `std` feature only enables the `Error` trait, `TryFrom<SystemTime>`, and methods that depend on `alloc`.

I had to extract the code from `TryFrom<SystemTime>` into a method on `DateTime<Utc>` to split the `now` feature from `std`.

djc: So what functionality does `now` enable when `std` is disabled?

pitdicker: Only `Utc::now()`. And the `clock` feature adds `Local::now()`.

djc: `Utc::now()` calls `SystemTime::now()`, so it's actually dependent on `std`, right?

pitdicker: Currently it does on all platforms.

pitdicker: Now I wonder if we shouldn't just remove the `now` feature and fold it into `std`.

djc: But on WASM we need other stuff for `now`, right?

pitdicker: You are completely right, I forgot. It wouldn't have to depend on std on that platform.

pitdicker:
Sorry that this is taking so many back-and-fort between us. But I'm really starting to like the changes here as good cleanups.

Cargo.toml
```diff
@@ -27,7 +27,7 @@ clock = [
   "dep:android-tzdata",
   "now",
 ]
-now = ["std"]
+now = []
```
djc, line 33: So I think we should revert this change, because `now` without `std` is pretty much meaningless.

djc: I think `now` should enable calling `Utc::now()` on most common platforms. If it doesn't, that feels broken.

pitdicker: It does. https://github.com/chronotope/chrono/pull/1552/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R474

djc: It seems surprising that the `cfg(feature = "std")` guard doesn't match the behavior of whether we enable/disable `no_std` exactly (for the library).

pitdicker: It's a choice we can make. Separating the features allows things like using `Utc::now` in combination with `wasmbind` without depending on the standard library.

djc: Okay, let's try it. Easier to add it later than to remove it.

PR #1553, opened 2024-04-04
CI: Reduce combinations in `cargo hack check`

pitdicker:
I'm not sure why but within the last week the combinations `cargo hack check` tries on our CI has increased 4x to 672. Cargo hack has seen a couple of new releases, so there may be a change or bugfix there.

Disabling the `oldtime` (unused) and `wasmbind` (not checked on linux anyway) features reduces it back to 192 combinations.

We can also now inform cargo hack that some features are mutually exclusive, or don't effect each other. The `rkyv` features are truly mutually exclusive. The `arbitrary` feature is far removed from serialization, and serde also doesn't seem to have anything in common with rkyv. So I think we can treat them as mutually exclusive.

This allows us to run `cargo hack check` for more meaningful combinations with the `rkyv-*` features while we only check 84 combinations. Room to grow again :smile:.

This should bring our CI time back below 5 minutes instead of close to 20 minutes.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -80,7 +80,8 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - run: |
           cargo hack check --feature-powerset --optional-deps arbitrary,serde \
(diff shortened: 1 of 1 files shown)


pitdicker:
4m3s :smile:.

PR #1554, opened 2024-04-04
Only package files needed for building and testing

pitdicker:
Compared to what we packaged before we now skip those files:
```
.git-ignore-revs
.github/codecov.yml
.github/dependabot.yml
.github/pull_request_template.md
.github/workflows/codecov.yml
.github/workflows/lint.yml
.github/workflows/test.yml
.gitignore
CHANGELOG.md
CITATION.cff
deny.toml
rustfmt.toml
taplo.toml
```

Also we had a leftover `appveyor.yml` file. Not sure why it was not deleted outright in #436.


djc (approved):
I think we should maybe ship `CITATION.cff`? Not sure how that gets used but I can see how that might be useful.

PR #1558, opened 2024-04-06
Optimize formatting a bit

pitdicker:
This takes the commits from #1163 that simplify and optimize our formatting code a bit. #1335 took the initial refactor from that PR, so now we can take the more interesting part.

Two commits make better use of `match` to only accept a formatting item if we also have the date/time/offset fields it needs, instead of using `map` inside a match arm.

The last commit introduces a couple of optimized number formatting functions. They are the trick that makes our formatting implementations for RFC 2822 and 3339 faster, so lets use them for regular formatting too :smile:.

### Benchmarks
```
     Running benches/chrono.rs (target/release/deps/chrono-413c84d256ba35f5)
bench_format            time:   [533.37 ns 535.14 ns 537.22 ns]
                        change: [-22.588% -21.943% -21.284%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) high mild
  3 (3.00%) high severe

bench_format_with_items time:   [388.35 ns 389.61 ns 391.09 ns]
                        change: [-29.520% -28.855% -28.271%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
```

### DIFF
--- a/src/format/formatting.rs
+++ b/src/format/formatting.rs
@@ -162,45 +162,34 @@ impl<'a, I: Iterator<Item = B> + Clone, B: Borrow<Item<'a>>> DelayedFormat<I> {
 
     #[cfg(feature = "alloc")]
     fn format_fixed(&self, w: &mut impl Write, spec: &Fixed) -> fmt::Result {
-        use self::Fixed::*;
+        use Fixed::*;
+        use InternalInternal::*;
 
-        let ret = match *spec {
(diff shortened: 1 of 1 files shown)


djc (approved):
Faster in less code, great work!

pitdicker:
Thank you :smile:.

PR #1559, opened 2024-04-06
Make `DelayedFormat` hold a generic offset

pitdicker:
Just learned that adding a type parameter is a backwards-compatible change if it has a default type, thanks to [RFC 0213](https://rust-lang.github.io/rfcs/0213-defaulted-type-params.html).

That means we can fix the biggest remaining performance problem with `DelayedFormat`, and make our formatting work without allocations! :tada:

In this PR I only added an extra generic parameter to `DelayedFormat`. Making our formatting work without allocating needs one more tricky thing that is best for another PR.

To quote from #1163:
> Our current formatting code has noticeable overhead (see #94). This is mostly caused by one design choice: if `DelayedFormat::new` is given an offset, it first formats a timezone name into a `String`. In case of `DateTime<FixedOffset>` and `DateTime<Local>` this involves formatting the offset, as there is no name available. This is responsible for 20~25% of the ca. 40% overhead in the `format_with_items` benchmark.

By taking the offset as a generic we can delay formatting the timezone name until it is needed, which often is never.

To keep the deprecated `format::{format, format_items}` functions working (which expect a `String` and `FixedOffset` instead of a generic offset) I had to add an `OffsetWrapper` type.

### Benchmarks
```
     Running benches/chrono.rs (target/release/deps/chrono-413c84d256ba35f5)
bench_format            time:   [467.24 ns 470.63 ns 475.26 ns]
                        change: [-16.270% -15.552% -14.816%] (p = 0.00 < 0.05)
                        Performance has improved.
  5 (5.00%) high mild
  5 (5.00%) high severe

bench_format_with_items time:   [323.83 ns 325.30 ns 326.94 ns]
                        change: [-18.296% -17.524% -16.769%] (p = 0.00 < 0.05)
                        Performance has improved.
  4 (4.00%) high mild
  4 (4.00%) high severe
```

### DIFF
--- a/src/format/formatting.rs
+++ b/src/format/formatting.rs
@@ -4,7 +4,7 @@
 //! Date and time formatting routines.
 
 #[cfg(all(feature = "alloc", not(feature = "std"), not(test)))]
-use alloc::string::{String, ToString};
+use alloc::string::String;
 #[cfg(feature = "alloc")]
 use core::borrow::Borrow;
 #[cfg(feature = "alloc")]
@@ -16,7 +16,7 @@ use crate::offset::Offset;
 #[cfg(any(feature = "alloc", feature = "serde"))]
 use crate::{Datelike, FixedOffset, NaiveDateTime, Timelike};
 #[cfg(feature = "alloc")]
-use crate::{NaiveDate, NaiveTime, Weekday};
+use crate::{NaiveDate, NaiveTime, Utc, Weekday};
 
 #[cfg(feature = "alloc")]
 use super::locales;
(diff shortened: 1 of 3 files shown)


src/format/formatting.rs
```diff
@@ -46,28 +46,17 @@ pub struct DelayedFormat<I> {
 }
 
 #[cfg(feature = "alloc")]
-impl<'a, I: Iterator<Item = B> + Clone, B: Borrow<Item<'a>>> DelayedFormat<I> {
+impl<'a, I, B> DelayedFormat<I>
+where
+    I: Iterator<Item = B> + Clone,
+    B: Borrow<Item<'a>>,
+{
```
pitdicker, line 53: As a note: `DelayedFormat::{new, new_with_locale}` are in an impl block without a generic offset parameter. Because they don't take an offset as argument it would fail during type inference otherwise.

`new_with_offset` is in an impl block with the new generic offset parameter.

PR #1569, opened 2024-04-12
Add `DateTime::to_timezone` and `FromOffset` trait

pitdicker:
We can't have a generic `From` or `Into` implementation that converts from one `Offset` type into another because the standard library has a blanket implementation of `From<T> for T`. For the same reason we can't have a generic `From` or `Into` implementation that converts from one `DateTime<Tz>` into another.

So we can and do have implementations to convert between `DateTime<Utc>`, `DateTime<FixedOffset>` and `DateTime<Local>`. But any `TimeZone` type outside chrono doesn't have this convenience, such as the one in chrono-tz.

I propose to add a `FromOffset` trait that `Offset` types can implement to convert from any generic `Offset`. And to add a `DateTime::to_timezone` method to conveniently convert a `DateTime<Tz>` into another type.

If we implement `FromOffset` in chrono-tz it we can do nice things such as converting from `FixedOffset` to a `Tz` time zone for a fixed offset, or eventually from `LocalOffset` (#1563) to a `Tz` with the same time zone if possible.


djc:
> I propose to add a `FromOffset` trait that `Offset` types can implement to convert from any generic `Offset`. And to add a `DateTime::to_timezone` method to conveniently convert a `DateTime<Tz>` into another type.
> 
> If we implement `FromOffset` in chrono-tz it we can do nice things such as converting from `FixedOffset` to a `Tz` time zone for a fixed offset, or eventually from `LocalOffset` (#1563) to a `Tz` with the same time zone if possible.

This all sounds kind of speculative and not as a current/near-future thing?

If so, I'd rather hold off on merging this now.

Alternatively stated: what do you envision to be the concrete near-term benefit for our users?

pitdicker:
I hope to prove it useful soon. Fine to hold off for now.

Issue #1571, opened 2024-04-15
0.4.37 semver-incompatibly removes trait bounds on `DateTime`


little-dude:
Hello,

The following code compile in 0.4.35 but breaks in 0.4.37:

```rust
use chrono::TimeZone;
use chrono::DateTime;

#[derive(Clone, Debug)]
pub struct MyDateTime<T: TimeZone>(DateTime<T>);

impl<T: TimeZone> Copy for MyDateTime<T> where T::Offset: Copy {}
```

With 0.4.37:
```
error[E0204]: the trait `Copy` cannot be implemented for this type
 --> src/main.rs:7:28
  |
5 | pub struct MyDateTime<T: TimeZone>(chrono::DateTime<T>);
  |                                    ------------------- this field does not implement `Copy`
6 |
7 | impl<T: TimeZone> Copy for MyDateTime<T> where T::Offset: Copy {}
  |                            ^^^^^^^^^^^^^
  |
note: the `Copy` impl for `DateTime<T>` requires that `T: Copy`
 --> src/main.rs:5:36
  |
5 | pub struct MyDateTime<T: TimeZone>(chrono::DateTime<T>);
  |                                    ^^^^^^^^^^^^^^^^^^^
help: consider further restricting this bound
  |
7 | impl<T: TimeZone + Copy> Copy for MyDateTime<T> where T::Offset: Copy {}
```
This is probably due to https://github.com/chronotope/chrono/pull/1492

djc:
Sorry for the regression! We'll look into this. I don't think we're going to just revert the changes for #1492, but let's see what we can do.

little-dude:
No worries, it's not a big deal for me. I just thought opening an issue might be good if other people run into this.

djc:
Right -- I think I'm inclined to let this one go unless there are multiple reporters for whom it turns out to be a larger problem. As I understand it, it mostly requires people add bounds that they should have been adding anyway.

pitdicker:
The manual `Copy` implementation that was deleted in #1492 was sound, it was only the comment that was incorrect. We had it because the type of `Tz` doesn't matter as it isn't a field of the `DateTime` struct.

For the `Send` implementation I'm not 100% sure it was sound.

pitdicker:
Reverting the change to `Copy` is easy, shall I just make a PR?

little-dude:
Yeah I agree that the previous `Copy` impl was better than the derived one. Wdyt @djc ?

[...]

PR #1572, opened 2024-04-15
manually implement `Copy` on `DateTime<Tz>`

little-dude:
`derive` generates the following impls:

```rust
impl<Tz: ::core::marker::Copy + TimeZone> ::core::marker::Copy for DateTime<Tz>
where
    Tz::Offset: ::core::marker::Copy,
{}

impl<Tz: ::core::clone::Clone + TimeZone> ::core::clone::Clone for DateTime<Tz>
where
    Tz::Offset: ::core::clone::Clone,
{
    #[inline]
    fn clone(&self) -> DateTime<Tz> {
        DateTime {
            datetime: ::core::clone::Clone::clone(&self.datetime),
            offset: ::core::clone::Clone::clone(&self.offset),
        }
    }
}
```

In the `Copy` impl, the `Tz: Copy` bound is un-necessary. Note that in the `Clone` impl, he `Tz: Clone` is also un-necessary, but it's implied by the `TimeZone` bound anyway, so there's not point having a manual implementation of `Clone`.

Fixes https://github.com/chronotope/chrono/issues/1571

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -46,7 +46,7 @@ mod tests;
 /// There are some constructors implemented here (the `from_*` methods), but
 /// the general-purpose constructors are all via the methods on the
 /// [`TimeZone`](./offset/trait.TimeZone.html) implementations.
-#[derive(Copy, Clone)]
+#[derive(Clone)]
 #[cfg_attr(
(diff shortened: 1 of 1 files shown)


pitdicker:
We were pretty much at the same time, thank you for working on it!

Issue #1577, opened 2024-04-29
NaiveDate::years_since wrong documentation


Taxalo:
The documentation on the `NaiveDate::years_since` says it `Returns None if base < self.`. But this is not correct. It is actually `Returns None if base > self.`. As the implementation is:

```rs
pub const fn years_since(&self, base: Self) -> Option<u32> {
        let mut years = self.year() - base.year(); 
        // Comparing tuples is not (yet) possible in const context. Instead we combine month and
        // day into one `u32` for easy comparison.
        if (self.month() << 5 | self.day()) < (base.month() << 5 | base.day()) {
            years -= 1;
        }

        match years >= 0 {
            true => Some(years as u32),
            false => None,
        }
    }
```

Base must be bigger than self to return None, take the example of _2022-01-01_ and _1998-01-01._ If base was 2022 and 1998 self it will be `1998 - 2022 < 0`. Which will return None from being negative.

PR #1578, opened 2024-04-29
#1577: Changed years_since documentation to match its implementation

Taxalo:
Fixes documentation on #1577 by changing the comments on both NaiveDate and DateTime.

No tests are needed since its only comments.


djc (approved):

pitdicker:
Thank you, good you fixed both :+1:.

PR #1582, opened 2024-05-23
Remove obsolete weird feature guard

djc:
This breaks CI on nightly because nightly is now checking for unknown features.


pitdicker (approved):
The line was last changed in https://github.com/chronotope/chrono/pull/358/commits/db8784f97cbe71a37f49f89d3cc95288f8dbc5cf. Seems to have been weird before that.

PR #1590, opened 2024-06-10
native/date: Improve DelayedFormat doc re Panics

behnam-oneschema:
Since `#[should_panic]` kind of checks are not available for doctesting, cannot add a test for it in the current structure. The panic behavior probably could use a unit test somewhere else, if it doesn't exist yet.


djc (approved):
Thanks!

PR #1595, opened 2024-07-08
Apply clippy suggestions from Rust 1.79

djc:


pitdicker (approved):
I was just looking into this also. :+1: when the CI is happy.

PR #1596, opened 2024-07-08
Use environment secrets for Codecov token

djc:
It seems like there are some unexpected differences between "Repository secrets" and "Environment secrets" for GitHub repositories. In another repo, this seemed to help. (I've copied the token into the new Environment.)

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -13,6 +13,7 @@ jobs:
   job_code_coverage:
     name: llvm-cov
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


pitdicker:
We can't easily test this I think? What do you think about adding it as a commit to https://github.com/chronotope/chrono/pull/1594?

djc:
It worked in this PR, right? Can rebase #1594 on top of this after it merges -- it certainly doesn't make anything worse.

pitdicker (approved):

PR #1598, opened 2024-07-16
docs: document century cutoff for %y

### LINKED ISSUE #1597: Century cutoff for %y undocumented
If I run
```rust
use chrono::NaiveDate;


fn main() {
    let res = NaiveDate::parse_from_str("01-01-69", "%d-%m-%y");
    println!("res: {:?}", res);
    let res = NaiveDate::parse_from_str("01-01-70", "%d-%m-%y");
    println!("res: {:?}", res);
}
```
I get
```
res: Ok(2069-01-01)
res: Ok(1970-01-01)
```

It seems that 70 is the cutoff, but the docs don't mention it https://docs.rs/chrono/latest/chrono/format/strftime/index.html#fn1

MarcoGorelli:
closes #1597

### DIFF
--- a/src/format/strftime.rs
+++ b/src/format/strftime.rs
@@ -99,6 +99,8 @@ Notes:
 
 [^1]: `%C`, `%y`:
    This is floor division, so 100 BCE (year number -99) will print `-1` and `99` respectively.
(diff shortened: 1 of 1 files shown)


djc (approved):
Thanks!

Issue #1469, opened 2024-02-28
Tracking issue: Converting the API to return `Result`s


pitdicker:
A list to keep score. Not yet complete, and only lists the public API.

Currently at 61/153.

### NaiveDate

- [x] `NaiveDate::from_ymd`: #1444
- [x] `NaiveDate::from_yo`: #1460
- [x] `NaiveDate::from_isoywd`: #1460
- [x] `NaiveDate::from_num_days_from_ce`: #1470
- [x] `NaiveDate::from_weekday_of_month`: #1512
- [ ] `NaiveDate::parse_from_str`
- [ ] `NaiveDate::parse_and_remainder`
- [x] `NaiveDate::and_hms`: #1410
- [x] `NaiveDate::and_hms_milli`: #1410
- [x] `NaiveDate::and_hms_micro`: #1410
- [x] `NaiveDate::and_hms_nano`: #1410
- [x] `NaiveDate::succ`: #1513
- [x] `NaiveDate::pred`: #1513
- [x] `NaiveDate::checked_add_days`: #1475
- [x] `NaiveDate::checked_sub_days`: #1475
- [x] `NaiveDate::checked_add_months`: #1537
- [x] `NaiveDate::checked_sub_months`: #1537
- [x] `NaiveDate::checked_add_signed`: #1514
- [x] `NaiveDate::checked_sub_signed`: #1514
- [x] `NaiveDate::with_year`: #1466
- [x] `NaiveDate::with_month`: #1509
- [x] `NaiveDate::with_day`: #1509
- [x] `NaiveDate::with_ordinal`: #1531

### NaiveTime

- [x] `NaiveTime::from_hms`: #1410
- [x] `NaiveTime::from_hms_micro`: #1410
- [x] `NaiveTime::from_hms_milli`: #1410
- [x] `NaiveTime::from_hms_nano`: #1410
- [x] `NaiveTime::from_num_seconds_from_midnight`: #1410
- [ ] `NaiveTime::parse_from_str`
- [ ] `NaiveTime::parse_and_remainder`
- [x] `NaiveTime::with_hour`: #1498
- [x] `NaiveTime::with_minute`: #1498
- [x] `NaiveTime::with_second`: #1498
- [x] `NaiveTime::with_nanosecond`: #1498

### NaiveDateTime

- [ ] `NaiveDateTime::and_local_timezone` (to `LocalResult`)
- [ ] `NaiveDateTime::parse_and_remainder`
[...]

PR #1600, opened 2024-07-18
Checked `NaiveWeek` methods

bragov4ik:
Currently, all `NaiveWeek`'s public methods panic on overflow. This PR allows to handle such scenarios.

Seems to solve https://github.com/chronotope/chrono/issues/1469 (`NaiveWeek` part).

* I'm not sure that `Result` makes sense here, since only overflow results in `None`. Thus `Option`.
* Adding `checked_*` methods in order to not break API

### DIFF
--- a/src/naive/mod.rs
+++ b/src/naive/mod.rs
@@ -60,13 +60,37 @@ impl NaiveWeek {
     #[inline]
     #[must_use]
     pub const fn first_day(&self) -> NaiveDate {
+        expect(self.checked_first_day(), "first weekday out of range for `NaiveDate`")
+    }
+
+    /// Returns a date representing the first day of the week or
+    /// `None` if the date is out of `NaiveDate`'s range
+    /// (more than ca. 262,000 years away from common era).
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use chrono::{NaiveDate, Weekday};
+    ///
+    /// let date = NaiveDate::MIN;
+    /// let week = date.week(Weekday::Mon);
+    /// if let Some(first_day) = week.checked_first_day() {
+    ///     assert!(first_day == date);
+    /// } else {
+    ///     // error handling code
+    ///     return;
+    /// };
+    /// ```
+    #[inline]
+    #[must_use]
+    pub const fn checked_first_day(&self) -> Option<NaiveDate> {
         let start = self.start.num_days_from_monday() as i32;
         let ref_day = self.date.weekday().num_days_from_monday() as i32;
         // Calculate the number of days to subtract from `self.date`.
         // Do not construct an intermediate date beyond `self.date`, because that may be out of
         // range if `date` is close to `NaiveDate::MAX`.
         let days = start - ref_day - if start > ref_day { 7 } else { 0 };
-        expect(self.date.add_days(days), "first weekday out of range for `NaiveDate`")
+        self.date.add_days(days)
     }
 
     /// Returns a date representing the last day of the week.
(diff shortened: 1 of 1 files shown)


djc:
This makes sense to me, thanks. In its current form the diffs get a little ugly, would you mind doing this with one commit per method and putting the panicking method above the `checked` method? Also probably prefer not to use the "recoverable"/"recover" wording which IMO is not used very often.

src/naive/mod.rs
```diff
@@ -159,6 +159,34 @@ impl NaiveWeek {
...
+    /// [checked_first_day](NaiveWeek::checked_first_day) and
+    /// [checked_last_day](NaiveWeek::checked_last_day) functions.
+    ///
+    /// Returns `None` if either of the boundaries are out of `NaiveDate`'s range
+    /// (more than ca. 262,000 years away from common era).
+    ///
+    ///
+    /// # Examples
+    ///
+    /// todo: examples with min/max
```
djc: Still `todo`?

src/naive/mod.rs
```diff
@@ -159,6 +159,34 @@ impl NaiveWeek {
...
+    /// use chrono::{NaiveDate, Weekday};
+    ///
+    /// let date = NaiveDate::from_ymd_opt(2022, 4, 18).unwrap();
+    /// let week = date.week(Weekday::Mon);
+    /// let days = week.checked_days().unwrap();
+    /// assert!(days.contains(&date));
+    /// ```
+    #[inline]
+    #[must_use]
+    pub const fn checked_days(&self) -> Option<RangeInclusive<NaiveDate>> {
```
djc, line 198: I'd expect this to be accompanied by a change elsewhere thats relying on this method except that it panics on `None`, why is that not the case here?

src/naive/mod.rs
```diff
@@ -75,11 +75,12 @@ impl NaiveWeek {
     ///
     /// let date = NaiveDate::MIN;
     /// let week = date.week(Weekday::Mon);
-    /// let Some(first_day) = week.checked_first_day() else {
+    /// if let Some(first_day) = week.checked_first_day() {
```
djc, line 78: Please squash these changes into the originating commit.

src/naive/mod.rs
bragov4ik: I don't understand 🙈
Do you mean adapt some other code within the library to use this new method, reference a code snipped that led to me making this PR, or something else?

src/naive/mod.rs
djc: Yes, exactly, I'd prefer to logically group all the changes via the principle of [atomic commits](https://www.aleksandrhovhannisyan.com/blog/atomic-git-commits/) -- this makes reviewing more efficient for me.

src/naive/mod.rs
bragov4ik: Changed `days()` as you described (but the downside is that now panic doesn't show which of the boundaries overflowed).
Didn't find usages of panicking functions within the crate for replacement (hopefully `rust-analyzer` searched well 😃 )

src/naive/mod.rs
```diff
@@ -6,22 +6,23 @@
...
 pub(crate) mod time;
 
-pub use self::date::{NaiveDate, NaiveDateDaysIterator, NaiveDateWeeksIterator};
 #[allow(deprecated)]
 pub use self::date::{MAX_DATE, MIN_DATE};
 #[allow(deprecated)]
 pub use self::datetime::{NaiveDateTime, MAX_DATETIME, MIN_DATETIME};
-pub use self::isoweek::IsoWeek;
-pub use self::time::NaiveTime;
+pub use self::{
```
djc, line 21: Please stick with the pre-existing import style of one module per line (that is, this should become three separate lines).

djc (approved):
LGTM modulo one nit, thanks! (Please fix it in the originating commit.)

src/naive/mod.rs
bragov4ik: Done!

PR #1613, opened 2024-09-13
Support ohos tzdata since ver.oh35

MirageLyu:
This is to support openHarmony OS(ohos) format timezone data. ohos has merged timezone files into single file 'tzdata' since recent version. pls review this, thanks.

### DIFF
--- a/src/offset/local/tz_info/timezone.rs
+++ b/src/offset/local/tz_info/timezone.rs
@@ -8,6 +8,9 @@ use std::{cmp::Ordering, fmt, str};
 use super::rule::{AlternateTime, TransitionRule};
 use super::{parser, Error, DAYS_PER_WEEK, SECONDS_PER_DAY};
 
+#[cfg(target_env = "ohos")]
+use crate::offset::local::tz_info::parser::Cursor;
+
 /// Time zone
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub(crate) struct TimeZone {
@@ -48,6 +51,12 @@ impl TimeZone {
             if let Ok(bytes) = android_tzdata::find_tz_data(tz_string) {
                 return Self::from_tz_data(&bytes);
             }
+        }
+
+        // ohos merge all file into tzdata since ver35
+        #[cfg(target_env = "ohos")]
+        {
+            return Self::from_tz_data(&find_ohos_tz_data(tz_string)?);
         }
 
         let mut chars = tz_string.chars();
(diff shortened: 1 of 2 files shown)


src/offset/local/tz_info/timezone.rs
```diff
@@ -629,6 +640,63 @@ fn find_tz_file(path: impl AsRef<Path>) -> Result<File, Error> {
...
+        }
+        idx += INDEX_CHUNK_SIZE as u32;
+    }
+
+    Err(Error::Io(io::ErrorKind::NotFound.into()))
+}
+
+#[cfg(target_env = "ohos")]
+fn from_tzdata_file(file: &mut File, tz_string: &str) -> Result<Vec<u8>, Error> {
+    let mut bytes: Vec<u8> = Vec::new();
```
djc, line 687: Nit: please ditch the unnecessary type annotation.

src/offset/local/tz_info/timezone.rs
```diff
@@ -629,6 +640,63 @@ fn find_tz_file(path: impl AsRef<Path>) -> Result<File, Error> {
     }
 }
 
+#[cfg(target_env = "ohos")]
+const TZDATA_PATH: &str = "/system/etc/zoneinfo/tzdata";
```
djc, line 644: Let's inline this into `find_ohos_tz_data()`.

src/offset/local/tz_info/timezone.rs
```diff
@@ -50,6 +53,14 @@ impl TimeZone {
             }
         }
 
+        // ohos merge all file into tzdata since ver35
+        #[cfg(target_env = "ohos")]
```
djc, line 57: Is there any point in falling through after this? If so, please add a comment making explicit why boths are useful.

src/offset/local/tz_info/timezone.rs
```diff
@@ -629,6 +640,63 @@ fn find_tz_file(path: impl AsRef<Path>) -> Result<File, Error> {
...
+    file.read_to_end(&mut bytes)?;
+    from_tzdata_bytes(&mut bytes, tz_string)
+}
+
+#[cfg(target_env = "ohos")]
+fn find_ohos_tz_data(tz_string: &str) -> Result<Vec<u8>, Error> {
+    if let Ok(mut file) = File::open(TZDATA_PATH) {
+        return from_tzdata_file(&mut file, tz_string);
+    }
+    Err(Error::Io(io::ErrorKind::NotFound.into()))
```
djc, line 697: There doesn't seem to be much point in synthesizing an `io::Error` here if we can just use the one from `File::open()`.

src/offset/local/tz_info/timezone.rs
```diff
@@ -629,6 +640,63 @@ fn find_tz_file(path: impl AsRef<Path>) -> Result<File, Error> {
...
+    if cursor.seek_after(index_offset_offset as usize).is_err() {
+        return Err(Error::Io(io::ErrorKind::NotFound.into()));
+    }
+    let mut idx = index_offset_offset;
+    while idx < data_offset_offset {
+        let index_buf = cursor.read_exact(ZONENAME_SIZE)?;
+        let offset = cursor.read_be_u32()?;
+        let length = cursor.read_be_u32()?;
+        let zone_name = str::from_utf8(index_buf)?.trim_end_matches('\0');
+        if zone_name == tz_string {
```
djc, line 670: Please invert this condition to reduce rightward drift in the block below.

src/offset/local/tz_info/timezone.rs
```diff
@@ -629,6 +640,63 @@ fn find_tz_file(path: impl AsRef<Path>) -> Result<File, Error> {
...
+    let mut cursor = Cursor::new(&bytes);
+    // version head
+    let _ = cursor.read_exact(VERSION_SIZE)?;
+    let index_offset_offset = cursor.read_be_u32()?;
+    let data_offset_offset = cursor.read_be_u32()?;
+    // final offset
+    let _ = cursor.read_be_u32()?;
+
+    if cursor.seek_after(index_offset_offset as usize).is_err() {
+        return Err(Error::Io(io::ErrorKind::NotFound.into()));
```
djc, line 662: Let's yield the error from `seek_after()` directly?

MirageLyu:
@djc Thanks for prompt and detailed review. I'm quite new to rust due to my work adjustment. I have modified the codes according to your advice, pls check again, thx!

src/offset/local/tz_info/timezone.rs
```diff
@@ -629,6 +638,58 @@ fn find_tz_file(path: impl AsRef<Path>) -> Result<File, Error> {
...
+    file.read_to_end(&mut bytes)?;
+    from_tzdata_bytes(&mut bytes, tz_string)
+}
+
+#[cfg(target_env = "ohos")]
+fn find_ohos_tz_data(tz_string: &str) -> Result<Vec<u8>, Error> {
+    const TZDATA_PATH: &str = "/system/etc/zoneinfo/tzdata";
+    match File::open(TZDATA_PATH) {
+        Ok(mut file) => from_tzdata_file(&mut file, tz_string),
+        Err(_err) => Err(Error::InvalidTzFile("invalid ohos tz file")),
```
djc, line 689: This seems wrong -- why aren't we returning `_err` here? The error here is not because the tz file is invalid but because we had trouble opening it.

src/offset/local/tz_info/timezone.rs
```diff
@@ -629,6 +638,58 @@ fn find_tz_file(path: impl AsRef<Path>) -> Result<File, Error> {
...
+            continue;
+        }
+        cursor.seek_after((data_offset_offset + offset) as usize)?;
+        return match cursor.read_exact(length as usize) {
+            Ok(result) => Ok(result.to_vec()),
+            Err(_err) => Err(Error::InvalidTzFile("invalid ohos tzdata chunk")),
+        };
+    }
+
+    Err(Error::Io(io::ErrorKind::NotFound.into()))
```
djc, line 674: `io::ErrorKind::NotFound` is about not being able to find a file, which is not the case here. Let's do a more appropriate error value?

MirageLyu: changed this to "InvalidTzString" error, because cannot find tzdata chunk via tz string

src/offset/local/tz_info/timezone.rs
```diff
@@ -629,6 +638,58 @@ fn find_tz_file(path: impl AsRef<Path>) -> Result<File, Error> {
...
+    file.read_to_end(&mut bytes)?;
+    from_tzdata_bytes(&mut bytes, tz_string)
+}
+
+#[cfg(target_env = "ohos")]
+fn find_ohos_tz_data(tz_string: &str) -> Result<Vec<u8>, Error> {
+    const TZDATA_PATH: &str = "/system/etc/zoneinfo/tzdata";
+    match File::open(TZDATA_PATH) {
+        Ok(mut file) => from_tzdata_file(&mut file, tz_string),
+        Err(_err) => Err(_err.into()),
```
djc, line 689:
```suggestion
        Err(err) => Err(err.into()),
```

djc (approved):

PR #1617, opened 2024-09-23
feat: Change to use `core::error::Error`

sorairolake:
Change to use [`core::error::Error`](https://doc.rust-lang.org/core/error/trait.Error.html) instead of [`std::error::Error`](https://doc.rust-lang.org/std/error/trait.Error.html). The MSRV is bumped up to 1.81.0 by this change.


djc:
I definitely don't want to take a 1.81 MSRV bump. We could make this an opt-in feature `core-error`, though?

PR #1618, opened 2024-09-30
Derive `PartialEq`,`Eq`,`Hash`,`Copy` and `Clone` on `NaiveWeek`

DSeeLP:
Derive `PartialEq`,`Eq`,`Hash`,`Copy` and `Clone` on `NaiveWeek`


djc (approved):

djc:
Thanks!

PR #1619, opened 2024-10-10
rkyv-08 feature

Awpteamoose:
Supersedes #1614

Not sure what to do about CI as currently the features are incompatible with rkyv 0.7, but I could use fully qualified paths for `rkyv_08::Archive` etc and then it should be ok?

Alternatively, could do smth like
```rust
// lib.rs
#[cfg(any(feature = "rkyv", feature = "rkyv-16", feature = "rkyv-32", feature = "rkyv-64"))]
pub(crate) use rkyv_07 as rkyv;

#[cfg(any(feature = "rkyv-08", feature = "rkyv-08-16", feature = "rkyv-08-32", feature = "rkyv-08-64"))]
pub(crate) use rkyv_08 as rkyv;

// elsewhere
use crate::rkyv::{Archive, Serialize, Deserialize};
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,11 @@ rkyv-16 = ["dep:rkyv", "rkyv?/size_16"]
 rkyv-32 = ["dep:rkyv", "rkyv?/size_32"]
 rkyv-64 = ["dep:rkyv", "rkyv?/size_64"]
 rkyv-validation = ["rkyv?/validation"]
+rkyv-08 = ["dep:rkyv-08", "rkyv-08/pointer_width_32"]
+rkyv-08-16 = ["dep:rkyv-08", "rkyv-08/pointer_width_16"]
+rkyv-08-32 = ["dep:rkyv-08", "rkyv-08/pointer_width_32"]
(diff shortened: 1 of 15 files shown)


djc:
> Not sure what to do about CI as currently the features are incompatible with rkyv 0.7, but I could use fully qualified paths for `rkyv_08::Archive` etc and then it should be ok?

How do you propose to handle additivity in the case there both rkyv 0.7 and 0.8 features are enabled?

Awpteamoose:
> How do you propose to handle additivity in the case there both rkyv 0.7 and 0.8 features are enabled?

I've thought about it a bit and I think it's doable - I don't actually need a separate set of pointer size features, but to be properly additive - the rkyv 0.8 `ArchivedT` types that get generated will have to all be renamed to smth like `Archived08T`, which doesn't really matter since you'd usually access them as `rkyv::Archived<T>` anyway. I will also have to use fully-qualified paths for rkyv everywhere.

vincent-sparks:
The recommended replacement for this crate (jiff) does not support rkyv at all.

PR #1621, opened 2024-10-14
Use Formatter::pad (instead of write_str) for Weekdays

### LINKED ISSUE #1620: `std::fmt::Display` implementation of `Weekday` does not honour width, alignment or fill
## Summary

The `Display` implementation of `Weekday` uses `Formatter::write_str`, which doesn't honour [width](https://doc.rust-lang.org/std/fmt/index.html#width) or [alignment and fill](https://doc.rust-lang.org/std/fmt/index.html#fillalignment).

## Reproducer

```rust
use chrono::Weekday;

fn main() {
    println!("Expected result: <{:X>10}>", format!("{}", Weekday::Mon));
    println!("Actual result: <{:X>10}>", Weekday::Mon);
}
```

([Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e9d6cafc549ad505e6f932697fbda305))

## Expected result

`<XXXXXXXMon>`

## Actual result

`<Mon>`

horazont:
That way, width, alignment and fill from the format string are respected.

Fixes #1620.

> If your feature is semver-compatible, please target the main branch;
> for semver-incompatible changes, please target the `0.5.x` branch.

Excellent question whether this is semver-compatible. I'm not 100% sure: On the one hand, this changes behaviour others may be relying on. On the other hand, it is fixing something which may be perceived as a bug. I'll let the reviewer judge and will retarget accordingly.

### DIFF
--- a/src/weekday.rs
+++ b/src/weekday.rs
@@ -171,7 +171,7 @@ impl Weekday {
 
 impl fmt::Display for Weekday {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        f.write_str(match *self {
+        f.pad(match *self {
             Weekday::Mon => "Mon",
             Weekday::Tue => "Tue",
             Weekday::Wed => "Wed",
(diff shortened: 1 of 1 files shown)


djc (approved):
I'm definitely counting this as semver-compatible. Thanks!

Issue #1625, opened 2024-11-07
Earliest and latest swapped on DST end when using Local timezone.


AVee:
When converting a NaiveDateTime that is on the end of DST into one in the Local timezone we get an Ambiguous result that contains the two possible results in the wrong order. Instead of (earliest, latest) the result contains (latest, earliest) .
See the code below for a reproduction.

```
use std::env;
use chrono::{DateTime, Local, MappedLocalTime, NaiveDate, NaiveDateTime, NaiveTime, TimeZone, Utc};

fn main() {
    // Set Posix Timezone CET/CEST (as found in /usr/share/zoneinfo/CET).
    // DST ends last sunday of october at 03:00 moving back to 02:00 
    // We set the TZ var here, but the same error occurs without any special settings on a linux system in Europe/Amsterdam
    env::set_var("TZ", "CET-1CEST,M3.5.0,M10.5.0/3");
    let date = NaiveDateTime::new(NaiveDate::from_ymd_opt(2024, 10, 27).unwrap(), NaiveTime::from_hms_opt(2, 0, 0).unwrap());
    let mapped_date = Local.from_local_datetime(&date);
    match mapped_date {
        MappedLocalTime::<DateTime<Local>>::Ambiguous(early, late) => {
            println!("Early before late: {}", early < late);
            println!("Early after late: {}", early > late);
            assert_eq!(early, Utc.with_ymd_and_hms(2024, 10, 27, 0, 0, 0).unwrap());
            assert_eq!(late, Utc.with_ymd_and_hms(2024, 10, 27, 1, 0, 0).unwrap());
        },
        MappedLocalTime::<DateTime<Local>>::Single(_) => assert!(false, "Expected ambiguous date"),
        MappedLocalTime::<DateTime<Local>>::None => assert!(false, "Expected ambiguous date"),
    }
}
```
([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=05fe017feab8ed0b384ed456b7a948ff))

So far I've seen this happen Linux when using default zone info from the system (set to Europe/Amsterdam) and on an ESP32 when setting the TZ environment variable. I haven't had the change to check the behavior on Windows yet.

It also doesn't seem to fail when using zones from chrono_tz or [tzfile](https://crates.io/crates/tzfile).

PR #1626, opened 2024-11-08
Correctly order earliest / latest for Ambiguous times in local TimeZone

AVee:
This swaps earliest / latest in Local Timezone handling as it seems to be consistently in the incorrect order, see issue #1625 . I think I changed all places where an instance of MappedLocalTime::Ambiguous is created in the unix part of offset/local. I haven't yet looked at Windows.

I've also added some tests to cover this, which should cover most cases.


src/offset/local/tz_info/timezone.rs
```diff
@@ -933,6 +936,338 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    #[allow(clippy::bool_assert_comparison)]
+    fn test_dst_backward_posix_tz() -> Result<(), Error> {
+        // Northern hemisphere DST (CET/CEST)
+        let tz: TimeZone = TimeZone::from_posix_tz("CET-1CEST,M3.5.0,M10.5.0/3")?;
```
djc, line 943: Nit: please drop all the type annotations for `tz` variables.

src/offset/local/tz_info/timezone.rs
```diff
@@ -933,6 +936,338 @@ mod tests {
...
+        assert_eq!(offsets.latest().unwrap().name.unwrap().as_bytes(), b"GMT");
+
+        Ok(())
+    }
+
+    #[test]
+    #[allow(clippy::bool_assert_comparison)]
+    fn test_dst_backward_tzfile() -> Result<(), Error> {
+        // Northern hemisphere DST (CET/CEST)
+        let data: [u8; 604] = [
```
djc, line 1003: These big binary blobs are a pain to manage. What value does this test have over and above the `test_dst_backward_posix_tz` tests? Do these tests actually gain us any coverage?

djc (commented):
Thanks for working on this, the code changes look good and I appreciate the added test coverage.

src/offset/local/tz_info/timezone.rs
AVee: Fair, I'll fix that.

djc:
> Any news on this?

I asked a question which has not been answered. I'm not making any semver-compatible changes in the near future.

PR #1627, opened 2024-11-09
chore: add `#[inline]` to `num_days`

CommanderStorm:
While looking at the code, I noticed something.
All `num_*` funktions (which do the same thing, but for other divisors) are marked `#[inline]`, except for `num_days`.

=> it looks not marking this method as `#[inline]` is a mistake.


djc (approved):
Thanks.

The wasm test failure is unrelated, so ignoring that.

PR #1632, opened 2024-11-26
Add impls for the Defmt logging crate

BenFordTytherington:
- Add impls for the main structs where it is obvious how to do so
- Gate this behind a new cargo feature called "defmt"


BenFordTytherington:
From Defmts perspective, implementing the trait here is a better solution due to being able to access all the private fields. Additionally, defmt 1.0 is planned to be released early next year, with hopefully no breaking changes compared to the current version.

PR #1634, opened 2024-11-29
Clippy

djc:


pitdicker (approved):
Good cleanups :+1:.
What do we have to do with unicode-ident?

djc:
It's a transitive dependency (via syn I think).

djc:
Not sure what's going on with llvm-cov here...

PR #1637, opened 2024-12-05
tests .to_rfc2822 panics

gabrielfalcao:


djc:
Why? What is the motivation here?

PR #1641, opened 2024-12-11
Fix capitalization

dannbbb1:
Description: Fixed the capitalization of "calendar" to "Calendar" to reflect proper noun formatting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@
 [codecov-link]: https://codecov.io/gh/chronotope/chrono
 
 Chrono aims to provide all functionality needed to do correct operations on dates and times in the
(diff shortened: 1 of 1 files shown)


djc:
I don't see how this is "proper noun formatting". It doesn't seem like an improvement to me.

dannbbb1:
Thank you for your feedback! I changed 'calendar' to 'Calendar' because, in the context of the text, it could be a proper noun (for example, in the name of an app or system), and I wanted to follow the appropriate rule. If that's not the case, please let me know.

djc:
If I search for "gregorian calendar" online, it seems like only pages that use title-casing capitalize Calendar. I don't think that fits here.

PR #1643, opened 2024-12-20
Added length function to Month so that I don't have to use time crate

DanielPoprawski:

### DIFF
--- a/src/month.rs
+++ b/src/month.rs
@@ -163,7 +163,7 @@ impl Month {
     }
 
     ///     Get the length of the month in a given year
-    pub const fn length(&self, year: i32) -> u8 {
+    pub const fn len(&self, year: i32) -> u8 {
         match *self {
             Month::January
             | Month::March
@@ -173,7 +173,7 @@ impl Month {
             | Month::October
             | Month::December => 31,
             Month::April | Month::June | Month::September | Month::November => 30,
-            Month::February if year % 4 == 0 && (year % 25 != 0 || year % 16 == 0) => 29,
+            Month::February if (year % 4 == 0) && (year % 100 != 0 || year % 400 == 0) => 29,
             Month::February => 28,
         }
     }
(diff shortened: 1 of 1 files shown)


src/month.rs
```diff
@@ -161,6 +161,22 @@ impl Month {
...
+        match *self {
+            Month::January
+            | Month::March
+            | Month::May
+            | Month::July
+            | Month::August
+            | Month::October
+            | Month::December => 31,
+            Month::April | Month::June | Month::September | Month::November => 30,
+            Month::February if year % 4 == 0 && (year % 25 != 0 || year % 16 == 0) => 29,
```
djc, line 176: The logic for `February` isn't fully accurate, and I think we have API for this on `Date` that you could repurpose.

src/month.rs
```diff
@@ -161,6 +161,22 @@ impl Month {
             Month::December => "December",
         }
     }
+
+    ///     Get the length of the month in a given year
+    pub const fn length(&self, year: i32) -> u8 {
```
djc, line 166: Not sure about this name. `len()` would be more idiomatic in Rust but maybe it should be a little more specific like `num_days()`?

DanielPoprawski: I think length is a pretty straightforward explanation of what the function does. I didn't like how num_days() sounds

src/month.rs
DanielPoprawski: I found the functions in question under NaiveDate and then under internals.rs but it seems to be meant for testing purposes only and it's unreachable from the month.rs file

djc: Feel free to expose whatever this needs as `pub(crate)`, I don't want to duplicate this logic here.

src/month.rs
djc: I disagree -- `len()`/`length()` is not very specific and not fully intuitive on `Month` IMO -- for example, it could yield a `Duration` instead. We already use the `num_days()` for a number of methods in chrono and it would be a good fit here.

DanielPoprawski (commented):
Fixed the month logic and changed the function name to len()

djc:
I'll do it myself.

Issue #1399, opened 2024-02-01
Parsing numeric literals can cause TooShort/Invalid


t-mart:
For a format/input string item that starts with a numeric literal, the various `parse_from_str` methods can return a `ParseError` error result of `TooShort` or `Invalid` if that literal comes directly after a numeric specifier/value with [variable parsing width (PW=∞)](https://github.com/chronotope/chrono/blob/main/src/format/mod.rs#L101).

In other words, this problem can occur when, for a consecutive pair of format items, `a` and `b`, `a` is a specifier value of variable width and `b` is a literal that starts with a digit:

- `%s1` *timestamp followed by literal digit*
- `%Y1` *year followed by literal digit*

Of particular note is that this same sequence of (specifier, literal) manifests as errors of two different variants:

## `TooShort` Example:

```rust
#[test]
fn test_timestamp() {
    use chrono::{DateTime, format::ParseErrorKind::TooShort};

    let literals = ['1', 'a'];
    for literal in literals {
        let input = format!("1{literal}");
        let fmt = format!("%s{literal}");

        let parsed = DateTime::parse_from_str(&input, &fmt);

        if literal.is_ascii_digit() {
            assert_eq!(parsed.unwrap_err().kind(), TooShort);
        } else {
            assert!(parsed.is_ok());
        }
    }
}
```

This same behavior exists for the other (PW=∞) specifiers, such as Year (e.g., `NaiveDate::parse_from_str("01-01-+100001", "%m-%d-%Y1")`.

## `Invalid` Example

Note that this case contains the same specifier-literal pair as above, but the error is different. Probably because the pair is *not* at the end of the strings.

```rust
#[test]
fn test_year2() {
    use chrono::{NaiveDate, format::ParseErrorKind::Invalid};

    let literals = ['1', 'a'];
    for literal in literals {
        let input = format!("+10000{literal}-01-01");
        let fmt = format!("%Y{literal}-%m-%d");

        let parsed = NaiveDate::parse_from_str(&input, &fmt);

        if literal.is_ascii_digit() {
            assert_eq!(parsed.unwrap_err().kind(), Invalid);
        } else {
            assert!(parsed.is_ok());
        }
    }
}
```

## Cause?

[...]

PR #1648, opened 2025-01-11
Allow nanoseconds unix epoch, and numeric literal after year format

titaneric:
This PR aims to resolve issue #560 and #1399 based on PR #561, and the multiple issues from vector such as https://github.com/vectordotdev/vrl/issues/117, https://github.com/vectordotdev/vrl/issues/790
Since there is no any respones from the author of #561, I would like to continue his/her work.

### DIFF
--- a/src/format/parse.rs
+++ b/src/format/parse.rs
@@ -412,12 +412,6 @@ where
                     );
                     max_width = numeric_bytes_available - next_size;
                 }
-                if substr.len() < min_width {
-                    return Err(TOO_SHORT);
-                }
-                if max_width == 0 {
-                    return Err(INVALID);
-                }
                 // println!("min_width: {:?}, max_width: {:?}", min_width, max_width);
                 let mut v = try_consume!(scan::number(substr, min_width, max_width));
                 if neg {
--- a/src/format/parse.rs
+++ b/src/format/parse.rs
@@ -1596,6 +1596,21 @@ mod tests {
             parsed!(nanosecond: 567_000_000, timestamp: 12_345_678_901_234),
         );
         check(
+            "12345678901234567",
+            &[num(Timestamp), internal_fixed(Nanosecond3NoDot)],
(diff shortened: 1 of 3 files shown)


titaneric:
Oops, I don't consider the no default features that would cause CI to fail. Let me check it.

titaneric:
Find `peakable` method from `Iterator` to implement the same idea in the beautiful way.

titaneric:
@meiamsome , @pitdicker , @djc,  @jszwedko , sorry to ping you here since you have discussed the topic in #561 , but I would like to hear any comments to my PR.

jszwedko:
> @meiamsome , @pitdicker , @djc, @jszwedko , sorry to ping you here since you have discussed the topic in #561 , but I would like to hear any comments to my PR.

I'll defer comments on the implementation since I'm not super familiar with chrono internals, but the tests look good to me. Thanks for attempting to fix this!

djc:
I don't like these changes much. They seem to be very focused on just passing the added tests and don't even try to clean up the surrounding code to understand that the underlying motivation is clearly addressed.

titaneric:
Present implementation heavily depends on previous attempted PR and one of the comment. I substract the next possible numerical literal to count the actual token size for either year or timestamp, that's two phase calculation but would only happen in the special case.

Are you saying that I need to do more work to the `parse_internal` function? Trying to calculate the token size with one-phase only?

djc:
> Are you saying that I need to do more work to the `parse_internal` function? Trying to calculate the token size with one-phase only?

I am saying that I as the maintainer don't find the presented changes an improvement in clarity to already pretty convoluted code. I won't pretend to know exactly how to solve it, which would take me a bunch more time which I'm not really motivated to spend unless some funding is available.

titaneric:
I have refactored the implementation a little bit, hopefully it could make the code more readable and concise.
Please review it whenever you are available.

Issue #1650, opened 2025-01-16
duration_round_up/ceil in DurationRound trait


MagnumTrader:
Hi,

I have a use case where i need to round  a DateTime upwards  towards a TimeDelta, like trunc but in reverse.

Is there any reason Ceil or Round_up is not implemented or should i submit a pull request for this and if so, what do you prefer, that i copy and inverse the  [duration_trunc](https://github.com/chronotope/chrono/blob/82bacb85bd87b31a097e5aa19cd1beac0f2965f8/src/round.rs#L215C1-L216C22) trait method 
or should i clean this module up using some local enum? example:
```rust
RoundingStrategy {
  Ceil,
  Trunc,
  Closest
}
```

best regards, Magnum

djc:
I don't think there is any particular reason these aren't implemented. Probably fine to submit a PR for `duration_round_up()`.

PR #1651, opened 2025-01-16
Feature/round_up method on DurationRound trait

MagnumTrader:
This PR Solves #1650. 

Implementation of duration_round_up method by inverting the duration_trunc method.

### DIFF
--- a/src/round.rs
+++ b/src/round.rs
@@ -183,6 +183,7 @@ pub trait DurationRound: Sized {
 
 impl<Tz: TimeZone> DurationRound for DateTime<Tz> {
     type Err = RoundingError;
+
     fn duration_round(self, duration: TimeDelta) -> Result<Self, Self::Err> {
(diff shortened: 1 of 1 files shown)


src/round.rs
```diff
@@ -177,6 +206,9 @@ impl DurationRound for NaiveDateTime {
     fn duration_trunc(self, duration: TimeDelta) -> Result<Self, Self::Err> {
         duration_trunc(self, self, duration)
     }
+    fn duration_round_up(self, duration: TimeDelta) -> Result<Self, Self::Err> {
```
djc, line 209: Nit: add an empty line before this method.

djc (commented):
Looks good modulo a small nit.

src/round.rs
MagnumTrader: Resolved Nit :)

djc:
Thanks!

PR #1652, opened 2025-01-16
⏰ Add `td_*seconds` serde `Serialize` and `Deserialize` functions for `TimeDelta`

### LINKED ISSUE #117: Add Serialize/Deserialize for Duration

tosti007:
Fixes https://github.com/chronotope/chrono/issues/117

This PR copy-pastes the current implementation for `DateTime` such that we have equal implementations for `DeltaTime`.

I opted to add the logic to the existing `serde.rs` file. However, I think this might be the wrong place and a better solution would be to move the `main.rs`'s `serde` module combined with this `serde.rs` file to a separate `src/serde.rs` file, thoughts?

PS:
I am just doing a drive-by PR on this crate, so I don't really know what the exact requirements are regarding testing ect. Please let me know!


src/datetime/serde.rs
```diff
@@ -23,6 +23,22 @@ pub struct MicroSecondsTimestampVisitor;
 #[derive(Debug)]
 pub struct MilliSecondsTimestampVisitor;
 
+#[doc(hidden)]
+#[derive(Debug)]
+pub struct SecondsTimedeltaVisitor;
```
djc, line 28: Please spell these as `TimeDelta` (with the uppercase `D`).

src/datetime/serde.rs
```diff
@@ -1206,6 +1222,1021 @@ pub mod ts_seconds_option {
...
+    /// let as_string = serde_json::to_string(&my_s)?;
+    /// assert_eq!(as_string, r#"{"delta":2018517}"#);
+    /// # Ok::<(), serde_json::Error>(())
+    /// ```
+    pub fn serialize<S>(td: &TimeDelta, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: ser::Serializer,
+    {
+        serializer.serialize_i64(td.num_nanoseconds().ok_or(ser::Error::custom(
+            "value out of range for a timedelta with nanosecond precision",
```
djc, line 1291: Let's capitalize `TimeDelta` in these error messages, too.

src/lib.rs
```diff
@@ -623,16 +623,29 @@ pub mod serde {
         E::custom(SerdeError::InvalidTimestamp(value))
     }
 
+    /// Create a custom `de::Error` with `SerdeError::InvalidTimestamp`.
+    pub(crate) fn invalid_td<E, T>(value: T) -> E
```
djc, line 627: Would prefer to spell this out as `invalid_time_delta`.

src/lib.rs
```diff
@@ -623,16 +623,29 @@ pub mod serde {
...
+    where
+        E: de::Error,
+        T: fmt::Display,
+    {
+        E::custom(SerdeError::InvalidTimedelta(value))
+    }
+
     enum SerdeError<T: fmt::Display> {
         InvalidTimestamp(T),
+        InvalidTimedelta(T),
```
djc, line 637: `InvalidTimeDelta`.

djc (commented):
I'd like to see at least one basic round-tripping test for each of these.

src/datetime/serde.rs
djc: Timestamp is one word in English, time delta is not, so there's a good reason to treat them differently. We should use `TimeDelta` consistently.

pistomat:
I can't wait for this PR to be merged!

tosti007:
> I can't wait for this PR to be merged!

Thanks for the ping, I totally forgot about this. I'll see if I can get it in a good state somewhere this week.

Cargo.toml
```diff
@@ -18,7 +18,7 @@ name = "chrono"
 
 [features]
 # Don't forget to adjust `ALL_NON_EXCLUSIVE_FEATURES` in CI scripts when adding a feature or an optional dependency.
-default = ["clock", "std", "oldtime", "wasmbind"]
+default = ["clock", "std", "oldtime", "wasmbind", "serde"]
```
djc, line 21: Don't change the default features.

eirnym:
@tosti007 yep, you can use cargo llvm-cov locally

djc:
I won't be taking any more changes to chrono in principle:

tosti007:
> I won't be taking any more changes to chrono in principle:
> 
>     * [Soft-deprecating chrono and chrono-tz #1768](https://github.com/chronotope/chrono/issues/1768)

Makes sense! Thanks for the work you've done on `chrono` and I'll close this PR then (since I am not really running into needing this anymore).

Issue #1628, opened 2024-11-11
Suggestion: Create public API for timezone from Posix strings and TZif files


AVee:
The handling of Local timezones includes code to parse both timezone information files and Posix timezone strings. This code is currently driven from either a TZ environment variable or a /etc/localtime symlink. But it seems there is no public API available to create these timezone manually from either a file or a string.

It would be useful to have a public API where you could do something like `TimeZone::from_file()` and `TimeZone::from_posix_string()`. This is probably mostly useful in (embedded) no-std scenarios where there is no concept of a 'configured' timezone on the OS level and timezone information comes from application specific configuration. There might be other uses too. A minor side benefit would be that it would allow for easy testing against for instance all timezone files as generated from the Iana database.

For handling files [tzfile](https://github.com/kennytm/tzfile) does exist, but it has some limitations because it doesn't handle Posix strings and therefore cannot do far future transitions. But more importantly, the required code is already present in Chrono anyway so it seems somewhat useless to duplicate it.

PR #1658, opened 2025-02-05
Use NaiveDateTime for internal tz_info methods.

AVee:
When looking into doing #1628 I found this comment:
`// should we pass NaiveDateTime all the way through to this fn?`
https://github.com/chronotope/chrono/blob/7cdca4b006fba93278ba3ef532495315cf6f8724/src/offset/local/tz_info/timezone.rs#L136

It seemed a sensible bit of cleanup to do regardless of other changes, so here's a separate pull request for it. 

This replaces passing a timestamp and year to several internal functions with passing a NaiveDateTime, making the function signature slightly clearer. The values originated from in the a NaiveDateTime in first place, so it basically just postpones the conversion.
Prep for #1628

### DIFF
--- a/src/offset/local/tz_info/rule.rs
+++ b/src/offset/local/tz_info/rule.rs
@@ -1,11 +1,11 @@
-use std::cmp::Ordering;
-
 use super::parser::Cursor;
 use super::timezone::{LocalTimeType, SECONDS_PER_WEEK};
 use super::{
     Error, CUMUL_DAY_IN_MONTHS_NORMAL_YEAR, DAYS_PER_WEEK, DAY_IN_MONTHS_NORMAL_YEAR,
     SECONDS_PER_DAY,
 };
+use crate::{Datelike, NaiveDateTime};
+use std::cmp::Ordering;
 
 /// Transition rule
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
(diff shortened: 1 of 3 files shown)


src/offset/local/tz_info/rule.rs
```diff
@@ -229,9 +228,11 @@ impl AlternateTime {
...
     fn find_local_time_type_from_local(
         &self,
-        local_time: i64,
-        current_year: i32,
+        local_time: NaiveDateTime,
     ) -> Result<crate::MappedLocalTime<LocalTimeType>, Error> {
+        let current_year = local_time.year();
+        let local_time = local_time.and_utc().timestamp();
+
         // Check if the current year is valid for the following computations
```
AVee, line 236: It should be impossible to fail the check below since (the current implementation of) NaiveDateTime has a max year of 262142. It could be removed, but perhaps we want to keep it in case NaiveDateTime ever gets changed?

djc (approved):
Thanks!

src/offset/local/tz_info/rule.rs
AVee: Iv'e removed the check. I looks like this code isn't currently covered, but I'm already working on #1628 and test there should cover this.

djc:
Thanks!

Issue #1663, opened 2025-02-23
Add padding size to Pad::Zero or Pad::Space


BigPapa314:
The uu_date command in [coreutils](https://github.com/uutils/coreutils/) uses chrono to parse the format string. In [#7334](https://github.com/uutils/coreutils/issues/7344) there is a report that it is needed to add the amount of padding to the format string.

# current behavior
`let format_items = StrftimeItems::new("+%03d");` will yield Item::Error.

# expected behavior
`let format_items = StrftimeItems::new("+%03d");` should yield Item::Numeric(Numeric::Day, Pad::Zero(3)).
And in `format_numeric()` replace the hard coded 1 char padding to the value in Pad::Space or Pad::Zero.

djc:
`Pad` is public API and I don't want to take a semver-incompatible change for this -- would be open to accomodate this use case within the contraints.

BigPapa314:
Ahh I missed that. 

Can you imagine a way to attach the padding width to the Item::Numeric enum without breaking the public API?

djc:
Well, `Numeric` is declared as `#[non_exhaustive]` at least, so it's probably possible to smuggle something in there?

BigPapa314:
Hmmm. We could add something like 
```rust
enum Numeric {
    //...
    Padded { numeric: Numeric, pad_count: usize }
}
``` 
But that's a hack.

Should I make a proposal or should we wait for 1.0 to add it to the Pad enum.

djc:
I guess we could try whether we can make `Pad` `#[non_exhaustive]` (and add a new variant) in a semver-compatible release. It's technically not compatible, but breakage might be pretty rare in practice.

BigPapa314:
Hmm, so what are the implications for dependent code:

1. The `Numeric` / `Pad` enums are retrieved from _chrono_ and forwarded to _chrono_ again. Like in _uu_date_. 
   * There is no impact at all no mater if we add a new variant or if we modify the existing one.
2. The enums are manually created and forwarded to _chrono_.
   * There would be no impact if we use the `#[non_exhaustive]` option.
   * There would be compile errors if we modify the current `Pad` and developers might need to init the `width` information.
      I guess there is no default initializer that we could use to init the `width` automatically to one.
3. The enums are retrieved from _chrono_ and handled in the dependent code.
   * The match statements will break without default branch when we add the `#[non_exhaustive]`.
      The dependent code has to add the new variants with the `width` field.
[...]

PR #1666, opened 2025-02-24
Add quarter (%q) date string specifier

drinkcat:
GNU date supports %q as a date string specifier. This adds support for that in chrono.

This is needed by uutils/coreutils for compability, and would help fix https://github.com/uutils/coreutils/issues/7333.

### DIFF
--- a/src/format/parse.rs
+++ b/src/format/parse.rs
@@ -343,6 +343,7 @@ where
                     IsoYear => (4, true, Parsed::set_isoyear),
                     IsoYearDiv100 => (2, false, Parsed::set_isoyear_div_100),
                     IsoYearMod100 => (2, false, Parsed::set_isoyear_mod_100),
+                    Quarter => (1, false, Parsed::set_quarter),
                     Month => (2, false, Parsed::set_month),
                     Day => (2, false, Parsed::set_day),
                     WeekFromSun => (2, false, Parsed::set_week_from_sun),
@@ -819,9 +820,16 @@ mod tests {
             parsed!(year_div_100: 12, year_mod_100: 34, isoyear_div_100: 56, isoyear_mod_100: 78),
         );
         check(
-            "1 2 3 4 5",
-            &[num(Month), num(Day), num(WeekFromSun), num(NumDaysFromSun), num(IsoWeek)],
-            parsed!(month: 1, day: 2, week_from_sun: 3, weekday: Weekday::Thu, isoweek: 5),
+            "1 1 2 3 4 5",
+            &[
+                num(Quarter),
+                num(Month),
+                num(Day),
+                num(WeekFromSun),
+                num(NumDaysFromSun),
+                num(IsoWeek),
+            ],
+            parsed!(quarter: 1, month: 1, day: 2, week_from_sun: 3, weekday: Weekday::Thu, isoweek: 5),
         );
         check(
             "6 7 89 01",
--- a/src/format/parsed.rs
+++ b/src/format/parsed.rs
@@ -140,6 +140,8 @@ pub struct Parsed {
     #[doc(hidden)]
     pub isoyear_mod_100: Option<i32>,
     #[doc(hidden)]
+    pub quarter: Option<u32>,
+    #[doc(hidden)]
     pub month: Option<u32>,
     #[doc(hidden)]
     pub week_from_sun: Option<u32>,
@@ -302,6 +304,23 @@ impl Parsed {
             return Err(OUT_OF_RANGE);
         }
         set_if_consistent(&mut self.isoyear_mod_100, value as i32)
+    }
+
+    /// Set the [`quarter`](Parsed::quarter) field to the given value.
+    ///
+    /// Quarter 1 starts in January.
+    ///
+    /// # Errors
+    ///
+    /// Returns `OUT_OF_RANGE` if `value` is not in the range 1-4.
+    ///
+    /// Returns `IMPOSSIBLE` if this field was already set to a different value.
+    #[inline]
+    pub fn set_quarter(&mut self, value: i64) -> ParseResult<()> {
+        if !(1..=4).contains(&value) {
+            return Err(OUT_OF_RANGE);
+        }
+        set_if_consistent(&mut self.quarter, value as u32)
     }
 
     /// Set the [`month`](Parsed::month) field to the given value.
@@ -698,7 +717,15 @@ impl Parsed {
             (_, _, _) => return Err(NOT_ENOUGH),
         };
 
-        if verified { Ok(parsed_date) } else { Err(IMPOSSIBLE) }
+        if !verified {
+            return Err(IMPOSSIBLE);
+        } else if let Some(parsed) = self.quarter {
+            if parsed != parsed_date.quarter() {
+                return Err(IMPOSSIBLE);
+            }
+        }
+
+        Ok(parsed_date)
     }
 
     /// Returns a parsed naive time out of given fields.
@@ -1013,6 +1040,14 @@ impl Parsed {
         self.isoyear_mod_100
     }
 
+    /// Get the `quarter` field if set.
+    ///
+    /// See also [`set_quarter()`](Parsed::set_quarter).
+    #[inline]
+    pub fn quarter(&self) -> Option<u32> {
+        self.quarter
+    }
+
     /// Get the `month` field if set.
     ///
     /// See also [`set_month()`](Parsed::set_month).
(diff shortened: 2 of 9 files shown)


src/format/parse.rs
```diff
@@ -819,9 +820,9 @@ mod tests {
             parsed!(year_div_100: 12, year_mod_100: 34, isoyear_div_100: 56, isoyear_mod_100: 78),
         );
         check(
-            "1 2 3 4 5",
-            &[num(Month), num(Day), num(WeekFromSun), num(NumDaysFromSun), num(IsoWeek)],
-            parsed!(month: 1, day: 2, week_from_sun: 3, weekday: Weekday::Thu, isoweek: 5),
+            "1 1 2 3 4 5",
+            &[num(Quarter), num(Month), num(Day), num(WeekFromSun), num(NumDaysFromSun), num(IsoWeek)],
```
drinkcat, line 824: Linter is not happy about this, but style is consistent with surrounding code.

src/format/parsed.rs
```diff
@@ -698,7 +717,13 @@ impl Parsed {
...
         };
 
-        if verified { Ok(parsed_date) } else { Err(IMPOSSIBLE) }
+        // verify the quarter if we have a good parsed date so far
+        let verify_quarter = |date: NaiveDate| {
+            let quarter = date.quarter();
+            self.quarter.unwrap_or(quarter) == quarter
+        };
+
+        if verified && verify_quarter(parsed_date) { Ok(parsed_date) } else { Err(IMPOSSIBLE) }
```
drinkcat, line 726: I'm very new to Rust, and perhaps not super happy with how I added this check. Not sure if there's some better way to do this.

djc: Don't think the closure is useful here. Suggest writing it like this:

```rust
if !verified {
    return Err(IMPOSSIBLE);
} else if let Some(parsed) = self.quarter {
    if parsed != parsed_date.quarter() {
        return Err(IMPOSSIBLE);
    }
}

Ok(parsed_date)
```

src/format/parse.rs
djc: Except the line length is past the limit. Just run `cargo fmt`.

djc (commented):
(Please squash any further changes into your original commit.)

src/format/parsed.rs
```diff
@@ -304,6 +306,23 @@ impl Parsed {
...
+    ///
+    /// Quarter 1 starts in January.
+    ///
+    /// # Errors
+    ///
+    /// Returns `OUT_OF_RANGE` if `value` is not in the range 1-4.
+    ///
+    /// Returns `IMPOSSIBLE` if this field was already set to a different value.
+    #[inline]
+    pub fn set_quarter(&mut self, value: i64) -> ParseResult<()> {
```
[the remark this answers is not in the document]
drinkcat: Are you talking about `set_quarter` or parsing in general?

If you're talking about parsing in general, for the coreutils case, I believe we only care about formatting (and not parsing), but it does feel nicer for the format to be symmetrical. But, indeed, most of the added code here actually deals with parsing...

src/format/parsed.rs
drinkcat: Yeah, made it too complicated. Will fix.

src/format/parsed.rs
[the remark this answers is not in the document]
djc: I mean `set_quarter()`. It seems like a really weird thing to want to do.

src/traits.rs
```diff
@@ -40,6 +40,11 @@ pub trait Datelike: Sized {
         if year < 1 { (false, (1 - year) as u32) } else { (true, year as u32) }
     }
 
+    /// Returns the quarter number starting from 1.
+    ///
+    /// The return value ranges from 1 to 4.
+    fn quarter(&self) -> u32;
```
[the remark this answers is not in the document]
drinkcat: Done, thanks!

src/format/parsed.rs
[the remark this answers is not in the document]
djc: Ahh, you're right.

djc (approved):
This mostly looks okay -- I think this is missing padding? I wouldn't be surprised if someone wanted to parse or format Q1 as `Q01`.

drinkcat:
Padding isn't missing, since this is a length-1 field.

e.g. `%q`, `%_q`, `%0q` all output `2` with GNU date. `chrono` does the same currently. (`%u` and `%w` -- days of the week also behave the same).

What _is_ missing though is the length specifier. GNU date support fields like `%2q`, `%_2q` (or `%05Y` FWIW), but the length specifier isn't supported at all in `chrono`, AFAICT. I can file another issue for this.

drinkcat:
Oh, and I'm confused by this:
```
Merging is blocked
Merge is not an allowed merge method in this repository.
This branch must not contain merge commits.
```

I did rebase on the lastest `main`, not sure what's wrong here?!

drinkcat:
(The padding issue is tracked in #1663 I think?)

drinkcat:
Thanks for the quick review!

djc:
(Just published 0.4.40 with this change.)

src/traits.rs
```diff
@@ -40,6 +40,14 @@ pub trait Datelike: Sized {
         if year < 1 { (false, (1 - year) as u32) } else { (true, year as u32) }
     }
 
+    /// Returns the quarter number starting from 1.
+    ///
+    /// The return value ranges from 1 to 4.
+    #[inline]
+    fn quarter(&self) -> u32 {
```
tustvold, line 47: Just as an FYI this resulted in our arrow-rs builds failing with the latest chrono patch release.

We have codepaths generic on `ChronoDateExt + Datelike` and this now results in trait ambiguity.

I don't really know if this counts as a breaking change or not... I feel it shouldn't, but it broke our build so maybe it is?? I don't know...

djc: See https://doc.rust-lang.org/cargo/reference/semver.html#trait-new-default-item. In general, with extension traits you're going to be somewhat more likely to be broken, especially if you pick very generic method names.

tustvold: So I am a little unsure how best to proceed here, in particular what changes arrow-rs should be making to ensure this doesn't occur again. Other than pinning chrono to a specific patch version, I can't see a tractable way to avoid a reoccurrence? 

Switching all callsites to use explicit [disambiguation syntax](https://doc.rust-lang.org/reference/expressions/call-expr.html#disambiguating-function-calls) when a chrono trait is in scope is not really a viable option given the extent to which the codebase makes use of traits and generics, and that isn't even considering our downstream consumers...

Maybe it is a case of do nothing now and only do something if it happens again, but I also wonder if chrono might consider not adding new methods to traits in patch releases, especially given how broad its user base is?

djc: I wouldn't worry about it too much, I think this will be relatively rare.

djc: > if chrono might consider not adding new methods to traits in patch releases, especially given how broad its user base is?

I'm not sure this is tenable. Unlike arrow which does regular incompatible releases, chrono is firmly entrenched as low-level library for which semver-incompatible bumps are pretty painful.

tustvold: > I'm not sure this is tenable. Unlike arrow which does regular incompatible releases, chrono is firmly entrenched as low-level library for which semver-incompatible bumps are pretty painful.

Fair we do major releases quarterly and that isn't without pain. I guess my thought was maybe that as a low-level library, chrono would err on the conservative side of making changes that might be breaking.

That being said there are ways to avoid adding methods to traits, e.g. adding new traits... IDK it's a sticky situation, I thought I'd just make our case 😄

djc: Sorry for that, maybe you should figure out some fast-track process for this kind of thing. Honestly a three-day process for little patch releases sounds very last decade to me.

tustvold: Sadly it is for legal reasons...

djc: If you want to take out a contract with me I can probably give you a 3-day heads up on non-urgent releases...

kenaniah: @djc why not just yank the release and push this as 0.5.0 like semver conventions would suggest?

djc: > @djc why not just yank the release and push this as 0.5.0 like semver conventions would suggest? 

Your understanding of semver conventions as practiced in the crates.io ecosystem seems limited at best, and you seemingly have no idea of the impact of semver-incompatible releases for foundational libraries like chrono. Please take your uninformed "why don't you just" elsewhere.

Xuanwo: > Sadly it is for legal reasons...

Hi, it is allowed by the ASF to publish hot fixes without a three-day voting period after a public discussion.

maxburke: @djc If chrono is "firmly entrenched as a low-level library", maybe it's time for it to graduate to version 1.x?

Regardless of whether publishing breaking changes to patch-level increments is considered OK by Cargo's semver specification, it displays a degree of immaturity for something as widely used as chrono to do that.

PR #1668, opened 2025-02-26
Add `subsec_micros` and `subsec_millis` methods to `TimeDelta`

ggoetz:
This PR revives @botahamec 's PR #1351, which has been abandoned, and adds `subsec_micros` and `subsec_millis` to the public API for `chrono::TimeDelta`, for consistency with the API of `std::time::Duration`.

Summary of changes:
   - update the docstring for `TimeDelta::subsec_nanos()` and add a test for the method.
   - add `subsec_millis` and `subsec_micros`, along with unit tests.

Fixes: #1348

### DIFF
--- a/src/time_delta.rs
+++ b/src/time_delta.rs
@@ -313,6 +313,24 @@ impl TimeDelta {
         if self.secs < 0 && self.nanos > 0 { self.nanos - NANOS_PER_SEC } else { self.nanos }
     }
 
+    /// Returns the number of microseconds in the fractional part of the duration.
+    ///
+    /// This is the number of microseconds such that
+    /// `subsec_micros() + num_seconds() * 1_000_000` is the truncated number of
+    /// microseconds in the duration.
+    pub const fn subsec_micros(&self) -> i32 {
+        self.subsec_nanos() / NANOS_PER_MICRO
+    }
+
+    /// Returns the number of milliseconds in the fractional part of the duration.
+    ///
+    /// This is the number of milliseconds such that
+    /// `subsec_millis() + num_seconds() * 1_000` is the truncated number of
+    /// milliseconds in the duration.
+    pub const fn subsec_millis(&self) -> i32 {
+        self.subsec_nanos() / NANOS_PER_MILLI
+    }
+
     /// Returns the total number of whole milliseconds in the `TimeDelta`.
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -660,6 +660,25 @@ pub mod rkyv {
     pub type ArchivedDuration = ArchivedTimeDelta;
 }
 
+/// The number of nanoseconds in a microsecond.
+pub const NANOS_PER_MICRO: i32 = 1000;
```
ggoetz, line 664: One question on the naming convention adopted here. This const is spelled "MICRO", whereas the rest of the codebase adopts "micros". Likewise, the const below uses "MILLI" vs "MILLIS" elsewhere, including on the other consts.

Should they be renamed to `NANOS_PER_MICROS` and `NANOS_PER_MILLIS`?

djc:
Honestly I'm not convinced by the argument that these consts should be made public just so two of them can make a single appearance in a docstring. IMO it would be fine to leave the constants alone and use literals to explain the methods.

ggoetz:
> Honestly I'm not convinced by the argument that these consts should be made public just so two of them can make a single appearance in a docstring. IMO it would be fine to leave the constants alone and use literals to explain the methods.

Makes sense to me -- and there was precedent with the previous docstring for `subsec_nanos` to only use a literal without exposing the corresponding const. I've reverted the commit that exposed the constants.

src/lib.rs
ggoetz: Not applicable anymore now that the corresponding commit has been reverted.

src/time_delta.rs
```diff
@@ -304,7 +304,9 @@ impl TimeDelta {
         if self.secs < 0 && self.nanos > 0 { self.secs + 1 } else { self.secs }
     }
 
-    /// Returns the number of nanoseconds such that
+    /// Returns the number of nanoseconds in the fractional part of the duration.
+    ///
+    /// This is the number of nanoseconds such that
```
djc, line 309: Sorry, by literal I meant to replace `NANOS_PER_SEC` with `1_000_000_000` here (and same for the new methods). Other than that this seems fine!

ggoetz: Updated the 3 comments and the PR description.

djc (approved):

Issue #1601, opened 2024-07-22
Add (try_)from_secs_f64, (try_)from_millis_f64, as_millis_f64 and as_secs_f64 to TimeDelta


bouk:
This allows converting a floating point representation of millis/secs into durations, just like std::time::Duration has.

djc:
I think this could make sense -- want to submit a PR?

PR #1671, opened 2025-02-28
Implement `as_seconds_f32` and `as_seconds_f64` for `TimeDelta`

ggoetz:
Implement methods for converting a  `TimeDelta` object into a fractional number of seconds, with single or double precision.

There are two possible choices for naming these methods: `as_secs_f**` or `as_seconds_f**`. The former is consistent with `core::time::Duration`, whereas the latter is consistent with existing `TimeDelta` methods (`num_seconds`, which is called `as_secs` in `core::time::Duration`). Consistency with the existing interface is chosen in this PR.

The reverse conversion (from fp seconds to a `TimeDelta`) is [tricky to get right](https://github.com/rust-lang/rust/blob/master/library/core/src/time.rs#L1466-L1544) without unexpected loss of precision, or corner cases causing incorrect rounding or overflow of the number of nanoseconds, so it's not included in this PR.

Summary of changes:
   - move `subsec_millis`, `subsec_micros`, `subsec_nanos` next to their corresponding `num_milliseconds`, `num_microseconds`, `num_nanoseconds` counterparts, which frees up a good location to add `as_secs_f32` and `add_secs_f64`;
   - implement `as_seconds_f32` and `as_seconds_f64` and associated unit tests.

Issue: #1601

### DIFF
--- a/src/time_delta.rs
+++ b/src/time_delta.rs
@@ -304,22 +304,14 @@ impl TimeDelta {
         if self.secs < 0 && self.nanos > 0 { self.secs + 1 } else { self.secs }
     }
 
-    /// Returns the number of nanoseconds in the fractional part of the duration.
-    ///
-    /// This is the number of nanoseconds such that
-    /// `subsec_nanos() + num_seconds() * 1_000_000_000` is the total number of
-    /// nanoseconds in the `TimeDelta`.
-    pub const fn subsec_nanos(&self) -> i32 {
-        if self.secs < 0 && self.nanos > 0 { self.nanos - NANOS_PER_SEC } else { self.nanos }
-    }
-
-    /// Returns the number of microseconds in the fractional part of the duration.
-    ///
-    /// This is the number of microseconds such that
-    /// `subsec_micros() + num_seconds() * 1_000_000` is the truncated number of
-    /// microseconds in the duration.
-    pub const fn subsec_micros(&self) -> i32 {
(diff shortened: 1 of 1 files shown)


ggoetz:
> Who's we/what's your motivation/use case?

Oh, I just wrote "we" in the PR description out of habit. The motivation is better parity with what `core::time::Duration` provides -- except with `chrono`'s support for negative duration. I'm working on a robotics codebase for which it's handy to have a conversion function from duration to fp numbers to calculate velocities, accelerations, etc, it'd be nice if `core::time::Duration` provided convenience methods for this.

> I think I'd prefer to move towards the core naming, want to rename the other method (retaining a deprecated alias for compatibility)?

Do you mean rename "num_seconds" to "as_secs"? This seems like a big departure from the current interface, and would make it inconsistent with "num_milliseconds", "num_microseconds", etc. Probably better kept as is for this PR? If you're interested in aligning with `core::time::Duration` better I'm happy to send a separate rename + deprecation PR.

In the meantime I'll rename the methods to `as_secs_f32`, `as_secs_f64` to better align with the standard library.

djc:
Okay, sorry, you've convinced me I was wrong -- let's stick with `as_seconds_f*` for this PR at least.

djc (approved):
Thanks!

(In the future, keeping the first line of your commit messages a little shorter would be nice -- typically below 72 chars is best.)

PR #1672, opened 2025-03-03
Add padding width

### LINKED ISSUE #1663: Add padding size to Pad::Zero or Pad::Space

BigPapa314:
Fixes #1663


BigPapa314:
Hmm the [codecov/project](https://github.com/chronotope/chrono/pull/1672/checks?check_run_id=38448497218) task seams to be stuck. Can someone have a look?

src/format/mod.rs
```diff
@@ -149,13 +151,71 @@ pub enum Numeric {
...
+        self
+    }
+
+    /// Gets the numeric and padding width from the numeric
+    ///
+    /// Should be removed if the padding width is added to the `Pad` enum.
+    pub fn unwrap_padding(&self) -> (&Self, usize) {
+        (self, 0)
+    }
+}
```
djc, line 217: Why is this useful? I don't think these should be part of the public API.

src/format/formatting.rs
```diff
@@ -127,89 +127,58 @@ impl<'a, I: Iterator<Item = B> + Clone, B: Borrow<Item<'a>>> DelayedFormat<I> {
...
-            w.write_char((b'0' + v) as char)
-        }
-
-        fn write_two(w: &mut impl Write, v: u8, pad: Pad) -> fmt::Result {
-            let ones = b'0' + v % 10;
-            match (v / 10, pad) {
-                (0, Pad::None) => {}
-                (0, Pad::Space) => w.write_char(' ')?,
-                (tens, _) => w.write_char((b'0' + tens) as char)?,
+        // unpack padding width if provided
```
djc, line 130: I want your diff to have minimal changes to the structure of the code, or to isolate structural (but non-functional) changes in a separate commit. Is that feasible? The current iteration seems pretty hard to review.

djc (commented):
Sorry for the very long delay...

(Also needs a rebase.)

src/format/formatting.rs
BigPapa314: The original code only had the oportunity to have exactly one or two padding signs. So there where special functions "write_one" and "write_two" to exactly do that. Because we now have a dynamic amount of padding signs it whould make the code more complex to have some kind of selection logic for "write_xxx" methods when we already had a method to write any amount of padding signs.

BigPapa314:
In general this PR should be a proposal how it whould look like if we add the padding width to the Numeric enum and not to the Pad enum where it whould belong to. My conclusion is:

Advantages:
- The API might be stable as the Numeric enum is #[non_exhaustive]

Drawbacks:
- The API only works with dynamic memory allocation
- The work with the padding width is cumbersome and needs more API changes that have to be removed when the padding width is moved to the Pad enum in the future

I think we should postpone the changes untill we can add the padding with to where it should be so that we can implement it in a minimal invasive and maximal compatible way.

djc:
> I think we should postpone the changes untill we can add the padding with to where it should be so that we can implement it in a minimal invasive and maximal compatible way.

Okay, but I have no concrete plans nor much motivation to work on chrono 0.5, so that might mean postponing more or less indefinitely.

PR #1673, opened 2025-03-09
Add `num_days_in_month` method to `Datelike` trait

aslilac:
The recent `num_days` addition to the `Month` type is great, but it could be even better if we took it just a little bit further.

This change adds a `num_days_in_month` function to the `Datelike` trait.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -1,4 +1,4 @@
-use crate::{IsoWeek, Weekday};
+use crate::{IsoWeek, Month, Weekday};
 
 /// The common set of methods for date component.
 ///
@@ -265,6 +265,18 @@ pub trait Datelike: Sized {
         ndays += ((year * 1461) >> 2) - div_100 + (div_100 >> 2);
         ndays + self.ordinal() as i32
     }
+
+    /// Get the length in days of the month
+    fn num_days_in_month(&self) -> u8 {
+        use num_traits::FromPrimitive;
+        // The value returned from `self.month()` is guaranteed to be in the
+        // range [1,12], which will never result in a `None` value here.
+        let month = Month::from_u32(self.month()).unwrap();
+        // `Month::num_days` will only return `None` if the provided year is out
+        // of range. Since we are passing it directly from a verified date, we
+        // know it is in range, and the result will never be `None`.
+        month.num_days(self.year()).unwrap()
+    }
 }
 
 /// The common set of methods for time component.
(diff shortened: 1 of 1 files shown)


src/date.rs
```diff
@@ -390,6 +390,10 @@ where
 }
 
 impl<Tz: TimeZone> Datelike for Date<Tz> {
+    #[inline]
+    fn leap_year(&self) -> bool {
```
djc, line 394: Let's call this `in_leap_year()`, and keep this addition in a separate commit.

src/month.rs
```diff
@@ -167,23 +167,7 @@ impl Month {
     ///
     /// Yields `None` if `year` is out of range for `NaiveDate`.
     pub fn num_days(&self, year: i32) -> Option<u8> {
-        Some(match *self {
```
djc, line 170: I definitely want to keep this logic in `Month`.

src/naive/date/mod.rs
```diff
@@ -1335,7 +1335,7 @@ impl NaiveDate {
     ///
     /// ```
     /// # use chrono::NaiveDate;
-    /// assert_eq!(NaiveDate::from_ymd_opt(2000, 1, 1).unwrap().leap_year(), true);
+    /// assert_eq!(NaiveDate::from_ymd_opt(2000, 1, 1).unwrap().mhm(), true);
```
djc: Not sure what this is about?

src/month.rs
```diff
@@ -245,6 +229,15 @@ impl num_traits::FromPrimitive for Month {
     }
 }
 
+impl<T: Datelike> From<&T> for Month {
+    fn from(value: &T) -> Self {
+        use num_traits::FromPrimitive;
+        // The month returned from the `Datelike` value is guaranteed to be in
+        // the range [1,12], which will never result in a `None` value here.
+        Month::from_u32(value.month()).unwrap()
```
djc, line 238: Move this logic into the `Datelike::num_days_in_month()` default impl.

djc (commented):
Please squash your commits into clean history:

- One for the `in_leap_year()` addition
- One for adding `num_days_in_month()`

src/date.rs
aslilac: it's already called `leap_year` on the `NaiveDate` type, and there are a bunch of other functions like this in the trait and on `NaiveDate` where the names match. wouldn't it feel a bit weird to break that convention here?

src/month.rs
aslilac: moving it here just avoids an `unwrap` call but I can move it back if you'd really prefer

src/date.rs
aslilac: I guess it doesn't really matter tho, because if the `num_days` logic is moved back to the `Month` type then this function isn't even needed for this PR anymore.

djc:
Thanks!

PR #1676, opened 2025-03-15
add `WeekdaySet`, a collection of `Weekday` that is `Copy`

Kinrany:
## Overview

Adds `WeekdaySet`: a type that represents a set of `Weekday`s, stored as a bitmask in one byte.

All functionality is in module `weekdays` behind a new feature `experimental_weekdays`.

The API aims to mirror that of standard collections and the underlying `BitFlags` where appropriate.

This PR is currently a draft that aims to demonstrate a basic implementation.

## Motivation

It is somewhat common to schedule events to happen every week, on certain days of the week.

Implementing a type like this on top of `chrono` plus a library like `bitflags` or `enumflags2` (~used in this PR as an implementation detail~) is easy, I've personally done this from scratch at least three times by now.

However there are a lot of minor API design decisions to make, making it unlikely that two hand-rolled implementations will be compatible with each other and easy to unify. Similarly, an external type is always going to be less ergonomic than a native `chrono` implementation.

One specific area where standardizing on a single type in `chrono` will help is serialization: `serde` (implemented here in `chrono`) and database types (implemented in database drivers).

On the other hand, I believe that the desired overall shape of the feature is very clear and not controversial. To put it in one sentence: we'd like to have a collection of `Weekday`s that is `Copy` and has the size of one byte while providing all the normal features of collections.

## Prior art

The most famous example is probably cron's fifth argument. For example, `0 0 * * 1,3,5` means "at 00:00:00 on Monday, Wednesday or Friday".

[iCalendar](https://en.wikipedia.org/wiki/ICalendar) ([RFC 5545](https://www.rfc-editor.org/rfc/rfc5545)) specifies field BYDAY that allows values like `MO,TU,SA,SU`.

`icu_calendar`, part of https://github.com/unicode-org/icu4x, uses `chrono` but implements its own type [WeekdaySet](https://docs.rs/icu_calendar/latest/icu_calendar/provider/struct.WeekdaySet.html).

[`weekdays`](https://crates.io/crates/weekdays) is a new package that focuses on this very issue.

## Future changes

The implementation could become slightly simpler in the future if `Weekday` was changed to use one bit for each weekday instead of the numbers from 0 to 6. This should make the conversions between `Weekday` and `WeekdaySet` slightly simpler and more efficient. This would be a breaking change of course.

### DIFF
--- a/src/weekdays.rs
+++ b/src/weekdays.rs
@@ -1,0 +1,515 @@
+use core::{
+    fmt::{self, Debug},
+    iter::FusedIterator,
+    ops::Not,
+};
+
+use crate::Weekday;
+
+/// A collection of `Weekday`s stored as a single byte.
+///
+/// This type is `Copy` and provides efficient set-like and slice-like operations.
+/// Many operations are `const` as well.
+///
+/// Implemented as a bitmask where bits 1-7 correspond to Monday-Sunday.
+#[derive(Clone, Copy, Default, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub struct WeekdaySet(u8); // Invariant: the 8-th bit is always 0.
+
+impl WeekdaySet {
+    /// Returns `Some(day)` if this collection contains exactly one day.
+    ///
+    /// Returns `None` otherwise.
+    ///
+    /// # Example
+    /// ```
+    /// # use chrono::WeekdaySet;
+    /// use chrono::Weekday::*;
+    /// assert_eq!(WeekdaySet::single(Mon).single_day(), Some(Mon));
+    /// assert_eq!(WeekdaySet::from_array([Mon, Tue]).single_day(), None);
+    /// assert_eq!(WeekdaySet::EMPTY.single_day(), None);
+    /// assert_eq!(WeekdaySet::ALL.single_day(), None);
+    /// ```
+    pub const fn single_day(self) -> Option<Weekday> {
+        match self {
+            Self(0b000_0001) => Some(Weekday::Mon),
+            Self(0b000_0010) => Some(Weekday::Tue),
+            Self(0b000_0100) => Some(Weekday::Wed),
+            Self(0b000_1000) => Some(Weekday::Thu),
+            Self(0b001_0000) => Some(Weekday::Fri),
+            Self(0b010_0000) => Some(Weekday::Sat),
+            Self(0b100_0000) => Some(Weekday::Sun),
+            _ => None,
+        }
+    }
+
+    /// Returns `true` if `other` contains all days in `self`.
+    ///
+    /// # Example
+    /// ```
+    /// # use chrono::WeekdaySet;
+    /// use chrono::Weekday::*;
+    /// assert!(WeekdaySet::single(Mon).is_subset(WeekdaySet::ALL));
+    /// assert!(!WeekdaySet::single(Mon).is_subset(WeekdaySet::EMPTY));
+    /// assert!(WeekdaySet::EMPTY.is_subset(WeekdaySet::single(Mon)));
+    /// ```
+    pub const fn is_subset(self, other: Self) -> bool {
+        self.intersection(other).0 == self.0
+    }
+
+    /// Adds a day to the collection.
+    ///
+    /// Returns `true` if the day was new to the collection.
+    ///
+    /// # Example
+    /// ```
+    /// # use chrono::WeekdaySet;
+    /// use chrono::Weekday::*;
+    /// let mut weekdays = WeekdaySet::single(Mon);
+    /// assert!(weekdays.insert(Tue));
+    /// assert!(!weekdays.insert(Tue));
+    /// ```
+    pub fn insert(&mut self, day: Weekday) -> bool {
+        if self.contains(day) {
+            return false;
+        }
+
+        self.0 |= Self::single(day).0;
+        true
+    }
+
+    /// Removes a day from the collection.
+    ///
+    /// Returns `true` if the collection did contain the day.
+    ///
+    /// # Example
+    /// ```
+    /// # use chrono::WeekdaySet;
+    /// use chrono::Weekday::*;
+    /// let mut weekdays = WeekdaySet::single(Mon);
+    /// assert!(weekdays.remove(Mon));
+    /// assert!(!weekdays.remove(Mon));
+    /// ```
+    pub fn remove(&mut self, day: Weekday) -> bool {
+        if self.contains(day) {
+            self.0 &= !Self::single(day).0;
+            return true;
+        }
+
+        false
+    }
+
+    /// Get the first day in the collection, starting from Monday.
+    ///
+    /// Returns `None` if the collection is empty.
+    ///
+    /// # Example
+    /// ```
+    /// # use chrono::WeekdaySet;
+    /// use chrono::Weekday::*;
+    /// assert_eq!(WeekdaySet::single(Mon).first(), Some(Mon));
+    /// assert_eq!(WeekdaySet::single(Tue).first(), Some(Tue));
+    /// assert_eq!(WeekdaySet::ALL.first(), Some(Mon));
+    /// assert_eq!(WeekdaySet::EMPTY.first(), None);
+    /// ```
+    pub const fn first(self) -> Option<Weekday> {
+        if self.is_empty() {
+            return None;
+        }
+
+        // Find the first non-zero bit.
+        let bit = 1 << self.0.trailing_zeros();
+
+        Self(bit).single_day()
+    }
+
+    /// Get the last day in the collection, starting from Sunday.
+    ///
+    /// Returns `None` if the collection is empty.
+    ///
+    /// # Example
+    /// ```
+    /// # use chrono::WeekdaySet;
+    /// use chrono::Weekday::*;
+    /// assert_eq!(WeekdaySet::single(Mon).last(), Some(Mon));
+    /// assert_eq!(WeekdaySet::single(Sun).last(), Some(Sun));
+    /// assert_eq!(WeekdaySet::from_array([Mon, Tue]).last(), Some(Tue));
+    /// assert_eq!(WeekdaySet::EMPTY.last(), None);
+    /// ```
+    pub fn last(self) -> Option<Weekday> {
+        if self.is_empty() {
+            return None;
+        }
+
+        // Find the last non-zero bit.
+        let bit = 1 << (7 - self.0.leading_zeros());
+
+        Self(bit).single_day()
+    }
+
+    /// Split the collection in two at the given day.
+    ///
+    /// Returns a tuple `(before, after)`. `before` contains all days starting from Monday
+    /// up to but __not__ including `weekday`. `after` contains all days starting from `weekday`
+    /// up to and including Sunday.
+    ///
+    /// # Example
+    /// ```ignore
+    /// # use chrono::WeekdaySet;
+    /// use chrono::Weekday::*;
+    /// let (before, after) = WeekdaySet::ALL.split_at(Fri);
+    /// assert_eq!(before, WeekdaySet::from_array([Mon, Tue, Wed, Thu]));
+    /// assert_eq!(after, WeekdaySet::from_array([Fri, Sat, Sun]));
+    /// ```
+    const fn split_at(self, weekday: Weekday) -> (Self, Self) {
+        let days_after = 0b1000_0000 - Self::single(weekday).0;
+        let days_before = days_after ^ 0b0111_1111;
+        (Self(self.0 & days_before), Self(self.0 & days_after))
+    }
+
+    /// Iterate over the `Weekday`s in the collection, starting from a given day
+    /// and wrapping around from Sunday to Monday.
+    ///
+    /// # Example
+    /// ```
+    /// # use chrono::WeekdaySet;
+    /// use chrono::Weekday::*;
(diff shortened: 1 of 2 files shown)


djc:
I'm open to merging something like this, but I don't want to take on a fat extra dependency for it. For chrono, I think hand-coding something like this (`struct WeekdaySet(u8)`) should be okay.

Kinrany:
I'll check back later in case there are any more MSRV-related failures.

I don't expect this to be the final version, but it should be close enough now. I removed `enumflags2`, and as a nice side effect there is now a number of methods that can all be `const`.

Kinrany:
I'm looking at examples of code I've written before and adding everything that I think would be universally useful and improve ergonomics of this type. But I would be perfectly happy to limit the first PR to the parts that are the least controversial and don't need more design/bikeshedding.

Kinrany:
I'm also force-pushing to keep the PR history somewhat clean, but only when I break checks. Let me know if this becomes an inconvenience or is a bad idea in general 😄

djc:
This should be a single commit, so force pushing is great. Keeping this somewhat small by limiting the number of traits implemented for this PR would be nice.

Kinrany:
Cool cool. I'll keep the local PR history for now to keep track of the additions and then squash once we're close to landing.

Though I've already added everything I could think of for now.

Improved ergonomics are a big reason people keep writing these types over just using `Vec<Weekday>` or `HashSet<Weekday>`, so _long term_ we'll probably still want a somewhat wide API, everything that has a single obviously correct and convenient implementation.

If compile times are an issue, hiding the type behind a feature would work great I think. I expect that there are many use cases that don't care about weekdays at all, and then there are use cases that need an arbitrary subset of these methods and impls.

I'll write up a list of things that I consider more questionable.

src/weekdays.rs
```diff
@@ -0,0 +1,723 @@
...
+    type Item = Weekday;
+    type IntoIter = WeekdaysIter;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.iter()
+    }
+}
+
+// impl Bit*<Weekday> for Weekdays
+impl BitOr<Weekday> for Weekdays {
```
Kinrany, line 614: This set of operations might be unnecessary. `days | day` looks better than `days | Weekdays::from(day)` and the meaning seems clear to me, but I may be biased.

src/weekdays.rs
```diff
@@ -0,0 +1,723 @@
...
+    }
+}
+
+// impl Bit*Assign<Weekday> for Weekdays
+impl BitOrAssign<Weekday> for Weekdays {
+    fn bitor_assign(&mut self, rhs: Weekday) {
+        *self |= Self::from(rhs);
+    }
+}
+impl BitAndAssign<Weekday> for Weekdays {
```
Kinrany, line 683: Some of these look a little weird in isolation.

`days &= day` means "remove all weekdays other than `day` from `days`". I would be hard pressed to come up with a use case for this one.

On the other hand, the meaning is still clear. And one would probably be surprised to _not_ be able to do this when other similar operations are available.

src/weekdays.rs
```diff
@@ -0,0 +1,723 @@
...
+    /// # Example
+    /// ```
+    /// # use chrono::Weekdays;
+    /// assert!(Weekdays::MON.is_subset(Weekdays::ALL));
+    /// assert!(!Weekdays::MON.is_subset(Weekdays::EMPTY));
+    /// assert!(Weekdays::EMPTY.is_subset(Weekdays::MON));
+    /// ```
+    pub const fn is_subset(self, other: Self) -> bool {
+        self.intersection(other).0 == self.0
+    }
```
djc, line 242: All methods that depend on other methods should be defined before them.

src/weekdays.rs
```diff
@@ -0,0 +1,723 @@
...
+    /// Returns `true` if `self` contains all days in `other`.
+    ///
+    /// # Example
+    /// ```
+    /// # use chrono::Weekdays;
+    /// assert!(Weekdays::ALL.is_superset(Weekdays::MON));
+    /// assert!(Weekdays::MON.is_superset(Weekdays::EMPTY));
+    /// assert!(!Weekdays::MON.is_superset(Weekdays::TUE));
+    /// ```
+    pub const fn is_superset(self, other: Self) -> bool {
```
djc, line 253: Let's drop this for now.

src/weekdays.rs
```diff
@@ -0,0 +1,723 @@
...
+    ///
+    /// # Example
+    /// ```
+    /// # use chrono::{Weekday, Weekdays};
+    /// let mut weekdays = Weekdays::MON;
+    /// assert!(weekdays.insert(Weekday::Tue));
+    /// assert!(!weekdays.insert(Weekday::Tue));
+    /// ```
+    pub fn insert(&mut self, day: Weekday) -> bool {
+        if self.contains(day) {
```
djc, line 269: Rewrite `insert()` and `remove()` to do an early return, dropping the `else` clause.

src/weekdays.rs
```diff
@@ -0,0 +1,723 @@
...
+    /// ```
+    /// # use chrono::{Weekday, Weekdays};
+    /// let weekdays = Weekdays::MON | Weekdays::WED | Weekdays::FRI;
+    /// let mut iter = weekdays.iter();
+    /// assert_eq!(iter.next(), Some(Weekday::Mon));
+    /// assert_eq!(iter.next(), Some(Weekday::Wed));
+    /// assert_eq!(iter.next(), Some(Weekday::Fri));
+    /// assert_eq!(iter.next(), None);
+    /// ```
+    pub const fn iter(self) -> WeekdaysIter {
```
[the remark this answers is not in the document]
djc: Let's do `iter_from()` instead.

src/weekdays.rs
```diff
@@ -0,0 +1,723 @@
...
+    ///
+    /// # Example
+    /// ```
+    /// # use chrono::{Weekday, Weekdays};
+    /// let weekdays = Weekdays::MON | Weekdays::WED | Weekdays::FRI;
+    /// let (before, after) = weekdays.split_at(Weekday::Wed);
+    /// assert_eq!(before, Weekdays::MON);
+    /// assert_eq!(after, Weekdays::WED | Weekdays::FRI);
+    /// ```
+    pub const fn split_at(self, weekday: Weekday) -> (Self, Self) {
```
djc, line 375: Let's drop this.

src/weekdays.rs
```diff
@@ -0,0 +1,723 @@
...
+    /// assert_eq!(iter.next(), Some(Weekday::Mon));
+    /// assert_eq!(iter.next(), None);
+    /// ```
+    pub const fn iter_from(self, start: Weekday) -> WeekdaysIterFrom {
+        WeekdaysIterFrom { days: self, start }
+    }
+
+    /// Convert the collection into a `Vec<Weekday>`.
+    #[cfg(feature = "std")]
+    pub fn to_vec(self) -> Vec<Weekday> {
```
djc, line 399: Drop this.

src/weekdays.rs
```diff
@@ -0,0 +1,723 @@
...
+        self.iter().collect()
+    }
+
+    /// Iterate over all 128 possible sets, from `EMPTY` to `ALL`.
+    pub fn iter_all() -> impl Iterator<Item = Self> {
```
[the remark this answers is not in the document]
djc: Make this private and `#[cfg(test)]` if we have tests today that need it.

src/weekdays.rs
```diff
@@ -0,0 +1,723 @@
...
+        (0b0000_0000..0b1000_0000).map(Self)
+    }
+
+    /// Panics if the 8-th bit of `self` is not 0.
+    #[cfg(test)]
+    fn assert_8th_bit_invariant(self) {
+        assert!(self.0 & 0b1000_0000 == 0, "the 8-th bit of {self:?} is not 0");
+    }
```
djc, line 412: This doesn't need to be a method.

src/weekdays.rs
```diff
@@ -0,0 +1,723 @@
...
+}
+
+/// An iterator over a collection of weekdays.
+///
+/// See `Weekdays::iter`.
+#[derive(Debug, Clone)]
+pub struct WeekdaysIter(pub Weekdays);
```
djc, line 419: Empty line between the type declaration and the `impl` please.

src/weekdays.rs
```diff
@@ -0,0 +1,723 @@
...
+        if self.0.is_empty() {
+            return None;
+        }
+
+        let next = self.0.first().expect("the collection is not empty");
+        self.0.remove(next);
+        Some(next)
+    }
+}
+impl DoubleEndedIterator for WeekdaysIter {
```
djc, line 433: Empty line between impl blocks everywhere.

src/weekdays.rs
```diff
@@ -0,0 +1,723 @@
...
+    }
+}
+impl ExactSizeIterator for WeekdaysIterFrom {
+    fn len(&self) -> usize {
+        self.days.len().into()
+    }
+}
+impl FusedIterator for WeekdaysIterFrom {}
+
+impl fmt::Display for Weekdays {
```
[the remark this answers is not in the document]
djc: I think the `Display` should be `[Mon, Tue, Wed]`-like.

src/weekdays.rs
```diff
@@ -0,0 +1,723 @@
...
+#[test]
+fn not_operation_preserves_8th_bit_invariant() {
+    for weekdays in Weekdays::iter_all() {
+        (!weekdays).assert_8th_bit_invariant();
+    }
+}
+
+impl From<Weekday> for Weekdays {
+    fn from(weekday: Weekday) -> Self {
+        Self::from_one(weekday)
```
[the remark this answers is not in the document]
djc: Let's drop it for now.

src/weekdays.rs
```diff
@@ -0,0 +1,723 @@
...
+    }
+}
+
+impl Extend<Weekday> for Weekdays {
+    fn extend<T: IntoIterator<Item = Weekday>>(&mut self, iter: T) {
+        for weekday in iter {
+            self.insert(weekday);
+        }
+    }
+}
```
djc, line 595: Drop this for now.

src/weekdays.rs
djc: Let's drop them for now.

src/weekdays.rs
```diff
@@ -0,0 +1,723 @@
...
+///
+/// # Example
+/// ```
+/// # use chrono::{Weekday, Weekdays};
+/// assert!(Weekdays::MON[Weekday::Mon]);
+/// assert!(Weekdays::ALL[Weekday::Mon]);
+/// assert!(!Weekdays::EMPTY[Weekday::Mon]);
+/// assert!(!Weekdays::TUE[Weekday::Mon]);
+/// ```
+impl Index<Weekday> for Weekdays {
```
djc, line 646: Drop this for now.

src/weekdays.rs
djc: Let's drop all the `BitX` impls for now and come back to them later.

src/weekdays.rs
```diff
@@ -0,0 +1,723 @@
...
+use crate::Weekday;
+
+/// A collection of `Weekday`s stored as a single byte.
+///
+/// This type is `Copy` and provides efficient set-like and slice-like operations.
+/// Many operations are `const` as well.
+///
+/// Implemented as a bitmask where bits 1-7 correspond to Monday-Sunday.
+#[derive(Clone, Copy, Default, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Weekdays(
```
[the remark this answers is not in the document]
Kinrany: I used to call these `WeekdaySet` at first too, but I'm now thinking that that would be confusing: it's natural for this type to include more than just set-like operations.

Using this type for counting weekdays is perhaps not as misleading as it looks: it makes no sense to count weekdays without uniqueness (that would be just `Days`), and thanks to being `Copy` and having `len()` this type does serve the closest possible purpose of counting unique weekdays.

src/weekdays.rs
Kinrany: Private method? It is used in `iter_some`.

src/weekdays.rs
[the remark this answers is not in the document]
djc: You haven't convinced me. 🙂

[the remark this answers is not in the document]
Kinrany: Haha, all right. To be honest I haven't even considered the similarity with the unit-of types, `Days` is the only one I ever needed. I'll rename.

src/weekdays.rs
Kinrany: It is used twice in `WeekdaySetIterFrom` and is complex enough that implementing and testing it separately makes sense to me, so I made it private and added a test that covers all possible arguments.

Kinrany:
All the removed code is in https://github.com/Kinrany/chrono/pull/1 for now.

Kinrany:
All resolved, except that I kept `split_at` as a private method.

I also added `from_array`, as we need _some_ way to construct literal values.

I see the code coverage warnings and this definitely needs a few more tests, but I'd like to make sure there are no more

djc:
Thanks, this looks like a good start. Going to make some tweaks in a followup PR.

wilfredjonathanjames:
Hey @Kinrany, is there any reason this wasn't made Serialize, Deserialize with the serde feature?

PR #1683, opened 2025-03-28
Add support for bincode v2

francisbr:
Following the latest major release of Bincode, the crate now relies on its own traits (Encode/Decode) it switched away from serde's De/Serialize. This PR is to add a feature flag here to optionally implement these traits on the DateTime structs.

The 2.0 bincode release: https://github.com/bincode-org/bincode/releases/tag/v2.0.0

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ __internal_bench = []
 [dependencies]
 num-traits = { version = "0.2", default-features = false }
 serde = { version = "1.0.99", default-features = false, optional = true }
+bincode = { version = "2.0.1", default-features = false, optional = true }
 pure-rust-locales = { version = "0.8", optional = true }
 rkyv = { version = "0.7.43", optional = true, default-features = false }
 arbitrary = { version = "1.0.0", features = ["derive"], optional = true }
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -40,6 +40,7 @@ __internal_bench = []
 [dependencies]
 num-traits = { version = "0.2", default-features = false }
 serde = { version = "1.0.99", default-features = false, optional = true }
+bincode = { version = "2.0.1", default-features = false, optional = true }
```
djc, line 43: This is a semver-incompatible change, so we can't merge this as is. All in all I'm not inclined to merge support for bincode 2 -- the upstream maintenance situation doesn't seem ideal given how long it has taken to release v2, and the stability just isn't there.

francisbr: I updated my implementation to be backward compatible with serde/bincode v1. You can see the tests I created to make sure encode & decoding between bincode v1 and v2 works.

As for the second part of you comment, I can't force anyone, but I feel just adding Traits impl behind a feature flag isn't too big of a compromise?

francisbr:
> Sorry, I'm still not inclined to merge this. There are plenty of competitive serialization formats (I like postcard, for example) that can rely on the existing serde traits, and merging this still means that I'm exposing a bunch of bincode API as part of the chrono public API, and the bincode maintainers have not earned enough of my trust for that.

Ok, I don't mind switching to postcard tbh. :)

However, we do need a more compact serialisation are you open to a new version for that(will mostlikely need a break semver upgrade (`0.5.0`?)

I am proposing this here: https://github.com/chronotope/chrono/pull/1684, I can implement it on the other struct if you like/want.

PR #1684, opened 2025-03-31
Compact serialization

francisbr:
Warning: Breaks chrono structs already serialized with non human readable serializer (postcard, bincode, ...) (problably want to to a "major update").

Here is a suggestion to implement a more compact serialization function for binary serializers. Encoding onle the timestamp (i64) & the nanoseconds (u32) instead of a full formatted string.

### DIFF
--- a/src/datetime/serde.rs
+++ b/src/datetime/serde.rs
@@ -1,11 +1,15 @@
 use core::fmt;
-use serde::{de, ser};
+use serde::{
+    de::{self, Error as _},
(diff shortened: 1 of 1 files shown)


djc:
> Warning: Breaks chrono structs already serialized with non human readable serializer (postcard, bincode, ...) (problably want to to a "major update").

Unfortunately this is a no-go. You can implement alternative serializers that must be enabled explicitly like we already have.

PR #1685, opened 2025-04-01
link - Update lib.rs

Hack666r:
fix old link to a new one


tgross35:
Fyi this user seems to be a bot, see e.g. https://github.com/rust-lang/libc/pull/4359 and https://github.com/MetisProtocol/metis-replica-node/pull/50

Issue #1686, opened 2025-04-03
NaiveWeek Eq trait inaccurate


Splashling1789:
The fact that the NaiveWeek Eq trait is derived and not manually implemented leads to inaccuracies when comparing weeks. For example:
`A = NaiveWeek {date: 03-04-2025, start: Monday}` and `B = NaiveWeek {date: 04-04-2025, start: Monday}` (In format day-month-year).

A and B are technically the same week. But when comparing it with the derived Eq trait, it will return false, as their dates are not equal.
A manual implementation of Eq, such as comparing the first day of each week `A.first_day() == B.first_day()` or something similar would resolve this issue and make week comparisons accurate.

This improvement would significantly enhance the reliability of this data type.

djc:
Ah, good point. Would you be able to submit a PR?

Splashling1789:
Done! :)

PR #1687, opened 2025-04-03
Implemented a consistent Eq trait for NaiveWeek.

Splashling1789:
NaiveWeek previously exhibited an inconsistency in the derived Eq and PartialEq traits, as discussed in Issue #1686. This issue arose because the derived implementations did not accurately compare weeks when the arbitrary date field differed.

To resolve this, I have implemented a custom PartialEq trait that ensures accurate week comparisons based on the first day of the week. The Eq trait remains derived, relying on the updated functionality of PartialEq. Additionally, I have included a test to validate this implementation and ensure its correctness.

### DIFF
--- a/src/naive/mod.rs
+++ b/src/naive/mod.rs
@@ -29,7 +29,7 @@ pub use self::internals::YearFlags as __BenchYearFlags;
 
 /// A week represented by a [`NaiveDate`] and a [`Weekday`] which is the first
 /// day of the week.
-#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+#[derive(Clone, Copy, Debug, Eq)]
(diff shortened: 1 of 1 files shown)


djc:
Can you fix the formatting and squash your commits into a single one? I'll deal with the (unrelated) MSRV issue separately.

djc:
So clippy is suggesting that this should also get a custom `Hash` implementation.

Splashling1789:
Ok, it should be fine by now

djc:
The no_std build is failing. Please rebase on current main to fix the MSRV issue.

Splashling1789:
Okay, I think everything is fine now. Please let me know if there’s anything else!

djc (approved):
Thanks!

PR #1691, opened 2025-04-05
TimeZone::from_posix_tz: Treat empty TZ variable as UTC

### LINKED ISSUE #1690: Empty `TZ` environment should use UTC, not local time
At least on Linux (glibc at least, musl as well AFAICT), `TZ=` forces UTC timezone, not the local timezone. That doesn't seem to be documented anywhere though...

Example C code:
```
#include <stdio.h>
#include <time.h>

int main(int argc, char *argv[]) {
    char outstr[200];
    time_t t;
    struct tm *tmp;

    t = time(NULL);
    tmp = localtime(&t);
    strftime(outstr, sizeof(outstr), "%c %Z", tmp);
    puts(outstr);
}
```

Example output. Setting `TZ=` is equivalent to setting `TZ=UTC`:
```
 $ ./test
Sat Apr  5 14:53:55 2025 CEST << that's /etc/localtime timezone
$ TZ= ./test
Sat Apr  5 12:53:57 2025 UTC
$ TZ=UTC ./test
Sat Apr  5 12:54:05 2025 UTC
```

Rust test program:
```
use chrono::Local;

fn main() {
    println!("{}", Local::now());
}
```

```
$ cargo run
2025-04-05 14:58:04.069014727 +02:00
$ TZ= cargo run
2025-04-05 14:58:06.276775336 +02:00 <<< This should probably be +02:00
$ TZ=UTC cargo run
2025-04-05 12:58:09.002152546 +00:00
```

This causes issues in uutils/coreutils: https://github.com/uutils/coreutils/issues/7498

drinkcat:
This is not technically POSIX, but glibc and musl behave that way: e.g. `TZ= date` returns a UTC date.

Fixes #1690.

### DIFF
--- a/src/offset/local/tz_info/timezone.rs
+++ b/src/offset/local/tz_info/timezone.rs
@@ -37,8 +37,9 @@ impl TimeZone {
 
     /// Construct a time zone from a POSIX TZ string, as described in [the POSIX documentation of the `TZ` environment variable](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html).
     fn from_posix_tz(tz_string: &str) -> Result<Self, Error> {
+        // It is commonly agreed (but not standard) that setting an empty `TZ=` uses UTC.
         if tz_string.is_empty() {
-            return Err(Error::InvalidTzString("empty TZ string"));
+            return Ok(Self::utc());
         }
 
         if tz_string == "localtime" {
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Thanks!

drinkcat:
@djc still not terribly urgent, but we'd really like to get this and #1693 for uutils/coreutils. Any chance we can get a new release? ,-) Thanks!!

djc:
> @djc still not terribly urgent, but we'd really like to get this and #1693 for uutils/coreutils. Any chance we can get a new release? ,-) Thanks!!

See #1698.

PR #1693, opened 2025-04-09
Add support for lossy format strings

### LINKED ISSUE #1692: Feature Request: Add non-panicking infallible format functions
Currently, there is no method in this crate that can format a timestamp according to an arbitrary string (e.g. user input). I propose a new set of functions that would ignore invalid sequences (i.e. copy them verbatim into the result). The only workaround I can find involves a lookaround regex (no worst case time complexity guarantees, potentially exponential) or a standard (O(m+n)) regex and a hack to allow for "overlapping" matches. If there's another way to do this that I haven't considered, let me know. If this functionality seems like something you'd want, I'm happy to open a PR.

Qelxiros:
closes #1692

@djc You described a `parse_lossy` function, but I ended up writing a `new_lossy` function. If you think it's better to set `lossy` at parse time, I'm happy to change it. However, it seems better to be able to set it for the `Iterator` directly.

### DIFF
--- a/src/format/strftime.rs
+++ b/src/format/strftime.rs
@@ -195,6 +195,7 @@ pub struct StrftimeItems<'a> {
     /// If the current specifier is composed of multiple formatting items (e.g. `%+`),
     /// `queue` stores a slice of `Item`s that have to be returned one by one.
     queue: &'static [Item<'static>],
+    lossy: bool,
     #[cfg(feature = "unstable-locales")]
     locale_str: &'a str,
     #[cfg(feature = "unstable-locales")]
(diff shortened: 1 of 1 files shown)


src/format/strftime.rs
```diff
@@ -454,16 +493,50 @@ impl<'a> StrftimeItems<'a> {
...
+
+                macro_rules! return_error {
+                    () => {
+                        return Some((remainder, error!()))
+                    };
+                    ($ch:expr) => {
+                        error!($ch, _);
+                        return_error!();
+                    };
+                }
```
djc, line 527: Sorry, I don't want all this macro mess here -- there's probably too much already. Can you figure out a better way?

Qelxiros: The most recent commit should be good. The coverage dropped because uncovered paths that used to be one line are now several.

src/format/strftime.rs
```diff
@@ -416,6 +455,23 @@ impl<'a> Iterator for StrftimeItems<'a> {
...
+    ) -> (&'b str, Item<'b>) {
+        if let Some(c) = ch {
+            if self.lossy {
+                *error_len -= c.len_utf8();
+            }
+        }
+        (
+            &original[*error_len..],
+            if self.lossy { Item::Literal(&original[..*error_len]) } else { Item::Error },
+        )
```
djc, line 472: I think this would be clearer if transposed:

```rust
if !self.lossy {
     return (&original[*error_len..], Item::Error);
}

if let Some(c) = ch {
     *error_len -= c.len_utf8();
}
(&original[*error_len..], Item::Literal(&original[..*error_len]))
```

djc (approved):
That's much better, thanks!

Wondering if `lossy` is the best name for this... maybe `lenient` or `copy_invalid`? Any thoughts/suggestions?

Qelxiros:
I agree that lossy isn't a great name, and lenient has parity with something like Java's `Calendar.setLenient`, so that seems like the best option. Let me know if you need anything else.

djc:
I don't think lenient formatting is important enough to earn API exposure on every type, please confine it to the format module.

Qelxiros:
Fixed.

djc (approved):

PR #1700, opened 2025-05-03
Add fuzzer for DateTime::parse_from_str

tyler92:
A new fuzz target has been implemented for `DateTime::parse_from_str`. Although the `format` string is typically constant, I believe it's worthwhile to randomize it to exercise corner cases that traditional unit tests may miss. This change improves fuzzing coverage by approximately 12%


tyler92:
I've simplified the code to avoid manual `Arbitrary` usage (libfuzzer supports it natively)

djc (approved):

djc:
Thanks!

Issue #1701, opened 2025-05-04
Local timezone in wasm always returns Single for MappedLocalTime


lutzky:
In Dublin, the time `2025-03-30T01:30` doesn't exist, as at 1AM the clock leaps forward to 2AM.

The following code returns different results on Linux and in WASM:

```rust
fn check_chrono_nonexistent_time() -> String {
    let explicit_dublin_tz: chrono_tz::Tz = "Europe/Dublin".parse().unwrap();
    let local_tz = chrono::Local;

    let nonexistent_time =
        chrono::NaiveDateTime::parse_from_str("2025-03-30T01:30", "%Y-%m-%dT%H:%M").unwrap();

    let explicit_dublin_tz_parse = explicit_dublin_tz.from_local_datetime(&nonexistent_time);
    let local_tz_parse = local_tz.from_local_datetime(&nonexistent_time);
    format!(
        "Explicit Dublin time parse: {explicit_dublin_tz_parse:?}\nLocal TZ parse: {local_tz_parse:?}",
    )
}
```

Linux (correct):

```
Explicit Dublin time parse: None
Local TZ parse: None
```

Web (incorrect):

```
Explicit Dublin time parse: None
Local TZ parse: Single(2025-03-30T01:30:00+01:00)
```

In my example I hardcode `Europe/Dublin` as the explicit timezone. However, this string can be fetched from JavaScript (e.g. using `js_sys`), it's `Intl.DateTimeFormat().resolvedOptions().timeZone`.

Now, of course, every user can implement that for themselves, but it would be neat if chrono did it on its own. However, this would add a dependency on `chrono_tz`, which is probably undesirable. I don't have a specific recommendation, but figured this is worth at least documenting in an issue.

djc:
Agreed this is crummy, but I'm not sure there's a (reasonable) solution.

lutzky:
While I also can't think of a reasonable solution, it might be worth warning the user about. A couple of ideas:

1. A warning in the documentation
2. Not sure if this is possible, but make `chrono::Local` deprecated for wasm, recommending instead to use an alias to it which would be called something like `chrono::LimitedDSTSupportLocal` - so we don't break anyone, and the classname makes it clear that it might be worth looking at the documentation to see the edge-case.

Happy to take a crack at it if it sounds good to you.

djc:
Let's start with a warning in the documentation.

lutzky:
Some details about what's causing this; the conversion in wasm happens here:

[...]

PR #1702, opened 2025-05-05
Add warning about MappedLocalTime and wasm

lutzky:
Warns about https://github.com/chronotope/chrono/issues/1701.

### DIFF
--- a/src/offset/mod.rs
+++ b/src/offset/mod.rs
@@ -46,6 +46,26 @@ pub use self::utc::Utc;
 ///   DST.
 /// * No result when the clock is turned forwards during a transition due to for example DST.
 ///
+/// <div class="warning">
+///
+/// In wasm, when using [`Local`], only the [`LocalResult::Single`] variant is returned.
+/// Specifically:
+///
+/// * When the clock is turned backwards, where `Ambiguous(earliest, latest)` would be expected,
+///   `Single(earliest)` is returned instead.
+/// * When the clock is turned forwards, where `None` would be expected, `Single(t)` is returned,
+///   with `t` being the requested local time represented as though there is no transition on that
+///   day (i.e. still "summer time")
+///
+/// This is caused because of limitations in the JavaScript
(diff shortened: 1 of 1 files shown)


src/offset/mod.rs
```diff
@@ -46,6 +46,21 @@ pub use self::utc::Utc;
...
+/// In wasm, when using [`Local`], only the [`LocalResult::Single`] variant is returned.
+/// Specifically:
+///
+/// * When the clock is turned backwards, where `Ambiguous(earliest, latest)` would be expected,
+///   `Single(earliest)` is returned instead.
+/// * When the clock is turned forwards, where `None` would be expected, `Single(t)` is returned,
+///   with `t` being the requested local time represented as though there is no transition on that
+///   day (i.e. still "summer time")
+///
+/// See further discussion and workarounds in <https://github.com/chronotope/chrono/issues/1701>.
```
djc, line 60: Can you add a sentence or two on why/how this happens? I think that would be helpful.

lutzky: I'd love to, but I couldn't really figure it out from a glance at the code. Could you point me in the right direction for "where do we actually ask the browser to do time conversion"?

djc: Presumably here?

https://github.com/chronotope/chrono/blob/main/src/offset/local/mod.rs#L61

lutzky: Ah, well, yep, it's in there all right. But it's kinda tautological:

https://github.com/chronotope/chrono/blob/361b9fc61eb9d495b41fa8de6296866323cb53e7/src/offset/local/mod.rs#L90-L91

djc: I guess it's fine to say that browsers always provide a single offset? Maybe even link to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTimezoneOffset for details.

djc: Maybe once we get [Temporal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal#browser_compatibility) we can do better in WASM?

djc (approved):

PR #1703, opened 2025-05-08
Fix wrong amount of micro/milliseconds

nmlt:
This is just a documentation change, so should be semver compatible.


djc (approved):
Thanks!

PR #1704, opened 2025-05-21
Migrate to core::error::Error

benbrittain:
Stabilized in rust 1.81, behind a feature flag so as to not bump the current MSRV of 1.61.


Cargo.toml
```diff
@@ -25,6 +25,7 @@ winapi = ["windows-link"]
 std = ["alloc"]
 clock = ["winapi", "iana-time-zone", "android-tzdata", "now"]
 now = ["std"]
+core_error = []
```
djc, line 28: Nit: replace the `_` with a `-` please? Otherwise this makes sense to me.

djc (approved):

Cargo.toml
benbrittain: done!

djc:
(FYI, your GitHub profile links to bwb.dev but its certificate expired like 2 years ago...)

PR #1705, opened 2025-06-03
Fix incorrect parsing of fixed-length second fractions

### LINKED ISSUE #1592: Err(ParseError(TooShort)) when parsing datetime with trailing 0
This was originally reported here: https://github.com/pola-rs/polars/issues/17167

To reproduce:
```rust
[package]
name = "scratch"
version = "0.1.0"
edition = "2021"

[dependencies]
chrono = "0.4.38"
```

`src/main.rs`
```rust
use chrono::NaiveDateTime;
fn main() {

    let result = NaiveDateTime::parse_from_str(
        "2024-06-03 20:02:48.6800000",
        "%Y-%m-%d %H:%M:%S%.6f0",
    );
    println!("{:?}", result);
    let result = NaiveDateTime::parse_from_str(
        "2024-06-03 20:02:48.680000",
        "%Y-%m-%d %H:%M:%S%.6f",
    );
    println!("{:?}", result);
}

```

prints
```
Err(ParseError(TooShort))
Ok(2024-06-03T20:02:48.680)
```

chris-leach:
When parsing, the "fixed-length decimal fraction" specifiers (`%.3f`, `%.6f`, `%.9f`) are currently treated identically to the variable-length specifier (`%.f`).

This causes issues when the fractional part is immediately followed by more digits signifying something completely different (you might think this bizarre, and I would agree, but I've seen upsetting things in vendor-specific log formats).

This PR changes the parser for those specifiers to consume only the exact number of digits specified, returning `TooShort` if there aren't enough. Test cases added for all three.

Fixes #1592.

### DIFF
--- a/src/format/parse.rs
+++ b/src/format/parse.rs
@@ -417,9 +417,30 @@ where
                         s = &s[2..];
                     }
 
-                    &Nanosecond | &Nanosecond3 | &Nanosecond6 | &Nanosecond9 => {
+                    &Nanosecond => {
                         if s.starts_with('.') {
                             let nano = try_consume!(scan::nanosecond(&s[1..]));
+                            parsed.set_nanosecond(nano)?;
+                        }
+                    }
+
+                    &Nanosecond3 => {
+                        if s.starts_with('.') {
+                            let nano = try_consume!(scan::nanosecond_fixed(&s[1..], 3));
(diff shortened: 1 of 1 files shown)


djc (approved):
Thanks!

PR #1712, opened 2025-07-08
Remove some `unsafe` on Windows

joshlf:


joshlf:
Oops, didn't realize that these were auto-generated. Is there a way to update the auto-generation to emit this code?

djc:
Not easily. Also not convinced I want to take on the extra dependency in a low-level crate like this.

joshlf:
> Not easily. Also not convinced I want to take on the extra dependency in a low-level crate like this.

Yeah no worries; feel free to just close.

PR #1721, opened 2025-08-12
Add `?Sized` bound to related methods of `DelayedFormat::write_to`

### LINKED ISSUE #1720: Change the signature of `DelayedFormat::write_to` from `impl Write` to `T: Write + ?Sized` to accepct `&mut dyn Write`
Currently, The `impl Write` in parameter position is equivalent to a generic bound `T: Write + Sized`, which prevents using `&mut dyn Write`.

```rust
pub fn write_to(&self, w: &mut impl Write) -> fmt::Result { ... }
```

Change the signature to `T: Write + ?Sized` is supposed to not be a breaking change. I can send a pr if you think this is okay.

**References**

https://doc.rust-lang.org/std/marker/trait.Sized.html

Huliiiiii:
Closes #1720


djc:
Thanks!

PR #1740, opened 2025-09-25
fix docs as there's no error being handled in MappedLocalTime::unwrap

2ndDerivative:
Self-explanatory, I think this is a leftover from some previous version


djc (approved):
Thanks!

djc:
Fixed the unrelated CI failures in #1741, please rebase.

PR #1655, opened 2025-01-23
Add optional defmt support

DBLouis:
This adds support for `defmt` behind a cargo feature.
Types in naive module have relatively "complex" `Debug` impl, I am unsure how to implement `defmt::Format` for those.

PR #1747, opened 2025-10-02
Add feature gated defmt support.

pebender:
This pull request adds defmt support gated behind Cargo.toml feature. It builds on the work of others and updates to the current chrono repo.

### DIFF
--- a/src/datetime/mod.rs
+++ b/src/datetime/mod.rs
@@ -1812,6 +1812,16 @@ impl<Tz: TimeZone> fmt::Debug for DateTime<Tz> {
     }
 }
 
+#[cfg(feature = "defmt")]
+impl<Tz: TimeZone> defmt::Format for DateTime<Tz>
+where
+    Tz::Offset: defmt::Format,
+{
+    fn format(&self, fmt: defmt::Formatter) {
+        defmt::write!(fmt, "{}{}", self.overflowing_naive_local(), self.offset);
+    }
+}
+
 // `fmt::Debug` is hand implemented for the `rkyv::Archive` variant of `DateTime` because
 // deriving a trait recursively does not propagate trait defined associated types with their own
 // constraints:
--- a/src/format/formatting.rs
+++ b/src/format/formatting.rs
@@ -474,6 +474,7 @@ impl OffsetFormat {
 ///
 /// See the `TimeZone::to_rfc3339_opts` function for usage.
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
 #[allow(clippy::manual_non_exhaustive)]
 pub enum SecondsFormat {
     /// Format whole seconds only, with no decimal point nor subseconds.
--- a/src/weekday.rs
+++ b/src/weekday.rs
@@ -38,6 +38,7 @@ use crate::OutOfRange;
 )]
 #[cfg_attr(feature = "rkyv-validation", archive(check_bytes))]
 #[cfg_attr(all(feature = "arbitrary", feature = "std"), derive(arbitrary::Arbitrary))]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub enum Weekday {
     /// Monday.
     Mon = 0,
@@ -253,6 +254,13 @@ impl fmt::Display for ParseWeekdayError {
 impl fmt::Debug for ParseWeekdayError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "ParseWeekdayError {{ .. }}")
+    }
+}
+
+#[cfg(feature = "defmt")]
+impl defmt::Format for ParseWeekdayError {
+    fn format(&self, fmt: defmt::Formatter) {
+        defmt::write!(fmt, "ParseWeekdayError {{ .. }}")
     }
 }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ name = "chrono"
 # Don't forget to adjust `ALL_NON_EXCLUSIVE_FEATURES` in CI scripts when adding a feature or an optional dependency.
 default = ["clock", "std", "oldtime", "wasmbind"]
 alloc = []
+defmt = ["dep:defmt", "pure-rust-locales?/defmt"]
 libc = []
 winapi = ["windows-link"]
 std = ["alloc"]
@@ -41,9 +42,10 @@ __internal_bench = []
 [dependencies]
 num-traits = { version = "0.2", default-features = false }
 serde = { version = "1.0.99", default-features = false, optional = true }
-pure-rust-locales = { version = "0.8", optional = true }
+pure-rust-locales = { version = "0.8.2", optional = true }
 rkyv = { version = "0.7.43", optional = true, default-features = false }
 arbitrary = { version = "1.0.0", features = ["derive"], optional = true }
+defmt = { version = "1.0.1", optional = true }
 
 [target.'cfg(all(target_arch = "wasm32", not(any(target_os = "emscripten", target_os = "wasi"))))'.dependencies]
 wasm-bindgen = { version = "0.2", optional = true }
--- a/src/date.rs
+++ b/src/date.rs
@@ -551,6 +551,16 @@ where
     }
 }
 
+#[cfg(feature = "defmt")]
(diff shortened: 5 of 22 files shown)


src/datetime/mod.rs
```diff
@@ -1817,6 +1817,16 @@ impl<Tz: TimeZone> fmt::Debug for DateTime<Tz> {
     }
 }
 
+#[cfg(feature = "defmt")]
+impl<Tz: TimeZone> defmt::Format for DateTime<Tz>
+where
+    Tz::Offset: defmt::Format,
+{
+    fn format(&self, fmt: defmt::Formatter) {
+        defmt::write!(fmt, "{}{}", self.overflowing_naive_local(), self.offset);
```
djc, line 1826: Instead of calling `overflowing_naive_local()`, should this forward to the `defmt::Format` impl for `NaiveDateTime`? Why/why not?

src/weekday.rs
```diff
@@ -256,6 +257,13 @@ impl fmt::Debug for ParseWeekdayError {
     }
 }
 
+#[cfg(feature = "defmt")]
+impl defmt::Format for ParseWeekdayError {
```
djc, line 261: Do you actually need a `defmt::Format` impl for these error types?

src/datetime/mod.rs
pebender: Because I cannot figure out how to do it.

NaiveDateTime has a method named format creates ambiguity with format from defmt::Format. The normal solution  would to cast to the Trait and call the function, e.g. <self.overflowing_naive_local() as defmt::Format>::format(fmt). However, defmt::Format is a dynamic trait, and Rust doesn't allow casts to dynamic Trait because it wants to know the size at compile time. If you know how to make it work, I would love to learn something new.

By the way, calling self.offset.format(fmt) works fine. However, since I couldn't figure out a way to make it work for self.overflowing_naive_local(), I saw little reason to split them apart.

src/weekday.rs
pebender: My general philosophy when added defmt::Format is that if the originator of the code felt there was a need for Debug, then there is a need for defmt::Format. When using the log crate for info!, warn!, error! etc, the Debug and Display traits are used. When using the defmt crate for info!, warn!, error! etc, the defmt::Format trait is used. So, if the author felt Debug was necessary, then defmt::Format is necessary.

Personally, I have find that lack of defmt::Format for error types is one of the first limitations I hit. Perhaps this is because errors are one of the first things that I send to the output when developing. The only thing I tend to send to the output before I get around to sending errors to the output is statements that say "you are got here", but those statements don't involve types. In time, I find that I am creating composite error types, When I get to that point, it is annoying to be required to create custom defmt::Format implementations to deal with the fact that the composite type wraps a error that doesn't support defmt.

djc: Fair enough.

src/datetime/mod.rs
pebender: I had thought you meant call the formatter using self.overflowing_naive_local().format(f) and self.offset.format(f).

The defmt::Format uses self.overflowing_naive_local() because it was copied from the fmt::Debug implementation which uses self.overflowing_naive_local(). I figured the safest thing was to use the same values as fmt::Debug rather than deviate and find out that there was some reason for using self.overflowing_naive_local() in fmt::Debug

pebender:
I have made the updates to reflect the 0.8.2 release of pure-rust-locales.

src/datetime/serde.rs
```diff
@@ -9,18 +9,22 @@ use crate::offset::{FixedOffset, Offset, TimeZone, Utc};
 
 #[doc(hidden)]
 #[derive(Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
```
djc: I don't think we need these.

src/datetime/tests.rs
```diff
@@ -1355,6 +1355,7 @@ fn test_datetime_is_send_and_copy() {
     }
 
     #[derive(Copy, Clone, Debug)]
+    #[cfg_attr(feature = "defmt", derive(defmt::Format))]
```
djc: This is test only, drop this.

src/format/formatting.rs
```diff
@@ -474,6 +474,7 @@ impl OffsetFormat {
 ///
 /// See the `TimeZone::to_rfc3339_opts` function for usage.
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
```
djc: I don't think this will be useful.

src/format/locales.rs
```diff
@@ -55,6 +55,7 @@ pub use pure_rust_locales::Locale;
 #[cfg(not(feature = "unstable-locales"))]
 mod unlocalized {
     #[derive(Copy, Clone, Debug)]
+    #[cfg_attr(feature = "defmt", derive(defmt::Format))]
```
djc: Drop this.

djc (commented):
Okay, I dislike the maximalist take of doing all the `Debug` impls. I think you should do the ones that you personally have a need for and then we can add more as needed.

src/datetime/serde.rs
pebender: Why do you think it is not needed? It is a public struct that needed Debug support at some point.

src/format/locales.rs
pebender: Agreed. I found some others that were types were scoped such that they are not accessible outside the crate and removed them as well.

src/datetime/serde.rs
djc: I think you're vastly overestimating how careful people are about adding `Debug` implementations. I understand your point that iterating on which implementations are necessary is annoying, but I disagree that any `Debug` derive or impl warrants a corresponding `Format` impl. I am maintaining chrono on a volunteer basis and you keep pushing back on my review feedback, causing multiple round trips that should have been unnecessary. I don't want to do a few more rounds of this.

I'm open to having impls for core public api types like `DateTime`, `NaiveDate`, etc.

I also understand the point about having them for error types.

I definitely don't want them for private types, for test-only types, or for tightly scoped types like these serde visitor impls, that are very unlikely to be reachable from the public API anyway.

I'll try one more review after this to see if we're close to something I can merge, please stop arguing now.

pebender: I appreciate that this is a voluntary 

As you can see from the comments and the code, I have taken your review feedback. I removed all the defmt that were added to non-public facing APIs. In addition, I removed a couple others that I believed where unlikely to be used even though they were public (e.g. iterators)

However, when the review feedback says "I don't like this" and "add them one-by-one rather than all at once to the public interface" with no technical or operational reason and I am going to explain my reason.

I don't understand why you consider that not taking feedback.

By the way, the chrono project has very few clippy directives. However, one of them is missing_debug_implementations. So, at some point, the attitude of the chrono project developers was to add Debug to everything not to add it as needed. I suspect their reason is the same reason as mine.

djc (approved):

src/datetime/serde.rs
djc: > As you can see from the comments and the code, I have taken your review feedback. I removed all the defmts that were added to non-public facing APIs. In addition, I removed a couple others that I believed where unlikely to be used even though they were public (e.g. iterators)

Thanks for that. I've merged this now.

> I don't understand why you consider this not taking feedback.

I just got frustrated by you restating your reasons with large walls of text.

> By the way, the chrono project has very few clippy directives in src/lib.rs. However, one of them is missing_debug_implementations. So, at some point, the attitude of the chrono project developers was to add Debug to everything not to add it as needed. I suspect their reason is the same reason as mine.

Good point, dropping that in #1753.

> However, I choose to take the time because it's more likely to help more people to have it in the mainline repository and release.

Appreciated, though I'll note that (compared to forking) this also just helps relieve yourself from future maintenance of the fork, whereas by contributing upstream you're asking a maintainer to indefinitely maintain additional surface. So I still think it's fairly asymmetrical.

pebender: Thank you for taking the time to review this pull request and for merging it.

I can get rather verbose. Sorry about that.

Yes, contributing it upstream means that that the maintenance workload shifts. I try to may pull requests either bug fixes for features other people appear to want (e.g. forked the repo to develop the feature as well).

By the way, I suspect you can close pull requests #1632 and #1655

PR #1748, opened 2025-10-03
Faster RFC 3339 parsing

djc:
Cleaned up version of

- #1738 

Thanks @sby1ce!


sby1ce:
Thanks for taking your time to do this, works great ❤️

Issue #1768, opened 2026-01-22
Soft-deprecating chrono and chrono-tz


djc:
In my [year in review post](https://dirkjan.ochtman.nl/writing/2026/01/09/reviewing-2025.html), I wrote:

> To balance out the ongoing adoption of orphaned crates, I'm inclined to wind down maintenance of chrono and chrono-tz in the coming months. The API design for these crates is quite dated, revising it would be a lot of effort, and with [jiff](https://crates.io/crates/jiff) there is now an alternative I feel comfortable recommending.

If someone wants to take over and they have a reasonable amount of reputation/references/standing, I will consider it (but strictly on my discretion).

daveman1010221:
Maybe I'm missing something, but Jiff doesn't appear to support rkyv. Seems like an odd "alternative".

djc:
I'm sure there is some way to work around that. Or maybe not, but I find these arguments to avoid doing so convincing:

https://github.com/BurntSushi/jiff/issues/537

daveman1010221:
Was just about to share that. lol

gdennie:
Is it possible to list all the test that would document the existing interface offered by chrono. We could then implement the tests and subsequently explore reformulating that interface as a major version bump.

djc:
> Is it possible to list all the test that would document the existing interface offered by chrono. We could then implement the tests and subsequently explore reformulating that interface as a major version bump.

I can't prevent you from working on that, but I won't be working on it. But keep in mind:

> If someone wants to take over and they have a reasonable amount of reputation/references/standing, I will consider it (but strictly on my discretion).

avsaase:
What does this mean for chrono-tz? Will it no longer be updated to reflect time zone database changes?

djc:
> What does this mean for chrono-tz? Will it no longer be updated to reflect time zone database changes?

Yes, that's right.

Oakchris1955:
I am not a fan of publishing new crates.io versions just to update, but I believe that the repository's README should be updated to at the very least include a link to this very issue.

Also, does chrono now match RustSec's [unmaintained definition](https://github.com/rustsec/advisory-db/blob/main/HOWTO_UNMAINTAINED.md)?

rjzak:
I'd be happy to help with the maintenance of chrono.

[...]

PR #1777, opened 2026-03-09
Add safe RFC 2822 formatter; deprecate panick-y to_rfc2822

ianthetechie:
I noticed (via a prod crash 😬) that the RFC 2822 string conversion will crash if the date is out of range (ish).

I also noticed that there was an effort to transform these crash-y APIs in https://github.com/chronotope/chrono/issues/1469. I've attempted to follow what seem like current conventions, but let me know if I've missed something.

I'm opening this against the 0.5.x branch as deprecations are technically semver breaking changes.

Open question to raise real quick as well (doesn't necessarily need to be addressed in this PR): RFC 2822 explicitly states that the year must be "1900 or later" 😂 Source: https://www.rfc-editor.org/rfc/rfc2822#section-3.3. `chrono` does not currently enforce this. I don't really have an opinion on whether it should, but raising that since I noticed.

### DIFF
--- a/bench/benches/chrono.rs
+++ b/bench/benches/chrono.rs
@@ -46,14 +46,14 @@ fn bench_datetime_from_str(c: &mut Criterion) {
     });
 }
 
-fn bench_datetime_to_rfc2822(c: &mut Criterion) {
+fn bench_datetime_to_rfc2822_opt(c: &mut Criterion) {
     let pst = FixedOffset::east(8 * 60 * 60).unwrap();
     let dt = pst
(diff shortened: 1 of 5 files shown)


djc:
Per

I don't think there will be a 0.5 anytime soon.

ianthetechie:
Ahh man, sorry I missed that @djc. Understood. Thanks for your maintenance of chrono and chrono-tz over the years! I'll have a look at jiff; seems like it will probably fit most of my use cases.

PR #1786, opened 2026-05-16
Bump defmt to 1.1.0

SimonIT:
I'm not sure if you would consider it to be semver-compatible


djc:
There is no need for this -- you can update to defmt 1.1 downstream.

SimonIT:
I did exactly that, but the build failed with the following error: error[E0277]: the trait bound `chrono::FixedOffset: defmt::Format` is not satisfied (https://github.com/vpetrigo/sntpc/actions/runs/25928678986/job/76221264763#step:4:138)

SimonIT:
Never mind, you're right. Sorry!

djc:
Feel free to submit a PR that includes a commit to add a derive (but don't remove any existing attributes/derives) and a commit to bump the version number.

PR #1787, opened 2026-05-24
fix(tz): reject TZ offset hour of 24 to avoid FixedOffset overflow

### LINKED ISSUE #1480: Panic with TZ="pBB24"

SAY-5:
Fixes #1480.

`parse_offset` in the POSIX TZ string parser accepted an offset hour of 24 (`0..=24`). A 24-hour offset is 86400 seconds, which does not fit in a `FixedOffset` (limited to ±23:59:59 / ±86399 seconds), so a TZ string such as `pBB24` parsed successfully but later panicked when the offset was converted to a `FixedOffset`.

This caps the offset hour at 23, matching the `FixedOffset` range, so an out-of-range offset is rejected at parse time with `InvalidTzString` instead of panicking downstream.

This change is semver-compatible, so it targets `main`.

Added `test_invalid_offset_hour` covering hour 24 (with and without sign) and the maximum valid `23:59:59` offset.

### DIFF
--- a/src/offset/local/tz_info/rule.rs
+++ b/src/offset/local/tz_info/rule.rs
@@ -357,7 +357,8 @@ fn parse_name<'a>(cursor: &mut Cursor<'a>) -> Result<&'a [u8], Error> {
 fn parse_offset(cursor: &mut Cursor) -> Result<i32, Error> {
     let (sign, hour, minute, second) = parse_signed_hhmmss(cursor)?;
 
-    if !(0..=24).contains(&hour) {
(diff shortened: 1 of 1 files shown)


djc (approved):
Fair enough.

PR #1795, opened 2026-06-12
Accept longer time zone designations and '_' in TZif files

### LINKED ISSUE #1793: Local silently falls back to UTC when /etc/localtime has a time-zone   designation outside 3–7 chars or containing '_'
### Summary

On a system whose `/etc/localtime` is a valid (libc/`zic`-accepted) `TZif` file with a time-zone *designation* longer than 7 characters — or containing characters like `_` — `chrono::Local` silently returns **UTC** instead of the real offset. The parser rejects the whole file, the error is swallowed, and `Local` falls back to UTC. Every chrono-based tool (e.g. `eza`) then displays UTC.

### Repro

`/etc/localtime` is a zone whose abbreviation is, e.g., `Solar_My_Little_House`:

```rust
fn main() {
    // prints `+00:00`; libc `date` / GNU `ls` print the correct offset on the same machine
    println!("{}", chrono::Local::now().offset());
}
```

`date` and GNU `ls` (which go through libc `localtime`) show the correct local time on the same machine — only chrono is wrong.

Parsing the same file with `tz-rs` (whose parser chrono's is derived from) surfaces the swallowed error:

```
invalid local time type: time zone designation must have between 3 and 7 characters
```

### Root cause

`src/offset/local/tz_info/timezone.rs`, `TimeZoneName::new`:

```rust
if !(3..=7).contains(&len) {
    return Err(Error::LocalTimeType("time zone name must have between 3 and 7 characters"));
}
// ...and the allowed set is only [0-9 A-Z a-z + -], so '_' is rejected too.
```

A single non-conforming **designation** makes the entire file parse fail → `Local` → UTC. The name is stored in a fixed `[u8; 8]`, which is why the 7-char ceiling exists.

### Why it's wrong

RFC 8536 / `tzfile(5)` say designations *SHOULD* be 3–6 characters — a recommendation, not a requirement. `zic` and glibc accept longer designations and a wider character set (e.g. `_`). A parser shouldn't discard valid **offset** data (the part that actually matters) over a cosmetic abbreviation, and certainly shouldn't silently degrade to UTC.

### Suggested fix

Make a non-conforming designation **non-fatal**: keep the offsets/transitions and store the name heap-allocated (or truncated / `None`) instead of returning `Err` for the whole timezone. At minimum, don't let `Local` fall back to UTC when the file is otherwise valid.

### Impact

Silent, wrong local time for all `chrono::Local` users on such systems — affecting every downstream tool. Surfaced via `eza` (eza-community/eza) showing UTC, root-caused to chrono.

### Environment

- chrono 0.4.45
- Linux, `TZ` unset, `/etc/localtime` a custom `TZif2` file with a 20-character designation containing `_`
- libc/coreutils read the same file correctly

W4HO-ham:
Fixes #1793.

`TimeZoneName::new` rejected any TZif time-zone designation outside 3–7 characters or outside `[0-9A-Za-z+-]`. But `zic` and glibc accept longer designations and a slightly wider character set (notably `_`), so on a system whose `/etc/localtime` uses such a (valid) designation, the entire file failed to parse and `chrono::Local` silently fell back to **UTC** — giving the wrong local time for every downstream tool.

This stores the designation in a 32-byte length-prefixed buffer (so `TimeZoneName` / `LocalTimeType` stay `Copy`, no API change), accepts up to 31 characters, and **truncates** longer designations rather than failing the parse — keeping the offset, which is what matters. It also allows `_`.

Validated against a real custom zone whose designation contains `_`: `chrono::Local::now().offset()` went from `+00:00` to the correct `-07:03:19`, matching `date '+%::z'` to the second. Adds regression tests covering a long designation, an underscore designation, and truncation of an over-long name.

### DIFF
--- a/src/offset/local/tz_info/timezone.rs
+++ b/src/offset/local/tz_info/timezone.rs
@@ -489,7 +489,7 @@ impl LeapSecond {
 #[derive(Copy, Clone, Eq, PartialEq)]
 struct TimeZoneName {
     /// Length-prefixed string buffer
-    bytes: [u8; 8],
+    bytes: [u8; 32],
 }
 
 impl TimeZoneName {
@@ -499,23 +499,28 @@ impl TimeZoneName {
     /// Time zone designations should consist of at least three (3) and no more than six (6) ASCII
     /// characters from the set of alphanumerics, “-”, and “+”. This is for compatibility with
     /// POSIX requirements for time zone abbreviations.
+    ///
+    /// That is a *should*, not a *must*: `zic` and glibc accept longer
+    /// designations and a slightly wider character set (notably `_`). Rejecting
+    /// an over-long designation would fail the whole TZif parse and make `Local`
+    /// fall back to UTC, so we keep the offset and simply truncate the name to
+    /// the 31-byte buffer, and additionally allow `_`.
     fn new(input: &[u8]) -> Result<Self, Error> {
-        let len = input.len();
+        // Truncate over-long designations rather than failing the parse.
+        let len = input.len().min(31);
 
-        if !(3..=7).contains(&len) {
-            return Err(Error::LocalTimeType(
-                "time zone name must have between 3 and 7 characters",
-            ));
+        if len < 3 {
+            return Err(Error::LocalTimeType("time zone name must have at least 3 characters"));
         }
 
-        let mut bytes = [0; 8];
-        bytes[0] = input.len() as u8;
+        let mut bytes = [0; 32];
+        bytes[0] = len as u8;
 
         let mut i = 0;
         while i < len {
             let b = input[i];
             match b {
-                b'0'..=b'9' | b'A'..=b'Z' | b'a'..=b'z' | b'+' | b'-' => {}
+                b'0'..=b'9' | b'A'..=b'Z' | b'a'..=b'z' | b'+' | b'-' | b'_' => {}
                 _ => return Err(Error::LocalTimeType("invalid characters in time zone name")),
             }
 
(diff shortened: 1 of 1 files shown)


src/offset/local/tz_info/timezone.rs
```diff
@@ -499,23 +499,28 @@ impl TimeZoneName {
     /// Time zone designations should consist of at least three (3) and no more than six (6) ASCII
     /// characters from the set of alphanumerics, “-”, and “+”. This is for compatibility with
     /// POSIX requirements for time zone abbreviations.
+    ///
+    /// That is a *should*, not a *must*: `zic` and glibc accept longer
+    /// designations and a slightly wider character set (notably `_`). Rejecting
+    /// such a designation here fails the whole TZif parse and makes `Local` fall
+    /// back to UTC, so we accept names up to 63 bytes and additionally allow `_`.
```
djc: Why 63?

W4HO-ham: I run a few different custom time zones all involving local solar time and ham radio for the longitude I'm in. The longest name I have so far is 29 chars, so a 32 char max seemed tight, so I chose 64. Well, minus 1 to accommodate the leading length byte.

djc: Suggest we use 31 and instead of failing the parse, cut off anything that's longer than that.

W4HO-ham: That sounds fine. Shall I just submit a fresh PR? (I'm new to this.)

W4HO-ham: Ahhh, ok, pushed to this same branch: switched to a 32-byte buffer (31-char names) and truncate anything longer instead of failing the parse, so the offset is always preserved. 73!

djc (approved):

djc:
Thanks!

W4HO-ham:
Thank you! 🚀

PR #1796, opened 2026-06-16
fix(tz): avoid i64 overflow in find_local_time_type_from_local

nvxbug:
find_local_time_type_from_local adds a transition's unix_leap_time straight to a local time type's UTC offset, but unix_leap_time is read verbatim from a TZif file (read_be_i64) and the offset is only screened against i32::MIN. A file with a transition near i64::MAX or i64::MIN and a same-signed offset overflows that i64 add while mapping any local datetime, so debug builds panic and release builds wrap and select the wrong offset. TZif data reaches this path when TZ points at a crafted file or a tampered /etc/localtime.

Switch the two transition_start/transition_end additions to saturating_add. The companion unix-time lookup already guards the same arithmetic with checked_add, so this keeps the two paths consistent. Saturating leaves the comparisons against a normal local timestamp correct and avoids handing the caller an error it would unwrap into a panic.

### DIFF
--- a/src/offset/local/tz_info/timezone.rs
+++ b/src/offset/local/tz_info/timezone.rs
@@ -237,8 +237,10 @@ impl<'a> TimeZoneRef<'a> {
 
                 // the end and start here refers to where the time starts prior to the transition
                 // and where it ends up after. not the temporal relationship.
-                let transition_end = transition.unix_leap_time + i64::from(after_ltt.ut_offset);
-                let transition_start = transition.unix_leap_time + i64::from(prev.ut_offset);
+                let transition_end =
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Fair enough. I guess this is only theoretical, because there are no real world TZif files that would hit this?

nvxbug:
Yeah, pretty much theoretica no real tzdata has transitions that far out, so normal systems never hit it. You'd only get there with a crafted TZif or a tampered /etc/localtime. Mostly I just wanted it to match the unix-time lookup, which already uses checked_add.
This is one of my first open-source contributions, so thanks for the quick review and merge — really appreciate it. Hoping to send more.

PR #1797, opened 2026-06-17
fix(tz): avoid out-of-bounds slice on partial tzdata index entry

nvxbug:
On Android and OpenHarmony the local time zone is read from a binary `tzdata` file. `TzDataIndexes::new` splits the index region with `buf.chunks(ENTRY_LEN)`, but when the region size isn't a multiple of `ENTRY_LEN` the final chunk is shorter than a full entry. The closure then slices `chunk[..TZ_NAME_LEN]` (and the offset/length fields past it) unconditionally, so a truncated or crafted `tzdata` file panics with an out-of-bounds slice index instead of returning an error.

Switch to `chunks_exact` so the trailing partial chunk is dropped. It can never hold a valid entry, and this matches the upstream ZoneInfoDb loader which walks a fixed count of full-size entries. Added a regression test that drives the index parser with a non-aligned index region.

### DIFF
--- a/src/offset/local/tz_data.rs
+++ b/src/offset/local/tz_data.rs
@@ -102,10 +102,12 @@ impl TzDataIndexes {
     fn new<const ENTRY_LEN: usize>(mut reader: impl Read, header: &TzDataHeader) -> Result<Self> {
         let mut buf = vec![0; header.data_offset.saturating_sub(header.index_offset) as usize];
         reader.read_exact(&mut buf)?;
-        // replace chunks with array_chunks when it's stable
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Note:

So I'm not super eager to continue reviewing PRs for theoretical issues like these.

nvxbug:
Makes sense. These are only reachable with crafted tzdata input, so they're firmly in the theoretical bucket, and I caught the wind-down note. I'll hold off on more of these. Thanks for taking it.

PR #1798, opened 2026-06-18
fix: clamp offset minute rounding to avoid invalid +24:00

nvxbug:
`OffsetFormat::format` rounds an offset's seconds to the nearest minute before printing at minute precision, which is the path `%z`, `%:z`, `to_rfc3339`, `to_rfc2822` and the `Display` of `DateTime<FixedOffset>` all take. `FixedOffset` accepts values up to 23:59:59, so any offset from 23:59:30 onward rounds to 1440 minutes and prints as `+24:00`. The offset hour is only valid as 00-23, and `parse_from_rfc3339` rejects `+24:00` with `OutOfRange`, so the formatted value no longer parses back.

Clamp the rounded minute count to `24 * 60 - 1` so the hour stays below 24 and those edge offsets print as `+23:59`. Keeping the clamp in the shared offset writer covers the `%z` family and both RFC helpers from one spot, and offsets that already fit keep rounding exactly as before.

### DIFF
--- a/src/format/formatting.rs
+++ b/src/format/formatting.rs
@@ -410,8 +410,10 @@ impl OffsetFormat {
                 OffsetPrecision::Hours
             }
             OffsetPrecision::Minutes | OffsetPrecision::OptionalMinutes => {
-                // Round seconds to the nearest minute.
-                let minutes = (off + 30) / 60;
+                // Round seconds to the nearest minute, but never roll over to 24:00. A
+                // `FixedOffset` may be as large as 23:59:59, and rounding that up would emit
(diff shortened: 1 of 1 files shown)


djc:
Do you not remember that chrono is soft-deprecated?

nvxbug:
Yeah, I know it's soft-deprecated. Figured this one was still worth flagging since it's a round-trip correctness bug (chrono emits +24:00, which its own parser then rejects) and the fix is a one-line clamp rather than anything new. It's a narrow edge though, only sub-minute offsets near 23:59:59 hit it, so if that's below the bar for maintenance-only I'm happy to close it. Your call.

djc (approved):

PR #1799, opened 2026-06-19
fix(tz): avoid usize overflow computing TZif data block sizes

nvxbug:
On a 32-bit target (i686, armv7/Android, wasm32), parsing a crafted TZif file:

    thread '...' panicked at src/offset/local/tz_info/parser.rs:
    attempt to multiply with overflow

`State::new` sizes each data block as `count * elem_size` in `usize`, where `count` is an unvalidated `u32` read straight from the file header. On a 32-bit `usize`, `type_count * 6` (and the transition/leap equivalents) wraps. In release the small wrapped length lets the read succeed on a tiny file, after which `Vec::with_capacity(type_count)` is asked for the raw count of hundreds of millions of records.

Reachable from `TZ=:<file>`, a tampered `/etc/localtime`, or the Android/OHOS tzdata loader. Switched the three block-size calculations to `checked_mul`. The added tests fail (overflow panic) on the 32-bit CI jobs without the fix.

### DIFF
--- a/src/offset/local/tz_info/parser.rs
+++ b/src/offset/local/tz_info/parser.rs
@@ -129,13 +129,20 @@ impl<'a> State<'a> {
             false => 8,
         };
 
+        // The header counts are unvalidated `u32`s read from the file. Their byte
+        // sizes are computed in `usize`, which is 32-bit on targets such as i686,
+        // armv7/Android and wasm32, so the products below overflow there.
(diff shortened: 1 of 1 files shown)


djc:
Given the soft-deprecated state, I don't care about crafted TZif files.

nvxbug:
Fair enough, that's a reasonable call given chrono's status. I'll leave this one.

PR #1801, opened 2026-06-22
fix: parse offset seconds in FromStr so Display round-trips

nvxbug:
FixedOffset's Display and Debug print a sub-minute offset as +HH:MM:SS, for example +00:00:30, but scan::timezone_offset only reads +HH:MM. That scanner backs FromStr for both FixedOffset and DateTime<FixedOffset> (through parse_rfc3339_relaxed), so a value chrono prints cannot be parsed back: DateTime<FixedOffset>::from_str rejects its own to_string() with TooLong on the trailing :SS, and FixedOffset::from_str silently drops the seconds and returns a truncated offset.

The fix teaches timezone_offset to consume an optional seconds component behind a new allow_seconds flag, applying the sign inside the scanner so negative offsets like -00:00:30 stay correct, and turns it on only for the two FromStr paths. Keeping it in the scanner fixes both round-trips in one place while strict parse_from_rfc3339, parse_from_rfc2822, and the %z family keep rejecting seconds. Regression tests cover the FixedOffset and DateTime paths across positive and negative sub-minute offsets.

### DIFF
--- a/src/datetime/tests.rs
+++ b/src/datetime/tests.rs
@@ -1013,6 +1013,22 @@ fn test_datetime_from_str() {
 }
 
 #[test]
+fn test_datetime_from_str_suboffset_roundtrip() {
+    // `Display` prints a sub-minute offset as `+HH:MM:SS`, so `FromStr` must read that
+    // back instead of choking on the trailing `:SS`.
+    for secs in [30, -30, 71608, -71608, 86399, -86399] {
+        let offset = FixedOffset::east_opt(secs).unwrap();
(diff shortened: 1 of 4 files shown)


djc:
I'm going to stop responding to these PRs for edge case bugs. There are probably a lot more; you should direct your efforts towards other implementations.

nvxbug:
Makes sense, I'll redirect to other things. Happy to close this if you'd rather not keep it on the pile, otherwise it's here if the round-trip ever matters.

PR #1803, opened 2026-06-28
Replace `target_arch = "wasm32"` with `target_family = "wasm"`

SuchAFuriousDeath:
Web-target cfg gates (the wasm-bindgen / js-sys code paths and the `wasm-pack test --node` harness) currently key on
`target_arch = "wasm32"`. As a result they silently disappear when building for `wasm64-unknown-unknown`: the `wasm-bindgen` and `js-sys` optional dependencies are not pulled in, `Utc::now()`, `Local`'s offset lookup, and the `js_sys::Date` conversions fall back to the non-web default (which calls `SystemTime::now()` and panics on wasm), and the `tests/wasm.rs` harness is compiled out.

Switching the gates to `target_family = "wasm"` makes the same web backend apply to any WebAssembly target, including `wasm64`. The existing `not(any(target_os = "emscripten", target_os = "wasi"[, "linux"]))` exclusions are preserved, so behavior on `wasm32-unknown-emscripten`, `wasm32-wasip*` and emscripten/WASI variants is unchanged.

`target_family = "wasm"` has been stable since Rust 1.54, well below chrono's MSRV (1.62).

`wasm64-unknown-unknown` is a Tier 3 target, so it cannot be added to CI, but the same code paths are still exercised by the existing `wasm-pack test --node --features wasmbind` job on `wasm32`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ rkyv = { version = "0.7.43", optional = true, default-features = false }
 arbitrary = { version = "1.0.0", features = ["derive"], optional = true }
 defmt = { version = "1.0.1", optional = true }
 
-[target.'cfg(all(target_arch = "wasm32", not(any(target_os = "emscripten", target_os = "wasi"))))'.dependencies]
+[target.'cfg(all(target_family = "wasm", not(any(target_os = "emscripten", target_os = "wasi"))))'.dependencies]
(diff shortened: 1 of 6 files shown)


djc (approved):

djc:
Thanks. Note that chrono is soft-deprecated:

SuchAFuriousDeath:
Oh. Thanks for the heads up then.

PR #1804, opened 2026-07-09
Remove unsound ExactSizeIterator for NaiveDate day/week iterators

### LINKED ISSUE #1757: `NaiveDateDaysIterator` (`NaiveDate::iter_days`) behaves unexpectedly when reversed
The iterator returned by `NaiveDate::iter_days` behaves unexpectedly when iterated in reverse. Consider:

```rust
#[test]
fn test() {
  let start = chrono::NaiveDate::from_ymd_opt(2025, 10, 10).unwrap();

  let mut next_seven_rev1: Vec<_> = start.iter_days().take(7).collect();
  next_seven_rev1.reverse();

  let next_seven_rev2: Vec<_> = start.iter_days().take(7).rev().collect();

  assert_eq!(next_seven_rev1, next_seven_rev2);
}
```

I would expect the assertion here to pass, but it does not:

```
assertion `left == right` failed
  left: [2025-10-16, 2025-10-15, 2025-10-14, 2025-10-13, 2025-10-12, 2025-10-11, 2025-10-10]
 right: [-258092-07-28]
```

---

My brief analysis of this follows. Take it with a grain of salt though, as I might be misunderstanding exactly how the various iterator traits are expected to interact.

From what I can tell, I think this behaviour is due to `NaiveDateDaysIterator` implementing `DoubleEndedIterator` and `ExactSizeIterator`, which the `Take` iterator [takes advantage of](https://doc.rust-lang.org/1.90.0/src/core/iter/adapters/take.rs.html#162), but:

- `start.iter_days().len()` reports the number of days until `NaiveDate::MAX`
- `start.iter_days().next_back()` is _not_ `NaiveDate::MAX`

And in particular, the latter seems to be the source of the issue. Per the docs on `DoubleEndedIterator`:

> It is important to note that both back and forth work on the same range, and do not cross: iteration is over when they meet in the middle.

So `start.iter_days().next_back()` should be equivalent to `start.iter_days().last()`, but this is not the case.

If I'm right, then the fix would seem to be to remember the starting date in `NaiveDateDaysIterator` and wrap around there instead of at `NaiveDate::MIN`, but that would break any current uses of `NaiveDate::iter_days().rev()`. So it might also necessitate adding some kind of `NaiveDate::iter_days_backwards()` API. In any case, this would be a backwards-incompatible change.

vidigoat:
Fixes #1757.

`NaiveDateDaysIterator` and `NaiveDateWeeksIterator` are unbounded in both directions: `next` walks forward toward `NaiveDate::MAX` and `next_back` walks backward toward `NaiveDate::MIN`. Both directions start at the same date and never meet, so they do not share a finite range.

Despite that, both types implement `ExactSizeIterator`, and their `size_hint` reports the *forward* distance to `NaiveDate::MAX` as an exact length. Any adapter that trusts that length while iterating from the back then produces nonsense. The reported case:

```rust
let start = NaiveDate::from_ymd_opt(2025, 10, 10).unwrap();
let mut forward: Vec<_> = start.iter_days().take(7).collect();
forward.reverse();
let backward: Vec<_> = start.iter_days().take(7).rev().collect();
assert_eq!(forward, backward);
```

fails because `Take::rev` calls `ExactSizeIterator::len()` to decide how many elements to skip from the back. `len()` returns ~the number of days between the start date and `NaiveDate::MAX`, so `rev` skips almost all of them and yields a single date near `NaiveDate::MIN` (`[-258092-07-28]`) instead of the seven expected days.

`DoubleEndedIterator` is well-defined for these iterators (stepping backward in time from the starting date, as documented and tested via `.rev().take(n)`); the `ExactSizeIterator` impls are the incorrect part, so this drops them. The documented reverse behaviour is unchanged, and `.take(n).rev()` now fails to compile instead of returning wrong values.

Removing a trait impl is technically a breaking change, so this targets `0.5.x` per the PR template. The maintainer indicated in #1757 they would accept a correctness fix here.

Added `test_day_iterator_rev` and `test_week_iterator_rev` covering the reverse walk and the `next`/`next_back` starting point for both iterators.


djc:
0.5 is not coming:

(Not sure why you changed the base branch?)

vidigoat:
Sorry for the churn on the base — you mentioned in #1757 you'd be open to this on 0.4.x, and since `main` is the 0.4.x line (with 0.5 retired per #1768), I retargeted it here so it lands on the active release line rather than the 0.5.x branch it was originally opened against. The diff is just the two `ExactSizeIterator` removals plus doc comments and the `rev()` regression tests. Happy to adjust if you'd rather handle it differently.

djc:
I don't think I was envisioning the `impl` would be removed when I wrote that? Also, I hadn't yet made the decision to soft-deprecate chrono. At this point, I don't think I want to take a fix that drops an `impl`. Maybe the impls can be improved to be better?

vidigoat:
Makes sense given the soft-deprecation — not worth an API break for this one. I'll close it out. The `.take(n).rev()` behavior is at least captured in #1757 now for anyone who runs into it. Thanks for taking a look so quickly!

PR #1813, opened 2026-08-14
Fast formatting

MassivePizza:
### Thanks for contributing to chrono!

If your feature is semver-compatible, please target the main branch;
for semver-incompatible changes, please target the `0.5.x` branch.

Please consider adding a test to ensure your bug fix/feature will not break in the future.

PR #1817, opened 2026-09-02
Fix `%C` for years outside 0..=9999

dylanpulver:
`YearDiv100` and `IsoYearDiv100` cast the century to `u8` before `write_two`, but `year().div_euclid(100)` spans `-2622..=2621`. The cast wraps, `write_two` has no upper bound, and `b'0' + tens` walks past `'9'`:

```
year     -99: %C = "I5"     (the docs' own example says -1)
year    -101: %C = "I4"
year   10000: %C = ":0"
year   12345: %C = "<3"
year -100000: %C = "24"     <- looks like a century, is not one
```

Sweeping every year `NaiveDate` accepts and comparing `%C` against `year.div_euclid(100)`: **514286 of 524286 disagree** on `main`, **0 of 524286** with this change. Only `0..=9999` was ever right.

`%y` is fine — `rem_euclid(100)` is always in range. This arrived in 0cfc405 ("Optimize number formatting"), so it is in every release from 0.4.38.

The fix routes both arms through `write_n`, the same fallback `write_year` already uses outside its fast path. Output for `0..=9999` is unchanged under `%C`, `%-C`, `%_C` and `%0C` (asserted). The sibling helper `write_hundreds` already guards `n >= 100`; `write_two` did not, so I added a debug assertion — every remaining caller is bounded below 100.

Adding just that guard is not enough on its own: with it, `format("%C").to_string()` panics with "a Display implementation returned an error" instead of printing garbage.

Existing coverage missed this because `test_date_format` has a "non-four-digit years" block that walks `%Y` from -12345 to 12345 and never once asks for `%C`. The three `%C` assertions on `main` are at years 2012, 2001 and 2015 (the last two in doc examples).

`cargo test --features "default unstable-locales rkyv-32 rkyv-validation serde arbitrary"` is green (328 lib + 329 doc), `cargo fmt --check` clean, on rustc 1.97.0.

Not addressed here: the same footnote claims year -99 prints `99` for `%y`, but floor division gives `1`. That looks like a docs error rather than a code one, so I left it.

Draft PR #1813 rewrites this region for speed and keeps the `u8` truncation, so the two will conflict textually if it lands.

AI disclosure: prepared with AI assistance (Claude Opus 5, `claude-opus-5`). The numbers above are from the commands as shown.


dylanpulver:
The three red jobs are not from this change:

- `cargo-deny` and `llvm-cov` fail the same way on #1812 and #1813.
- `rust_versions (beta)` fails on three doctests that use `core::i32::MAX`, in `naive/date/mod.rs` and `naive/datetime/mod.rs`. This PR only touches `src/format/formatting.rs`, and the lib tests in that same job pass 328/328.

djc:
See

I'm not taking any more changes unless there's something really serious.

PR #1820, opened 2026-09-13
fix docs as there's no error being handled in MappedLocalTime::unwrap

2ndDerivative:
Rebase of #1740 

Sorry if I screwed up the git operations there, I don't use rebase a lot


2ndDerivative:
@djc you had reviewed the original, this is just a rebase off that

djc:
Yes, but in the meantime chrono has stopped taking changes:

PR #1822, opened 2026-09-21
Implement Iterator::size_hint for WeekdaySetIter

shoemoney:
### Bug

`WeekdaySetIter` (src/weekday_set.rs:360) implements `ExactSizeIterator` by overriding only `len()`, and never overrides `Iterator::size_hint`. It therefore reports the default `(0, None)` while `len()` returns the exact count, which violates the documented `ExactSizeIterator` requirement that `size_hint` return the exact size.

On main before this change:

```
WeekdaySet::ALL.iter(Mon).size_hint()  == (0, None)   // len() == 7
WeekdaySet::single(Wed).iter(Mon).size_hint() == (0, None)   // len() == 1
```

Observable consequence: `WeekdaySet::ALL.iter(Mon).collect::<Vec<_>>()` grows 4 -> 8 instead of making one exact allocation of 7, because `Vec`'s `FromIterator` preallocates from `size_hint().0`. Any adapter that propagates `size_hint` (`zip`, `chain`, `take`) or user code doing `Vec::with_capacity(iter.size_hint().0)` sees 0. `WeekdaySet::iter` is public API and the only way to enumerate a `WeekdaySet`.

This is the same iterator-contract class as #1809, but the `WeekdaySet` iterator was never given a `size_hint` at all.

### Fix

Purely additive: add `size_hint` to the existing `impl Iterator for WeekdaySetIter`. No API break, no trait impl removed. `self.days.len()` is already the exact remaining count (it is what `ExactSizeIterator::len` delegates to), and `next`/`next_back` each remove exactly one day, so the hint stays exact under mixed forward/backward consumption.

### Test

`iter_size_hint_matches_len` in the existing inline `mod tests` walks all 128 `WeekdaySet` values and all 7 start days, alternating `next()` and `next_back()`, asserting `size_hint() == (len(), Some(len()))` after every step down to `(0, Some(0))`.

Before the fix (test only, source pristine):

```
test weekday_set::tests::iter_size_hint_matches_len ... FAILED

thread 'weekday_set::tests::iter_size_hint_matches_len' panicked at src/weekday_set.rs:508:9:
assertion `left == right` failed
  left: (0, None)
 right: (7, Some(7))

test result: FAILED. 3 passed; 1 failed; 0 ignored; 0 measured; 289 filtered out
```

After:

```
test weekday_set::tests::debug_prints_8th_bit_if_not_zero ... ok
test weekday_set::tests::bitwise_set_operations_preserve_8th_bit_invariant ... ok
test weekday_set::tests::iter_size_hint_matches_len ... ok
test weekday_set::tests::split_at_is_equivalent_to_iterating ... ok

test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 289 filtered out
```

`cargo test`, `cargo test --no-default-features`, `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` all pass.

### Deliberately not fixed

I did not audit the other iterators in the crate for the same gap, and I did not add a `TrustedLen` impl (that is nightly-only and a larger commitment than this fix).


djc:
chrono is soft-deprecated and not accepting any more changes in principle:

shoemoney:
Understood, closing. Thanks for the pointer to #1768.

